Issue #777, opened 2017-05-28
public contact page - settings button and indicator


tomholub:
Check out "Public Contact Page" link on the bottom of settings

this will become one of the tiles in cryptup settings, but it's missing a tile icon

Please make a PR where you add it to the right of NEW EMAIL

mondo989:
@tomholub got ya. 
Will the 'Public Contact Page' also be included in the Advanced Settings as well?

tomholub:
I think just the basic ones

tomholub:
@mondo989 for pull requests, it's ok to work only in the Chrome directory. I have a script to copy it into Firefox as needed, and I can do that as I merge the commit.

Must be a hassle to keep them both in sync manually. It's fine to just work on Chrome only.

tomholub:
@mondo989 one note, except in cases where it makes a lot of sense (eg the drop here overlay), please don't put user facing text into CSS. It will make it hard to make it multi-lingual down the road. Thanks!

PR #781, opened 2017-05-29
add public profile icon to basic settings

mondo989:
Issue #777 
screen shot 2017-05-29 at 12 56 58 am

We may want to use a different icon for the public profile.
(Layout is ready)


tomholub:
I wonder if we should call it contact page

The most descriptive would be "encrypted contact page" but that is too long

Now that I'm looking at it, it looks a bit scary because it's a privacy
product - "public profile" is the kind of phrase that could freak people out

--
Tom

mondo989:
I'll update it as a 'Contact Page' agree that the most descriptive wording may be a bit too verbose. 

I also have a similar concern, having Public Profile in the basic settings.
Being that it is clearly presented with the other important settings functions (including sending encrypted email), it almost implies that we're giving everyone public profiles. 

I just played around a little bit, but what if we put it (public profile) as a smaller feature in the corner or somewhere not as prominent as each icon in the settings is. 
By having it smaller and toggleable, it might make it clear that this is a secondary feature and something that can easily be turned off or on. Quickly seeing the toggle is turned off in the settings could put me a bit at ease. 

Just thinking out loud. What do you think?

tomholub:
I actually want people to use this a lot. It's the best marketing tool if
it works well. So whatever will make it prominent

--
Tom James Holub <http://holub.me/>

mondo989:
@tomholub Got ya. 
To increase the visibility of it, what if we were to present the 'Public Profile' in a way that stands out to the user (bottom left corner). Having it here breaks the norm of the settings flow.

Playing around with some ideas... What do you think about some of these drafts?

a) Basic settings, including public profile in the bottom left corner.
basic-a

b) Advanced settings has "My Keys" and "Security" content inside of collapsible accordions. (Also has public profile toggleable).
advanced-a

c) Advanced settings toggled open & closed.
advanced-b

tomholub:
It should definitely be on the basic settings page. I see what you did there with the toggle and it makes sense, but there's nothing to group the toggle with. 

Now that I'm looking at it "Encrypted Contact Page" is a good name after all. It's just too long for the tiles, but elsewhere it's fine and it's very clear.

A minor issue with the toggle is that I won't turn it on until I understand what it is, and clicking the (?) just to understand it sounds like an extra. I have to set up a few things before it will be put on so it makes sense to put that all on one page. Instead of a toggle, there will just be a "status doughnut" and the whole thing will be a button.

What I like about this is that it's the only element in red color when people install it, so it will definitely catch their eye. But once they set it up, it will be green so it won't be unnecessarily getting attention.

like below (example), except it will say "**Encrypted Contact Page**"

image

mondo989:
Hey Tom,

I like the red element's contrasting, it does a good job with us for indicating something is wrong/stop/danger/attention. Also, I agree with your logic behind the "status donut," as opposed to toggling actions.  

Although, I do see a value in providing additional context in regards to 'Public Contact Page' or 'Encrypted Contact Page.' 

Initially, I was thinking of showing the "?" icon, and on hover showing a "tool tip" to provide some additional information. As myself advocating for beginners, I'd imagine that not everyone would understand the context or use case of it ( It being 'Public Contact Page' or 'Encrypted Contact Page' ). Although the wording is quite explicit. 

After thinking a bit more about the tool tip/question marks, I think this may work...

screen shot 2017-06-02 at 7 12 08 pm

We provide a mini popup with some context, and a close icon to remove it. 
(Probably modifying the copy a bit) 
We still keep the status donut the way it is. 

What do you think?

mondo989:
@tomholub Also, in regards to all element's positioning. 
I feel like our top navbar is becoming overcrowded with content, especially if users are using smaller browsers...

(elements)
screen shot 2017-06-02 at 7 20 52 pm
(spacing is a bit exaggerated) 

screen shot 2017-06-02 at 7 46 59 pm
(Screenshot is at 960px)

screen shot 2017-06-02 at 7 50 15 pm



Wanted to play around with the positioning of elements just a tad. 

screen shot 2017-06-02 at 7 43 10 pm

screen shot 2017-06-02 at 7 42 02 pm

I like it at the bottom left/right corner, it reduces the cognitive load of the user on the nav. 
Being alone in the corner also helps to bring attention/differentiate it. 

Let me know any feedback here too.

tomholub:
I kind of like this one 

image

tomholub:
but keeping it all on the top for now is also ok with me. The version on the bottom may clash if we wanted to keep it on the advanced page

tomholub:
I mean, in "Additional Settings"

mondo989:
Cool I'll add it to the top

tomholub:
Also, the black bubble, when did you plan to show it? Does it show somehow automatically? It may not be necessary here, but I may like to use similar bubbles in other places.

mondo989:
The black bubble will show on hover

mondo989:
screen shot 2017-06-05 at 11 44 04 pm


What do you think about accordions?

tomholub:
I we were to use them, I would probably remove the concept of Basic/Additional settings, and just have everything on one page instead. But that's a bigger project.

The bubble on hover makes sense, I would just remove the icon in it, just text will work.

mondo989:
For advanced, how do you feel about having it bottom right?
I can adjust the spacing of it a bit to give it some breathing room.

screen shot 2017-06-05 at 11 54 50 pm


image

tomholub:
Seems like keeping it on the top is the most universal :)

--
Tom James Holub <http://holub.me/>

tomholub:
I would probably put it on the top, and by the end of the year we may do a
bigger rehaul of the settings.

--
Tom James Holub <http://holub.me/>

mondo989:
The rehaul could be nice as the top is a bit crowded and I'll make the updates.

tomholub:
Yup, save that for later :)

Once the contact page button is ready to merge, let me know.

mondo989:
Just about there...

mondo989:
Ok here's the 

screen shot 2017-06-06 at 12 55 04 am
recent...

PR #1103, opened 2018-05-26
Update Secure Compose button for Gmail's Redesign

mondo989:
Issue #1076
- Updates Secure Compose button for Gmail's Redesign

may-26-2018 17-19-23

may-26-2018 18-41-11
-alt


mondo989:
@tomholub let me know if we need anything updated here..

PR #1152, opened 2018-06-29
Update settings pages responsiveness

mondo989:
- Initial commit of updating search page's responsiveness. 
I imagine there's a few more bugs needing fine tuning, but it seems to be decent start overall.
For now, I'm including the updates in a separate stylesheet, in the future it can be included in the `settings.css`

jun-29-2018 14-41-13


tomholub:
That looks good!

Issue #1001, opened 2018-02-15
backspace doesn't re-evaluate recipients - leaving compose in unusable state


tomholub:
1) add recipient that has pgp
2) add second one that doesn't (gray email)
3) password box pops up
4) backspace-remove last recipient
5) password box does not go away

tomholub:
The work here should be in `js/common/composer.ts`. It's a messy file to be working at if you are new to the codebase, but with a `console.log` here and a `debugger;` there, you should be able to find your way around it enough to fix this issue relatively elegantly.

evanhellmuth:
Claiming this issue.

PR #1526, opened 2019-02-08
check if password box etc. should be removed recipient deletion via backspace

evanhellmuth:
Solves https://github.com/FlowCrypt/flowcrypt-browser/issues/1001

Considered using `this.removeReceiver(htmlElement)` but it expects the recipient close "x" element to be passed which is a bit clunky, and`this.showHidePwdOrPubkeyContainerAndColorSendBtn();` is already used a few times in the same manner.


evanhellmuth:
@tomholub

tomholub:
LGTM

PR #1536, opened 2019-02-12
 pubkey import form should accept just fingerprint

RobertZ2011:


tomholub:
It's going in the right direction, but I suspect it will have side effects. Side effects are not good. detectBlocks is used for a lot of things that may not expect this, particularly when parsing emails, particularly in some instances, like:
```
Hello
PGP is a good way to encrypt email.
That's all
```
It's just a guess, I have to test it.

RobertZ2011:
Tom,

I see what you're saying. Would it be possible to do the fingerprint detection in `Ui.replaceRenderableBlocks`? Because that function returns nothing if it only finds a single text block, which makes it seem like a better starting point.

tomholub:
Details will be discussed privately.

Issue #1257, opened 2018-09-25
ui to diagnose faulty keys


tomholub:
Eg start with this key:

```
-----BEGIN PGP PUBLIC KEY BLOCK-----
Version: FlowCrypt 5.7.1 Gmail Encryption flowcrypt.com
Comment: Seamlessly send, receive and search encrypted email

xsFNBFmak3MBEACvNGDbXKGGInrAFb/FFSR8TeWSyEuQpf4xNmz1J0E1Rmr0
q7zvEKK98H7LAGNchKoj21fOO9tp8IBquh+jKdu/Kd3/B26hX+6rmiLFe4l6
YGUA+L6+dYS3IDc68WmhIP1m/9zXyUVkRBDEXBJzEqaI9HIqfLeN7fm9jmby
LElFhxhB1wpXV0lBJ8OOdfAQvpTr3StAcC32XZ1J0cPaDClROwW143ouU4Fj
sePevhawcKbwwzJrP8juGxZCTo5JBl2Z1+Ry0bQRe2Fch/wQdJts1lGJr+H7
OSvFtQlGzSjzamE7TWaOHhKoh7gCn6j9ThKfJt/OudSmPuGrBnKV6rLmRIKQ
VUpIPLWhEOFoPZfUdD7imO6HknxLEIngfPcqE7XSvDtoAo8yAZoFDFdv0ifZ
CuryFlzOQlXcO1+S8mY2zvVEjjXmLuZCrfGT+Ixor8Hzbw5n4SMeQsGZv6DK
6Zf4ZR4z7JvvTFxdeCwJY50He0ggE3kz006ZqDQTKIV/rsCBJtpqYOQSOVw2
OkXNoxVxdDy/L1IZzrbKzDReV1dt9pnQBmStpAuSw8ngOEz64iQQjkVkII3O
JOeOhkwsSSV2xKJpWDM5dSzZPSnbFIXw6/pZKo/pg5OtlOUjh9ZhKg2lUG0X
Z0nl2wvdwYUYYXpebQeqKmtOgzmCh1Gd+xUYQQARAQABwsFfBCABCAAJBQJZ
mpOMAh0CAAoJENNJLS+L4OHw/5IP/1gSkYKkESvCB4UPSeODNi1agEYzNeO+
MOL/EbQXaVO60fhhp8dtykKtSIFVADyPqSmuK2/rYXJBgqFbU5PFmghI+R/b
[...]

PR #1538, opened 2019-02-12
 ui to diagnose faulty keys

tomholub:
addresses #1257


tomholub:
@schmorrison please address the test fails. You should be able to get the same result when you run `npm run-script test`

```
> flowcrypt-browser@6.6.2 test_tslint /home/runner/flowcrypt-browser
> tslint --project ./tsconfig.json --config ./conf/tslint.yaml && ./node_modules/.bin/tslint --project ./conf/tsconfig.content_scripts.json --config ./conf/tslint.yaml


ERROR: /home/runner/flowcrypt-browser/extension/chrome/settings/modules/compatability.ts[95, 27]: Unsafe use of expression of type 'any'.
ERROR: /home/runner/flowcrypt-browser/extension/chrome/settings/modules/compatability.ts[118, 67]: Unsafe use of expression of type 'any'.
ERROR: /home/runner/flowcrypt-browser/extension/chrome/settings/modules/compatability.ts[118, 73]: Unsafe use of expression of type 'any'.
ERROR: /home/runner/flowcrypt-browser/extension/chrome/settings/modules/compatability.ts[147, 61]: Unsafe use of expression of type 'any'.
ERROR: /home/runner/flowcrypt-browser/extension/chrome/settings/modules/compatability.ts[151, 67]: Unsafe use of expression of type 'any'.
ERROR: /home/runner/flowcrypt-browser/extension/chrome/settings/modules/compatability.ts[199, 7]: Promises must be handled appropriately
ERROR: /home/runner/flowcrypt-browser/extension/chrome/settings/modules/compatability.ts[201, 45]: Unsafe use of expression of type 'any'.
```

Unsafe use means when you do something like:
```js
let someString: any;
someString = 'hello world';
someString.split(' ');
```
I don't know if this exact snippet would trigger it, but in general, an 'any' could be anything, eg a number, and doing number.split would cause an exception: not safe.

It's safer to use the `unknown` type with typeguards, or you could cast it to not be any if you know what you're doing.

tomholub:
Handled promises means you have to `await` every promise, else you lose rejections that happen in it.

schmorrison:
Getting a testing error. The file in question is in gitignore file, presumably its provided on the CI test system.

```
  Uncaught exception in build/test/test/source/test.js
  Error: ENOENT: no such file or directory, open 'test/test-secrets.json'
  Object.<anonymous> (test/source/util/index.ts:20:41)
  Object.extensions.(anonymous function) [as .js] (node_modules/require-precompiled/index.js:16:3)
```

schmorrison:
I made another change to the typings to handle one of the `unsafe use of any` errors.

tomholub:
closes #1257

tomholub:
Correct, the secrets file is provided by ci. You will also get one once we sign NDA and stuff like that.

PR #1543, opened 2019-02-14
fixes #929 : Use Gmail api to fetch user email aliases

### LINKED ISSUE #929: use Gmail api for aliases
https://developers.google.com/gmail/api/v1/reference/users/settings/sendAs/get

vincentp:
Anything that needs to be done to setup tests so I can write some? I get:
`Error: ENOENT: no such file or directory, open 'test/test-secrets.json'`

Also, we could filter out emails on "verificationStatus" (user might not have validated his alias) or leave that to users to decide. What do you think?

### DIFF
--- a/extension/js/common/api/api.ts
+++ b/extension/js/common/api/api.ts
@@ -194,8 +194,8 @@ export namespace R { // responses
   export type GmailDraftUpdate = {};
   export type GmailDraftGet = { id: string, message: GmailMsg };
   export type GmailDraftSend = {};
-  export type GmailAliases = { sendAs: GmailAlias[] };
-  export type GmailAlias = { sendAsEmail: string, displayName: string, replyToAddress: string, signature: string, isDefault: boolean, treatAsAlias: boolean, verificationStatus: string };
+  export type GmailAliases = { sendAs: GmailAliases$sendAs[] };
+  type GmailAliases$sendAs = { sendAsEmail: string, displayName: string, replyToAddress: string, signature: string, isDefault: boolean, treatAsAlias: boolean, verificationStatus: string };
 
   export type OpenId = { // 'name' is the full name, picture is url
     at_hash: string; exp: number; iat: number; sub: string; aud: string; azp: string; iss: "https://accounts.google.com";
--- a/extension/js/common/api/api.ts
+++ b/extension/js/common/api/api.ts
@@ -194,6 +194,8 @@ export namespace R { // responses
   export type GmailDraftUpdate = {};
   export type GmailDraftGet = { id: string, message: GmailMsg };
   export type GmailDraftSend = {};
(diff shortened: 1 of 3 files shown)


extension/js/common/settings.ts
```diff
@@ -21,18 +21,13 @@ export class Settings {
...
-        return results.filter(email => !Value.is(email).in(Settings.ignoreEmailAliases));
-      }
-      for (const from of headers.from) {
-        results.push(Str.parseEmail(from).email);
-        query += ' -from:"' + Str.parseEmail(from).email + '"';
-      }
-    }
+    const response = await Google.gmail.fetchAcctAliases(acctEmail);
+
+    const aliases = response.sendAs.map(alias => {
```
tomholub: can shorten to to `map(alias => alias.sendAsEmail)`

Also can remove the empty lines

extension/js/common/api/api.ts
```diff
@@ -194,6 +194,8 @@ export namespace R { // responses
   export type GmailDraftUpdate = {};
   export type GmailDraftGet = { id: string, message: GmailMsg };
   export type GmailDraftSend = {};
+  export type GmailAliases = { sendAs: GmailAlias[] };
+  export type GmailAlias = { sendAsEmail: string, displayName: string, replyToAddress: string, signature: string, isDefault: boolean, treatAsAlias: boolean, verificationStatus: string };
```
tomholub: Looks good but don't export `GmailAlias`, only `GmailAliases`. Actually, I have a convention for this: I recommend to rename `GmailAlias` to `GmailAliases$sendAs`.

extension/js/common/settings.ts
```diff
@@ -21,18 +21,13 @@ export class Settings {
...
+      return alias.sendAsEmail;
+    });
+
+    return aliases.filter(email => !Value.is(email).in(Settings.ignoreEmailAliases));
```
tomholub: Actually you could probably remove `ignoreEmailAliases` completely (not needed anymore if we get aliases from authoritative source) and instead start filtering on whether aliases were verified as you mentioned.

tomholub:
As for the tests, you will have to wait for our CI tests to pick up this branch (I'll allow it manually). You won't be able to run them locally due to the missing file.

tomholub:
@vincentp Please confirm that you release this (and any following) contributions under MIT so that I can merge them. Thanks!

vincentp:
MIT confirmed, thanks for the feedback :)

tomholub:
This was not bad. If you want to take a look at other issues marked as `actionable`, we should probably start paying you. How long did this take?

tomholub:
Minor feedback: `verificationStatus` type should ideally be union of literals, eg `'accepted' | 'new' | 'whatever'` rather than just a `string`.

PR #1549, opened 2019-02-15
Issue 998 add aliases on reply form

### LINKED ISSUE #998: choose reply-from address on a reply
ref https://mail.google.com/mail/u/0/#inbox/1618228ad8d86433

vincentp:
Fixes #998 show email aliases on the reply form.
Added From/To to be consistent with Gmail.
Also switch position, From is coming first (as in Gmail).

One thing we should do is to pre-select the email to was sent to in the sender dropdown, how can I access this variable?

screenshot 2019-02-15 15 55 54
screenshot 2019-02-15 16 22 50


tomholub:
In general, consistency with Gmail is a good thing, but we don't need to do everything the same way they do it. In particular, I would not like to add `From` and `To` because they take up space and make design more difficult. Eg this looks broken, there is no order to how all the elements are laid out:

image

One way to simplify this would be to include `From: ` in each option of the dropdown.

Also vertical spacing is off, the two rows are now too close to one another, and too far from outer edges. Please don't re-arrange elements in the layout unless you take great care to make sure everything is well laid out and aligned when you're done, we cannot ship a broken-looking layout to our users.

1) revert positioning to how it was before
2) Remove the From: and To:
3) Make sure that the from field is hidden by default in reply, and can be displayed by clicking a down-chevron. Maybe this wasn't that clear from the issue, now that I'm re-reading it. The down chevron should have a title eg `Choose sending address`, and I would recommend to position to float on the right side, roughly where the cog icon would have been. You can use the `chevron-left.svg` icon, css-rotated to point down. Hide the chevron when you render the select box (not reversible). Make sure to call `this.resizeReplyBox();` after user clicks the chevron and you have rendered the from-selector. Use `this.getSender()` as a default option.
4)

vincentp:
It felt like "Compose" and "Secure Compose" should get the same layout within Gmail but I get your points! I've updated the logic/layout to follow your suggestions. There is no layout change, only the down chevron added.
I also had to switch the logic within getSender, as if I change the sender address, it would take it from this.v.from otherwise.
Let me know if I need to adjust anything!
screenshot 2019-02-19 14 59 36

tomholub:
It looks much better! I will do a proper review tomorrow.

extension/js/common/composer.ts
```diff
@@ -1528,29 +1528,50 @@ export class Composer {
...
+    const addresses = this.app.storageGetAddresses();
+    if (addresses.length > 1) {
+      const showAliasChevronHtml = '<img id="show_sender_aliases_options" src="/img/svgs/chevron-left.svg" title="Choose sending address">';
+      const inputAddrContainer = $('#input_addresses_container');
+      Xss.sanitizeAppend(inputAddrContainer, showAliasChevronHtml);
+      inputAddrContainer.find('#show_sender_aliases_options').click(Ui.event.handle(() => this.renderSenderAliasesOptions('reply'), this.getErrHandlers(`show sending address options`)));
+    }
+  }
+
+  private renderSenderAliasesOptions(emailContext: 'new' | 'reply') {
```
tomholub: Instead of passing `emailContext`, you can check `this.v.isReplyBox`. When true, it's reply, else it's new message.

tomholub:
All good now, thanks!

PR #1550, opened 2019-02-16
Issue 308 open sans fonts local

schmorrison:
This merge makes Open Sans fonts locally hosted so no requests are made to `https://fonts.gstatic.com`.

Because the font URLs listed in the original CSS file do some magic to provide different font files depending on the browser, using `https://google-webfonts-helper.herokuapp.com` I copied the generated CSS that provides the same Open Sans font family sets, ie regular, light, and semibold, and included the font files provided.

Because Chrome is the only browser supported, it may be superfluous to use the additional font file (woff). Depending on your preference I will happily make the change.


tomholub:
Generally speaking, if a dependency for the Chrome extension was only needed for old versions of browsers, we could drop it. I guess the safari here could be an exception.

PR #1551, opened 2019-02-16
fixes #1547 and addresses changes cited in comments on in #1257

### LINKED ISSUE #1547: improve compatibility page behavior
See comments:
https://github.com/FlowCrypt/flowcrypt-browser/issues/1257#issuecomment-462912446
https://github.com/FlowCrypt/flowcrypt-browser/issues/1257#issuecomment-462913216

schmorrison:
executes all tests in the following cases:
1) either private or public key
2) private key is not decrypted
3) key has no valid encryption packet
4) the actual key is scambled but the integrity check is restored as suggested by the error

instead of using the public key, I used a try/catch in the encrypt-decrypt test. Im still hazy on what information is common to both pub/prv. 

Also, and this I will need you to evaluate, I provided a private key, then I moved a few lines of the key block around, I corrected the integrity check, and provided my password. But the sign/verify test would still complete successfully. I don't know the anatomy of these keys, so I am probably not the best person to come up with test-cases at this time, but I feel like it is something i should mention incase it is relevant.

### DIFF
--- a/extension/chrome/settings/modules/compatability.ts
+++ b/extension/chrome/settings/modules/compatability.ts
@@ -25,22 +25,26 @@ Catch.try(async () => {
 
   const testEncryptDecrypt = async (key: OpenPGP.key.Key): Promise<string[]> => {
     const output: string[] = [];
-    const encryptedMsg = await openpgp.encrypt({
-      message: openpgp.message.fromText(encryptionText),
-      publicKeys: key.toPublic(),
-      armor: true,
-      passwords: [encryptionPassphrase]
-    });
-    output.push(`Encryption with key was successful`);
-    if (key.isPrivate() && key.isDecrypted()) {
-      const decryptedMsg = await openpgp.decrypt({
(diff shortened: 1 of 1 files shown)


tomholub:
> Also, and this I will need you to evaluate, I provided a private key, then I moved a few lines of the key block around, I corrected the integrity check, and provided my password. But the sign/verify test would still complete successfully. I don't know the anatomy of these keys, so I am probably not the best person to come up with test-cases at this time, but I feel like it is something i should mention incase it is relevant.

Is this consistent? If you could provide a test private key in its original form, and one that has been editted this way, for me to play with.. I'll take a look (you can create another issue).

tomholub:
On that note, when working with user interfaces, if you see large boxes of similar size and similar placement, they should typically be the same size (in some dimension) and aligned, else the interface looks broken as above. 

I did this with both the `pubkey -> test result` pair as well as the `passphrase -> button` pair, but it was really the pubkey - test result that was visually way off, placement wise.

PR #1553, opened 2019-02-17
list the blog entries returned by /feed endpoint.

schmorrison:
Let me know if you'd like to see it implemented in some other way.

### DIFF
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -150,6 +150,8 @@ Catch.try(async () => {
         $('.hide_if_setup_not_done').css('display', 'none');
       }
     }
+
+    await renderBlogPostList();
   };
 
   const checkFcAcctAndSubscriptionAndContactPage = async () => {
@@ -307,6 +309,38 @@ Catch.try(async () => {
       await Store.passphraseSave('session', acctEmail!, $(target).attr('longid')!, undefined);
       reload(true);
     }));
+  };
+
+  const renderBlogPostList = async () => {
+    const errorRetrievingPosts = () => {
+      Xss.sanitizeAppend('.blog_post_list', `<div class="line>Error retrieving blog posts...try again later.</div>\n`);
+    };
+    try {
+      const res = await Api.ajax({
+        url: 'https://flowcrypt.com/feed',
+        method: 'GET',
+        dataType: 'json',
+        async: true
+      }, Catch.stackTrace());
+
+      interface BlogPostData {
+        url: string;
+        title: string;
+        date: string;
+      }
+      if (res instanceof Array) {
+        (res as BlogPostData[]).map(post => {
+          $('.blog_post_list').append(
+            `<div class="line"><a href="https://flowcrypt.com${Xss.escape(post.url)}" target="_blank">${Xss.escape(post.title.trim())}</a> (${Xss.escape(post.date.trim())})</div> \n`);
+        });
+      } else {
+        errorRetrievingPosts();
+      }
+    } catch (err) {
+      console.error(err);
+      errorRetrievingPosts();
+      return;
+    }
   };
 
   // tslint:disable-next-line:no-unsafe-any
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -151,12 +151,12 @@ Catch.try(async () => {
       }
     }
 
-    Api.retreiveBlogPosts().then((res: { url: string, date: string, title: string }[]) => {
-      for (const post of res) {
+    Api.retreiveBlogPosts().then(blogs => {
+      for (const post of blogs) {
         $('.blog_post_list').append(
           `<div class="line"><a href="https://flowcrypt.com${Xss.escape(post.url)}" target="_blank">${Xss.escape(post.title.trim())}</a> (${Xss.escape(post.date.trim())})</div> \n`);
       }
-    }).catch((e: Error) => Api.err.isSignificant(e) ? Catch.handleErr(e) : undefined);
+    }).catch(e => Api.err.isSignificant(e) ? Catch.handleErr(e) : undefined);
   };
 
   const checkFcAcctAndSubscriptionAndContactPage = async () => {
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -151,7 +151,12 @@ Catch.try(async () => {
       }
     }
 
-    await renderBlogPostList();
+    Api.retreiveBlogPosts().then((res: { url: string, date: string, title: string }[]) => {
+      for (const post of res) {
+        $('.blog_post_list').append(
+          `<div class="line"><a href="https://flowcrypt.com${Xss.escape(post.url)}" target="_blank">${Xss.escape(post.title.trim())}</a> (${Xss.escape(post.date.trim())})</div> \n`);
+      }
+    }).catch((e: Error) => Api.err.isSignificant(e) ? Catch.handleErr(e) : undefined);
   };
 
   const checkFcAcctAndSubscriptionAndContactPage = async () => {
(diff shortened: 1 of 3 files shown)


schmorrison:
Cannot figure out why the test is failing, doesn't seem to be related to any changes I made here, but I am not sure if where I am executing `renderBlogPostList` is the problem.

I think it'd be useful to get the test file so I can address these issues before I issue a PR.

tomholub:
To get the test file, I'll need the NDA signed - did you receive any email about it?

extension/chrome/settings/index.ts
```diff
@@ -309,6 +311,38 @@ Catch.try(async () => {
     }));
   };
 
+  const renderBlogPostList = async () => {
+    const errorRetrievingPosts = () => {
+      Xss.sanitizeAppend('.blog_post_list', `<div class="line>Error retrieving blog posts...try again later.</div>\n`);
+    };
+    try {
+      const res = await Api.ajax({
```
tomholub: this should be as a method somewhere in the `Api` class. Also, we typically put short object literals on one line eg `{url: 'https://flowcrypt.com/feed', method: 'GET', dataType: 'json', async: true}`. You can probably omit the method and async (I think), so that would be `{url: 'https://flowcrypt.com/feed',  dataType: 'json'}`.

extension/chrome/settings/index.ts
```diff
@@ -309,6 +311,38 @@ Catch.try(async () => {
...
+        url: 'https://flowcrypt.com/feed',
+        method: 'GET',
+        dataType: 'json',
+        async: true
+      }, Catch.stackTrace());
+
+      interface BlogPostData {
```
tomholub: Again this would be in the `api.ts` file like the rest of API definitions.

extension/chrome/settings/index.ts
```diff
@@ -309,6 +311,38 @@ Catch.try(async () => {
...
+        url: string;
+        title: string;
+        date: string;
+      }
+      if (res instanceof Array) {
```
tomholub: You don't need to test this, if there is error, it will be thrown and reported, so we can deal with it.

extension/chrome/settings/index.ts
```diff
@@ -309,6 +311,38 @@ Catch.try(async () => {
     }));
   };
 
+  const renderBlogPostList = async () => {
+    const errorRetrievingPosts = () => {
+      Xss.sanitizeAppend('.blog_post_list', `<div class="line>Error retrieving blog posts...try again later.</div>\n`);
+    };
+    try {
```
tomholub: You won't need the try/catch here either because we will be handling it one step up with the `.catch()`

extension/chrome/settings/index.ts
```diff
@@ -309,6 +311,38 @@ Catch.try(async () => {
...
+        async: true
+      }, Catch.stackTrace());
+
+      interface BlogPostData {
+        url: string;
+        title: string;
+        date: string;
+      }
+      if (res instanceof Array) {
+        (res as BlogPostData[]).map(post => {
```
tomholub: Don't cast, the return method in the Api class would already have a return type, or you would cast it there.

extension/chrome/settings/index.ts
```diff
@@ -309,6 +311,38 @@ Catch.try(async () => {
...
+          $('.blog_post_list').append(
+            `<div class="line"><a href="https://flowcrypt.com${Xss.escape(post.url)}" target="_blank">${Xss.escape(post.title.trim())}</a> (${Xss.escape(post.date.trim())})</div> \n`);
```
tomholub: Xss escaping looks correct :)

extension/chrome/settings/index.ts
```diff
@@ -309,6 +311,38 @@ Catch.try(async () => {
...
+        });
+      } else {
```
tomholub: All of the lines from here down will end up being redundant. Actually, because it's just blog post, you don't even need to render errors here - if the network is down, you can just skip displaying posts - they are not that important to always show. But if you wanted to handle them, maybe a better place would be in the new `.catch()` block above, because then you only have to write it one time, and don't need to have a method for it.

extension/chrome/settings/index.ts
tomholub: Also, please don't use map for looping when you can do
```js
for (const post of res) {
  // ...
}
```

tomholub:
I'm running the tests again to see if this was just a flaky test or a real issue.

schmorrison:
Hey tom, I had signed the NDA a couple days ago. I assumed you got an email of the signed contract like I had.

I'll work on these changes today.

tomholub:
My bad, let me take a look

tomholub:
I was brainless enough to sign up for the doc sign service with an email that I didn't have a DNS configured for yet. Could you forward me the signed NDA please? Sorry about this.

extension/chrome/settings/index.ts
```diff
@@ -151,7 +151,12 @@ Catch.try(async () => {
       }
     }
 
-    await renderBlogPostList();
+    Api.retreiveBlogPosts().then((res: { url: string, date: string, title: string }[]) => {
```
tomholub: Much better! Now, don't clarify the type of `res` here when receiving it. That should be clarified in the api file with `retreiveBlogPosts = async (): Promise<R.FcBlogPosts> => {`

Then here you only do `Api.retreiveBlogPosts().then(res => {`

This is so that if we later want to change the structure of the api response, you don't have to go hunting for all usages of this function where you would have to change the blog post structure. There will only be one authoritative place defining the response type, in the api file.

tomholub: Also I recommend to use more descriptive `Api.retreiveBlogPosts().then(blogs => {`

extension/chrome/settings/index.ts
```diff
@@ -151,7 +151,12 @@ Catch.try(async () => {
...
+      for (const post of res) {
+        $('.blog_post_list').append(
+          `<div class="line"><a href="https://flowcrypt.com${Xss.escape(post.url)}" target="_blank">${Xss.escape(post.title.trim())}</a> (${Xss.escape(post.date.trim())})</div> \n`);
+      }
+    }).catch((e: Error) => Api.err.isSignificant(e) ? Catch.handleErr(e) : undefined);
```
tomholub: Don't clarify `e` as `Error` because in JS, anything could be thrown. Narrowing it down to `Error` gives you a false sense of security. I could be a string or a null, for all you know. Do: `.catch(e => Api.err.isSignificant(e) ? Catch.handleErr(e) : undefined);`

extension/js/common/api/api.ts
```diff
@@ -806,4 +806,8 @@ export class Api {
     apiFcCall: (path: string, vals: Dict<any>, fmt: ReqFmt = 'JSON') => Api.internal.apiCall(Api.fc.url('api'), path, vals, fmt, undefined, { 'api-version': '3' }),
   };
 
+  public static retreiveBlogPosts = async () => {
```
tomholub: Looks good, just add the types as discussed above, follow the style of other API methods.

extension/js/common/api/api.ts
```diff
@@ -806,8 +807,8 @@ export class Api {
     apiFcCall: (path: string, vals: Dict<any>, fmt: ReqFmt = 'JSON') => Api.internal.apiCall(Api.fc.url('api'), path, vals, fmt, undefined, { 'api-version': '3' }),
   };
 
-  public static retreiveBlogPosts = async () => {
-    return await Api.ajax({ url: 'https://flowcrypt.com/feed', dataType: 'json' }, Catch.stackTrace());
+  public static retreiveBlogPosts = async (): Promise<R.FcBlogPost[]> => {
+    return Api.ajax({ url: 'https://flowcrypt.com/feed', dataType: 'json' }, Catch.stackTrace());
```
tomholub: Looks good. Here, having or not having await makes no difference, it will behave the same, because it's an async function anyway - it will always return a promise regardless.

Issue #1414, opened 2018-12-14
option to see original in flowcrypt inbox


tomholub:
This one could be done very easily and smoothly

tomholub:
Came up when I wanted to copy signed message including armor

tomholub:
deprecating

tomholub:
not deprecating, this is about the inbox in flowcrypt settings

tomholub:
Basically just sort of replicate what that "show original" button does in Gmail, but show it directly in the convo, not in new window, and only affect currently opened convo, and don't remember the choice

tomholub:
In Settings -> Inbox

schmorrison:
Hey Tom, I started working on this but I'm going to need a primer or explanation on what you want to see when you say

> Basically just sort of replicate what that "show original" button does in Gmail, but show it directly in the convo, not in new window, and only affect currently opened convo, and don't remember the choice

I found the stuff in `browser.ts` and the `gmail_element_replacer.ts`, and I know that in `js/content_scripts/webmail/*.ts` are the files in the webmail bundle which is loaded by the extension in a gmail inbox.

So I can see that clicking the `See Original` button just executes a click on the `In new Window` button in gmail, like so:
```
this.addfcConvoIcon(convoUpperIcons, this.factory.btnWithoutFc(), '.show_original_conversation', () => {
    convoUpperIcons.find('.gZ').click();
});
```

But if you could explain how you might like the layout to be organized, or some styling pointers for this. I started towards just replacing the decrpyted PGP UI blocks with the original PGP message block wrapped in a `pre` tag to handle the newlines, (maybe you have a helper for that already). Then providing a button that will replace the 'original messages' with the decrypted content. For my development of this I created some messages with the public key attached, the way it works on gmail is a PGP message block with a PGP Public Key in a file attachment. The change I made simply prints the PGP public key block directly after the PGP message block, but I can change it to do attachments alternatively if youd prefer

Best regards,

Sam Morrison

tomholub:
Basically you should re-load and re-render the whole conversation, but don't do any decrypting any decrypting this time. (this is easier than trying to replace already rendered blocks)

So I would imagine in `renderMsg` you would find a way to skip `Ui.renderableMsgBlock` and instead treat and render each block as `text`. That would be a good start, from there, you would see what else needs work, if any.

tomholub:
[...]

PR #1565, opened 2019-02-21
Fixes 1414 - add a 'show original' button to fc inbox

schmorrison:
- Adds a 'Show Original' button to the FlowCrypt inbox
- Only shows the button if there are PGP messages in the thread
- When button is click:
  - page is reloaded with a 'showOriginal' url param set to 'true'
  - PGP blocks are not rendered but instead shown as text wrapped in a pre tag
  - The button is changed to show 'See Decrypted' now
  - When 'See Decrypted' is clicked page is reloaded again but with the 'showOriginal' url param omitted

### DIFF
--- a/extension/chrome/settings/inbox/inbox.ts
+++ b/extension/chrome/settings/inbox/inbox.ts
@@ -21,7 +21,7 @@ Catch.try(async () => {
   const acctEmail = Env.urlParamRequire.string(uncheckedUrlParams, 'acctEmail');
   const labelId = uncheckedUrlParams.labelId ? String(uncheckedUrlParams.labelId) : 'INBOX';
   const threadId = Env.urlParamRequire.optionalString(uncheckedUrlParams, 'threadId');
-  const showOriginal = Env.urlParamRequire.optionalString(uncheckedUrlParams, 'showOriginal') === 'true' ? true : false;
+  const showOriginal = uncheckedUrlParams.showOriginal === true;
   let threadHasPGPMessage = false;
 
   let emailProvider;
@@ -350,15 +350,9 @@ Catch.try(async () => {
 
       if (threadHasPGPMessage) {
         $(".action_see_original_message").css('display', 'inline-block');
+        $(".action_see_original_message").click(Ui.event.handle(() => loadUrl({ acctEmail, threadId, showOriginal: !showOriginal })));
         if (showOriginal) {
           $(".action_see_original_message").text('See Decrypted');
-          $(".action_see_original_message").click(Ui.event.prevent('double', async (self) => {
-            loadUrl({ "acctEmail": acctEmail, "threadId": threadId });
-          }));
-        } else {
-          $(".action_see_original_message").click(Ui.event.prevent('double', async (self) => {
-            loadUrl({ "acctEmail": acctEmail, "threadId": threadId, "showOriginal": "true" });
-          }));
         }
       }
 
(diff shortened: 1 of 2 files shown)


schmorrison:
Forgot to run local test. Fixing and Ill push again

tomholub:
You don't always have to run tests locally, that's what these automated tests are for. Locally they sometimes take ages anyway. Since you're working on your own branch, it's quite normal to have to fix up tests in another commit.

tomholub:
However this let/const should show up automatically in the IDE

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -17,10 +17,12 @@ import { Buf } from '../../../js/common/core/buf.js';
 
 Catch.try(async () => {
 
-  const uncheckedUrlParams = Env.urlParams(['acctEmail', 'labelId', 'threadId']);
+  const uncheckedUrlParams = Env.urlParams(['acctEmail', 'labelId', 'threadId', 'showOriginal']);
   const acctEmail = Env.urlParamRequire.string(uncheckedUrlParams, 'acctEmail');
   const labelId = uncheckedUrlParams.labelId ? String(uncheckedUrlParams.labelId) : 'INBOX';
   const threadId = Env.urlParamRequire.optionalString(uncheckedUrlParams, 'threadId');
+  let showOriginal = Env.urlParamRequire.optionalString(uncheckedUrlParams, 'showOriginal') === 'true' ? true : false;
```
tomholub: You can use booleans instead of strings. Here you would then do
```js
const showOriginal = uncheckedUrlParams.showOriginal === true;
```

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -345,6 +347,22 @@ Catch.try(async () => {
       for (const m of thread.messages) {
         await renderMsg(m);
       }
+
+      if (threadHasPGPMessage) {
+        $(".action_see_original_message").css('display', 'inline-block');
+        if (showOriginal) {
+          $(".action_see_original_message").text('See Decrypted');
+          $(".action_see_original_message").click(Ui.event.prevent('double', async (self) => {
+            loadUrl({ "acctEmail": acctEmail, "threadId": threadId });
```
tomholub: Should be shortened to (that's what the lint was probably complaining about)
```js
loadUrl({ acctEmail, threadId });
```
Since it's so short, might as well:
```js
$(".action_see_original_message").click(Ui.event.prevent('double', () => loadUrl({ acctEmail, threadId })));
```
The `double` is trying to prevent double clicks, but here it wouln't do any harm, so you can do:
```js
$(".action_see_original_message").click(Ui.event.handle(() => loadUrl({ acctEmail, threadId })));
```
Just to make sure errors are handled / reported.

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -345,6 +347,22 @@ Catch.try(async () => {
...
+          }));
+        } else {
+          $(".action_see_original_message").click(Ui.event.prevent('double', async (self) => {
+            loadUrl({ "acctEmail": acctEmail, "threadId": threadId, "showOriginal": "true" });
```
tomholub: same here

tomholub: Actually you could probably simplify the two further on one line:
```js
$(".action_see_original_message").click(Ui.event.handle(() => loadUrl({ acctEmail, threadId, showOriginal: !showOriginal })));
```

tomholub:
#1414

tomholub:
Please also mention the issue number in the description of the PR, or somewhere in comments, eg `#1414` as I did above, so that the PR and the issue get linked together.

Issue #1564, opened 2019-02-20
standard message -> reply -> click "secure reply" -> delete message: broken


tomholub:
Leaves an empty broken reply that does not let you do anything until you navigate away from the conversation, then come back.

![image](https://user-images.githubusercontent.com/6306961/53123430-f4123300-3550-11e9-9fb1-d8efff8c1b3b.png)

Instead of hiding the encrypted reply box, we should delete the draft then reload the secure reply frame.

tomholub:
Something similar happens with `click "secure reply" and discard it, and click the "green" reply on the current email`.

There will be no reply on the bottom, and none will show when you click the green "reply" flowcrypt icon. 

Reloading the frame instead of removing it should resolve this too.

PR #1569, opened 2019-02-22
Reload reply iframe instead of removing it

evanhellmuth:
Fixes https://github.com/FlowCrypt/flowcrypt-browser/issues/1564

Note that this UX differs from that of deleting a normal gmail reply:
Normal reply --> delete draft --> editor closes and user needs to click reply again
Secure reply --> delete draft --> text disappears, any extra recipients disappear, but editor stays open.

This was the ask in https://github.com/FlowCrypt/flowcrypt-browser/issues/1564, and I think it's a fine UX, just wanted to make the difference explicit.

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -299,7 +299,6 @@ export class Composer {
     }, this.getErrHandlers(`change footer`)));
     $('.delete_draft').click(Ui.event.handle(async () => {
       await this.draftDelete();
-
       // Reload iframe so we don't leave users without a reply UI.
       window.location.reload();
     }, this.getErrHandlers('delete draft')));
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -299,7 +299,9 @@ export class Composer {
     }, this.getErrHandlers(`change footer`)));
     $('.delete_draft').click(Ui.event.handle(async () => {
       await this.draftDelete();
-      this.app.closeMsg();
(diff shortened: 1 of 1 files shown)


tomholub:
Thanks for pointing out the difference! It's not ideal but it gets the job done without being horribly annoying to the user.

tomholub:
Remember to use the `issue-1564-reply-securereply-delete-flow-fix` branch name format to have your changes tested by automated tests.

evanhellmuth:
@tomholub after changes UI now displays the following screenshot after deleting a secure reply. Matches Gmail UI perfectly (really clear directions from you, thanks for that):
screen shot 2019-02-24 at 6 51 03 pm

extension/js/common/composer.ts
```diff
@@ -299,9 +299,9 @@ export class Composer {
     }, this.getErrHandlers(`change footer`)));
     $('.delete_draft').click(Ui.event.handle(async () => {
       await this.draftDelete();
-
       // Reload iframe so we don't leave users without a reply UI.
-      window.location.reload();
+      this.v.skipClickPrompt = false;
+      window.location.href = Env.urlCreate(window.location.protocol + '//' + window.location.hostname + window.location.pathname, this.v);
```
tomholub: You could use `window.location.href.split('?')[0]` which is shorter but your is cleaner so it's just a matter of taste. Looks good.

Issue #1568, opened 2019-02-22
clicking account icon in flowcrypt inbox to let me choose acct


tomholub:
Currently does nothing

![image](https://user-images.githubusercontent.com/6306961/53245876-f5557400-36a6-11e9-93f7-a8ff242f9a96.png)

Should also use cursor: pointer

Reuse functionality from `settings/index.ts`

PR #1580, opened 2019-02-25
Port functionality from settings.ts to inbox.ts, updating markup where necessary

evanhellmuth:
Solves https://github.com/FlowCrypt/flowcrypt-browser/issues/1568

The html tag splitting stylistic changes such as https://github.com/FlowCrypt/flowcrypt-browser/compare/issue-1568-clickable-fc-inbox-acct-icon?expand=1#diff-68138ce182225d2609ef1661bf5b5535R22 were being made automatically on save - I assume this is eslint's doing? If not, it may be something particular to my editor (I started using VS Code for FC work but I'm a bit of a noob at the moment). If it's something I should look into, let me know and I can find and eliminate the root cause.


tomholub:
I don't like the html splitting either, but it's something my IDE does. I think it's the in-built auto-format feature. I would still like it auto-formatted, but this particular way it likes to split html tags is weird, so if you want to investigate which setting is causing that and nuke it in the project settings, please do.

(I don't think it's tslint, tslint doesn't do formatting, only linting, and only on TS/JS files)

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -454,6 +460,17 @@ Catch.try(async () => {
     }));
   };
 
+  const menuAcctHtml = (email: string, picture = '/img/svgs/profile-icon.svg') => {
```
tomholub: Instead of copy/pasting, please put it into a common place (with a better name). I would put it into `js/common/settings.ts` as a class method like the rest.

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -464,4 +481,16 @@ Catch.try(async () => {
       await renderInbox(labelId);
     }
   }
+
+  const acctEmails = await Store.acctEmailsGet();
+  const acctStorages = await Store.getAccounts(acctEmails, ['picture']);
+  for (const email of acctEmails) {
```
tomholub: Same thing with this functionality. Instead of copy-pasting it all, better move it to a common place, package it into a function and re-use. You could put most of the code plus the `menuAcctHtml` into a single appropriatelly named function in the `Settings` class, then re-use.

extension/js/common/browser.ts
```diff
@@ -214,6 +214,14 @@ export class Env {
     return ['gmail', 'inbox']; // async because storage may be involved in the future
   }
 
+  public static getBaseUrl = () => {
```
evanhellmuth: Pulled URL-getting logic into Env for convenience and code cleanliness.

extension/js/common/settings.ts
```diff
@@ -372,4 +372,31 @@ export class Settings {
...
+      Xss.sanitizePrepend('#alt-accounts', menuAcctHtml(email, acctStorages[email].picture, page === 'inbox.htm'));
+    }
+    $('#alt-accounts img.profile-img').on('error', Ui.event.handle(self => {
+      $(self).off().attr('src', '/img/svgs/profile-icon.svg');
+    }));
+    $('.action_select_account').click(Ui.event.handle(target => {
+      const acctEmail = $(target).find('.contains_email').text();
+      const acctStorage = acctStorages[acctEmail];
+      window.location.href = acctStorage.setup_done
+        ? Env.urlCreate(page, { acctEmail })
```
evanhellmuth: Let me know your style policy on the ternary operator.

evanhellmuth:
@tomholub extracted logic to `Settings.ts` and handled the not-yet-set-up-account case. Take a look!

extension/js/common/settings.ts
tomholub: Currently, if it fits on one line, do `cond ? a : b` on a single line. If it doesn't, do a normal multi-line if/else, including curly brackets.

tomholub: You could shorten it by doing if/else inside:
```js
Env.urlCreate(acctStorage.setup_done ? xx : yy, { acctEmail })
```

extension/js/common/browser.ts
tomholub: absolutely!

tomholub:
I'll go ahead and merge it. Later I may add a tslint rule to disallow multiline ternary operators, at which point I may reformat the line above.

PR #1582, opened 2019-02-28
Issue #428 message password security

### LINKED ISSUE #428: message password security
```diff
- check that the password is not included in the hint..
+ check that the password is not included in the intro and subject
add a strength bar, but don't enforce
```

schmorrison:
Fixes #428 

I have got to say, unravelling what is going on in the composer class is not a simple task, it could sure do with some heavy commenting. However I think I have figured out issue/feature:

1. `check that the password is not included in the hint.` -- To check that the password was not included in the hint, I used a `indexOf` check while preparing the password protected email text/html. There may be something more suitable, but it works fine in the very basic circumstances I have been working with. It may also be worth checking the subject line and the intro for the password too, so let me know if you want me to do that
2. `add a strength bar, but don't enforce` -- Added a password strength bar with some very rough styling. I added a callback to the `renderPasswordStrength` function so I could manage setting the colour of the bar as suggested in task 4 below. ***I know the styleguide dictates an aversion to callbacks, but I think this circumstance (assuming I understand your request) warrants it***
3. `also check that this password is different than all known users passphrases!` -- To check that the password is not the same as any passphrase, I found a function in Store that returns passphrases for all the keys. This seems like a really easy way to test that the password is not the same as any private key passphrase. Its probably more efficient then trying to decrypt each private key as well, but that could be tested. Let me know if there is a limitation to checking the password in this fashion.
4. `strength bar should show not red at about 1/100th entropy level compared to pass phrases` -- I will need additional explanation to do this. I am not entirely sure what the task is requesting. But with the callback I added above, there should be some room to make this customisation fairly easily.


schmorrison:
I will do these changes this evening. Thanks for the clarification Tom, I didn't realize that I didn't understand certain points until after I was well underway.

1. I will make sure to never add UI elements unless explicitly asked, and if I feel one may be implied I will always confirm it before working on it.
2. So the callback was related to the request on point 4. Yes I did copy/paste, but only because I didn't know exactly how you wanted to modify the return value ie. `strength bar should show not red at about 1/100th entropy level compared to pass phrases`. I assumed this meant that the strength bar would need to be coloured differently than the existing function does it.

I had no problem using the `renderPasswordStrength` function I just thought that based on the request 4 I would be required to change the colours of the elements outside of the existing function. And because `renderPasswordStrength` doesn't return any values, the optional callback was the cleanest way I figured to change the jquery CSS calls arbitrarily.

When assigning an issue I will, at least for a while, take a look over the code to get a sense of the required changes. If it looks too complex or obscure I will either move on, or ask you for some clarity.

schmorrison:
Builds/tests fine locally, might be a "flakey build" like before.

tomholub:
Absolutely, it's better to ask more questions early on!

extension/js/common/composer.ts
```diff
@@ -1596,7 +1597,7 @@ export class Composer {
     const html = [];
     if (intro) {
       text.push(intro + '\n');
-      html.push(intro.replace(/\n/, '<br>') + '<br><br>');
+      html.push(intro.replace(/\n/g, '<br>') + '<br><br>');
```
tomholub: Good find!

extension/js/common/composer.ts
```diff
@@ -1610,6 +1611,22 @@ export class Composer {
     return { 'text/plain': text.join('\n'), 'text/html': html.join('\n') };
   }
 
+  private pwdProtectedEmailIntro = async (pwd: Pwd, subj: string) => {
+    const keysAndPass = await Store.keysGetAllWithPassphrases(this.v.acctEmail);
+    for (const pass of keysAndPass.passphrases) {
```
tomholub: Could be simplified as
```js
const { passphrases } = await Store.keysGetAllWithPassphrases(this.v.acctEmail);
if (passphrases.indexOf(pwd.answer) !== -1) {
  throw ...;
}
```

extension/chrome/elements/compose.htm
```diff
@@ -68,8 +68,22 @@ <h1>New Secure Message</h1>
...
+            </div>
+          </div>
+          <div style="display: inline-block;">
+            <div class="password_feedback"><span class="password_result"></span> (time to crack: <span class="password_time"></span>)
+              <ul></ul>
+            </div>
+          </div>
+        </div>
+
         <div class="question_copy">If they do use encryption, <a href="#" class="add_pubkey" data-test="action-open-add-pubkey-dialog">add
```
tomholub: please remove the css margin-top from question_copy (in the css file), it doesn't play well with the new design (gap too wide)

extension/js/common/composer.ts
```diff
@@ -811,6 +811,7 @@ export class Composer {
       // admin_code stays locally and helps the sender extend life of the message or delete it
       const { short, admin_code } = await Api.fc.messageUpload(encryptedBody['text/plain']!, subs.active ? 'uuid' : undefined);
       const storage = await Store.getAcct(this.v.acctEmail, ['outgoing_language']);
+      await this.pwdProtectedEmailIntro(pwd, subj);
```
tomholub: I would call it `throwIfPwdNotAcceptable`

tomholub:
Trying to re-run tests

tomholub:
Once you pull the latest commit from master, the tests should get fixed. Sorry about the hiccup.

tomholub:
Todo for me:
 - [ ] remove css
 - [ ] reduce difficulty
 - [ ] remove time to crack
 - [ ] change / expand error messages

tomholub:
@michael-volynets next week could you please evaluate, re-base and review this PR?

tomholub:
@michael-volynets please have a look after the current issue

michael-volynets:
+3h
The result: 
image

I'm not really sure about this code:
```TS
    estimateStrength: (zxcvbnResultGuesses: number, decreaseTimeToCrack?: number): PasswordStrengthResult => {
      const timeToCrack = zxcvbnResultGuesses / (Pgp.PASSWORD_GUESSES_PER_SECOND / (decreaseTimeToCrack ? decreaseTimeToCrack * 5000000000 : 1));
```
I don't know exactly how this works so I just devided `timeToCrach` with that value.
If you know a better solution please tell me, thanks!

extension/js/common/composer.ts
```diff
@@ -84,6 +84,8 @@ export class Composer {
   private composeWindowIsMaximized = false;
   private refBodyHeight?: number;
   private urlParams: ComposerUrlParams;
+  private keyImportUI = new KeyImportUi({});
+  private removeValidationElements: (() => void) | undefined;
```
tomholub: should be called `rmPwdValidationElements` - because otherwise it's hard to guess what kind of validation this refers to

Issue #1575, opened 2019-02-24
allow pasting public key straight into recipients box


tomholub:
Sometimes, users are brilliant in their ignorance

> Given a public key - how do i send to the owner an ecrypted message?  Tried to copy the given key into the recipients bar, but it doesnt let me

This sounds like a great thing to allow, and it was smart of the user to even try.

We should have a way to detect when a public key was pasted into recipient field, parse it, figure out what is the primary email, save the pubkey into contacts and add the primary email as a recipient in the compose box.

evanhellmuth:
@tomholub is it safe to assume that the '@' character will never appear in a public key?

tomholub:
Good question! Mostly, but not really. You can have custom comments.
```
-----BEGIN ... -----
Comment: Yo my email is me@domain.com

Ufdtf/4Ty....
----END ... -----
```

It's however safe to assume that if a string has `-----BEGIN` in it, it _may_ be a pubkey.

If there's a way to detect that, and it turns out begin string is in there, then you could run the pasted content through `PgpMsg.stripPublicKeys`, and see if it yields any pubkeys. If it does, discard any non-pubkey content, grab primary email from each found pubkey and add them to recipients (save pubkeys to contacts, too - but only if the no pubkey recorded for that particular email. Do not update / replace pubkeys in storage when the user does this).

You could try to do this in `parseRenderRecipients` (sorry about all the log statements, trying to catch a nasty bug). You could evaluate `inputTo` before it gets lowercased. I don't know if this will give you newlines, I'm afraid not, so you may need to listen to clipboard paste events and process it there (with prevent default) before it ever gets to `parseRenderRecipients`.

It may be a bit tricky to implement. Last resort solution would be to change `input_to` to a multiline, with a single actual line.. that may give you the newlines without too much hassle. You'd need to play with css for a while to make it look and behave the same.

evanhellmuth:
@tomholub started digging in and I can't figure out how to get an email from a pubkey. There are plenty of ways to go from email --> pubkey but not the reverse. 

Is there a FC API call I can make, or should I be searching the current user's contacts? I could forge ahead with the contacts approach, but want to get your input first in case that would be a waste of time. Let me know!

tomholub:
[...]

PR #1584, opened 2019-03-04
Leave empty reply empty on message send

evanhellmuth:
Remove automatic adding of "Re: <subject>" in browser.ts and detect undefined and use empty string instead upon send in composer.ts

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -628,7 +628,7 @@ export class Composer {
   private extractProcessSendMsg = async () => {
     try {
       const recipients = this.getRecipientsFromDom();
-      const subject = this.v.subject || String($('#input_subject').val()); // replies have subject in url params
+      const subject = this.v.subject || ($('#input_subject').val() === undefined ? '' : String($('#input_subject').val())); // replies have subject in url params
       const plaintext = this.extractAsText('input_text');
       await this.throwIfFormNotReady(recipients);
       this.S.now('send_btn_span').text('Loading');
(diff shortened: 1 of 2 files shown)


extension/js/common/composer.ts
```diff
@@ -628,7 +628,7 @@ export class Composer {
   private extractProcessSendMsg = async () => {
     try {
       const recipients = this.getRecipientsFromDom();
-      const subject = this.v.subject || String($('#input_subject').val()); // replies have subject in url params
+      const subject = this.v.subject || ($('#input_subject').val() === undefined ? '' : String($('#input_subject').val())); // replies have subject in url params
```
tomholub: We could test for truthiness here eg `$('#input_subject').val() ? String($('#input_subject').val()) : '')` or shorter `String($('#input_subject').val() || '')`. 

This is a nitpick of course, I'll approve it as is.

tomholub:
#1575

extension/js/common/composer.ts
evanhellmuth: We can't use `String($('#input_subject').val())` as a boolean because `String(undefined)` evaluates to `"undefined"`, which is not falsy!

tomholub: What you say is correct, but that's not what I proposed :)
`String($('#input_subject').val())`
vs
`String($('#input_subject').val() || '')`

evanhellmuth: Ah misread it. That would be a little more elegant!

PR #1593, opened 2019-03-11
Add a button to import binary keyfile

### LINKED ISSUE #1134: allow importing binary keyring to contacts
ref `pubring.pkr`

schmorrison:
Fixes #1134 . Adds a button to the contacts page that allows choosing a file, if the file selected is a valid PGP key file in binary format, the armored key is written to the public key input field.

### DIFF
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -6,7 +6,7 @@ import { VERSION } from '../../../js/common/core/const.js';
 import { Catch } from '../../../js/common/platform/catch.js';
 import { Store } from '../../../js/common/platform/store.js';
 import { Att } from '../../../js/common/core/att.js';
-import { Xss, Ui, XssSafeFactory, Env, Browser } from '../../../js/common/browser.js';
+import { Xss, Ui, XssSafeFactory, AttUI, Env, Browser } from '../../../js/common/browser.js';
 import { BrowserMsg } from '../../../js/common/extension.js';
 import { Pgp } from '../../../js/common/core/pgp.js';
 import { requireOpenpgp } from '../../../js/common/platform/require.js';
@@ -33,6 +33,48 @@ Catch.try(async () => {
 
   BrowserMsg.listen(tabId); // set_css
 
+  const attUI = new AttUI(() => Promise.resolve({ size_mb: 5, size: 5 * 1024 * 1024, count: 1 }));
+  attUI.initAttDialog('fineuploader', 'fineuploader_button');
+  attUI.setAttAddedCb(async (file: Att) => {
+    (file !== null && typeof file !== 'undefined' && typeof file.id !== 'undefined' && file.id !== '') ? attUI.clearAtt(file.id) : attUI.clearAtts();
+    await handleKeyfileSelected(file);
+  });
+
+  const handleKeyfileSelected = async (file: Att) => {
+    try {
+      const rawFile = file.getData();
+      const stringFile = new TextDecoder("utf-8").decode(rawFile);
+      const { blocks } = await Pgp.armor.detectBlocks(stringFile);
+
+      const keys: OpenPGP.key.Key[] = [];
+      if (blocks !== null && blocks.length > 0 && blocks.some(block => block.type === "publicKey")) {
+        for (const block of blocks) {
+          if (block.type === "publicKey") {
+            const result = await openpgp.key.readArmored(block.content);
+            if (result.err !== null && typeof result.err !== 'undefined') {
+              throw result.err;
+            }
+            keys.push(...result.keys);
+          }
+        }
+      } else {
+        const result = await openpgp.key.read(rawFile);
+        if (result.err !== null && typeof result.err !== 'undefined') {
+          throw result.err;
+        }
+        keys.push(...result.keys);
+      }
+      const input = $('#bulk_import .input_pubkey');
+      for (const key of keys) {
+        input.val(`${input.val()}\n\n${key.armor()}`);
+      }
+
+      $('#bulk_import .action_process').trigger('click');
+    } catch (err) {
+      Catch.handleErr(err);
+    }
+  };
+
   const renderContactList = async () => {
     const contacts = await Store.dbContactSearch(undefined, { has_pgp: true });
 
@@ -51,7 +93,7 @@ Catch.try(async () => {
       $('#bulk_import .input_pubkey').val('').css('display', 'inline-block');
       $('#bulk_import .action_process').css('display', 'inline-block');
       $('#bulk_import #processed').text('').css('display', 'none');
-      $('#file_import .action_upload_keyfile').css('display', 'inline-block');
+      $('#file_import #fineuploader_button').css('display', 'inline-block');
       $('#page_back_button').click(Ui.event.handle(() => renderContactList()));
     }));
 
@@ -117,7 +159,7 @@ Catch.try(async () => {
       $('#bulk_import .input_pubkey').val('').css('display', 'inline-block');
       $('#bulk_import .action_process').css('display', 'inline-block');
       $('#bulk_import #processed').text('').css('display', 'none');
-      $('#file_import .action_upload_keyfile').css('display', 'inline-block');
+      $('#file_import #fineuploader_button').css('display', 'inline-block');
       $('#page_back_button').click(Ui.event.handle(() => renderContactList()));
     }));
 
@@ -127,30 +169,7 @@ Catch.try(async () => {
         await Ui.modal.warning('Could not find any new public keys');
       } else {
         $('#bulk_import #processed').html(replacedHtmlSafe).css('display', 'block'); // xss-safe-factory
-        $('#bulk_import .input_pubkey, #bulk_import .action_process, #file_import .action_upload_keyfile').css('display', 'none');
-      }
-    }));
-
-    $('#file_import div.action_open_keyfile').click(Ui.event.prevent('double', async (self) => {
-      $('#file_import  #keyfile_input').trigger('click');
-    }));
-
-    $('#file_import #keyfile_input').change(Ui.event.handle(async (self, event: JQuery.Event<HTMLInputElement, null>) => {
-      if (event !== null && typeof event !== 'undefined') {
-        if (event.target.files !== null && typeof event.target.files !== 'undefined' && event.target.files.length > 0) {
-          try {
-            const fileReader = new FileReader();
-            fileReader.onload = async () => {
-              const binaryKey = new Uint8Array(fileReader.result as ArrayBuffer);
-              const key = await openpgp.key.read(binaryKey);
-              console.log(key);
-              // $('#bulk_import .input_pubkey').val(key.armor());
-            };
-            fileReader.readAsArrayBuffer(event.target.files[0]);
-          } catch (err) {
-            Catch.handleErr(err);
-          }
-        }
+        $('#bulk_import .input_pubkey, #bulk_import .action_process, #file_import #fineuploader_button').css('display', 'none');
       }
     }));
 
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -33,49 +33,106 @@ Catch.try(async () => {
     await handleKeyfileSelected(file);
   });
 
+  const renderViewPublicKey = async (viewPubkeyButton: HTMLElement) => {
+    const [contact] = await Store.dbContactGet(undefined, [$(viewPubkeyButton).closest('tr').attr('email')!]); // defined above
+    $('.hide_when_rendering_subpage').css('display', 'none');
+    Xss.sanitizeRender('h1', `${backBtn}${space}${contact!.email}`); // should exist - from list of contacts
+    if (contact!.client === 'cryptup') {
+      Xss.sanitizeAppend('h1', '&nbsp;&nbsp;&nbsp;&nbsp;<img src="/img/logo/flowcrypt-logo-19-19.png" />');
+    } else {
+      Xss.sanitizeAppend('h1', '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;');
+    }
+    $('#view_contact .key_dump').text(contact!.pubkey!); // should exist - from list of contacts && should have pgp - filtered
+    $('#view_contact .key_fingerprint').text(contact!.fingerprint!); // should exist - from list of contacts && should have pgp - filtered
+    $('#view_contact .key_words').text(contact!.keywords!); // should exist - from list of contacts && should have pgp - filtered
+    $('#view_contact').css('display', 'block');
+    $('#page_back_button').click(Ui.event.handle(() => renderContactList()));
+  };
+
+  const renderChangePublicKey = (changePubkeyButton: HTMLElement) => {
+    $('.hide_when_rendering_subpage').css('display', 'none');
+    const email = $(changePubkeyButton).closest('tr').attr('email')!;
+    Xss.sanitizeRender('h1', `${backBtn}${space}${Xss.escape(email)}${space}(edit)`);
+    $('#edit_contact').css('display', 'block');
+    $('#edit_contact .input_pubkey').val('').attr('email', email);
+    $('#page_back_button').click(Ui.event.handle(() => renderContactList()));
+  };
+
+  const actionSaveEditedPublicKey = async () => {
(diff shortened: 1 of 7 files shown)


tomholub:
Now that I see the issue, I would like to clarify (sorry I didn't think about it earlier!) - we could also auto-detect if the file has armored keys in it and process that too. The format is usually armored pubkeys concatenated with a newline.

That way it will work regardless if I use binary keyring or armored keyring.
(there may be several armored blocks, and each armored block may have several keys within it)

extension/chrome/settings/modules/contacts.ts
```diff
@@ -115,7 +117,27 @@ Catch.try(async () => {
...
+
+    $('#file_import .action_binary_keyfile').off().click(Ui.event.prevent('double', async target => {
+      $('#binary_file').click();
+    }));
+
+    $('#binary_file').change(Ui.event.handle(async (self, event: JQuery.Event<HTMLInputElement, null>) => {
+      if (event !== null && typeof event !== 'undefined') {
+        if (event.target.files !== null && typeof event.target.files !== 'undefined' && event.target.files.length > 0) {
+          const fileReader = new FileReader();
+          fileReader.onload = async () => {
```
tomholub: As is, exceptions that happen inside this function will not get reported. Please add try/catch with `Catch.handleErr` in the catch block.

extension/chrome/settings/modules/contacts.ts
```diff
@@ -115,7 +117,27 @@ Catch.try(async () => {
...
+            if (fileReader.result instanceof ArrayBuffer) {
```
tomholub: Thanks for staying safe. In this particular case, since the spec says it must be ArrayBuffer, you can cast it as

```js
new Uint8Array(fileReader.result as ArrayBuffer);  // readAsArrayBuffer must produce ArrayBuffer
```

extension/js/common/core/pgp.ts
```diff
@@ -246,6 +246,10 @@ export class Pgp {
       const { keys: [key] } = await openpgp.key.readArmored(armoredKey);
       return key;
     },
+    readBinary: async (binaryKey: Uint8Array) => {
+      const { keys: [key] } = await openpgp.key.read(binaryKey);
```
tomholub: I really like that you followed the style above! The trouble with this particular usecase is that we should import all public keys in the file, and there may be many. So returning one file is not so useful, and returning many is inconsistent with `Pgp.key.read`.

`Pgp.key.read` is used to read internal already-stored keys which are guaranteed to be one key per armored block, but keys imported from external sources may contain several, so in this case I would recommend to remove this method and use the code directly.

In general, adding an appropriate method, similar to other methods around, would be the right thing to do. In this particular case, the intended usecase does not match that well.

Maybe we should actually rename `.read` to `.readOne` to make it really clear, but that's for another day.

extension/chrome/settings/modules/contacts.ts
```diff
@@ -115,7 +117,27 @@ Catch.try(async () => {
         await Ui.modal.warning('Could not find any new public keys');
       } else {
         $('#bulk_import #processed').html(replacedHtmlSafe).css('display', 'block'); // xss-safe-factory
-        $('#bulk_import .input_pubkey, #bulk_import .action_process').css('display', 'none');
+        $('#bulk_import .input_pubkey, #bulk_import .action_process, #file_import .action_binary_keyfile').css('display', 'none');
+      }
+    }));
+
+    $('#file_import .action_binary_keyfile').off().click(Ui.event.prevent('double', async target => {
+      $('#binary_file').click();
```
tomholub: Please add comment in the code to add a note about what this does / why is it useful.

tomholub: (eg why is a click on the other element needed, as opposed to just having one element. Eg is the other one hidden for visual reasons? etc)

tomholub:
Test failure: this branch was not forked from latest master, and it's missing a test fix that should hopefully prevent this fail.

extension/chrome/settings/modules/contacts.ts
schmorrison: No problem, I wasn't really thinking. For some reason I had it in my head that the input element would provide text input element. I forgot that it is just a button.

extension/js/common/core/pgp.ts
schmorrison: Okey dokey. So I should use `openpgp.read()` directly?

 I can do that, but can you advise how you want me to handle after obtaining the array of keys. Currently I am writing the single returned key into the pubkey input text box and the user will then press the `Show Contents` button. Now assuming the keyring has multiple keys should these be handled in another way. I assume the `Ui.replaceRenderableMsgBlocks` will not work for a concatenated list of key blocks.

Ive been looking over things again, I am not quite clear on how the contact/PGP key is being stored.

extension/chrome/settings/modules/contacts.ts
schmorrison: Actually I am finding applying the button styles to the input element really very uncooperative. So I am hiding the file input and triggering a click on the file input element when the styled button is clicked. Then I have an event handler onchange for the input element.

schmorrison:
> Now that I see the issue, I would like to clarify (sorry I didn't think about it earlier!) - we could also auto-detect if the file has armored keys in it and process that too. **The format is usually armored pubkeys concatenated with a newline.**
> 
> That way it will work regardless if I use binary keyring file or armored keyring file.
> (there may be several armored blocks, and each armored block may have several keys within it)

@tomholub I did some more work on this and I have had 2 findings:

1. I have been unable to find a GPG command that will export PGP public keys as a list `concatenated with a newline`, GPG seems to only export the keys in a single key block (though GPG will import a file with a list of key blocks). I'm not sure what program will export a public keyring in the format you describe, but I am trying to handle it in FC

2. Given a file of armored public keys in the format you describe, `openpgp.readArmored` only seems to read the first armored block and ignores the rest, but it can read the correct number of keys in the format exported by `gpg --export --armor`. So I have decided to try using `Pgp.armor.detectBlocks` on a keyring in the format you describe, obviously all the key blocks can be detected, at which point I can read each key from the resulting blocks content field.

I am not sure exactly how you would like me to proceed from here. I could write each key into the `input_pubkey` field and obtain the 'Add To Contacts' blocks for the user to select which keys to add. Seems like the easiest option from my end.

tomholub:
> 1. I have been unable to find a GPG command that will export PGP public keys as a list concatenated with a newline, GPG seems to only export the keys in a single key block (though GPG will import a file with a list of key blocks). I'm not sure what program will export a public keyring in the format you describe, but I am trying to handle it in FC

Some software does that, including FlowCrypt (try to export the contacts)

> 2. Given a file of armored public keys in the format you describe, openpgp.readArmored only seems to read the first armored block and ignores the rest, but it can read the correct number of keys in the format exported by gpg --export --armor. So I have decided to try using Pgp.armor.detectBlocks on a keyring in the format you describe, obviously all the key blocks can be detected, at which point I can read each key from the resulting blocks content field.

Correct, you will need to detect the blocks and process these.

> I am not sure exactly how you would like me to proceed from here. I could write each key into the input_pubkey field and obtain the 'Add To Contacts' blocks for the user to select which keys to add. Seems like the easiest option from my end.

That's what I would do too.

extension/chrome/settings/modules/contacts.ts
```diff
@@ -23,6 +33,48 @@ Catch.try(async () => {
...
+  });
+
+  const handleKeyfileSelected = async (file: Att) => {
+    try {
+      const rawFile = file.getData();
+      const stringFile = new TextDecoder("utf-8").decode(rawFile);
+      const { blocks } = await Pgp.armor.detectBlocks(stringFile);
+
+      const keys: OpenPGP.key.Key[] = [];
+      if (blocks !== null && blocks.length > 0 && blocks.some(block => block.type === "publicKey")) {
```
tomholub: Check the return type, `blocks` will never be null - no need to check for it.

extension/chrome/settings/modules/contacts.ts
```diff
@@ -23,6 +33,48 @@ Catch.try(async () => {
...
+        for (const block of blocks) {
+          if (block.type === "publicKey") {
+            const result = await openpgp.key.readArmored(block.content);
+            if (result.err !== null && typeof result.err !== 'undefined') {
```
tomholub: can simplify to `if (result.err)`.

However, I think it would be better to ignore these errors. If at least one pubkey can be processed, I'd consider that a success.

extension/chrome/settings/modules/contacts.ts
```diff
@@ -23,6 +33,48 @@ Catch.try(async () => {
 
   BrowserMsg.listen(tabId); // set_css
 
+  const attUI = new AttUI(() => Promise.resolve({ size_mb: 5, size: 5 * 1024 * 1024, count: 1 }));
+  attUI.initAttDialog('fineuploader', 'fineuploader_button');
+  attUI.setAttAddedCb(async (file: Att) => {
+    (file !== null && typeof file !== 'undefined' && typeof file.id !== 'undefined' && file.id !== '') ? attUI.clearAtt(file.id) : attUI.clearAtts();
```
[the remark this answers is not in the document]
tomholub: Actually, does clearing / canceling attachments serve any purpose here? Does the functionality change when you remove it?

extension/js/common/browser.ts
```diff
@@ -1239,6 +1239,14 @@ export class AttUI {
     return atts;
   }
 
+  clearAtt = (id: string) => {
+    delete this.attachedFiles[id];
+  }
```
tomholub: This method is a carbon copy of `cancelAtt`

extension/chrome/settings/modules/contacts.ts
```diff
@@ -23,6 +33,48 @@ Catch.try(async () => {
...
+            if (result.err !== null && typeof result.err !== 'undefined') {
+              throw result.err;
+            }
+            keys.push(...result.keys);
+          }
+        }
+      } else {
+        const result = await openpgp.key.read(rawFile);
+        if (result.err !== null && typeof result.err !== 'undefined') {
+          throw result.err;
```
tomholub: Again I would probably not catch errors here, just collect keys.

If you do catch errors, then you should be showing them to the user. `Ui.modal.error` can help you do that. As it is now, errors will just get reported to us (using handleErr) without informing the user, which is not useful anyway.

extension/chrome/settings/modules/contacts.ts
schmorrison: Ok. Its a habit, but if you would prefer I ignore null checks when the typings don't provide a null I will ignore them. It will make my code more concise for sure.

tomholub: TypeScript is quite good with null and undefined types. If the type is wrong, I would rather that it throws an exception, so that we know to update the type.

extension/chrome/settings/modules/contacts.ts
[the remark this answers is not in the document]
schmorrison: What was happening, if I decided I wanted to change the file imported (picked the wrong file, or a file that didn't have any PGP keys) clicking the import button again it would throw an error modal about having too many attachments. So since I am not actually uploading the attachments, I thought I could clear the attachment variable as soon as I had the file object in my callback.

image

extension/js/common/browser.ts
schmorrison: `cancelAtt` is a private method, I didn't want to change the access key of your methods. But I can change and use `cancelAtt` if you would prefer. Or I could call `cancelAtt` from with in `clearAtt`.

schmorrison: Since I am just clearing all attachments I will remove this method.

extension/chrome/settings/modules/contacts.ts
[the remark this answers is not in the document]
tomholub: Got it, makes sense.

extension/js/common/browser.ts
tomholub: Please go ahead and change cancelAtt to public, that is better than creating effectively the same method that is public.

You should not need `cancelAtts` next to it, the already existing method should do what you need.

extension/js/common/core/pgp.ts
tomholub: Sorry I couldn't respond earlier. As it is now it looks better (seems you have removed it and managed to answer your questions :) ).

extension/chrome/settings/modules/contacts.ts
tomholub: No problem there, just when you do something that doesn't have a clear purpose from the code, it's good to add a comment to explain the intent. I do see that you have since replaced it with code similar to what we do elsewhere, presumably resolving the issues.

extension/chrome/settings/modules/contacts.ts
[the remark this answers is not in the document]
tomholub: In this case you should be able to just use `cancelAtt` (made public) with the id without all the checks in front of it (if I'm not wrong).

extension/chrome/settings/modules/contacts.ts
schmorrison: So I changed it a bit. There is a little duplication now, which I can make a helper method for but it probably isn't required.

    if (result.err) {
        let e = `Got error processing armored key:\n`;
        result.err.map(v => e += `-> ${v.message}\n`);
        await Ui.modal.error(e);
        throw result.err;
    }

tomholub: You could collect errors the way you collect parsed keys into an erray, and then put this block once at the end.
Also I would not be re-throwing the error.

extension/chrome/settings/modules/contacts.ts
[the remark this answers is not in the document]
schmorrison: Slight problem with the above, the `attUI.setAttAddedCb` is returning a file object with an 'undefined' id field. This is the object I am having returned.

```
{
    bytes: Uint8Array[.......],
    cid: undefined,
    getData: () => {....},
    hasData: () => {....},
    id: undefined,
    inline: true,
    length: 27115,
    msgId: undefined,
    name: "pbr.asc",
    setData: (bytes) => {....},
    treatAs: () => {....},
    treatAsValue: undefined,
    type: "text/plain",
    url: undefined
}
```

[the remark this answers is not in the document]
tomholub: Got it. You have found an api where the naming is not the best(on a second look), because the attachment id represents a different thing than the file upload id.

In this case, the simplest thing is to do `clearAtts` as you did, but I would call it `cancelAllAtts()` to resemble original naming, and make it public. Then `cancelAtt(id)` would stay private.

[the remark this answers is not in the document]
tomholub: Having you work on this uncovers some issues with my own code that's not so clear, I'll make an issue for this.

extension/js/common/platform/require.ts
```diff
@@ -31,6 +33,11 @@ type Codec = { encode: (text: string, mode: 'fatal' | 'html') => string, decode:
 
 export const requireOpenpgp = (): typeof OpenPGP => {
   try {
+    if (typeof openpgp !== 'undefined') { // in certain environments, eg browser content scripts, openpgp is not included (not all functions below need it)
```
tomholub: You can remove the if/else here, thanks to the try/catch block

extension/chrome/settings/modules/contacts.ts
```diff
@@ -23,6 +26,54 @@ Catch.try(async () => {
...
+      const rawFile = file.getData();
+      const stringFile = new TextDecoder("utf-8").decode(rawFile);
+      const { blocks } = await Pgp.armor.detectBlocks(stringFile);
+
+      const keys: OpenPGP.key.Key[] = [];
+      if (blocks.length > 0 && blocks.some(block => block.type === "publicKey")) {
+        for (const block of blocks) {
+          if (block.type === "publicKey") {
+            const result = await openpgp.key.readArmored(block.content);
+            if (result.err) {
```
tomholub: Can collect errs into array like you do with keys, to prevent code duplication, then check errs at the end with a single err handling block.

extension/js/common/browser.ts
```diff
@@ -1239,6 +1239,10 @@ export class AttUI {
     return atts;
   }
 
+  clearAtts = () => {
```
tomholub: Rename to `cancelAllAtts` as communicated earlier.

tomholub:
Tried a larger file: 

image

tomholub:
Maybe that's for a separate issue

Issue #941, opened 2017-11-06
drafts always re-uses original recipients


tomholub:
changes to recipients should be also saved

tomholub:
The way to reproduce:
1. You are among several recipients of an encrypted mail.
2. You hit Reply (not Reply All)
3. Start typing response. Notice the To field has only one recipient - the original sender.
4. Switch to another folder, e.g. Sent mail. Your draft is saved.
5. Return to Inbox and click on your draft.
6. The To field will load all recipients instead of only the original sender.

tomholub:
ref https://mail.google.com/mail/u/1/#inbox/15f9139b950ccec4

tomholub:
(reserving issue for interviews)

tomholub:
The solution would be to make sure encrypted drafts also get saved with recipients, and this info is then surfaced when loading the drafts. Work will involve editting `common/composer.js` and `elements/pgp_block.js`

tomholub:
(reserving issue for interviews)

PR #1611, opened 2019-03-30
'parseRenderRecipients' returns boolean if reciepent is added and save draft depends on that value

michael-volynets:


tomholub:
#941

Issue #714, opened 2017-04-19
pop out/maximize secure compose window


tomholub:
This will involve adding a maximize icon to the left of close icon in our Secure Compose window. Behavior similar to what Gmail does.

You can use maximize or minimize icons from https://feathericons.com/ - please choose style that you think fits better.

![image](https://user-images.githubusercontent.com/6306961/55668388-3d94b400-5859-11e9-9f09-38862814ed32.png)

--------------------------

Functionality similar to Gmail: when I click on it, it window will get maximized from small:

![image](https://user-images.githubusercontent.com/6306961/55578967-a7e90f80-5706-11e9-8ff0-fdae892d584c.png)

to big (make sure to also lift it a bit from the bottom, like Gmail does):

![image](https://user-images.githubusercontent.com/6306961/55578972-acadc380-5706-11e9-9e5a-9c7aa67f5a8f.png)

And back to normal size if I click on it again.

tomholub:
https://mail.google.com/mail/u/1/#inbox/15cef8079148cf42

tomholub:
ref https://mail.google.com/mail/u/0/#inbox/163016d3c86314d7

michabbb:
it´s only css, so please don´t wait too long with that... thanks !

tomholub:
ref https://mail.google.com/mail/u/0/#inbox/163730a1a78123df

tomholub:
reserving this as a screening issue

tomholub:
Lyova is taking it

tomholub:
Available again.

michael-volynets:
@tomholub do you know how I can add class to the element out of iFrame?
e.g. here 
![image](https://user-images.githubusercontent.com/49104848/57875304-48107900-781b-11e9-968e-7523306f5aae.png)

I was trying to use BrowserMsg.send (I created some new methods such as add_class and remove_class)
but, I don't know why, its behavior is very weird. Sometimes it works, sometimes it doesn't.

I need to add the class synchronously.
Maybe you know some way to do it.
Thanks

tomholub:
> I was trying to use BrowserMsg.send (I created some new methods such as add_class and remove_class)

That is probably what you have to do, I don't know of any other way

> but, I don't know why, its behavior is very weird. Sometimes it works, sometimes it doesn't.

[...]

PR #1627, opened 2019-04-04
issue #714 pop out/maximize secure compose window

lyova98:


tomholub:
@lyova98  Hello, this does not address the issue, which was to **add** (not replace) a fourth button that would come with new functionality to pop out / maximize the compose window, similar to what Gmail does:

image


image

And back to normal size if I click on it again.

lyova98:
@tomholub as I understand the task is change the maximize icon to the left of close icon □

tomholub:
No, the task is to add a new icon, similar to the one in Gmail, and implement similar functionality to what Gmail does.

lyova98:
@tomholub so I have changed the icon  □  to like in gmail, now I need to add functionality like in gmail for resize window icon like in gmail right?
https://prnt.sc/n7k6s1

Issue #1072, opened 2018-04-29
Recipient is there in the email draft, but FlowCrypt thinks it's not


chuacw:
This is hard to reproduce, but occasionally, when I write a new email, and I've added a new recipient, write portions of the email, then save it as draft, some time later (after closing and restarting the browser, or changing to a tab, then coming back to Gmail), and reopening the draft to work on it.

When trying to send the email, FlowCrypt will complain that there's no recipient, but the recipient is there in the draft!

tomholub:
I'm adding actionable tag in the sense that we should triage it.

Possible sources of trouble: 
 - maybe there is a race condition filling up recipients when loading from draft and evaluating recipients
 - maybe we fixed it since. The report is from about time before we aggressively started to use async / await and sometimes weird race conditions would happen.

contributor: try this a few times, if you cannot reproduce it, please close.

PR #1632, opened 2019-04-06
issue-1072 Load recipients while draft is loading

michael-volynets:
#1072

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -514,7 +514,7 @@ export class Composer {
           this.S.cached('input_to').focus();
           this.S.cached('input_to').val(headers.to.join(','));
           this.S.cached('input_text').focus();
-          this.parseRenderRecipients('harshRecipientErrs');
+          this.parseRenderRecipients('harshRecipientErrs').catch(Catch.reportErr);
         }
         if (headers && headers.from) {
           this.S.now('input_from').val(headers.from);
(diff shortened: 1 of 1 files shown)


tomholub:
@michael-volynets when you install `tslint` extension in vscode (and restart vscode), it should show you red squiggles in `composer.ts` around line 517.
```
ERROR: /home/runner/flowcrypt-browser/extension/js/common/composer.ts[517, 11]: Promises must be handled appropriately
```

It's complaining because `parseRenderRecipients` is async and returns a promise. If you don't `await` it, it will silently drop errors (plus you may run into concurrency issues). It wants you to `await` it.

extension/js/common/composer.ts
```diff
@@ -514,6 +514,7 @@ export class Composer {
           this.S.cached('input_to').focus();
           this.S.cached('input_to').val(headers.to.join(','));
           this.S.cached('input_text').focus();
+          this.parseRenderRecipients('harshRecipientErrs');
```
tomholub: > ERROR: /home/runner/flowcrypt-browser/extension/js/common/composer.ts[517, 11]: Promises must be handled appropriately

This call is missing an `await`

michael-volynets:
Hi, Tom,
I wanted to add there "await", but if I add await there, the composer is
loading more time than without it,
I mean without "await" everything works fine, composer loads faster, but
when it's loaded, we see loading icons on recipients list.

tomholub:
Got it. If you don't want to wait for it to finish, you should add `.catch(Catch.reportErr)`. This way, if errors ever happen, they will be properly reported. This will make it stop complaining, too.

michael-volynets:
Oh, Okay, Got It

Issue #535, opened 2017-02-26
Message and attachments overlap 


mondo989:
![Uploading Screen Shot 2017-02-26 at 1.40.58 AM.png…]()

tomholub:
Don't see the image but I think I know what you mean

tomholub:
First of all, I think the attachments should be styled as `flow:right` or `flow:letf`. Then they should stack to the left of each other, until they form a second line from the bottom.

That way at least they all fit on the screen.

mondo989:
I like how it never attached :o

tomholub:
Repro: add 10 attachments to a reply with a short message body and notice it overflow

![image](https://user-images.githubusercontent.com/6306961/53565446-70a8a100-3b51-11e9-8e48-6971e08976a8.png)

Only grab this issue if you are very comfortable changing a bunch of html/css in addition to (possibly) a bit of js. 

I'm thinking the attachments should be in their own table row, which should have no height if there are no attachments. Then the attachments inside should be `inline-block` so that they go to the right of one another, until they form a second row and the containing `td` expands. That way they don't overlap with text.

PR #1633, opened 2019-04-06
Issue-535 Fixed owerlap when some files uploaded

michael-volynets:
#535

### DIFF
--- a/extension/js/common/browser.ts
+++ b/extension/js/common/browser.ts
@@ -1186,7 +1186,7 @@ export class AttUI {
     this.getLimits = getLimits;
   }
 
-  initAttDialog = (elId: string, btnId: string) => {
+  initAttDialog = (elId: string, btnId: string, onSuccessUpload?: () => Promise<void>) => {
     $('#qq-template').load(this.templatePath, () => {
       const config = {
         autoUpload: false,
@@ -1197,7 +1197,7 @@ export class AttUI {
           extraDropzones: $('#input_text'),
         },
         callbacks: {
-          onSubmitted: (uploadFileId: string, name: string) => this.processNewAtt(uploadFileId, name).catch(Catch.reportErr),
+          onSubmitted: (uploadFileId: string, name: string) => this.processNewAtt(uploadFileId, name, onSuccessUpload).catch(Catch.reportErr),
           onCancel: (uploadFileId: string) => Catch.try(() => this.cancelAtt(uploadFileId))(),
         },
       };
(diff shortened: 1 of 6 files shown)


tomholub:
Thanks! I will review it. For issues that change dexign or UX, please include a screen shot after the change in the PR. That will be super helpful (in the future).

michael-volynets:
Okay, Got it,

image

tomholub:
I just realized what the issue is, and here is how I resolved it before:
```js
  /**
   * On Firefox, we have to manage textbox height manually. Only applies to composing new messages
   * (else ff will keep expanding body element beyond frame view)
   * A decade old firefox bug is the culprit: https://bugzilla.mozilla.org/show_bug.cgi?id=202081
   *
   * @param updateRefBodyHeight - set to true to take a new snapshot of intended html body height
   */
  private setInputTextHeightManuallyIfNeeded = (updateRefBodyHeight: boolean = false) => {
    if (!this.urlParams.isReplyBox && Catch.browser().name === 'firefox') {
      let cellHeightExceptText = 0;
      for (const cell of this.S.cached('all_cells_except_text')) {
        cellHeightExceptText += $(cell).is(':visible') ? ($(cell).parent('tr').height() || 0) + 1 : 0; // add a 1px border height for each table row
      }
      if (updateRefBodyHeight || !this.refBodyHeight) {
        this.refBodyHeight = this.S.cached('body').height() || 605;
      }
      this.S.cached('input_text').css('height', this.refBodyHeight - cellHeightExceptText);
    }
}
```
You'll just need to update that function. I'm guessing at the end you will subtract attachment box height also, and make sure this function is also called when attachments are added or removed.

tomholub:
For the attachments themselves, it would be better if the the attachments themselves had variable width based on the file name (with some maximum). Something like this:

image

Now that I'm looking at it, it would be great if the `X` was also centered with the borders: 

image

Also the space between attachments should be the same vertically as horizontally:

image

Thanks!

michael-volynets:
The result after all Changes:
image

Checked on Firefox and it works fine

extension/js/common/composer.ts
```diff
@@ -1543,7 +1544,7 @@ export class Composer {
       }
     }, this.getErrHandlers(`focus on recipient field`))).children().click(() => false);
     this.resizeInputTo();
-    this.attach.initAttDialog('fineuploader', 'fineuploader_button');
+    this.attach.initAttDialog('fineuploader', 'fineuploader_button', async () => this.setInputTextHeightManuallyIfNeeded());
```
tomholub: Would it be possible to re-use `this.attach.setAttAddedCb(...)` ? That way we would not need to add a new `onSuccess` callback above.

extension/js/common/browser.ts
```diff
@@ -1266,6 +1266,7 @@ export class AttUI {
         return;
       }
       this.attachedFiles[uploadFileId] = newFile;
+      onSuccess!().catch(Catch.reportErr);
```
tomholub: You will get exception if `onSuccess` was not defined (see other note, adding callback may not be necessary).

extension/js/common/composer.ts
michael-volynets: oh, yes, I didn't see that method.

tomholub:
Normally you don't need to test on Firefox too much. It's mostly design changes that can come out a bit unexpected on FF, and this one was an especially tricky one, where the behavior on FF is very different. Normally this won't be an issue.

michael-volynets:
Ok, got it

Issue #1047, opened 2018-04-01
"Retrieving message.." should show percentage


tomholub:
ref comp https://mail.google.com/mail/u/2/#label/tests%2Fattachment/1622eaa286f90737

tomholub:
We already have callbacks for % progress on our AJAX api calls, you just have to surface it to the UI. Take a look at uploading files during compose, or downloading a file in `attachment.ts`

tomholub:
This is the relevant part to start updating: 
https://github.com/FlowCrypt/flowcrypt-browser/blob/61295de4cc472a7a965e3fa60b2921637eee9494/extension/chrome/elements/pgp_block.ts#L487

Here is an example of how to show progress as a file is downloaded (elsewhere): 
https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/chrome/elements/attachment.ts#L140

Within `extractArmoredBlock`, in some situations we call it in `full` format, where it only gets metadata and body, and then we download attachments separately (which should be much larger than body, body should be negligible in such case). Other times we call it in `raw` format where it gets a mime message that may potentially be big. It should show percentage progress in either case.

To debug this, you can install Enigmail and send yourself an encrypted message in PGP/MIME format and a 1MB+ attachment.

PR #1645, opened 2019-04-10
#1047 Added percent of loading while retriving message

michael-volynets:


michael-volynets:
Here is a solution only for "full" format because I haven't found yet how I can do for "raw" format

extension/js/common/api/google.ts
```diff
@@ -286,8 +286,17 @@ export class Google extends Api {
...
-    fetchAtts: async (acctEmail: string, atts: Att[]) => {
-      const responses = await Promise.all(atts.map(a => Google.gmail.attGet(acctEmail, a.msgId!, a.id!)));
+    fetchAtts: async (acctEmail: string, atts: Att[], progressCb?: ProgressCb) => {
+      const loadedAr: Array<number> = [];
+      // 1.33 is a coefficient we need to multiply because total size we need to download is larger than all files together
+      const total = atts.map(x => x.length).reduce((a, b) => a + b) * 1.33;
+      const responses = await Promise.all(atts.map((a, index) => Google.gmail.attGet(acctEmail, a.msgId!, a.id!, (progress, loaded, s) => {
+        loadedAr[index] = loaded || 0;
+        const totalLoaded = loadedAr.reduce((a, b) => a + b);
+        if (progressCb) {
```
tomholub: This is the correct way to do it - check if `progressCb` exists instead of forcing it with `!`. Thanks!

Issue #1649, opened 2019-04-15
right-clicking should work on any part of the attachment


tomholub:
Right now user has to right-click specifically the link, which is not obvious.

Instead we should allow to right click anywhere within the attachment frame (the whole thing would be a link) and it should also show a `pointer` style of cursor.

The text should say:
> Right-click here and choose Save Link As to save encrypted file.

PR #1651, opened 2019-04-15
issue-1649 Changed download text and made the whole block clickable

michael-volynets:
How it looks like:

image

### DIFF
--- a/extension/js/common/browser.ts
+++ b/extension/js/common/browser.ts
@@ -63,13 +63,10 @@ export class Browser {
       a.href = window.URL.createObjectURL(blob);
       a.download = Xss.escape(att.name);
       if (renderIn) {
-        a.textContent = 'DECRYPTED FILE';
-        a.style.cssText = 'font-size: 16px; font-weight: bold;';
-        Xss.sanitizeRender(
-          renderIn,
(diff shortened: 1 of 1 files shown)


tomholub:
If you take a look at the test failure at https://semaphoreci.com/flowcrypt/flowcrypt-browser/branches/issue-1649-right-clicking-on-attachment/builds/1 , you'll see:
```
> flowcrypt-browser@6.7.4 test_patterns /home/runner/flowcrypt-browser
> node build/test/test/source/patterns.js

unchecked xss in extension/js/common/browser.ts:66:
        a.innerHTML = '<div>Right-click here and choose \'Save Link As\' to save encrypted file</div>';

patterns.ts: Found 1 unhandled patterns, exiting
```
Our code checkers don't like to see innerHTML because it is easy to cause xss vulnerability (not in your case, but in general).
Since you are adding the content there straight from a string (and not from variable) you can add ` // xss-direct` comment at the end which should suppress the code check on that line.

michael-volynets:
oh, okay, got it, 
I forgot to run tests before doing the pull request.

tomholub:
No worries, that's what CI is for.

tomholub:
#1649

Issue #1625, opened 2019-04-04
update featherlight to latest version


tomholub:
maybe they have since resolved the following err that I keep seeing:
```
TypeError: Cannot read property 'naturalWidth' of undefined
at Featherlight.onResize (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/featherlight/featherlight.js:591:31)
at Featherlight.i [as onResize] (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/jquery.min.js:2:86266)
at Featherlight.<anonymous> (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/featherlight/featherlight.js:96:45)
at Function.each (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/jquery.min.js:2:2573)
at globalEventHandler (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/featherlight/featherlight.js:94:5)
at dispatch (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/jquery.min.js:2:41772)
at y.handle (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/jquery.min.js:2:39791)
```
I believe I've made some changes to featherlight css we use (uncommented `.featherlight .featherlight-close-icon`), so if you also update the css, pay attention to see if anything broke or changed in regards to styling, and maybe uncomment it in the same way (or do something appropriate in our own css without changing featherlight css file, for easier updates).

Alternatively we could just ignore the new css version and keep using our own - up to you.

tomholub:
@michael-volynets You said you encountered the error above on the new version. How did you manage to make the error show? I could never reproduce it. Can you still reproduce it?

PR #1660, opened 2019-04-18
Issue #1625 Changed version from 1.7.0 to 1.7.13

michael-volynets:
I've checked this method because it uses featherlight:
`Ui.renderOverlayPromptAwaitUserChoice({ finishSetup: {}, later: { color: 'gray' } }, msg);`

Everything looks the same way

#1625

### DIFF
--- a/extension/css/cryptup.css
+++ b/extension/css/cryptup.css
@@ -509,7 +509,8 @@ div#cryptup_dialog iframe.medium { height: 500px; margin-top: -300px; }
 div#cryptup_dialog iframe.tall { height: 780px; margin-top: -390px; }
 
 .featherlight { z-index: 8 !important; }
-.featherlight .featherlight-close-icon { background: none !important; border: none !important; }
+.featherlight .featherlight-close-icon { background: none !important; border: none !important; width: 30px !important; 
+  font-size: 24px !important; opacity: 0.5 !important; line-height: 36px !important; }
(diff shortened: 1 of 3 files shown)


tomholub:
Look at a possible missing close button when you open any settings sub-page (eg help form)

tomholub:
Or maybe the tag changed so automated tests cannot find the button

michael-volynets:
Yes, There aren't 'data-test' attributes, 
I've added them but I did two commits, I can merge them in one if it's better

tomholub:
No problem, I always squash the PR into one commit when I merge it into master.

tomholub:
I just messed up something on the backend, investigating. It may affect these tests too.

michael-volynets:
I checked the test logs and there are errors only related to 'data-test' attributes.

michael-volynets:
Perhaps would be better to add that attribute in another place?
Because if we update this library the tests will fail again
What do you think?

tomholub:
Sure, it would be better if you can figure out how.

michael-volynets:
I just realized that the page doesn't load at all 

There is the error message you have written in issue #1625.

They haven't fixed the issue yet,
Let's keep the current version of featherlight?

tomholub:
I'll take a look at it later, you can leave this PR as is

tomholub:
Also fixing css to be as before (need larger X button):

From:
image

To:
image

Issue #996, opened 2018-02-08
investigate missing gmail message id


tomholub:
```
acarrel-certificat-cabincetcarrel.asc
acarrel-certificat-inovalex.asc
Clef publique Vincent Baudouin Inovalex.asc
```

maybe investigate any small .asc file?

ref https://mail.google.com/mail/u/0/#sent/16157c5612b54ba2

tomholub:
This may now work properly, but needs to be tested.

Repro: create three files with the names above, and put an armored pubkey into each. Just plain text files. Send them to flowcrypt.compatibility@gmail.com attached in standard non-encrypted email, one per email.

See if FlowCrypt picks up on the files and renders the import pubkey UI.

If all rendered well, please close. If some are ignored, consult @tomholub on how to proceed.

michael-volynets:
In general, it renders well but sometimes I see "Unknown Message Id" error or there isn't attachment block in the message.
I tried to investigate that but I haven't found the reason for that behavior. 
See the attachments:
![image](https://user-images.githubusercontent.com/49104848/56213256-5cb6f100-6064-11e9-93b4-1ebadfed9ecc.png)

tomholub:
Normally, the content script will try to figure out which Gmail message id are you viewing. Based on that, it can pull the message attachments from the Gmail API, investigate the attachments, and determine if they should be treated as public keys, or an encrypted message, encrypted attachment, plain attachment, ...

It's not always needed to check the API. But sometimes it is. If I send you a file named `something.asc`, it's hard to say if it's just a plain text file, pubkey, encrypted file or a message.

We parse message ID from the dom (google provides it in Gmail dom). Looks like in some situations it will fail, which would be definitely good to investigate!

michael-volynets:
Do you know why there is this comment?
`  private determineMsgId = (innerMsgEl: HTMLElement | JQueryEl) => { // todo - test and use data-message-id with Gmail API once available
    return $(innerMsgEl).parents(this.sel.msgOuter).attr('data-legacy-message-id') || '';
  }`

Sometimes It cannot find data "data-legacy-message-id" but there is always ''data-message-id".
Can I replace them? (data-legacy-message-id -> data-message-id)

tomholub:
Gmail is moving away from the legacy id format and into the new id format since last year.

The problem was, last time I checked, the Gmail API was still returning the old (legacy) format, so I couldn't make the switch without side effects.

Example api response:\
[...]

PR #1664, opened 2019-04-19
Changed attribute name for getting message id

michael-volynets:
I've tested decoding email message, replying to some encoded email message and everything is okay.
Those two methods work with "data-message-id"


michael-volynets:
Those tests really help :)

tomholub:
#996

Issue #1659, opened 2019-04-18
Recipients duplicates after draft saved


michael-volynets:
Steps to reproduce:
1) Open Reply Box with the recipient in it
2) Save Draft
3) Reload the page

![image](https://user-images.githubusercontent.com/49104848/56359200-8c433600-61e9-11e9-936f-1dd545181036.png)

tomholub:
Thanks! You can self-assign this when you have time.

PR #1668, opened 2019-04-22
Issue 1659 recipients dublicate

michael-volynets:
#1659

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -508,7 +508,9 @@ export class Composer {
       if (result.success) {
         this.S.cached('prompt').css({ display: 'none' });
         Xss.sanitizeRender(this.S.cached('input_text'), await Xss.htmlSanitizeKeepBasicTags(result.content.toUtfStr().replace(/\n/g, '<br>')));
-        if (headers && headers.to && headers.to.length) {
+        if (this.urlParams.isReplyBox) {
+          await this.renderReplyMsgComposeTable();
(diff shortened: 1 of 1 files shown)


extension/js/common/composer.ts
```diff
@@ -508,7 +508,9 @@ export class Composer {
       if (result.success) {
         this.S.cached('prompt').css({ display: 'none' });
         Xss.sanitizeRender(this.S.cached('input_text'), await Xss.htmlSanitizeKeepBasicTags(result.content.toUtfStr().replace(/\n/g, '<br>')));
-        if (headers && headers.to && headers.to.length) {
+        if (this.urlParams.isReplyBox) {
+          await this.renderReplyMsgComposeTable();
+        } else if (headers && headers.to && headers.to.length) {
           this.S.cached('input_to').focus();
```
tomholub: Thanks for the PR. Looking at the code I would imagine that loading recipients from the draft would no longer work, and instead it will load recipient based on the previous message.

I haven't tested it, but you can:
1) click reply
2) start typing draft
3) add a new recipient
4) wait for draft to save
5) refresh page, wait for draft to load
6) I think the new recipient will be missing

Please test this scenario and please fix it if it needs fixing. If it works well, let me know so I can merge it as is.

Thanks!

michael-volynets: Yes, Everything works well. Every Recipient is loaded.

tomholub: Perfect, thanks!

Issue #421, opened 2017-01-29
clicking on recipient pubkey fetch refresh is broken


tomholub:
To reproduce

0) open secure compose
1) go offline
2) add email address as recipient that didn't use FlowCrypt before, and have it attempt loading pubkey
3) error, offline
4) press the `retry` icon, which will (wrongly) remove the recipient. Instead, it's supposed to retry fetching recipient pubkey (as if the email was just entered).
5) Additionally, there should be an `X` button like in other situations that should do what the refresh button does now (remove the recipient)

Currently looks like this (missing the `X` icon)
![image](https://user-images.githubusercontent.com/6306961/55279831-2adc2580-5315-11e9-9195-0a5230c3eeca.png)

Solution:
 - make the refresh button actually retry loading recipient info
 - add an `X` button that removes the recipient

tomholub:
1) go offline
2) add never used before email and have it attempt loading pubkey
3) error, offline
4) retry button should retry loading.. plus add `X` button for removing

tomholub:
currently retry button removes the recipient - wrong

PR #1670, opened 2019-04-25
Issue 421 fetch refresh

michael-volynets:
How it looks like:
image


tomholub:
Thanks for the screenshot! Will take a look

tomholub:
#421 
(I'm adding the issue number inside a comment because without it, this PR doesn't get nicely linked to the issue on GitHub. You can do that in next PR)

Issue #1530, opened 2019-02-08
there should be a reply button on the bottom of convo


tomholub:

tomholub:
When replying to plain text messages, currently the only way to switch to encrypted is on the very top right of the conversation.

There should be a subtle button or link to reply with encrypted message on the bottom as well.

tomholub:
Todo in another issue: 

Make the lock much bigger so that it's obvious. Maybe make both the arrow and the lock the same size and put them next to one another - we can try combinations of `[lock][arrow]` or `[arrow][lock]`, depending on what seems more obvious to the user.

![image](https://user-images.githubusercontent.com/6306961/57691799-ac251880-7634-11e9-9623-1b4efadb3b3d.png)

tomholub:
Todo in another issue:

Make the reply button behave the same as surrounding buttons (circle background and a tooltip).

![image](https://user-images.githubusercontent.com/6306961/57691876-dbd42080-7634-11e9-8cb6-ee31b619fc05.png)

When I do this, usually I try to give it the same html classes that other surrounding buttons have, then Gmail will do the work. Then you add `data-tooltip="Secure Reply"`. I quickly tested this but it didn't behave well, please see if you can make it work.

If not, we would have to re-implement a tooltip (with css like here https://www.w3schools.com/css/tryit.asp?filename=trycss_tooltip ) and the background circle effect ourselves. But that is not ideal because users may be using a different Gmail theme. Still, it's better than nothing.

PR #1677, opened 2019-04-27
Issue 1530 reply button non encrypted email

michael-volynets:
#1530 

image


tomholub:
This is not bad. I was imagining to put the encrypted reply button on the bottom, but actually I think what you have done is better.

Maybe in this case we can stop adding the "secure reply" button on the top right, as it's redundant.

Secondly, the new button behaves weirdly when hovered, it will get bigger/smaller as I hover, disturbing surrounding buttons.

Thirdly, it seems to look a bit out of place (I know it's our own icon :) ), maybe we could redesign it. Maybe we could use the original Gmail icon, but in green color, and with a lock next to it. Then it will blend with the rest of Gmail better.

I can supply the icon if you like.

michael-volynets:
Okay, got it.
I will do it. Provide me the icon if you can

tomholub:
Just noting here that we talked over email that you'll give it a try to create that icon

michael-volynets:
Should the reply button be in all messages blocks or only in the last one?
image

Because  In the code I found that it tries to add it to each message block

tomholub:
You've done a great job here!

I especially like that I can click "encrypted reply", then change my mind and click "normal reply" instead, and it will work - I think it didn't work before and I really like to see it working. Not sure what you did there, but it really helped.

If you click encrypted -> normal reply -> encrypted -> normal (change your mind several times), it will get stuck on encrypted, similar to before. But that's not so bad, this situation should be a lot rarer.

How do you make svg icons? Which software do you use? It's good but the lock turned out very small. I'll open another issue for this.

Lastly, it's missing a title that normally shows over other gmail buttons when you hover over them. I'll add the title after I merge it.

PR #1683, opened 2019-05-03
Issue 1648 recognize backup

michael-volynets:
image

### DIFF
--- a/extension/chrome/elements/backup.ts
+++ b/extension/chrome/elements/backup.ts
@@ -65,7 +65,7 @@ Catch.try(async () => {
             $(".line.pass_phrase_test").addClass('green').text("Your pass phrase matches!");
             sendResizeMsg();
         } else {
-            await Ui.modal.warning('Pass phrase did not match. Please try again. If you forgot your pass phrase, please change it, so that do don\'t get' +
+            await Ui.modal.warning('Pass phrase did not match. Please try again. If you forgot your pass phrase, please change it, so that you don\'t get' +
                 'locked out of your encrypted messages.');
         }
     }));
--- a/extension/chrome/elements/backup.ts
+++ b/extension/chrome/elements/backup.ts
@@ -3,10 +3,11 @@
 'use strict'
 
 import { Catch } from '../../js/common/platform/catch.js';
-import { Ui, Env, Xss } from '../../js/common/browser.js';
+import { Ui, Env } from '../../js/common/browser.js';
 import { Str } from '../../js/common/core/common.js';
 import { mnemonic } from '../../js/common/core/mnemonic.js';
 import { Pgp } from '../../js/common/core/pgp.js';
+import { BrowserMsg } from '../../js/common/extension.js';
 import { Store } from '../../js/common/platform/store.js';
 
 declare const openpgp: typeof OpenPGP;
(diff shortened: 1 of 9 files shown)


tomholub:
Awesome, thanks! I'll review it tomorrow.

extension/chrome/elements/backup.ts
```diff
@@ -0,0 +1,75 @@
...
+        const desiredHeight = $('#backup_block').height()!;
+        BrowserMsg.send.setCss(parentTabId, { selector: `iframe#${frameId}`, css: { height: `${desiredHeight}px` } });
+    };
+
+    $('.action_test_pass').click(Ui.event.handle(async target => {
+        if (await Pgp.key.decrypt(key, [String($('#pass_phrase').val())]) === true) {
+            $(".line.pass_phrase_test").addClass('green').text("Your pass phrase matches!");
+            sendResizeMsg();
+        } else {
+            await Ui.modal.warning('Pass phrase did not match. Please try again. If you forgot your pass phrase, please change it, so that do don\'t get' +
```
tomholub: typo: ` so that do don\'t get` -> ` so that you don\'t get`

extension/js/common/browser.ts
```diff
@@ -827,6 +827,10 @@ export class XssSafeFactory {
     return this.frameSrc(this.extUrl('chrome/elements/pgp_pubkey.htm'), { frameId: this.newId(), armoredPubkey, minimized: Boolean(isOutgoind), });
   }
 
+  srcBackupIframe = (armoredPubkey: string, isOutgoind?: boolean) => {
```
tomholub: typo: `isOutgoind` -> `isOutgoing`

But it should be removed anyway, it's not used and not useful for private keys.

It's useful for public keys, because those may be sent over email. If a public key is outgoing, that means I sent it to someone else, then it's a waste of space to print out all the details about it. Because private keys are never outgoing (you would not be sending them to other people), `isOutgoing` is not useful for private key backups.

Please remove, thanks!

extension/js/common/browser.ts
```diff
@@ -895,6 +899,10 @@ export class XssSafeFactory {
     return this.iframe(this.srcPgpPubkeyIframe(armoredPubkey, isOutgoing), ['pgp_block']);
   }
 
+  embeddedBackup = (armoredPubkey: string, isOutgoing?: boolean) => {
```
tomholub: also please remove `isOutgoing`

extension/js/common/core/att.ts
```diff
@@ -97,7 +99,7 @@ export class Att {
     }
   }
 
-  public static pgpNamePatterns = () => ['*.pgp', '*.gpg', '*.asc', 'noname', 'message', 'PGPMIME version identification', ''];
+  public static readonly attachmentsPattern = /^(((cryptup|flowcrypt)-backup-[a-z]+\.key)|(.+\.pgp)|(.+\.gpg)|(.+\.asc)|(noname)|(message)|(PGPMIME version identification)|())$/gm;
```
tomholub: Looks good, it's a complicated pattern but it simplifies the logic overall, thanks!

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -221,7 +221,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     for (const a of attMetas) {
       const treatAs = a.treatAs();
       // todo - [same name + not processed].first() ... What if attachment metas are out of order compared to how gmail shows it? And have the same name?
-      const attSel = this.filterAtts(attsContainerInner.children().not('.attachment_processed'), [a.name]).first();
+      const attSel = this.filterAtts(attsContainerInner.children().not('.attachment_processed'), new RegExp(a.name)).first();
```
tomholub: I think you may need to regex-escape. I suspect this will fail with a filename `something[hello].txt`. Maybe it would even throw exception on bad regex like `some[thing.txt`, not sure. You can test it to see for yourself, and then probably use `Str.regexEscape` on the string before passing it to `new RegExp()`. After escaping, you could also add `^` on the beginning and `$` at the end just to be sure.

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -248,6 +248,8 @@ export class GmailElementReplacer implements WebmailElementReplacer {
             }
           } else if (treatAs === 'publicKey') { // todo - pubkey should be fetched in pgp_pubkey.js
             nRenderedAtts = await this.renderPublicKeyFromFile(a, attsContainerInner, msgEl, isOutgoing, attSel, nRenderedAtts);
+          } else if (treatAs === 'backup') {
+            nRenderedAtts = await this.renderBackupFromFile(a, attsContainerInner, msgEl, isOutgoing, attSel, nRenderedAtts);
```
tomholub: Please remove `isOutgoing` here as well.

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -323,21 +325,29 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
+  private renderBackupFromFile = async (attMeta: Att, attsContainerInner: JQueryEl, msgEl: JQueryEl, isOutgoing: boolean, attSel: JQueryEl, nRenderedAtts: number) => {
+    let downloadedAtt: R.GmailAtt;
+    try {
+      downloadedAtt = await Google.gmail.attGet(this.acctEmail, attMeta.msgId!, attMeta.id!); // .id! is present when fetched from api
+    } catch (e) {
+      attsContainerInner.show().addClass('attachment_processed').find('.attachment_loader').text('Please reload page');
+      nRenderedAtts++;
+      return nRenderedAtts;
+    }
+    const openpgpType = await BrowserMsg.send.bg.await.pgpMsgType({ rawBytesStr: downloadedAtt.data.toRawBytesStr() });
```
tomholub: Here you are doing type detection of the content based on the first few bytes of the attachment.

This is useful for public keys received from outside, because our regex to detect them is quite vague, and sometimes we're not sure if it's pubkey or not from the name alone.

But for backups (private keys), the regex is very specific, so it's unlikely any normal file will accidentally end up being wrongly treated as a backup. You can skip this detection and just treat the file as a backup in all cases.

extension/js/common/browser.ts
michael-volynets: Got it, Thanks!

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: oh, yes, I forgot that these situations could have happened. 
Thanks!

tomholub:
#1683

Issue #534, opened 2017-02-26
Drafts don't save attachments


mondo989:
Having CryptUp save drafts is awesome. 

Although I noticed that it did not save the attachment.

tomholub:
Ha. Didn't think of that.

tomholub:
Contributor: 
 - when saving draft, also upload attachments (unless the attachment did not change).
 - When removing an attachment, also remove it from the gmail draft.
 - When adding attachment, trigger a new draft save.
 - When loading msg from draft, also load attachments.

This is a prerequisite to #1211

michael-volynets:
Hello @tomholub, 

Is there a way to convert Att to a file? I save 'Att's in a draft but I can't turn it back to the 'File'

michael-volynets:
I've resolved the issue with the file.

tomholub:
Sorry for responding slowly. I'm not really sure which `File` you meant. If it's something to pass to Ajax, you may want to look at what the `Api.s3` api call does.

tomholub:
I'm marking it as not actionable, the code complication is not quite worth it. Maybe later.

tomholub:
related PR run into issues, may be lacking Gmail API

PR #1688, opened 2019-05-07
issue-534 Save and load attachments in drafts (stale)

michael-volynets:
#534 

There is an issue when you upload a large file, seems like that the composer is frozen.
it happens because it sends a request to Gmail to save the draft and it takes a lot of time,
Do you have any idea have it can be fixed?

### DIFF
--- a/extension/js/common/browser.ts
+++ b/extension/js/common/browser.ts
@@ -1191,6 +1191,7 @@ export class AttUI {
   private uploader: any = undefined;
   private attAddedCb?: (r: Att) => Promise<void>;
   private attRemovedCb?: () => void;
+  private initializedCb?: () => void;
 
   constructor(getLimits: () => Promise<AttLimits>) {
     this.getLimits = getLimits;
@@ -1212,6 +1213,9 @@ export class AttUI {
         },
       };
       this.uploader = new qq.FineUploader(config); // tslint:disable-line:no-unsafe-any
+      if (this.initializedCb) {
+        this.initializedCb();
+      }
     });
   }
 
@@ -1257,6 +1261,23 @@ export class AttUI {
 
   clearAllAtts = () => {
     this.attachedFiles = {};
+  }
+
+  setAtts = (atts: Array<Att>) => {
+    const files: Array<File> = [];
+    for (const att of atts) {
+      const data = att.getData() as Uint8Array;
+      files.push(new File([data], att.name, { type: att.type }));
+    }
+    const addFiles = () => {
+      this.uploader.reset(); // tslint:disable-line:no-unsafe-any
+      this.uploader.addFiles(files); // tslint:disable-line:no-unsafe-any
+    };
+    if (this.uploader) {
+      addFiles();
+    } else {
+      this.initializedCb = addFiles;
+    }
   }
 
   private cancelAtt = (uploadFileId: string) => {
@@ -1309,7 +1330,6 @@ export class AttUI {
       reader.readAsArrayBuffer(this.attachedFiles[uploadFileId]);
     });
   }
-
 }
 
 /*
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -430,7 +430,9 @@ export class Composer {
       try {
         this.S.cached('send_btn_note').text('Saving');
         const primaryKi = await this.app.storageGetKey(this.urlParams.acctEmail);
-        const encrypted = await PgpMsg.encrypt({ pubkeys: [primaryKi.public], data: Buf.fromUtfStr(this.extractAsText('input_text')), armor: true }) as OpenPGP.EncryptArmorResult;
+        const data = Buf.fromUtfStr(this.extractAsText('input_text'));
+        const atts = await this.attach.collectAtts();
+        const encrypted = await PgpMsg.encrypt({ pubkeys: [primaryKi.public], data, armor: true }) as OpenPGP.EncryptArmorResult;
         let body: string;
         if (this.urlParams.threadId) { // reply draft
           body = `[cryptup:link:draft_reply:${this.urlParams.threadId}]\n\n${encrypted.data}`;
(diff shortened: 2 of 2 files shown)


michael-volynets:
@tomholub Perhaps would be better to not save attachments larger than 500KB?
What do you think?

tomholub:
I'll have to test it out today to see exactly what do you mean.

tomholub:
I have tested it today and the composer seemed to function as usual while it was sending.

However I did notice the browser get slow and choppy. That's weird, I would have to investigate it. I was using a 4 mb file.

But one error you could look into:

1) I added first 4mb file
3) added second 4 mb file
4) started editing subject, and it started showing `Not saved (error)`

```
[Request Entity Too Large: 413 when PUT-ing https://www.googleapis.com/gmail/v1/users/me/drafts/r7592961122171182584 string: message]
Google.gmailCall@moz-extension://feb74b72-561a-4474-a9d3-8ffafde0330a/js/common/api/google.js:52:16


url: moz-extension://feb74b72-561a-4474-a9d3-8ffafde0330a/chrome/elements/compose.htm?frameId=frame_uZuxWBeYcu&isReplyBox=___cu_false___&placement=gmail&acctEmail=flowcrypt.compatibility%40gmail.com&parentTabId=1%3A0


provided ajax call stack:
Google.gmailCall@moz-extension://feb74b72-561a-4474-a9d3-8ffafde0330a/js/common/api/google.js:52:16


url: moz-extension://feb74b72-561a-4474-a9d3-8ffafde0330a/chrome/elements/compose.htm?frameId=frame_uZuxWBeYcu&isReplyBox=___cu_false___&placement=gmail&acctEmail=flowcrypt.compatibility%40gmail.com&parentTabId=1%3A0

    at <async> ajax (/extension/js/common/api/api.ts:618:23)
    at <async> apiGoogleCallRetryAuthErrorOneTime (/extension/js/common/api/google.ts:557:53)
    at <async> gmailCall (/extension/js/common/api/google.ts:46:28)
```

image

tomholub:
`Request entity too large` is a weird error when I'm just editing the subject. Is it trying to upload the attachments again each time? It should only need to upload an attachment one time, and then I should be able to edit the subject without editing the attachment.

michael-volynets:
Yes, I think it uploads each time because when I send the request to the Gmail API to save draft I paste there Mime message with all attachments in it. 

       const mimeMsg = await Mime.encode(body, { To: to, From: this.getSender(), Subject: subject }, atts);
        if (!this.urlParams.draftId) {
          const { id } = await this.app.emailProviderDraftCreate(this.urlParams.acctEmail, mimeMsg, this.urlParams.threadId);
          this.S.cached('send_btn_note').text('Saved');
          this.urlParams.draftId = id;
          await this.app.storageSetDraftMeta(true, id, this.urlParams.threadId, to, String(this.S.cached('input_subject').val()));
          // recursing one more time, because we need the draftId we get from this reply in the message itself
          // essentially everytime we save draft for the first time, we have to save it twice
          // currentlySavingDraft will remain true for now
          await this.draftSave(true); // forceSave = true
        } else {
          await this.app.emailProviderDraftUpdate(this.urlParams.draftId, mimeMsg);
          this.S.cached('send_btn_note').text('Saved');
        }

I guess Gmail API should have a solution for these situations, I will check it out.

tomholub:
I thought there was a way to independently edit attachments in a draft. But the more I look into it, the more it looks like maybe there isn't.

Limiting it to small attachments is also problematic, because then the user experience is inconsistent (sometimes attachments get saved, sometimes not).

I don't know of a good way to resolve this. If I add a 20mb attachment and keep typing, I could easily end up uploading 1000mb of data by the time I'm done writing a long message, just to update the message text each time.. that makes no sense.

I'm afraid we'll have to drop this, unless you can find a way to do it seamlessly.

tomholub:
One idea is to prompt user when they want to close the window, if there are attachments, and say `Message was saved as draft without attachments. Add attachments to draft before closing? [yes] [no]`

This would solve new messages but not replies, because it's quite normal to start writing a draft, then change your mind and click on another conversation or inbox etc.. so again it's inconsistent. But maybe better than nothing. I don't know. What do you think?

michael-volynets:
I think it is a good idea if I don't find another way to do it.
But sometimes we can catch the error message ('Request entity too large'), doesn't matter whether we save it before closing or not. 
I don't know yet what should we do in that case even if we save attachments before closing.

tomholub:
Gmail has a special api for sending large messages over 5mb, where you prepend it with `/upload` and change the format a bit, like it's done when sending messages. So that's not a problem.

My bigger worry is that if we end up only doing this when user clicks "close", then we are complicating the codebase for relatively little benefit.

Maybe it's worth it, maybe not - not sure.

michael-volynets:
@tomholub I have tested another email client to see how they resolved this situation,
They don't save a draft while the user is typing a text, they save a draft when you press `CTRL + S` or `CMD + S`, or when you want to close the message there is a dialog  that asks you if you want to save this message to drafts.

If I upload a large file (~15MB) they show this loader:
image because it takes a while.
If I upload a super large file (~40MB) Gmail throws me an error and they just show it to the user:
image

So, I think the best way to do it is to show the dialog that asks the user to save the draft with attachments. If the user clicks `YES` then we will show our loader and after all of that, we can close the compose window if the draft saving process has been finished without errors.

Can I start work on the issue in this way?

tomholub:
I think this would be easier to do if we used IMAP, because I'm pretty sure IMAP allows to edit parts of message independently (maybe your client is doing that if you choose to upload attachment and keep editing content). We could do that in our Android app one day.

Anyway, with Gmail API, since we only can always upload the whole message, the only time that it makes sense to do this is when user is trying to close the window. If we upload attachment before that, any edit will delete the attachments..

 - we should not be prompting the user while he is still composing the message, that is intrusive
 - when saving a draft and user did not add attachments, it should say `Draft saved` when done (now it just says `Saved`). If the user has added attachments, then after auto-save success it should say `Draft saved without attachments`.
 - when user wants to close the window, maybe we should automatically show a modal loader `Uploading draft attachments, please wait... [Save draft without attachments instead]`. If user doesn't do anything, attachments will save and it will close. If user clicks cancel, it will cancel the upload request, and make another one to save draft without attachments, then close window.

This still doesn't fix message replies. If user tries to navigate away, draft attachments will be lost (unsaved). I don't think we can prevent navigating away? And if we can, it may turn out intrusive or potentially clashing with whatever Gmail does.  So this may create another inconsistency.

I'm inclined to wait on this issue. The amount of effort does not seem to pay of for the relatively small benefit - I think there are more pressing issues considering how complicated the result will be. Maybe if we think about it a bit more, we'll come out with something simpler.

tomholub:
I'm closing this until Gmail has a better API

Issue #1648, opened 2019-04-14
recognize prv backup and show details


tomholub:
#1621

Similar to pubkey
allow user to test their pass phrase, say the pass phrase is needed, etc

tomholub:
Michael,

I get many emails similar to:
> How do I open the backup? I tried to open it but don't know how.

To that I respond:

> Hello,
> you don't have to view or open the backup. Feel free to just archive it.
> To verify that your account is backed up properly, please go to FlowCrypt Settings -> Backup. It should say "your account is backed up correctly".
> It is however important to record your pass phrase properly. To verify that you remember it correctly, please go to FlowCrypt Settings -> Security -> Test my pass phrase.
> After that, please note your pass phrase down on a paper. It will be useful if your device ever stops working and you need to get set up on a new one.

-------------

Solution: let's start displaying details about backups, similar to how we recognize and render details about attached public keys.
The work will be somewhere around `replaceAtts`
`pgpNamePatterns` could maybe start returning one big `RegEx` instead of a list of strings. Then `filterAtts` could be simplified a lot.
Example, backups can be recognized with: `(cryptup|flowcrypt)-backup-[a-z]+\.key`
Once you recognize a backup file, you treat it similar to a pubkey. But instead of `pgp_pubkey.htm` frame you would insert `pgp_backup.htm` (a new iframe type).

--------------

The frame will look something like this:
```
This message includes a Public Key backup for some@email.com.

Private key is needed to open your encrypted messages.

Your pass phrase is needed if you ever need import this backup in the future (for example, when setting up FlowCrypt on another device, or when your main device stops working). 

Test that you remember your pass phrase correctly:
[ passphrase _______________  ][ø]   [ TEST PASSPHRASE]  // green btn

KeyWords: SOUP PADDLE MERGE INSIDE SKILL ATTACK // pull from the key
(Fingerprint: 5C42 AD03 B8AA 743A FB13 7C1B CFF3 DA2D BA9C A673) // pull from key

// bottom
```
Then, on the very bottom, it will either say 
```
This key is already imported.
```
or
```
This private key was not imported. We suggest to import all backups so that you can read all incoming encrypted emails.

[IMPORT MISSING PRIVATE KEY] // green btn
```
-------------------------

You can figure out if the key was already imported by extracting primary `longid` from the backup, and trying to find a key with the same `longid` in local private key storage:

```js
const backupKeyLongid = await Pgp.key.longid(backup);
[...]

PR #1699, opened 2019-05-12
Issue 1648 corrections recognize backup

michael-volynets:
#1648


extension/js/common/core/mime.ts
```diff
@@ -71,6 +71,13 @@ export class Mime {
         decoded.signature = decoded.signature || file.getData().toUtfStr();
       } else if (treatAs === 'publicKey') {
         blocks.push(...Pgp.armor.detectBlocks(file.getData().toUtfStr()).blocks);
+      } else if (treatAs === 'backup') {
+        const backupBlocks = Pgp.armor.detectBlocks(file.getData().toUtfStr()).blocks;
+        // Need to set type to the backup because Pgp detects it as a private key
```
tomholub: Thanks for the comment, I was wondering what is going on. Actually, `privateKey` block and `backup` are the same thing. I didn't realize we already had a block named `privateKey`. I will refactor it a bit after merging.

tomholub:
Thanks for fixing this quickly, I appreciate it!

tomholub:
Next time I will be more careful during reviews. That way unwanted change will not get to master, and that way I don't have to hurry you to fix it, and then you can enjoy your Sundays :)

Thanks again.

Issue #802, opened 2017-06-05
pubkey import form should accept just the fingerprint


tomholub:
Eg people may try to paste one of the following:

PGP: 62BF F048 0077 1341 FEB3 B75C 76DC B628 6C3A 87CA
62BF F048 0077 1341 FEB3 B75C 76DC B628 6C3A 87CA
PGP: 62BF F048 0077 1341 FEB3 B75C 76DC B628 6C3A 87CA
0x76dcb6286c3a87ca
0x62BFF04800771341FEB3B75C76DCB6286C3A87CA
76dcb6286c3a87ca
76DCB6286C3A87CA
0x76DCB6286C3A87CA

etc

show a few examples of what should be put in there

This especially affects people new to encryption, because other people just put fingerprints in their emails

tomholub:
- For this use the lookup api eg `GET https://flowcrypt.com/attester/lookup/ADAC279C95093207`
- make sure to convert it to standard format eg 76DCB6286C3A87CA before getting `https://flowcrypt.com/attester/lookup/..` - although this API accepts all sorts of weird and spaced formats, soon it won't.

tomholub:
The form in question is at FlowCrypt Settings -> Additional Settings -> Contacts -> import.

tomholub:
claimed

tomholub:
Available as an interview issue again because the solution above was rejected.

Please don't make the same mistake as the attempt above, which made unnecessary changes too deeply in the internals of the codebase without checking if it breaks any functionality.

The solution to this issue should be relatively simple and contained mostly within the contacts screen. You should be quite certain that what you do does not affect functionality elsewhere in the software.

michael-volynets:
@tomholub Where can I get fingerprints to test it out?

tomholub:
You can fing fingerprints in FlowCrypt Settings -> Additional Settings -> Contacts, where you click to view a public key and it will show fingerprint.

![image](https://user-images.githubusercontent.com/6306961/57703236-30849500-764f-11e9-99d5-38538251e420.png)

`0x76dcb6286c3a87ca` is a longid. On this case, it was created by lowercasing the fingerprint, trimming to last 16 characters, then prefixing `0x`.

As you can see I have listed many different formats of fingerprints and longids, because different software and different people format it differently.

[...]

PR #1705, opened 2019-05-14
Issue 802 accept fingerprint pubkey import

michael-volynets:
#802


michael-volynets:
Thanks, I remembered it
I will try to not repeat this mistake again.

Issue #1701, opened 2019-05-14
reply button icon larger lock


tomholub:
Make the lock much bigger so that it's obvious. Maybe make both the arrow and the lock the same size and put them next to one another - we can try combinations of `[lock][arrow]` or `[arrow][lock]`, depending on what seems more obvious to the user.

![image](https://user-images.githubusercontent.com/6306961/57691799-ac251880-7634-11e9-9623-1b4efadb3b3d.png)

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/1530#issuecomment-492184955_

tomholub:
Before making a PR, it may be better to just produce an edited screenshot of roughly what would the result look like (for the two combinations I mentioned, or another style if you can think of something). When we settle on the design, then you can make a PR. Thanks!

The goal is to make it more obvious to a first time user.

michael-volynets:
I've tried a lot of variants of the icon but if arrow and lock are the same size it doesn't look really good because there isn't a lot of place
to make lock bigger I've done this thing, what do you think?
![image](https://user-images.githubusercontent.com/49104848/57775230-377dd700-7725-11e9-8443-2251ec072a40.png)

michael-volynets:
I used two main colors of Flowcrypt logo

tomholub:
I like the two colors!

That's definitely better, but maybe not quite there yet. I feel like the lock needs to be quite prominent, some people don't have the best sight, some use large res screens..

Something like:

![image](https://user-images.githubusercontent.com/6306961/57776053-11990800-770e-11e9-932c-3509c1076b41.png)

.
.
.

Or:

![image](https://user-images.githubusercontent.com/6306961/57776214-69d00a00-770e-11e9-9ca8-d4d64adae4cb.png)

Does that look ok to you?

I would actually prefer the lock even bigger..

![image](https://user-images.githubusercontent.com/6306961/57776368-c92e1a00-770e-11e9-8b98-691e855da8f5.png)

I don't mind if our icon is wider compared to other icons.

tomholub:
On the screens above I actually squashed the "tail" of the reply icon horizontally, to have lower width, down to about 33% width.

michael-volynets:
What do you think about this one?
![image](https://user-images.githubusercontent.com/49104848/57783289-8ed87300-7736-11e9-8650-a0ea5036cf43.png)

tomholub:
[...]

PR #1706, opened 2019-05-15
issue-1701 change reply icon

michael-volynets:
#1701


tomholub:
It's good to mention the issue number in the issue body as well (notice I edited your post above) so that it would show on the issue page:

image

Without it, github doesn't connect the two.

PR #1714, opened 2019-05-21
Issue 1701 change reply icon

michael-volynets:
#1701


tomholub:
@michael-volynets what does this PR fix? I'd like it here for the record (the issue only mentions design). Thanks!

michael-volynets:
@tomholub There was a bug when there is only one message in a conversation.
The reply button wasn't shown correctly. It was in the center of the message

Here is how it looks like now:
image

tomholub:
Got it, thanks!

Issue #1685, opened 2019-05-07
auth popup needed errors during backup / setup?


tomholub:
not 100% sure if this is auth popup error, but it looks like one. I'm not exactly sure how this could be reproduced, my best guess would be to:

- connect to gmail
- new key
- enter new pass phrases
- abandon the setup
- change password on google
- then come back and click "continue" which will create the backup

----------------
> i am trying to setup flowcrypt on my computer but i am being told that it failed to backup my key. below is a copy of the technical data relating to the error that i am facing.
```
Below are technical details about the error. This may be useful for debugging.

Error string: Error: Unauthorized: 401 when POST-ing https://www.googleapis.com/upload/gmail/v1/users/me/messages/send?uploadType=multipart string: not-a-json

Error stack:     at Function.GoogleAuth.apiGoogleCallRetryAuthErrorOneTime (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/api/google.js:600:42)
    at async Function.Google.gmailCall (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/api/google.js:52:10)
    at async doBackupOnEmailProvider (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/settings/modules/backup.js:219:13)
    at async setupCreateSimpleAutomaticInboxBackup (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/settings/modules/backup.js:353:5)
    at async chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/settings/modules/backup.js:395:6

url: chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/settings/modules/backup.htm?action=setup&acctEmail=seaneoin22%40gmail.com


provided ajax call stack:
    at Function.GoogleAuth.apiGoogleCallRetryAuthErrorOneTime (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/api/google.js:600:42)
    at async Function.Google.gmailCall (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/api/google.js:52:10)
    at async doBackupOnEmailProvider (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/settings/modules/backup.js:219:13)
[...]

PR #1716, opened 2019-05-21
issue-1685 added confirmation modal window

michael-volynets:
#1685 

Perhaps would be better to use `modal.info` function, what do you think?


tomholub:
Yes, looking at the code, if you used `.info` and just informed the user of what will happen, that would simplify it. Otherwise looks good.

Issue #524, opened 2017-02-22
click on reply fist -> click add to contacts later -> gray should refresh


tomholub:

tomholub:
1) receive a public key from a person who does not have public key recorded on Attester
2) click reply, to find that their email shows in gray
3) click "SAVE CONTACT" in the rendered public key above (in their original message)
4) notice the recipient stays gray. It would be great if it could turn green.

I don't know what's a good way to do this, maybe a broadcast browser message using `BrowserMsg` that the reply box could be listening to, and when received, check contacts storage for all recipients currently marked as gray, and re-render the recipients if new pubkey found.

michael-volynets:
@tomholub I can't understand what this code does:
```
        const element = $(this.factory.btnReply()).insertBefore($(lastReplyBtn).children().last());
        if (isEncrypted) {
          element.click(Ui.event.prevent('double', Catch.try(this.setReplyBoxEditable)));
        } else {
          element.click(Ui.event.prevent('double', Catch.try(() => this.replaceStandardReplyBox(true, true)))); // xss-safe-factory
        }
```

I see when the message isn't encrypted but there is a public key file, the reply button (in the top-right corner) doesn't work. 

If I replace the whole `if...else` statement with
```
          element.click(Ui.event.prevent('double', Catch.try(this.setReplyBoxEditable)));
```
everything works well on encrypted and non encrypted messages.

tomholub:
Thanks for asking! The main difference seems to be that it's using `this.replaceStandardReplyBox(true, true)` where the second `true` is `force` (which default to `false`). Which is used here:
```js
const doReplace = Boolean(convoRootEl.find('iframe.pgp_block').filter(':visible').length || (convoRootEl.is(':visible') && force));
```
Without `force` set as true, it would not replace the standard gmail reply box if the conversation has no encrypted elements in it yet (normal plain conversation).

So that would be something for you to double check:
1) open standard plain conversation
2) click encrypted reply
3) if all is good, you can replace it with your version

If you do replace it, please link in the PR to this comment so that I know to test it very thoroughly myself.

tomholub:
Works very well. This will help usability a lot, especially for new users who need to write to someone who uses non-FlowCrypt software.

PR #1737, opened 2019-06-03
issue-524 change recipients color immediatly if they were added to co…

michael-volynets:
…ntacts

#524

Issue #524 
Time: ~4 hrs
It took 2h  for investigating how BrowserMsg works.

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -441,10 +441,10 @@ export class Composer {
   private checkReciepientsKeys = async () => {
     for (const recipientEl of $('.recipients span.no_pgp')) {
       const email = $(recipientEl).text().trim();
-      const res = await this.lookupPubkeyFromDbOrKeyserverAndUpdateDbIfneeded(email);
-      if (res !== 'fail') {
+      const [dbContact] = await this.app.storageContactGet([email]);
+      if (dbContact) {
         $(recipientEl).removeClass('no_pgp');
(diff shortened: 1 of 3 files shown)


tomholub:
Please also check email I've sent you a few days ago, to remember to note how much time each PR took. Thanks!

michael-volynets:
@tomholub  Yes, I've done that

extension/js/common/composer.ts
```diff
@@ -436,6 +438,17 @@ export class Composer {
     }
   }
 
+  private checkReciepientsKeys = async () => {
+    for (const recipientEl of $('.recipients span.no_pgp')) {
+      const email = $(recipientEl).text().trim();
+      const res = await this.lookupPubkeyFromDbOrKeyserverAndUpdateDbIfneeded(email);
```
tomholub: Here I recommend to check db directly, without checking attester. After all, the imported key should by now already be in the local db.

(as implemented now, If you have 3 recipients who are all gray, and add one pubkey, two of them will be re-checked on Attester unnecessarily)

Something like (please test):

```js
const [dbContact] = await this.app.storageContactGet([email]);
if(dbContact) {
  $(recipientEl).removeClass('no_pgp');
  await this.renderPubkeyResult(recipientEl, email, dbContact);
}
```

tomholub:
> It took 2h for investigating how BrowserMsg works.

It sometimes takes me longer than that, especially when I need to make changes to it :smile:  Thanks for the note.

Issue #1733, opened 2019-05-31
hitting escape should dismiss passphrase dialog in gmail


tomholub:
![image](https://user-images.githubusercontent.com/6306961/58695255-75653700-8384-11e9-8a62-1190329eb482.png)

https://flowcrypt.com/docs/guide/common-tasks/require-pass-phrase.html

michael-volynets:
Maybe would be better to use here featherlight-dialog? What do you think?

tomholub:
You got me thinking. Why not. On the other hand, now I'm thinking to maybe stop using featherlight in the settings and start doing what we do in Gmail :laughing:  

I like featherlight. But I keep getting errors reports from it, and I don't feel like debugging it.

https://github.com/FlowCrypt/flowcrypt-browser/issues/1625
https://github.com/FlowCrypt/flowcrypt-browser/issues/432

Although it's not a big dependency, it's yeat another 650 lines of code, and all we really need is just an iframe in a box (already have), a close button on top right (easy to add) and an escape handler (a few lines). We could reuse some of the CSS from featherlight.

We have to review all of our code for security (including dependencies) - the fewer code to review, the better, so now I'm thinking to potentially remove featherlight in the future.

michael-volynets:
If you want to remove featherlight at all I think we need to create our own dialogs structure.
So now, I'm going to add `ESC` handler for the passphrase dialog

tomholub:
What do you mean with "our own dialogs structure"?

PR #1745, opened 2019-06-07
issue-1733 Added ESC handler and close dialog when user clicked on th…

michael-volynets:
…e background

Issue #1733 
Time: 6 hrs

I got found the issue when some user firstly clicks on the background of dialog and then hitting the ESC button  The handler doesn't work because IFrame isn't on the focus. so I had to find another solution and it's to close Dialog when the user clicks on the background

### DIFF
--- a/extension/chrome/elements/passphrase.ts
+++ b/extension/chrome/elements/passphrase.ts
@@ -76,10 +76,16 @@ Catch.try(async () => {
     $('#passphrase').focus();
   };
 
-  $('.action_close').click(Ui.event.handle(() => {
+  const closeDialog = () => {
     BrowserMsg.send.passphraseEntry('broadcast', { entered: false });
     BrowserMsg.send.closeDialog(parentTabId);
-  }));
+  };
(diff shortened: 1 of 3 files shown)


extension/chrome/elements/passphrase.ts
```diff
@@ -101,8 +107,7 @@ Catch.try(async () => {
       }
     }
     if (atLeastOneMatched) {
-      BrowserMsg.send.passphraseEntry('broadcast', { entered: true });
-      BrowserMsg.send.closeDialog(parentTabId);
+      closeDialog();
```
tomholub: Notice that entered: true was now replaced with entered: false (meaning pass phrase was not entered). The iframe that is waiting for the passphrase may get confused. This should be tested.

michael-volynets: Ohh, Thanks, I missed that

tomholub:
Please don't close referenced issues yet, i want to close them myself after thorough testing. Thanks!

Issue #1667, opened 2019-04-22
signature verification from several non-FlowCrypt software senders


tomholub:
All signed emails from Kraken fail verification. Seems to work for Thunderbird.

tomholub:
Another one from outlook
>  I sent a SIGNED email from outlook to gmail, and when I viewed it with gmail, flowcrypt said the signature isn't valid.  I send signed emails frequently and I know my Outlook is signing them properly.  I checked to make sure flowcrypt had the correct public key, and it did. While searching for an explaination, I stumbled across an article that said flowcrypt can't send emails that are signed AND encypted yet, and this made me realize that flowcrypt isn't usable for me yet.  It's on the right track though!

Another regular problems from GPG Suite:
> I have been testing your Flowcrypt extension (quite excellent!) and have noticed that when message are validly signed by the GPGTools GPG Suite using Apple Mail on Mojave, your plugin shows those messages as failing the signature check. I have used Apple Mail and GPG Suite to sign this message; I suspect it will show a failed signature. Not sure what the issue is.
> Signing messages with Flowcrypt does not cause this issue, nor does sending PGP messages using Protonmail’s built in PGP ability. ProtonMail’s built in PGP validly decodes and recognizes the signature. It seems to just be something to do with Apple Mail and GPG Suite and your plugin specifically.

tomholub:
Lastly, ProtonMail. See in `flowcrypt.compatibility` email:
 - `signed message from proton mail containing 1234, plain text`
 - `signed message from proton mail containing 1234, html`

tomholub:
Enigmail also failing. #1736

tomholub:
The problem is that FlowCrypt is verifying the text itself, while the signatures are actually over a mime section.

Eg it will try to verify `1234`, when it should be verifying:
```
Content-Type: multipart/mixed;boundary=---------------------453436af916092d2719b9101113b6353

-----------------------453436af916092d2719b9101113b6353
Content-Transfer-Encoding: quoted-printable
Content-Type: text/plain;charset=utf-8

1234
-----------------------453436af916092d2719b9101113b6353--
```
For context, whole message:

```
Delivered-To: flowcrypt.compatibility@gmail.com
Received: by 2002:ac8:87d:0:0:0:0:0 with SMTP id x58csp566979qth;
        Thu, 9 May 2019 03:08:15 -0700 (PDT)
[...]

PR #1750, opened 2019-06-11
fixed detached signatures

tomholub:
#1667 

~ 8 hrs

Took about as long as I'd expect, signature verification can be tricky to debug.


tomholub:
- remember to submit fixed types

PR #1755, opened 2019-06-15
issue #1754 Corrected styles for reply box, set height 100% if it isn…

### LINKED ISSUE #1751: attachments in the reply box aren't shown correctly
When you click on `Ask sender to update` the attachments in the reply box doesn't align correctly.

image

It could have happened because of flex-box. A while ago we changed styles for showing attachments and it could have become a reason for this bug

michael-volynets:
…'t reply box

Time Spent: 2h 30m

fixed without any issues

Also, in this pull request, I've fixed #1751


tomholub:
References #1754

tomholub:
Sorry, this was messy - next time I will rebase it instead.

Issue #1691, opened 2019-05-09
prevent window closing when message is being sent


tomholub:
1) compose message
2) add a large attachment
3) click send
4) message is being sent slowly.. now click `X` to close the secure compose window

The window will go away, which is bad.

It would be better to show a confirmation modal that says

> A message is currently being sent. Closing the compose window may abort sending the message.
> 
> Abort sending?
> 
> [yes] [no]

PR #1765, opened 2019-06-20
issue #1691 check if the message is being sent and show confirm modal

michael-volynets:
Issue #1691 

Time Spent: 1h

Done without any issues

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1647,12 +1647,8 @@ export class Composer {
       this.renderSenderAliasesOptionsToggle();
     } else {
       $('.close_new_message').click(Ui.event.handle(async () => {
-        if (!this.isSendMessageInProgress) {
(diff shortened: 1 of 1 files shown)


extension/js/common/composer.ts
```diff
@@ -1641,7 +1646,15 @@ export class Composer {
       }
       this.renderSenderAliasesOptionsToggle();
     } else {
-      $('.close_new_message').click(Ui.event.handle(() => this.app.closeMsg(), this.getErrHandlers(`close message`)));
+      $('.close_new_message').click(Ui.event.handle(async () => {
+        if (!this.isSendMessageInProgress) {
```
tomholub: Looks good, could be simplified as:
```js
if (!this.isSendMessageInProgress || await Ui.modal.confirm('A message is currently being sent. Closing the compose window may abort sending the message.\nAbort sending?')) {
  this.app.closeMsg();
}
```

michael-volynets: I've changed it. 
I didn't do it because I a bit don't like long lines but if it's okay for you I will try to write the code this way.

Issue #1717, opened 2019-05-21
async pull list of email aliases from Gmail every time


tomholub:
Currently, if user installs FlowCrypt and then adds an alias later, they will have to manually go to settings and have it refreshed for the alias to show in compose window.

It would be better if the extension did it automatically (in the background) each time I open secure compose window.

If it does find a new alias, or an alias was removed, it should save the new info to storage and a `modal.confirm` should show saying `Your email aliases on Gmail have refreshed since the last time you used FlowCrypt. Reload the compose window now?`.

I would not worry about preserving state or the draft, because in 99% of situations this will happen within 1 second of when the window opened, probably before the user has really done anything with it.

tomholub:
Of course, after reloading, it should be able to show the new aliases from the storage right away.

michael-volynets:
I don't exactly understand what is an alias and how I can test it? 
Could you please explain me a little more.
Thanks

tomholub:
See Gmail Settings -> Accounts.

![image](https://user-images.githubusercontent.com/6306961/59882665-6598ab80-93a2-11e9-924d-32e5e6b615f3.png)

![image](https://user-images.githubusercontent.com/6306961/59882692-76492180-93a2-11e9-9b15-d6478d902c4d.png)

All of these are alternative send-from addresses. So I can have an account `acct@gmail.com` but send emails as `something@domain.com` for example.

The it lets me choose the sending address:

![image](https://user-images.githubusercontent.com/6306961/59882781-b1e3eb80-93a2-11e9-803a-641f4394b05e.png)

The issue is, if I just added a new email alias, and open secure compose, it will not notice and not show the new alias. You would have to go to FlowCrypt Settings -> Additional -> Experimental -> Refresh aliases, and after that open secure compose.

That's not a very good ux, so when someone adds or removes an alias, it would be better if the compose window noticed.

For this you should:
 - when secure compose is opened, async check aliases on gmail api
 - if a different set of aliases is found, save them to storage (`addresses`), then reload the window
 - if call fails, ignore it

michael-volynets:
Got it.
Thanks! It will help me to work on this issue!

PR #1766, opened 2019-06-21
Issue 1717 pull list email aliases

michael-volynets:
Issue #1717 

Time Spent: 5h

is it okay if I'm reloading composer this way?:
```javascript
          this.app.closeMsg();
          BrowserMsg.send.openNewMessage(this.urlParams.parentTabId);
```
Or there is a better way?

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1647,12 +1647,8 @@ export class Composer {
       this.renderSenderAliasesOptionsToggle();
     } else {
       $('.close_new_message').click(Ui.event.handle(async () => {
-        if (!this.isSendMessageInProgress) {
+        if (!this.isSendMessageInProgress || await Ui.modal.confirm('A message is currently being sent. Closing the compose window may abort sending the message.\nAbort sending?')) {
           this.app.closeMsg();
-        } else {
-          if (await Ui.modal.confirm('A message is currently being sent. Closing the compose window may abort sending the message.\nAbort sending?')) {
-            this.app.closeMsg();
-          }
         }
       }, this.getErrHandlers(`close message`)));
       $('.minimize_new_message').click(Ui.event.handle(this.minimizeComposerWindow));
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -151,6 +151,8 @@ export class Composer {
   private refBodyHeight?: number;
   private urlParams: ComposerUrlParams;
 
+  private isSendMessageInProgress = false;
+
   constructor(appFunctions: ComposerAppFunctionsInterface, urlParams: ComposerUrlParams, initSubs: Subscription) {
     this.attach = new AttUI(() => this.getMaxAttSizeAndOversizeNotice());
     this.app = appFunctions;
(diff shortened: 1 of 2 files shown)


tomholub:
> Issue #1717
> 
> Time Spent: 5h
> 
> is it okay if I'm reloading composer this way?:
> 
> ```js
>           this.app.closeMsg();
>           BrowserMsg.send.openNewMessage(this.urlParams.parentTabId);
> ```
> 
> Or there is a better way?

I would just do `window.location.reload()`. It will reload the frame. I think that's better.

extension/js/common/composer.ts
```diff
@@ -1707,6 +1718,20 @@ export class Composer {
     }
   }
 
+  private async checkEmailAliases() {
+    if (!this.urlParams.isReplyBox) {
+      const addresses = Value.arr.unique((await Settings.fetchAcctAliasesFromGmail(this.urlParams.acctEmail)).concat(this.urlParams.acctEmail));
+      const storedAdresses = (await Store.getAcct(this.urlParams.acctEmail, ['addresses'])).addresses!;
```
tomholub: A bit dangerous. Type of `(await Store.getAcct(this.urlParams.acctEmail, ['addresses'])).addresses` is `string[] | undefined`. You don't know which one it is unless you test it. Maybe you got `string[]` when you were testing it, but what if sometimes you get `undefined`?

A safer way to deal with this is to add `|| []` instead of `!`. This way you know that either you got an array, or you got undefined but will work with an empty array. Then below the undefined would get fixed after it pulls from Gmail.

extension/js/common/composer.ts
```diff
@@ -1707,6 +1718,20 @@ export class Composer {
...
+      if (addresses.sort().join() !== storedAdresses.sort().join()) { // This way of comparation two arrays works only for not object arrays
```
tomholub: If the forced `!` expression above was actually undefined, here you would get an error.

extension/js/common/composer.ts
```diff
@@ -1707,6 +1718,20 @@ export class Composer {
...
+        await Store.setAcct(this.urlParams.acctEmail, { addresses });
+        if (await Ui.modal.confirm('Your email aliases on Gmail have refreshed since the last time you used FlowCrypt.\nReload the compose window now?')) {
+          this.app.closeMsg();
+          BrowserMsg.send.openNewMessage(this.urlParams.parentTabId);
```
tomholub: `window.location.reload()` will work better instead of these two lines.

extension/js/common/composer.ts
michael-volynets: Okay, got it, I've corrected it. 
I did it that way because I thought that there is always should be at least one element.

tomholub: There should be. But who knows, maybe a bug in other code could cause it to be undefined.

extension/js/common/composer.ts
michael-volynets: Oh, Yes, it is definitely better. 
I forgot that after `location.reload` the whole window won't reload, it reloads the only iframe
Thanks!

tomholub:
The test fails are not your fault

michael-volynets:
Has something happened with CI?

tomholub:
Google oauth screen acting up, I'm on it

Issue #1646, opened 2019-04-12
showing reply info tag in sent reply pwd msg


tomholub:
1) reply to someone who has not installed flowcrypt
2) use password protection
3) be on a paid plan
4) send and wait for the sent message to render:

![image](https://user-images.githubusercontent.com/6306961/56051054-059bde00-5d3d-11e9-9046-ec0ec9c2ba34.png)

It should not be included in the send success render

michael-volynets:
Can you please explain to me what this function does?
What is the reply token and why it's used for?

```TypeScript
 private addReplyTokenToMsgBodyIfNeeded = async (recipients: string[], subject: string, plaintext: string, challenge: Pwd | undefined, subscription: Subscription): Promise<string> => {
    if (!challenge || !subscription.active) {
      return plaintext;
    }
....
```

tomholub:
"challenge" is another word for message password (a challenge password). Used when emailing to users who don't have any public key. 

Recipient can open such messages online.

We have functionality that allows the recipient to reply. For this, the webpage needs a reply token (which is then passed to our server. This way only someone who can decrypt a message can then reply to it). 

This functionality is only available to paying users, thus the "subscription.active". 

The problem of this issue is that when this gets induced, sender ends up immediately seeing it in the sent reply, which is weird.

If you then reload the convo, I think it doesn't show anymore. So we have some functionality that strips it out, which you can use for rendering sent message too. 

Let me know if I should clarify further

michael-volynets:
I got it,
Thanks for the explanation!

PR #1780, opened 2019-06-27
issue #1646 Remove info tag from the text that will be shown in reply…

michael-volynets:
… message

Issue #1646 

Time Spent: 3h
Fixed without problems. Just needed to clarify some things I asked in the task.

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1807,9 +1807,7 @@ export class Composer {
 
   // This function is needed for removing things that don't need to be displayed in the replyed message
   // we have only div element with 'cryptup-reply'class
-  private formatTextForReply = (plainText: string): string => {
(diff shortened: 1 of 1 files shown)


extension/js/common/composer.ts
```diff
@@ -1570,6 +1570,7 @@ export class Composer {
     this.S.cached('compose_table').css('display', 'none');
     this.S.cached('reply_msg_successful').find('div.replied_from').text(this.getSender());
     this.S.cached('reply_msg_successful').find('div.replied_to span').text(msg.headers.To.replace(/,/g, ', '));
+    plaintext = this.formatTextForReply(plaintext);
```
tomholub: You can use `PgpMsg.stripFcTeplyToken` instead.

michael-volynets: oh, I didn't know there is that function

PR #1781, opened 2019-06-27
issue #1575 Added ability to paste pub key directly in recipients box

michael-volynets:
Issue #1575 

Time Spent: 3h
Done without issues because all your comments in the task helped a lot

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1632,6 +1632,9 @@ export class Composer {
           return; // key is invalid
         }
         const { keys: [key] } = await Pgp.key.parse(normalizedPub);
+        if (!key.users.length) { // there can be no users
+          return;
+        }
         const keyUser = Str.parseEmail(key.users[0]);
         if (!await Store.dbContactGet(undefined, [keyUser.email])) {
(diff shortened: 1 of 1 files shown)


extension/js/common/composer.ts
```diff
@@ -1620,6 +1621,24 @@ export class Composer {
...
+        const textData = event.originalEvent.clipboardData.getData('text/plain');
+        const keyImportUi = new KeyImportUi({ checkEncryption: true });
+        let normalizedPub: string;
+        try {
+          normalizedPub = await keyImportUi.checkPub(textData);
+        } catch (e) {
+          return; // key is invalid
+        }
+        const { keys: [key] } = await Pgp.key.parse(normalizedPub);
+        const keyUser = Str.parseEmail(key.users[0]);
```
tomholub: Everything looks done very well. The only issue is that some keys may be lacking users in it. We don't support such keys. But in this particular case, you would probably get an exception because `key.users[0]` may be undefined. Therefore I would add one more check and return early if you find `!key.users.length`.

michael-volynets: Okay, Got it, thanks.

I thought there is always should be a user.

michael-volynets:
I've checked the failed test and the reason is:
`set up global browsers and config Rejected promise returned by test`

Is this my change triggered it?

tomholub:
Nope, you're good. I'll debug the tests later.

PR #1782, opened 2019-06-27
refactor parseEmail to validate email

tomholub:
#1779 
2 hrs
Used in more places than I thought


tomholub:
+1 hr debugging issues

Issue #1239, opened 2018-08-17
support encrypted subjects by enigmail


tomholub:
> I've just noticed the new Thunderbird/Enigmail settings for replacing the Subject line with "Subject: Encrypted Message" is not supported by FlowCrypt.

> Any chance of that happening? It's now really hard to follow conversations as they all have the same Subject line under FlowCrypt

That sounds like a good thing to have, I'll keep you posted as we investigate this. I won't be able to actually get it done before September though.

I'd imagine Gmail will group them in the same conversation regardless of what we do. We could, however, show the subject on top of each block of encrypted text in such situations. That may somewhat help.

> Yes I was wondering about that too. Having the decrypted subject in the body would help

> I'd assume you could fix this 'properly' in your android mail app, but chrome is a tad harder

tomholub:
We already know how to extract this:
```js
            } else if (Mime.getNodeType(node) === 'text/rfc822-headers') {
              // todo - surface and render encrypted headers
            } else {
```
So it's just a matter of finding out how to render these. I would just render it inside `pgp_block`, maybe in bold and with a line separating the subject and the text.

tomholub:
Send yourself a message from Enigmail with encrypted subject to test this.

tomholub:
The result will probably be that on Gmail, it will still show `Encrypted Message` on top of the conversation, and then on top of each `pgp_block` message will be a subject that was extracted from within the encrypted message. I think that would be an ok result.

tomholub:
Very roughly, something like this

![image](https://user-images.githubusercontent.com/6306961/60109704-5c209200-975a-11e9-862e-8d9837006327.png)

PR #1784, opened 2019-06-28
issue #1239 Add subject to MimeContent, added method to create subjec…

michael-volynets:
…t html content

Issue #1239 

Time Spent: ~2h 30m

Done without any problems but I don't know if I correctly get Subject line in `mime.ts` file. 
Maybe there is a better way to do it

### DIFF
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -307,12 +307,22 @@ Catch.try(async () => {
     } else {
       renderText('Formatting...');
       const decoded = await Mime.decode(decryptedBytes);
+      decryptedContent = '';
+      if (decoded.subject) {
+        decryptedContent = createEncryptedSubject(decoded.subject);
+        isHtml = true;
+      }
       if (typeof decoded.html !== 'undefined') {
-        decryptedContent = decoded.html;
+        decryptedContent += decoded.html;
         isHtml = true;
       } else if (typeof decoded.text !== 'undefined') {
-        decryptedContent = decoded.text;
+        if (isHtml) {
+          decryptedContent += Xss.htmlSanitizeAndStripAllTags(decoded.text, '\n');
+        } else {
+          decryptedContent += decoded.text;
+        }
       }
+
       for (const att of decoded.atts) {
         if (att.treatAs() !== 'publicKey') {
           renderableAtts.push(att);
@@ -483,6 +493,13 @@ Catch.try(async () => {
       }
     }
   };
+
+  const createEncryptedSubject = (subject: string) => {
+    return `<div style="font-size: 14px;"> Encrypted Subject:
+              <b> ${subject}</b>
+            </div>
+            <hr/>`;
+  }
 
   const appendCollapsedQuotedContentButton = (message: string, isHtml: boolean = false) => {
     const pgpBlk = $("#pgp_block");
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -318,9 +318,9 @@ Catch.try(async () => {
       }
       for (const att of decoded.atts) {
         if (att.treatAs() !== 'publicKey') {
-          renderableAtts.push(att)
+          renderableAtts.push(att);
         } else {
-          publicKeys.push(att.getData().toUtfStr())
+          publicKeys.push(att.getData().toUtfStr());
         }
       }
     }
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -317,7 +317,7 @@ Catch.try(async () => {
         isHtml = true;
       } else if (typeof decoded.text !== 'undefined') {
         if (isHtml) {
-          decryptedContent += Xss.htmlSanitizeAndStripAllTags(decoded.text, '\n');
+          decryptedContent += Xss.escape(decoded.text);
         } else {
           decryptedContent += decoded.text;
         }
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -499,7 +499,7 @@ Catch.try(async () => {
               <b> ${subject}</b>
             </div>
             <hr/>`;
-  }
+  };
 
   const appendCollapsedQuotedContentButton = (message: string, isHtml: boolean = false) => {
     const pgpBlk = $("#pgp_block");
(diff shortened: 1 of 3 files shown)


tomholub:
Can you please also share a screen shot? Thanks!

michael-volynets:
Sure, I forget to attach it
image

Also, tslint tests were failed. Wait a minute I'll fix them

michael-volynets:
Fixed

extension/js/common/core/mime.ts
```diff
@@ -156,7 +157,10 @@ export class Mime {
             } else if (Mime.getNodeType(node) === 'text/plain' && !Mime.getNodeFilename(node)) {
               mimeContent.text = Mime.getNodeContentAsUtfStr(node);
             } else if (Mime.getNodeType(node) === 'text/rfc822-headers') {
-              // todo - surface and render encrypted headers
+              const subjectsLines = Mime.getNodeContentAsUtfStr(node).match(/(?<=Subject:\s).*/gi);
```
tomholub: There should be `node.headers` property or something similar. Maybe `node.headers.Subject[0]`

michael-volynets:
I think it's incorrect what I'm doing here:
```TS
decryptedContent += Xss.htmlSanitizeAndStripAllTags(decoded.text, '\n');
```

It will remove the text which is incorrect.

extension/chrome/elements/pgp_block.ts
```diff
@@ -484,6 +494,13 @@ Catch.try(async () => {
     }
   };
 
+  const createEncryptedSubject = (subject: string) => {
+    return `<div style="font-size: 14px;"> Encrypted Subject:
+              <b> ${subject}</b>
+            </div>
+            <hr/>`;
```
tomholub: Please give the `div` a `border-bottom: 1px cacaca` (or similar color) instead. The hr looks ugly without further formatting.

(also, since it's a tiny bit of html, I'd put it on one line)

michael-volynets:
Now it works correctly

extension/js/common/core/mime.ts
michael-volynets: No, unfortunately,  there isn't the subject header. There are only 'Content-Type' and 'Content-Disposition headers'.

michael-volynets: But I found it in _parentNode. 
image

I will try to access to it

extension/chrome/elements/pgp_block.ts
michael-volynets: I tried border bottom but it didn't work because we call `await renderContent(message.html(), false);`  and it sanitizes  HTML and removes the border. 
I think I can add border somehow but it will complicate the codebase

michael-volynets: I think I can add `encryptedSubject` parameter to this function 
```TS 
const separateQuotedContentAndRenderText = async (decryptedContent: string, isHtml: boolean)
```
 and in the end just prepend to the rendered message.
I think it won't complicate the codebase a lot. What do you think?

tomholub: Got it. We can, for now, leave the `<hr>` but I have to look into the sanitization. It should not be stripping away css styles. I'll make an issue for myself to look at it.

tomholub: > I think I can add `encryptedSubject` parameter to this function
> 
> ```ts
> const separateQuotedContentAndRenderText = async (decryptedContent: string, isHtml: boolean)
> ```
> 
> and in the end just prepend to the rendered message.
> I think it won't complicate the codebase a lot. What do you think?

It doesn't really matter where you put it as long as it's minimal and easy to follow.

Any code that is minimal and easy to follow is good code.

Go ahead & try it & evaluate it & see.

michael-volynets: Okay

michael-volynets:
I haven't found a really good way to simplify it yet. I just changed on `if` expression to ternary operator. 

I think we need to divide this function into some other functions, it should simplify it.
I can create another task for that.

extension/js/common/core/mime.ts
```diff
@@ -156,7 +157,9 @@ export class Mime {
             } else if (Mime.getNodeType(node) === 'text/plain' && !Mime.getNodeFilename(node)) {
               mimeContent.text = Mime.getNodeContentAsUtfStr(node);
             } else if (Mime.getNodeType(node) === 'text/rfc822-headers') {
-              // todo - surface and render encrypted headers
+              if (node._parentNode && node._parentNode.headers.subject) {
```
tomholub: I'm not 100% sure here if `node._parentNode.headers.subject.length` should also be checked? Else you may get `undefined.value`

If you see other code that doesn't, than it's probably ok.

extension/chrome/elements/pgp_block.ts
```diff
@@ -313,12 +313,12 @@ Catch.try(async () => {
...
         decryptedContent = decoded.text;
       }
+      if (decoded.subject) {
+        decryptedContent = getEncryptedSubjectText(decoded.subject, isHtml) + decryptedContent;
+      }
       for (const att of decoded.atts) {
-        if (att.treatAs() !== 'publicKey') {
-          renderableAtts.push(att);
-        } else {
+        att.treatAs() !== 'publicKey' ? renderableAtts.push(att) :
```
tomholub: I like the succinctness, however please try to avoid ternary operators when not working with the return value, eg:

```js
// not good, traditional `if` is better, even though it takes up more space
x ? doOne() : doTwo();

// good
return x ? doOne() : doTwo();

// good
const y = x ? doOne() : doTwo();

// good
doSomething(x ? one() : two());
```

I checked the style guide and it doesn't mention it. Sorry about that! Please feel free to updated with my samples above: https://github.com/FlowCrypt/flowcrypt-browser/wiki/FlowCrypt-TypeScript-Style-Guide

tomholub: It would be great to have a tslint for this, but I'm not sure there is one.

michael-volynets: oh, okay, got it. I'll try to not make this mistake again. 
Thanks!

extension/js/common/core/mime.ts
michael-volynets: yes, there is the same code above:
```TS
          for (const name of Object.keys(parser.node.headers)) {
            mimeContent.headers[name] = parser.node.headers[name][0].value;
          }
```

michael-volynets: Should I add `if` expression to both?

tomholub: I think not. If it worked well until now, it should work well the way you wrote it too. Thanks!

Issue #873, opened 2017-07-07
quote previous message / reply


tomholub:
#1621

https://mail.google.com/mail/u/1/#inbox/15d12319eadac1f4
relies on #92

tomholub:
https://mail.google.com/mail/u/1/#sent/15f07b45ccf9ca34

tomholub:
ref https://mail.google.com/mail/u/0/#inbox/1606516537801891

tomholub:
@michael-volynets You can do similar thing that we do for replies, with a little more standard intro line:

- Html content should be converted to text before being included in the quoted part
- hen quoted with `On 2019-02-20 at 19:48 someone@hello.com wrote:`
- and `>` on each line below

tomholub:
Also, you should probably reuse the UX from quoted messages (`pgp_block`) on the bottom of the reply window as well, to mimic what Gmail is doing:

![image](https://user-images.githubusercontent.com/6306961/60113600-a9ecc880-9761-11e9-84ec-1a8148e84361.png)

Clicking it will make it go away, and then I see the quoted content which I can remove or edit if I want.

michael-volynets:
just to make sure,
If I correctly understood, I need to add 3 dots to the bottom of the reply box and after clicking on it, it should show previous message (with header line you wrote and ) that can be edited? (Like Gmail does)

michael-volynets:
@tomholub I should let you know that this issue will take more time than I expected.
Because I need to handle replied messages along with drafts. 
For Example, if I clicked reply button then the draft shouldn't be saved immediately, it should be saved only when I clicked on those 3 dots or typed something, and it should be saved along with the quoted replied message in the end. 
Then when I opened a draft again, I need to check if there is a replied part, if yes - I need to hide it and show 3 dots.
There also can be a case, when I have a draft but there isn't replied part (I just removed it), in this case, I don't need to show 3 dots.
Or when the replied message has no text (e.g. it has attachments only)
I hope you understood what I mean.

I tested how Gmail does it. and basically, I understand what I should do that.

This really interesting task, thanks for assigning :)

michael-volynets:
But I think there is another easier way, is to do this similar to forwarding messages (without hiding). 
When the user clicks reply then the replyed message is immediately appended in quoted text with header line to the reply window. (without 3 dots).

This way will be easier and faster but without hiding replied content. 
I think the first way (with hiding) is better for UX but it can take much more time than without hiding.
[...]

PR #1791, opened 2019-07-02
Issue #873 quote previous reply

tomholub:
I had to create a PR so that I can make a review with comments, it's clearer that way
#873

### DIFF
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -83,7 +83,7 @@ Catch.try(async () => {
     draftId = storage.drafts_reply[threadId]; // there may be a draft we want to load
   }
 
-  const processedUrlParams = { acctEmail, draftId, threadId, threadMsgId, subject, from, to, frameId, tabId, isReplyBox, skipClickPrompt, parentTabId, disableDraftSaving, debug };
+  const processedUrlParams = { acctEmail, draftId, threadId, subject, from, to, frameId, tabId, isReplyBox, skipClickPrompt, parentTabId, disableDraftSaving, debug };
   const closeMsg = () => {
     $('body').attr('data-test-state', 'closed'); // used by automated tests
     if (isReplyBox) {
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -202,13 +202,13 @@ Catch.try(async () => {
     emailProviderDetermineReplyMsgHeaderVariables: async (progressCb?: ProgressCb) => {
       try {
         const thread = await Google.gmail.threadGet(acctEmail, threadId, 'full', progressCb);
-        if (thread.messages && thread.messages.length > 0) {
-          const threadMsgIdLast = Google.gmail.findHeader(thread.messages[thread.messages.length - 1], 'Message-ID') || '';
-          const threadMsgRefsLast = Google.gmail.findHeader(thread.messages[thread.messages.length - 1], 'In-Reply-To') || '';
-          return { lastMsgId: thread.messages[thread.messages.length - 1].id, headers: { 'In-Reply-To': threadMsgIdLast, 'References': threadMsgRefsLast + ' ' + threadMsgIdLast } };
-        } else {
+        const lastMsg = (thread.messages || []).reverse().find(m => !m.labelIds || !m.labelIds.includes('TRASH'));
+        if (!lastMsg) {
           return;
         }
+        const threadMsgIdLast = Google.gmail.findHeader(lastMsg, 'Message-ID') || '';
+        const threadMsgRefsLast = Google.gmail.findHeader(lastMsg, 'In-Reply-To') || '';
+        return { lastMsgId: lastMsg.id, headers: { 'In-Reply-To': threadMsgIdLast, 'References': threadMsgRefsLast + ' ' + threadMsgIdLast } };
       } catch (e) {
         if (Api.err.isAuthPopupNeeded(e)) {
           BrowserMsg.send.notificationShowAuthPopupNeeded(parentTabId, { acctEmail });
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -199,9 +199,9 @@ Catch.try(async () => {
         }
       });
     },
-    emailProviderDetermineReplyMsgHeaderVariables: async () => {
+    emailProviderDetermineReplyMsgHeaderVariables: async (progressCb?: ProgressCb) => {
       try {
-        const thread = await Google.gmail.threadGet(acctEmail, threadId, 'full');
+        const thread = await Google.gmail.threadGet(acctEmail, threadId, 'full', progressCb);
         if (thread.messages && thread.messages.length > 0) {
           const threadMsgIdLast = Google.gmail.findHeader(thread.messages[thread.messages.length - 1], 'Message-ID') || '';
           const threadMsgRefsLast = Google.gmail.findHeader(thread.messages[thread.messages.length - 1], 'In-Reply-To') || '';
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -83,7 +83,7 @@ Catch.try(async () => {
     draftId = storage.drafts_reply[threadId]; // there may be a draft we want to load
   }
 
-  const processedUrlParams = { acctEmail, draftId, threadId, subject, from, to, frameId, tabId, isReplyBox, skipClickPrompt, parentTabId, disableDraftSaving, debug };
+  const processedUrlParams = { acctEmail, draftId, threadId, threadMsgId, subject, from, to, frameId, tabId, isReplyBox, skipClickPrompt, parentTabId, disableDraftSaving, debug };
   const closeMsg = () => {
     $('body').attr('data-test-state', 'closed'); // used by automated tests
     if (isReplyBox) {
--- a/extension/chrome/elements/reply_pubkey_mismatch.ts
+++ b/extension/chrome/elements/reply_pubkey_mismatch.ts
@@ -23,6 +23,7 @@ Catch.try(async () => {
   const subject = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'subject') || '';
   const frameId = Assert.urlParamRequire.string(uncheckedUrlParams, 'frameId');
   const threadId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadId') || '';
+  const threadMsgId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadMsgId') || '';
   const to = uncheckedUrlParams.to ? String(uncheckedUrlParams.to).split(',') : [];
   const debug = uncheckedUrlParams.debug === true;
 
@@ -32,7 +33,7 @@ Catch.try(async () => {
   const appFunctions = Composer.defaultAppFunctions();
   const tabId = await BrowserMsg.requiredTabId();
   const processedUrlParams = {
-    acctEmail, draftId: '', threadId, subject, from, to, frameId, tabId, debug,
+    acctEmail, draftId: '', threadId, subject, from, to, frameId, threadMsgId, tabId, debug,
     isReplyBox: true, skipClickPrompt: false, // do not skip, would cause errors. This page is using custom template w/o a prompt
     parentTabId, disableDraftSaving: true
   };
--- a/extension/chrome/elements/reply_pubkey_mismatch.ts
+++ b/extension/chrome/elements/reply_pubkey_mismatch.ts
@@ -23,7 +23,6 @@ Catch.try(async () => {
   const subject = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'subject') || '';
   const frameId = Assert.urlParamRequire.string(uncheckedUrlParams, 'frameId');
   const threadId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadId') || '';
-  const threadMsgId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadMsgId') || '';
   const to = uncheckedUrlParams.to ? String(uncheckedUrlParams.to).split(',') : [];
   const debug = uncheckedUrlParams.debug === true;
 
@@ -33,7 +32,7 @@ Catch.try(async () => {
   const appFunctions = Composer.defaultAppFunctions();
   const tabId = await BrowserMsg.requiredTabId();
   const processedUrlParams = {
-    acctEmail, draftId: '', threadId, subject, from, to, frameId, threadMsgId, tabId, debug,
+    acctEmail, draftId: '', threadId, subject, from, to, frameId, tabId, debug,
     isReplyBox: true, skipClickPrompt: false, // do not skip, would cause errors. This page is using custom template w/o a prompt
     parentTabId, disableDraftSaving: true
   };
--- a/extension/js/common/api/google.ts
+++ b/extension/js/common/api/google.ts
@@ -39,7 +39,7 @@ export namespace GmailRes { // responses
   export type GmailMsg$payload$body = { attachmentId: string, size: number, data?: string };
   export type GmailMsg$payload$part = { body?: GmailMsg$payload$body, filename?: string, mimeType?: string, headers?: GmailMsg$header[] };
   export type GmailMsg$payload = { parts?: GmailMsg$payload$part[], headers?: GmailMsg$header[], mimeType?: string, body?: GmailMsg$payload$body };
-  export type GmailMsg$labelId = 'INBOX' | 'UNREAD' | 'CATEGORY_PERSONAL' | 'IMPORTANT' | 'SENT' | 'CATEGORY_UPDATES';
+  export type GmailMsg$labelId = 'INBOX' | 'UNREAD' | 'CATEGORY_PERSONAL' | 'IMPORTANT' | 'SENT' | 'CATEGORY_UPDATES' | 'TRASH';
   export type GmailMsg = {
     id: string; historyId: string; threadId?: string | null; payload: GmailMsg$payload; internalDate?: number | string;
     labelIds?: GmailMsg$labelId[]; snippet?: string; raw?: string;
--- a/extension/js/common/api/google.ts
+++ b/extension/js/common/api/google.ts
@@ -145,9 +145,10 @@ export class Google extends EmailProviderApi {
       r.emailAddress = r.emailAddress.toLowerCase();
       return r;
     },
-    threadGet: (acctEmail: string, threadId: string, format?: GmailResponseFormat): Promise<GmailRes.GmailThread> => Google.gmailCall(acctEmail, 'GET', `threads/${threadId}`, {
-      format,
-    }),
+    threadGet: (acctEmail: string, threadId: string, format?: GmailResponseFormat, progressCb?: ProgressCb): Promise<GmailRes.GmailThread> =>
+      Google.gmailCall(acctEmail, 'GET', `threads/${threadId}`, {
+        format,
+      }, { download: progressCb }),
     threadList: (acctEmail: string, labelId: string): Promise<GmailRes.GmailThreadList> => Google.gmailCall(acctEmail, 'GET', `threads`, {
       labelIds: labelId !== 'ALL' ? labelId : undefined,
       includeSpamTrash: Boolean(labelId === 'SPAM' || labelId === 'TRASH'),
@@ -195,9 +196,10 @@ export class Google extends EmailProviderApi {
      * because strings over 1 MB may fail to get to/from bg page. A way to mitigate that would be to pass `R.GmailMsg$raw` prop
      * as a Buf instead of a string.
     */
-    msgGet: async (acctEmail: string, msgId: string, format: GmailResponseFormat): Promise<GmailRes.GmailMsg> => Google.gmailCall(acctEmail, 'GET', `messages/${msgId}`, {
-      format: format || 'full'
-    }),
+    msgGet: async (acctEmail: string, msgId: string, format: GmailResponseFormat, progressCb?: ProgressCb): Promise<GmailRes.GmailMsg> =>
+      Google.gmailCall(acctEmail, 'GET', `messages/${msgId}`, {
+        format: format || 'full'
+      }, { download: progressCb }),
     msgsGet: (acctEmail: string, msgIds: string[], format: GmailResponseFormat): Promise<GmailRes.GmailMsg[]> => {
       return Promise.all(msgIds.map(id => Google.gmail.msgGet(acctEmail, id, format)));
     },
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1180,8 +1180,9 @@ export class Composer {
     this.S.cached('input_to').val(this.urlParams.to.join(',') + (this.urlParams.to.length ? ',' : '')); // the comma causes the last email to be get evaluated
     await this.renderComposeTable();
     if (this.canReadEmails) {
-      // not awaited because it blocks whole the compose box and it can take a lot of time depends on replied/forwarded message
-      this.addExpandingButton(method).catch(Catch.reportErr);
+      if (!this.urlParams.draftId) { // if there is a draft, don't attempt to pull quoted content. It's assumed to be already present in the draft
+        this.addExpandingButton(method).catch(Catch.reportErr); // not awaited because can take a long time & blocks rendering
+      }
     } else {
       Xss.sanitizeRender(this.S.cached('prompt'),
         `${Lang.compose.needReadAccessToReply}<br/><br/><br/>
@@ -1921,16 +1922,7 @@ export class Composer {
         }
         const sentDate = new Date(String(this.messageToReplyOrForward.headers.date));
         this.msgExpandingHTMLPart = '<br><br>' + this.generateHTMLRepliedPart(this.messageToReplyOrForward.text, sentDate, this.messageToReplyOrForward.headers.from);
-        if (!this.urlParams.draftId) {
-          this.setExpandingTextAfterClick(this.msgExpandingHTMLPart);
-        } else {
-          const currentHTML = this.S.cached('input_text').html();
-          if (currentHTML.endsWith(this.msgExpandingHTMLPart)) {
-            Xss.sanitizeRender(this.S.cached('input_text'), currentHTML.substring(0, currentHTML.length - this.msgExpandingHTMLPart.length));
-            this.setExpandingTextAfterClick(this.msgExpandingHTMLPart);
-            this.resizeComposeBox();
-          }
-        }
+        this.setExpandingTextAfterClick(this.msgExpandingHTMLPart);
       }
     }
   }
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1368,11 +1368,8 @@ export class Composer {
 
   private setQuoteLoaderProgress = (text: string) => this.S.cached('icon_show_prev_msg').find('#loader').text(text);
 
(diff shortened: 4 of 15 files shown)


extension/chrome/elements/compose.ts
```diff
@@ -83,7 +83,7 @@ Catch.try(async () => {
     draftId = storage.drafts_reply[threadId]; // there may be a draft we want to load
   }
 
-  const processedUrlParams = { acctEmail, draftId, threadId, subject, from, to, frameId, tabId, isReplyBox, skipClickPrompt, parentTabId, disableDraftSaving, debug };
+  const processedUrlParams = { acctEmail, draftId, threadId, threadMsgId, subject, from, to, frameId, tabId, isReplyBox, skipClickPrompt, parentTabId, disableDraftSaving, debug };
```
tomholub: The threadMsgId is used as a backup mechanism in case we could not parse threadId. In fact I want to remove & deprecate its use, but didn't have a chance to do it yet.

The issue is that it's not guaranteed to be there, you should go by threadId.

extension/chrome/elements/reply_pubkey_mismatch.ts
```diff
@@ -23,6 +23,7 @@ Catch.try(async () => {
   const subject = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'subject') || '';
   const frameId = Assert.urlParamRequire.string(uncheckedUrlParams, 'frameId');
   const threadId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadId') || '';
+  const threadMsgId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadMsgId') || '';
```
tomholub: Sorry - as above - better remove.

extension/chrome/elements/reply_pubkey_mismatch.ts
```diff
@@ -32,7 +33,7 @@ Catch.try(async () => {
   const appFunctions = Composer.defaultAppFunctions();
   const tabId = await BrowserMsg.requiredTabId();
   const processedUrlParams = {
-    acctEmail, draftId: '', threadId, subject, from, to, frameId, tabId, debug,
+    acctEmail, draftId: '', threadId, subject, from, to, frameId, threadMsgId, tabId, debug,
```
tomholub: Sorry - as above - better remove.

extension/js/common/composer.ts
```diff
@@ -67,6 +67,7 @@ export type ComposerUrlParams = {
   tabId: string;
   acctEmail: string;
   threadId: string;
+  threadMsgId: string;
```
tomholub: Sorry - as above - better remove.

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
         if (method === 'forward') {
           this.urlParams.subject = 'Fwd: ' + this.urlParams.subject;
           await this.retrieveDecryptAddForwardedMsg(determined.lastMsgId);
+        } else {
+          try {
+            const { raw } = await Google.gmail.msgGet(this.urlParams.acctEmail, this.urlParams.threadMsgId, 'raw');
```
tomholub: Although `threadMsgId` usually points to the last message, this is not guaranteed and I'd like to stop passing it into compose one day.

You can use `threadId` to pull a list of messages, then grab the last one. Please look at what forward is doing, and re-use (some of) the code.

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+            const msgMimeContent = await Mime.process(Buf.fromBase64UrlStr(raw!));
+            if (!msgMimeContent.from || !msgMimeContent.headers.date || !msgMimeContent.blocks || !msgMimeContent.blocks.length) {
+              return;
```
tomholub: This is a fair precaution, though I cannot imagine it would actually happen. Maybe indeed `!msgMimeContent.blocks.length` could be true for empty messages, not sure.

But it's ok to check, just some thoughts.

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+            }
+            if (!this.urlParams.draftId) {
```
tomholub: This took me a while to realize the reason for the check. Please add a comment
```js
// if there is a draftId, then the quoted message was already pulled before and added into the draft
```

extension/js/common/composer.ts
```diff
@@ -1827,6 +1873,24 @@ export class Composer {
     }
   }
 
+  private quoteText(text: string) {
+    return text.split('\n').map(l => '<br/>&gt; ' + l).join('\n');
+  }
+
+  private formatDate(date: Date) {
```
tomholub: maybe shorten to 
```js
date.toISOString().replace(/T/, ' ').replace(/:[^:]+$/, '')
// "2019-07-02 08:15"
```
Then you could add a method as `Str.fromDate` or something like that.

extension/js/common/core/mime.ts
```diff
@@ -15,7 +15,7 @@ const MimeBuilder = requireMimeBuilder();  // tslint:disable-line:variable-name
 const Iso88592 = requireIso88592();  // tslint:disable-line:variable-name
 
 type MimeContentHeader = string | { address: string; name: string; }[];
-type MimeContent = {
+export type MimeContent = {
```
tomholub: It's not clear that the export is needed, at least based on the code above. It seems it would work without the export.

Nothing wrong with exporting, but if it's not used elsewhere, better keep it local, so that it's easier to discover when it's unused (local unused variable goes gray, exported does not)

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+          try {
+            const { raw } = await Google.gmail.msgGet(this.urlParams.acctEmail, this.urlParams.threadMsgId, 'raw');
+            const msgMimeContent = await Mime.process(Buf.fromBase64UrlStr(raw!));
+            if (!msgMimeContent.from || !msgMimeContent.headers.date || !msgMimeContent.blocks || !msgMimeContent.blocks.length) {
+              return;
+            }
+            if (!this.urlParams.draftId) {
+              const blocksToDisplay = msgMimeContent.blocks.filter(b => b.type === 'encryptedMsg' ||
+                b.type === 'plainText' || b.type === 'plainHtml');
+              const contentToDisplay: Array<string> = [];
```
tomholub: All of this below looks good, but it's somewhat duplicated. Very similar functionality is in `retrieveDecryptAddForwardedMsg`, so these two should be merged together and put into one function (the pulling and decrypting of the message, which just returns plain text).

I think your method does it slightly better, because you are cycling through all of the blocks with `Mime.process`, so you'll be sure to catch the needed content.

On the other hand, you are pulling the raw message. If it contains a 15 mb attachment, that may take a while. So you may need to add a loading % sign, maybe next to the three dots, eg "Loading previous message quote... 18%". That means the three dots should probably be visible while this happens, but have opacity 50% and clicking it should not do anything (maybe should also have `cursor: default`). Then when it loads, opacity of (...) goes to 1 and it can be clicked if user wants to.

Either that, or adopt the faster but less precise method from forwarding messages, which may just be good enough. I don't know, your call - you can test it on `flowcrypt.compatibility` emails to see how many of them (that are not marked with a `fix` label) load without issues. You should be able to load almost any plain, plain html, encrypted, or signed message.

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+              for (const block of blocksToDisplay) {
+                if (block.type === 'encryptedMsg') {
+                  const armored = Pgp.armor.clip(block.content as string);
+                  if (armored) {
+                    const decrypted = await PgpMsg.decrypt({
+                      kisWithPp: await Store.keysGetAllWithPassphrases(this.urlParams.acctEmail),
+                      encryptedData: Buf.fromUtfStr(armored)
+                    });
+                    if (decrypted.success) {
+                      contentToDisplay.push(decrypted.content.toUtfStr());
```
tomholub:
```js
} else {
  contentToDisplay.push(`\n(Failed to decrypt quote from previous message because: ${decrypted.error.type}: ${decrypted.error.message})\n`);
}
```
or something like that. Easier to debug issues

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+            if (!msgMimeContent.from || !msgMimeContent.headers.date || !msgMimeContent.blocks || !msgMimeContent.blocks.length) {
+              return;
+            }
+            if (!this.urlParams.draftId) {
+              const blocksToDisplay = msgMimeContent.blocks.filter(b => b.type === 'encryptedMsg' ||
+                b.type === 'plainText' || b.type === 'plainHtml');
+              const contentToDisplay: Array<string> = [];
+              for (const block of blocksToDisplay) {
+                if (block.type === 'encryptedMsg') {
+                  const armored = Pgp.armor.clip(block.content as string);
```
tomholub: You're already getting and `encryptedMsg` block, it should no longer need clipping.

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+                  if (armored) {
```
tomholub: Here the if should not be needed.

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+                    const decrypted = await PgpMsg.decrypt({
+                      kisWithPp: await Store.keysGetAllWithPassphrases(this.urlParams.acctEmail),
+                      encryptedData: Buf.fromUtfStr(armored)
+                    });
+                    if (decrypted.success) {
+                      contentToDisplay.push(decrypted.content.toUtfStr());
+                    }
+                  }
+                } else if (block.type === 'plainHtml') {
+                  contentToDisplay.push(Xss.escape(block.content as string));
```
tomholub: Escaping would result in ugly html tags in the resulting quote text, I think. Can do `Xss.htmlSanitizeAndStripAllTags(content, '\n')` so that you actually convert the html to text, with proper (guessed) newlines.

Let me know if I misunderstood

extension/js/common/composer.ts
```diff
@@ -1201,6 +1208,45 @@ export class Composer {
...
+                } else {
+                  contentToDisplay.push(block.content as string);
+                }
+              }
+              const sentDate = new Date(msgMimeContent.headers.date as string);
```
tomholub: Safer to do `String(x)` (if the type is `String | other things`) or alternatively `x || ''` (if the type is `string | undefined`) rather than casting it, when possible / appropriate.

extension/js/common/composer.ts
michael-volynets: Pgp.armor.clip returns `string | undefined`, because of this, I added `if` expression

extension/js/common/composer.ts
michael-volynets: Okay, thanks. `Xss.htmlSanitizeAndStripAllTags(content, '\n')` would be better here

extension/js/common/composer.ts
michael-volynets: Got it, Thanks!

extension/js/common/composer.ts
michael-volynets: Okay, I will use threadId to get a list and then pick up the last message from the list

extension/chrome/elements/compose.ts
michael-volynets: Okay, I will remove `threadMsgId`. Thanks!

extension/js/common/composer.ts
tomholub: I see it now. Once you remove the clip, you can remove this too.

extension/js/common/core/mime.ts
michael-volynets: oh, I forget to remove it,
I tried to use `Mime.decode` before but I didn' work well and I removed it, and forget to remove `export`.
Thanks

extension/chrome/elements/compose.ts
michael-volynets: in the code when we forward a message I can see you get last msgId using `await this.app.emailProviderDetermineReplyMsgHeaderVariables();`.
I think I can use this one too instead of getting all the messages in the thread. Am I correct?

tomholub: Yes, you can do it similarly to what forward does.

michael-volynets:
@tomholub I almost did it, it quotes replied message but it works laggy because of loading message and decrypting it. I need to think about how can I fix it.
So you can review it but don't merge it yet

tomholub:
You may need to add a % loader next to it, since it also needs to download the message, which can be big and take time. You could just show loader up to 60% (= 100% dowloaded), and then show it stuck at 60 until decrypting is done too.

I also saw that there was still the large date formatting function, if possible please replace.

michael-volynets:
I also thought about loaders. 
I think they can fix the problem but I think would be better to start loading the message to composer even user didn't click the reply button and when they click on the button (if they click) the replied part might show immediately. The minus of this way that the user might not click to the reply button but the request for the message will be sent.

tomholub:
Some messages can be up to 35mb if you are pulling the raw message. That's a lot of things to pull.

michael-volynets:
Wow, okay, then that's a bad idea to pull it without clicking on reply button

extension/js/common/composer.ts
```diff
@@ -1149,58 +1163,43 @@ export class Composer {
...
-          await this.retrieveDecryptAddForwardedMsg(determined.lastMsgId);
+          await this.appendForwardedMsg(this.messageToReplyOrForward.text);
+        } else {
+          if (!this.messageToReplyOrForward.headers.from || !this.messageToReplyOrForward.headers.date) {
+            return;
+          }
+          const sentDate = new Date(String(this.messageToReplyOrForward.headers.date));
+          this.msgExpandingHTMLPart = '<br><br>' + this.generateHTMLRepliedPart(this.messageToReplyOrForward.text, sentDate, this.messageToReplyOrForward.headers.from);
+          if (!this.urlParams.draftId) {
+            this.showExpandButton(this.msgExpandingHTMLPart);
```
tomholub: No need to resize compose box here?
(I'm not sure)

extension/js/common/composer.ts
```diff
@@ -1653,8 +1714,8 @@ export class Composer {
     }, this.getErrHandlers(`focus on recipient field`))).children().click(() => false);
     this.resizeInputTo();
     this.attach.initAttDialog('fineuploader', 'fineuploader_button');
-    this.attach.setAttAddedCb(async () => this.setInputTextHeightManuallyIfNeeded());
-    this.attach.setAttRemovedCb(() => this.setInputTextHeightManuallyIfNeeded());
+    this.attach.setAttAddedCb(async () => { this.setInputTextHeightManuallyIfNeeded(); this.resizeComposeBox(); });
```
tomholub: when there is more than one command, please format it as multiline (max one statement; per line).

extension/js/common/composer.ts
```diff
@@ -1653,8 +1714,8 @@ export class Composer {
...
+    this.attach.setAttRemovedCb(() => { this.setInputTextHeightManuallyIfNeeded(); this.resizeComposeBox(); });
```
tomholub: format as multiline block

extension/js/common/core/mime.ts
```diff
@@ -42,10 +42,16 @@ export type MsgBlock = {
   attMeta?: AttMeta; // only in plainAtt, encryptedAtt, decryptedAtt, encryptedAttLink (not sure if always)
   decryptErr?: DecryptError; // only in decryptErr block, always
 };
+export type MimeProccesedMsg = {
```
tomholub: Yes, why not. Thanks

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -125,11 +125,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
         const lastReplyBtn = $(convoReplyBtnsToReplaceArr.pop()!);
         $(lastReplyBtn).addClass('inserted');
         const element = $(this.factory.btnReply()).insertBefore($(lastReplyBtn).children().last());
-        if (isEncrypted) {
-          element.click(Ui.event.prevent('double', Catch.try(this.setReplyBoxEditable)));
-        } else {
-          element.click(Ui.event.prevent('double', Catch.try(() => this.replaceStandardReplyBox(true, true)))); // xss-safe-factory
-        }
+        element.click(Ui.event.prevent('double', Catch.try(this.setReplyBoxEditable)));
```
tomholub: Why this change? I think the `true, true` is a way to force replacement of the standard reply box even when the conversation is otherwise not encrypted (eg clicking "encrypted reply") on a regular conversation.

tomholub:
When this is all done, and we have added tests for the behavior, we'll need to refactor it. The composer file is getting too big. I'm thinking all draft functionality will go into some `ComposerDraft` class, all quoting functionality would go into a `ComposerQuote` class, and finally a `ComposerContacts` class. I think that covers a big chunk of the functionality.

Don't do it yet, until we test all this more thoroughly.

tomholub:
That will be fore another issue / another PR.

extension/js/common/composer.ts
michael-volynets: resizeComposeBox is below, after `else` statement

extension/js/common/composer.ts
michael-volynets: Okay, got it

extension/js/common/composer.ts
tomholub: Right, but that one will only run if you enter the `else` branch. How about in this `if` branch evaluates true?

michael-volynets:
Issue # 873

Time Spent: ~ 2d 4h
There were some issues, like hiding quoted part when a draft is loaded or when the message is loading very long.
I think I could have made it faster because sometimes I removed the code that I had written because I started writing in the incorrect way from the beginning but in general I've obtained a lot of experience with the composer by working on this task. 
Thanks

extension/js/common/composer.ts
```diff
@@ -1383,21 +1390,26 @@ export class Composer {
...
-        const { raw } = await Google.gmail.msgGet(this.urlParams.acctEmail, determined.lastMsgId, 'raw');
+        const { raw } = await Google.gmail.msgGet(this.urlParams.acctEmail, determined.lastMsgId, 'raw',
+          (progress: number) => this.renderGettingPreviousMessageProgress(30 + progress * 0.3));
         const message = await Mime.process(Buf.fromBase64UrlStr(raw!));
         const readableBlocks = message.blocks
           .filter(b => b.type === 'encryptedMsg' || b.type === 'plainText' || b.type === 'plainHtml');
+        const encryptedCount = readableBlocks.filter(b => b.type === 'encryptedMsg').length;
         const decryptedAndFormatedContent: string[] = [];
-        for (const block of readableBlocks) {
+        for (const [index, block] of readableBlocks.entries()) {
```
tomholub: Interesting, I haven't seen that before. That may come handy. I usually did a traditional for loop with i++, but this may be better if I really need both.

tomholub:
Re-running tests

extension/js/common/composer.ts
michael-volynets: Honestly, I'm using this first time :) 
I don't like traditional `for` statement because it's hard to read.

I'm 100% sure that traditional `for` would be better for productivity but the difference won't be very big (less the 1ms I think).

tomholub: I'm pretty sure the performance difference is meaningless

extension/js/common/api/google.ts
```diff
@@ -145,9 +145,10 @@ export class Google extends EmailProviderApi {
       r.emailAddress = r.emailAddress.toLowerCase();
       return r;
     },
-    threadGet: (acctEmail: string, threadId: string, format?: GmailResponseFormat): Promise<GmailRes.GmailThread> => Google.gmailCall(acctEmail, 'GET', `threads/${threadId}`, {
-      format,
-    }),
+    threadGet: (acctEmail: string, threadId: string, format?: GmailResponseFormat, progressCb?: ProgressCb): Promise<GmailRes.GmailThread> =>
+      Google.gmailCall(acctEmail, 'GET', `threads/${threadId}`, {
+        format,
+      }, { download: progressCb }),
```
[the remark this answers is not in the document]
michael-volynets: I was testing it and it took almost 30% percent of whole the loading time.

tomholub:
Approved

michael-volynets:
So then I can switch out on the new task? Or shall I start trying to write some tests?

tomholub:
Actually, since you mentioned it. This would be a good opportunity to try and write some tests. I will test and merge this PR. You can make an issue to create tests in another PR.

tomholub:
You will want to craft a composer window URL that includes the thread of some other message in it, so that it attempts to reply to that message. Then you test behavior - three dots will load, once loaded, clicking three dots causes text to appear. Do this once for an encrypted message (standard), once for a pgp/mime message, and once for a plain message. That way you'll end up with three tests. This would fall under `compose.ts` tests.

Remember, do not use id= or class= selectors for tests. Each testable element has a special tag attribute for tests, eg `data-test="action-expand-quoted-text"`, so you'll need to add these into the markup.

tomholub:
@michael-volynets I need one change - pulling the message should not block user from starting to type their message.

I mentioned it here, but unfortunately this info gets lost easily in the review comments:

> So you may need to add a loading % sign, maybe next to the three dots, eg "Loading previous message quote... 18%". That means the three dots should probably be visible while this happens, but have opacity 50% and clicking it should not do anything (maybe should also have cursor: default). Then when it loads, opacity of (...) goes to 1 and it can be clicked if user wants to.

image

For the screenshot above I used `opacity: 0.6` while it's loading.

tomholub:
The reason is that as is, it may cause a degraded user experience. I want to start typing my reply, not watch a loader (if the message was 20mb big, I may be staring at the loader for a while). If you do it as in the screenshot above, you're allowing the user to do their work while this is loading.

That means you'll also have to resolve a situation when user hits send but it's still loading. You can add a check for this in `throwIfFormNotReady`

michael-volynets:
yes sorry I did not notice that. I agree with you. And I like the way you came up with.
I think it will be a quick change but I have little reworked forwarding messages.
Now when you forward a message, it also shows a loader before the forwarded message is loaded and I can not do the same thing with forwarding messages because there are no three points.

I think would be better if the solution will be more general? I mean that the same code will work with Forwarding and replying messages.
Maybe we can show the loader inside the reply window but not inside three dots container.

tomholub:
I'd the three dots and the loader for forwarded message too, the same way. When done loading (forward), just auto-click it.

tomholub:
Eg the logic is the same, except the auto-click (alternatively don't register a click handler at all and just do the actual append, whatever is easier in code.

michael-volynets:
Okay, got it, I will do it

extension/js/common/composer.ts
```diff
@@ -625,6 +624,9 @@ export class Composer {
     if (String(this.S.cached('input_to').val()).length) { // evaluate any recipient errors earlier treated as gentle
       await this.parseRenderRecipients('harshRecipientErrs');
     }
+    if (this.S.cached('icon_show_prev_msg').hasClass('progress')) {
```
tomholub: Good. I tend to save state in the DOM sometimes. I know the React community and others consider it a bad practice, but it has worked for me. Thanks for following the style of surrounding code!

extension/js/common/composer.ts
```diff
@@ -1409,7 +1382,9 @@ export class Composer {
               kisWithPp: await Store.keysGetAllWithPassphrases(this.urlParams.acctEmail),
               encryptedData: Buf.fromUtfStr(stringContent)
             });
-            this.renderGettingPreviousMessageProgress(60 + (40 / encryptedCount) * (index + 1));
+            if (progressCb) {
+              progressCb(60 + (40 / encryptedCount) * (index + 1));
+            }
             if (decrptResult.success) {
```
tomholub: Typo :) `decrptResult`

extension/js/common/composer.ts
```diff
@@ -625,6 +624,9 @@ export class Composer {
     if (String(this.S.cached('input_to').val()).length) { // evaluate any recipient errors earlier treated as gentle
       await this.parseRenderRecipients('harshRecipientErrs');
     }
+    if (this.S.cached('icon_show_prev_msg').hasClass('progress')) {
+      throw new ComposerNotReadyError('Retrieving previous message, please wait.');
```
tomholub: Whoops I deleted the previous comment. Thanks for following the surrounding code style, even if checking state from DOM is not considered the best practice, it has worked for me.

tomholub:
Actually other than the typo, code is good. I'll go ahead and test it.

tomholub:
There is one gotcha to cover - missing pass phrase.

Go to FlowCrypt Settings -> Security -> Require pass phrase.

Then try to reply to a message. I'll cover this one since I already have the code open.

tomholub:
I'm just fixing a few details, hard to say if I broke it by editing the code or it was already that way.

reply -> don't manual expand -> write something -> safe draft -> reopen draft (missing newlines):

image

michael-volynets:
Okay, Thanks. 
Yes, I forgot to about passphrases. If you haven't covered it yet I can continue to cover that

tomholub:
It's done, you can review the commits above

tomholub:
No worries, the pass phrases are easy to forget that they may be missing.

One of my test browsers has settings to forget pass phrase (forgot if it's FF or chrome), so that I would bump into it sometimes as I test.

tomholub:
For the last commit https://github.com/FlowCrypt/flowcrypt-browser/pull/1791/commits/cac7315334b7f3e2d225b3035f35ae811d920258 I have decided to turn it off completely when loading from a draft.

The reason is that comparing our own version of `this.msgExpandingHTMLPart` with `this.S.cached('input_text').html()` is likely to cause unexpected behavior. For example, once you insert HTML into the dom, it's up to the browser to manipulate it any way it wants. Eg what if the browser normalizes `<br>` tags to `<br/>` or the other way around etc.. it's not really possible to compare strings that way reliably once they touched the dom.

Instead I make the loaded draft just render and include the quote, which seems like a decent UX compromise for drafts.

tomholub:
Also missed an important things (but maybe I didn't mention it) - to actually include the quoted part in the sent message :laughing: I'm pushing a commit soon.

tomholub:
Also updating so that the `On...` header is part of quoted content in `pgp_block`

image

Before the change it would still show as part of the response.

tomholub:
Ok I think I'm done. @michael-volynets please review my code in case you spot any issues or have comments.

tomholub:
You can review a range of commits (eg only review my changes instead of re-reviewing everything) by clicking the filter:

image

Then shift + clicking my commits.

tomholub:
Tom +3 hours testing & improvements & polishing

michael-volynets:
> Also missed an important thing (but maybe I didn't mention it) - to actually include the quoted part in the sent message  I'm pushing a commit soon.

Ahahah  :laughing:, I think I was so excited working on the other UI things so forgot the main goal of the task.

I've reviewed the code you wrote and I didn't find any issues.
But there is one minor thing:
`private decryptPreviousMsg = async (index: number, encryptedCount: number, encryptedData: Buf, progressCb?: ProgressCb): Promise<string>`, 
I don't think that's a great idea to create the function containing these parameters `index, encryptedCount` and maybe `progressCb` because it isn't necessary, I would leave progress all the calculations of progress in `getAndDecryptPreviousMessage` function.

If you confirm that we need to change it I can change it now because it's opened in my VScode right now 

Also, there are some things I have learnt from your code 
Thanks for the help.

tomholub:
> and maybe `progressCb` because it isn't necessary, I

Thanks for the review. I didn't 100% understand the progress calculation, so I just copy/pasted and it shows. That make sense, please update it.

michael-volynets:
I've changed it

Issue #811, opened 2017-06-09
signed messages verification missing pubkey ui


tomholub:
- attested keys should load automatically if not present, if the email is attasted for the same key as the one that signed the email
 - if not attested but same key, offer a button
 - if attested or not attested and it's a different key, show the mismatch & offer custom import

tomholub:
should offer to download key https://mail.google.com/mail/u/1/#inbox/15ce838fc14f9c04

mbwmbw1337:
When will this be released? Seeing the issue still.

tomholub:
It's now available on Firefox. I'll go ahead and make it available on new Chrome installations. That means if you uninstall & reinstall a few hours from now, you'll get the update. Over the course of the next week, we'll slowly increase the rollout until it reaches everyone.

Version `6.8.8`

mbwmbw1337:
> It's now available on Firefox. I'll go ahead and make it available on new Chrome installations. That means if you uninstall & reinstall a few hours from now, you'll get the update. Over the next week, we'll slowly increase the chrome rollout until it reaches everyone.
> 
> Version `6.8.8`

Thank you. I'll validate in a few hours. Surprisingly I had this pop up on a few emails recently and was about to submit a bug when I found your issue already opened. Thank you for resolving so quickly!

tomholub:
For what it's worth, it's an issue from 2017

¯\_(ツ)_/¯

PR #1795, opened 2019-07-03
issue #811-verify-load-pubkey branch 2

tomholub:
total 4 hrs at this point
#811


tomholub:
+1 hr debugging content scripts :(

Issue #1810, opened 2019-07-08
Make more tests for actual content when quoting


tomholub:
Make more tests for actual content
 - encrypted html
- plain html
 - encrypted text

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/1794#issuecomment-509184710_

tomholub:
+ add error handling when loading quote

PR #1813, opened 2019-07-08
Issue 1810 more tests for quoting

michael-volynets:
Issue #1810 

Time Spent: ~4h

In general, went without issues. but there were small things I didn't know. (e.g. I didn't know that the iframe composer src path from 'flowcrypt-consumer' is different from 'flowcrypt-consumer-mock' (even if both messages are present in both consumers))


michael-volynets:
I something went wrong with CI? I didn't make any changes there

tomholub:
Yeah, not your fault. I had to increase the instance size for tests.

tomholub:
Basically, our browsers were crashing during ci tests, sometimes. I starts about 15 browsers at the same time.

Increasing ram from 4gb to 8gb seems to help. Let's see.

michael-volynets:
Okay, got it.
Thanks

tomholub:
> but there were small things I didn't know. (e.g. I didn't know that the iframe composer src path from 'flowcrypt-consumer' is different from 'flowcrypt-consumer-mock' (even if both messages are present in both consumers))

That's interesting. I would think the url should be exactly the same. If it doesn't work with the url exactly the same, maybe that's a bug that should be fixed.

tomholub:
I bug in the mock api, maybe.

michael-volynets:
Hmm, I can double check it. Maybe, I pasted src incorrectly

extension/js/common/composer.ts
```diff
@@ -1919,7 +1919,14 @@ export class Composer {
       this.S.cached('icon_show_prev_msg').show().addClass('progress');
       Xss.sanitizeAppend(this.S.cached('icon_show_prev_msg'), '<div id="loader">0%</div>');
       this.resizeComposeBox();
-      this.messageToReplyOrForward = await this.getAndDecryptPreviousMessage((progress) => this.setQuoteLoaderProgress(progress + '%'));
+      try {
+        this.messageToReplyOrForward = await this.getAndDecryptPreviousMessage((progress) => this.setQuoteLoaderProgress(progress + '%'));
+      } catch (e) {
+        Catch.reportErr(e);
```
tomholub: Here you can do:
```js
if(Api.err.isSignificant(e)) {
  Catch.reportErr(e);
}
```
That way it will only report errors that are not server 5xx, network errors and similar (eg stuff that we should actually fix).

tomholub:
You don't have to double check it now. But next time you see that, you can share a sample that works and sample that doesn't etc, I can take a look.

extension/js/common/composer.ts
```diff
@@ -1919,7 +1919,14 @@ export class Composer {
...
+        await Ui.modal.error(`The error occured while getting ${method === 'reply' ? 'replied' : 'forwarded'} message. The messasge won't be included in new message.`);
```
tomholub: Here at the end you can add `\n${Api.err.eli5(e)}`

eli5 = "Explain like I'm 5". It will add user-friendly explanation of what exactly went wrong.

michael-volynets:
Okay, got it

extension/js/common/composer.ts
michael-volynets: Yeah, I forgot to add this. Thanks!

extension/js/common/composer.ts
michael-volynets: aha, I was thinking about why it's called `eli5` 
Now, I know :)
I've added it

tomholub:
Looks good

michael-volynets:
I don't know if you remove this line accidentally or not 
```TS
        this.S.cached('icon_show_prev_msg').remove();
```

but in the composer, there will be loader with 0% loaded. I don't think it's a good way to show it.
image

michael-volynets:
Or you wanted to try to reload the previous message by clicking on it?

tomholub:
Sorry, that needs a fix. I wanted to reset the loader so that it would be the same as if user never clicked it. Then they can click the three dots again to try again.

tomholub:
Thanks for checking

Issue #1797, opened 2019-07-04
Multi-frame (passphrase) tests for loading msg quote


tomholub:
#873

Looks like there will be more tests to write:
 - Reply -> enter pass phrase dialog -> enter -> ok -> expand -> text present
 - Reply -> enter passphrase dialog -> cancel -> click loader to cancel -> text not present

But these are more complex because they involve several frames, so I propose to solve them in another issue. But once you do these, you will know about all there is to know about our automated tests, and should be able to write tests of any type.

Moving to a new issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/1794#issuecomment-508410544_

michael-volynets:
Hi @tomholub 

I was trying to create the quoting test with the passphrase but I don't know why the passphrase modal wasn't shown in the test window. Seems like there is some error and in the console, I can see the `Not Found` error. Here is a screenshot of the issue:
![image](https://user-images.githubusercontent.com/49104848/60962298-a129f400-a316-11e9-86c9-98ba08db5d13.png)

And also here is the code of the test
```TS
    ava.test.only('compose[global:compatibility] - reply - pass phrase dialog - ok', testWithNewBrowser(async (t, browser) => {
      const appendUrl = 'isReplyBox=___cu_true___&threadId=16b584ed95837510&skipClickPrompt=___cu_false___&ignoreDraft=___cu_false___' +
        '&threadMsgId=16b584ed95837510&to=flowcrypt.compatibility%40gmail.com&from=flowcrypt.compatibility%40gmail.com' +
        '&subject=Re%3A%20testing%20quotes';
      const messageBeginingText =
        [
          'On 2019-06-14 at 23:24, flowcrypt.compatibility@gmail.com wrote:',
          '> This is some message',
          '>',
          '> and below is the quote',
          '>',
          '> > this is the quote',
          '> > still the quote',
          '> > third line',
          '> >> double quote',
          '> >> again double quote'
        ].join('\n');
      const pp = Config.key('flowcrypt.compatibility.1pp1').passphrase;
      await BrowserRecipe.setUpCommonAcct(t, browser, 'compatibility');
      const settingsPage = await browser.newPage(t, Url.extensionSettings());
      await SettingsPageRecipe.changePassphraseRequirement(settingsPage, pp, 'session');
[...]

PR #1816, opened 2019-07-10
Issue #1797 Created two tests for quoting reply part with passphrase

michael-volynets:
Issue #1797 

Time Spent: 7h

I had some troubles with PassPhrase modal window, It didn't show in the test.
Then I tried to use `inboxPage` and it worked.

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1376,7 +1376,6 @@ export class Composer {
     if (decryptRes.success) {
       return decryptRes.content.toUtfStr();
     } else if (decryptRes.error && decryptRes.error.type === 'need_passphrase') {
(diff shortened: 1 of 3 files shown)


extension/js/common/composer.ts
```diff
@@ -1376,6 +1376,7 @@ export class Composer {
     if (decryptRes.success) {
       return decryptRes.content.toUtfStr();
     } else if (decryptRes.error && decryptRes.error.type === 'need_passphrase') {
+      console.log('passphrase needed');
```
tomholub: Please remove

michael-volynets: Oh, forgot to remove, thanks

Issue #1817, opened 2019-07-10
replies no longer reuse the same thread on the other end


tomholub:
![image](https://user-images.githubusercontent.com/6306961/60980309-c113e480-a323-11e9-97c2-a3a9387f1ea4.png)

This was supposed to be a much longer conversation.

When testing, beware that just because it shows in the same thread on the sender's end does not mean it will show on the same thread on the recipient's end. You need to be looking at the recipient's end.

tomholub:
Please look into this first thing tomorrow, it's a new regression, quite important one.

tomholub:
The problem could either be due to my edit:

https://github.com/FlowCrypt/flowcrypt-browser/pull/1791/files#diff-71cc61c524dc2e78c03a5747baa55f1aR202

Or it could be due to the error handling:

https://github.com/FlowCrypt/flowcrypt-browser/pull/1791/files#diff-f549a749e46085d5ea6e7b9a01bb0270R1435

The error handling causes undefined to be returned even if we already have reply headers, which needs to be fixed either way. If we already have reply headers, they should not get lost of there is error getting the quote.

tomholub:
Another issue, headers should be passed regardless if text is present

https://github.com/FlowCrypt/flowcrypt-browser/pull/1791/files#diff-f549a749e46085d5ea6e7b9a01bb0270R1914

tomholub:
I probably found the actual issue, it's adding the headers when forwarding instead of when replying

```js
     if (method === 'forward') {
        this.additionalMsgHeaders['In-Reply-To'] = this.messageToReplyOrForward.headers.inReplyToHeaders;
        this.additionalMsgHeaders.References = this.messageToReplyOrForward.headers.references;
        this.urlParams.subject = 'Fwd: ' + this.urlParams.subject;
        this.S.cached('icon_show_prev_msg').remove();
        await this.appendForwardedMsg(this.messageToReplyOrForward.text);
      } else {
        if (!this.messageToReplyOrForward.headers.from || !this.messageToReplyOrForward.headers.date) {
          return;
        }
        const sentDate = new Date(String(this.messageToReplyOrForward.headers.date));
        this.msgExpandingHTMLPart = '<br><br>' + this.generateHTMLRepliedPart(this.messageToReplyOrForward.text, sentDate, this.messageToReplyOrForward.headers.from);
        this.setExpandingTextAfterClick(this.msgExpandingHTMLPart);
      }
    }
```
[...]

Issue #1818, opened 2019-07-10
test reply headers


tomholub:
After #1817 and #1774 is done.

michael-volynets:
Here I should check if `In-Reply-To` header contains in some message of the whole thread. 

I want to test reply headers in this way:

1. Get messages by `threadId`
2. Get `In-Reply-To` header Value and try to find the message from the thread that contains this header
3. If there is the message - then everything is okay - user is replying to existing msg. 

This header is formatted in `RFC2822` standard. Example of  `In-Reply-To` header value:
`<CAKbuLTpiB6g9kBSnBuKVQ6WLuJF=sNdPrrMWCLJhOaEcb4jNzA@mail.gmail.com>`
I don't know what have I to do with this string. I found some info about `RFC2822` format but there isn't any info about msgId.

Maybe you know what I can do with that string?

tomholub:
That whole think is the message ID. If you want it very simple, I think you would just grab all raw mime messages by `threadId` from the mock data, then just check that at least one of the `In-Reply-To` values is contained somewhere in the MIME message.

You may need to remove the `<` and `>`, but whatever is inside of it, you should be able to find in the raw message.

See if that works. If it does, it would keep it simple.

tomholub:
Correction, there can only be one `In-Reply-To` value (eg you can only reply to one specific message, I think).

But there are often many `References` (the whole thread, basically).

For now we will focus on `In-Reply-To` as it is somewhat simpler. For `References`, it's similar except there can be several space-separated values:
```
References: <id1> <id2> <id3>
```
Mime headers are subject to various encodings. Maybe my suggestion above (just searching raw mime string) is too crude. But you should not be manually parsing based on the RFC either. There are libraries for that. I saw that you used a library to parse MIME messages. That same library should be giving you parsed and decoded headers, which you can compare.

PR #1821, opened 2019-07-10
issue #1817 Add replying headers before adding three dots

michael-volynets:
Issue #1817 
Time Spent: 40-50min

went smoothly


tomholub:
I will try to be a bit more careful on the next review, since I let this slip through.

Luckily I only deployed last two versions to about 10% of users so far (partial deploy).

michael-volynets:
Yeah, I also will try to test code before PR

tomholub:
I think the real answer is to invest more heavily in automated tests, here is an example: #1818 that way we can be quote confident we din't break anything major when they all pass.

PR #1823, opened 2019-07-11
Issue #1819 custom ks

tomholub:
#1819 4 hrs

The `Rules` class needs a sha1 implementation, which is available in OpenPGP.js, which is not available everywhere in the plugin (not available in `pgp_block.ts` for performance reasons). Since we use Rules class more in this PR, I needed to give it a native sha1 implementation that is not dependent on OpenPGP.js.

Also I took a shortcut to hardcode a keyserver into the `Rules` class instead of serving it from the backend, that should be fixed in the future.

### DIFF
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -16,8 +16,8 @@ import { Google, GmailResponseFormat, GoogleAuth } from '../../js/common/api/goo
 import { Buf } from '../../js/common/core/buf.js';
 import { BackendRes, Backend } from '../../js/common/api/backend.js';
 import { Assert } from '../../js/common/assert.js';
-import { Attester } from '../../js/common/api/attester.js';
 import { Xss } from '../../js/common/platform/xss.js';
+import { Keyserver } from '../../js/common/api/keyserver.js';
 
 Catch.try(async () => {
 
@@ -220,7 +220,7 @@ Catch.try(async () => {
           render(`Found the right pubkey ${signerLongid} on keyserver, but will not use it because you have conflicting pubkey ${senderContactByEmail.longid} loaded.`, () => undefined);
           return;
         } // ---> and user doesn't have pubkey for that email addr
-        const { pubkey, pgpClient } = await Attester.lookupEmail(senderEmail);
+        const { pubkey, pgpClient } = await Keyserver.lookupEmail(acctEmail, senderEmail);
         if (!pubkey) {
           render(`Missing pubkey ${signerLongid}`, () => undefined);
           return;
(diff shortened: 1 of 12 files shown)


extension/chrome/elements/pgp_block.ts
```diff
@@ -235,7 +235,7 @@ Catch.try(async () => {
         await Store.dbContactSave(undefined, await Store.dbContactObj({ email: senderEmail, pubkey, client: pgpClient })); // <= TOFU auto-import
         render('Fetched pubkey, click to verify', () => window.location.reload());
       } else { // don't know who sent it
-        const { pubkey, pgpClient } = await Attester.lookupEmail(signerLongid);
+        const { pubkey, pgpClient } = await Keyserver.lookupLongid(acctEmail, signerLongid);
```
tomholub: The original `lookupEmail` was actually a typo, that still originally worked well because Attester accepts longids or email addresses interchangeably.

extension/chrome/settings/modules/keyserver.ts
```diff
@@ -9,10 +9,11 @@ import { Ui, Env } from '../../../js/common/browser.js';
 import { BrowserMsg } from '../../../js/common/extension.js';
 import { Settings } from '../../../js/common/settings.js';
 import { Api } from '../../../js/common/api/api.js';
-import { Attester, AttesterRes } from '../../../js/common/api/attester.js';
+import { Attester } from '../../../js/common/api/attester.js';
 import { Pgp } from '../../../js/common/core/pgp.js';
 import { Assert } from '../../../js/common/assert.js';
 import { Xss } from '../../../js/common/platform/xss.js';
+import { AttesterRes } from '../../../js/common/api/keyserver.js';
```
tomholub: AttesterRes should be renamed to KeyserverRes

extension/js/common/api/keyserver.ts
```diff
@@ -0,0 +1,53 @@
...
+import { Rules } from '../rules.js';
+import { Attester } from './attester.js';
+import { Sks } from './sks.js';
+import { Dict } from '../core/common.js';
+
+export type PgpClient = 'flowcrypt' | 'pgp-other' | null;
+export type PubkeySearchResult = { pubkey: string | null; pgpClient: PgpClient };
+
+export namespace AttesterRes { // responses
+  export type AttTestWelcome = { sent: boolean };
```
tomholub: AttTestWelcome should go back to Attester file

extension/js/common/api/sks.ts
```diff
@@ -0,0 +1,45 @@
...
+      if (Api.err.isNotFound(e)) {
+        return undefined;
+      }
+      throw e;
+    }
+  }
+
+  public static lookupEmail = async (server: string, email: string): Promise<PubkeySearchResult> => {
+    const index = await Sks.get(server, `pks/lookup?search=${encodeURIComponent(email)}&fingerprint=on&exact=on&options=mr&op=index`);
+    console.log(index);
```
tomholub: remove log

extension/js/common/api/sks.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (!index) {
+      return { pubkey: null, pgpClient: null }; // tslint:disable-line:no-null-keyword
+    }
+    const match = index.match(/^pub:[A-F0-9]{24}([A-F0-9]{16}):[0-9:]+:$/m); // in particular cannot end with :r, meaning revoked
+    console.log(match);
```
tomholub: remove log

extension/chrome/settings/modules/keyserver.ts
tomholub: not renamed, just moved

tomholub:
Default settings

image

Custom Keyserver

image

Issue #1829, opened 2019-07-11
refactor encrypted message verification


tomholub:
would need some refactoring to reuse fewer OpenPGP.js internals

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/1772#issuecomment-510585635_

PR #1830, opened 2019-07-11
#1829 refactor message verify

tomholub:
#1829
20 minutes

read stream to end before attempting to verify


tomholub:
+ 2 hours debugging hanging streams

PR #1832, opened 2019-07-12
Issue 1774 gmail api mock evaluate sent message

michael-volynets:
Issue #1774 

Time Spent: 8-9h
There were some issues with drafts.
The `threadId` is cleaned only when the result of draft saving is not found
```TS
        } else if (!this.urlParams.draftId && Api.err.isNotFound(e)) {
          // not found - creating draft on a thread that does not exist
          this.urlParams.threadId = ''; // forget there was a threadId
```

I would rather set `threadId` to empty string here:
```TS
      const determined = await this.app.emailProviderDetermineReplyMsgHeaderVariables();
      if (determined) {
        ........
        } else {
            this.urlParams.threadId = '' ;       
        }
```

What do you think?


michael-volynets:
Also, I will do some cosmetical refactoring after I fix the issue with enterprise mock

tomholub:
If you just unconditionally return `return { id: 'mockfakesend' };` like before again (as a test) around this line:

https://github.com/FlowCrypt/flowcrypt-browser/pull/1832/files#diff-773d66f438e8c3726d6ccf0b46afe07fL131

Does the enterprise mock start working?

Likely, when throwing the 404, the client doesn't get the right CORS headers, which will throw off enterprise mock but not consumer.

tomholub:
Normally done here:

https://github.com/FlowCrypt/flowcrypt-browser/blob/448d659fe94d7ce64042776ca0c34b478aa5ee7e/test/source/mock/api.ts#L139

Maybe when throwing it will take a different code path

michael-volynets:
Yes, that's worked.
Thanks!

test/source/mock/google-api-mock.ts
```diff
@@ -15,6 +18,7 @@ const isPost = (r: IncomingMessage) => r.method === 'POST';
 const isPut = (r: IncomingMessage) => r.method === 'PUT';
 const isDelete = (r: IncomingMessage) => r.method === 'DELETE';
 const parseResourceId = (url: string) => url.match(/\/([a-zA-Z0-9\-_]+)(\?|$)/)![1];
+const availableEmails: Array<string> = ['human+manualcopypgp@flowcrypt.com', 'human@flowcrypt.com', 'human+nopgp@flowcrypt.com'];
```
tomholub: Should call `allowedRecipients`

test/source/mock/google-api-mock.ts
```diff
@@ -120,19 +124,39 @@ export const startGoogleApiMock = async (logger: (line: string) => void) => {
         const id = parseResourceId(req.url!);
         const msgs = new Data(acct).getMessagesByThread(id);
         if (!msgs.length) {
-          throw new HttpClientErr(`MOCK thread not found for ${acct}: ${id}`);
+          throw new HttpClientErr(`MOCK thread not found for ${acct}: ${id}`, 404);
```
tomholub: I was intentionally returning error 400 so that we can notice this. It means we have to either edit the test to not use a resource that's missing in the mock data, or update the mock data. You could try to use another http error if you want that's more obvious. Maybe 501 Not Implemented.

But if you were testing something that was missing intentionally.. that's a different story. Maybe we could have a list of ids where we actually know they are missing (on purpose) and give err 404 for these, but still give other err for the rest.

test/source/mock/google-api-mock.ts
michael-volynets: Agree, `allowedRecipients` sounds better

tomholub:
> ```ts
> iables()
> ```

That's ok with me, let's do it as a separate PR after this one is merged.

test/source/mock/google-api-mock.ts
michael-volynets: I can't remember exactly why I have changed it to `404`. I think I was testing Gmail API and on that request, it returned 404 for me.

I've checked the tests and seems like they work well with `400` status code.

michael-volynets: let's see what's CI returns for us.

michael-volynets: Yes, tests are passed

michael-volynets:
> > Issue #1774
> > Time Spent: 8-9h
> > There were some issues with drafts.
> > The `threadId` is cleaned only when the result of draft saving is not found
> > ```ts
> >         } else if (!this.urlParams.draftId && Api.err.isNotFound(e)) {
> >           // not found - creating draft on a thread that does not exist
> >           this.urlParams.threadId = ''; // forget there was a threadId
> > ```
> > 
> > 
> > I would rather set `threadId` to empty string here:
> > ```ts
> >       const determined = await this.app.emailProviderDetermineReplyMsgHeaderVariables();
> >       if (determined) {
> >         ........
> >         } else {
> >             this.urlParams.threadId = '' ;       
> >         }
> > ```
> > 
> > 
> > What do you think?
> 
> That's ok with me, let's do it as a separate PR after this one is merged.

I will create an issue, so I don't forget

Issue #791, opened 2017-06-01
add names to outgoing email, from: and to:


tomholub:
Action steps:
1) When composing email, we load contacts from the database, along with their pubkey.
    - If the contact has a `.name` property that is truthy, then outgoing email headers should contain the name, for example: `To: Recipient name <their@email.com>`
    - If not truthy, continue doing what was done until now, eg: `To: their@email.com`.
2) Do same thing with the `From: ` header. Use `full_name` from local storage (but only if truthy - same thing applies - it may not be set, in which case continue to use previous behavior).

Related to #147, but this one is about mime headers. Users who want this: 
https://mail.google.com/mail/u/1/#inbox/15c6201c3bf60782
https://mail.google.com/mail/u/0/#inbox/FMfcgxvzLFDKgdhLrNKqmKlrJngWwBKj

tomholub:
Radek will take a shot

tomholub:
Hints:
- `extractProcessSendMsg` will probably have to be updated
  - In particular, `collectAllAvailablePublicKeys` should probably be renamed to `loadContacts` and just return `Contact[]` or something similar.

Which means that the signatures of almost all methods in `composers.ts` will have to change, and it's not obvious what's the best way to approach it. Maybe this is not an appropriate issue for a new contributor.

tomholub:
@michael-volynets this is relatively involved, so you should ask when unsure

michael-volynets:
> If the contact has a .name property that is truthy, then outgoing email headers should contain the name, for example: To: Recipient name <their@email.com>
If not truthy, continue doing what was done until now, eg: To: their@email.com.

Maybe I don't understand exactly but what do you mean by `truthy?
How can I check if the contact is truthy or not?

michael-volynets:
Do you want me to add some RegEx to check if the name is correct?

tomholub:
Checking for truthiness means you check if something evaluates to `true`.
```js
if(contact.name) {   // contact.name is truthy
  to = `${contact.name} <${contact.email}>`;
} else {
  to = contact.email;
}
```

michael-volynets:
oh, okay, got it
I will remember that :)

PR #1834, opened 2019-07-12
Issue #791 Added names in email headers

michael-volynets:
Issue #791 

Time Spent: 1h 40m
Went smoothly

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -904,6 +904,7 @@ export class Composer {
     if (this.S.cached('icon_pubkey').is('.active')) {
       msg.atts.push(Att.keyinfoAsPubkeyAtt(await this.app.storageGetKey(this.urlParams.acctEmail)));
     }
+    await this.addNamesToMsg(msg);
     let msgSentRes: GmailRes.GmailMsgSend;
     try {
       this.isSendMessageInProgress = true;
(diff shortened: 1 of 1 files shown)


extension/js/common/composer.ts
```diff
@@ -1949,6 +1950,24 @@ export class Composer {
     }
   }
 
+  private addNamesToMsg = async (msg: SendableMsg): Promise<void> => {
+    const contacts = (await this.app.storageContactGet(msg.to)).filter(contact => {
```
tomholub: I feel like this whole section could be a lot simpler
```js
    const contacts = (await this.app.storageContactGet(msg.to)).filter(contact => {
      return contact && contact.name;
    }) as Array<Contact>; // the filter methods filters not null/undefiend contacts
    msg.to = msg.to.map(email => {
      const contact = contacts.find(c => c.email === email);
      if (contact) {
        return `${contact.name} <${email}>`;
      } else {
        return email;
      }
    });
```
Something like
```js
    msg.to = await Promise.all(msg.to.map(async email => {
      const [contact] = await this.app.storageContactGet([email]);
      return contact && contact.name ? `${contact.name} <${email}>` : email;
    }));
```
Or, to stay closer to our style guides:
```js
for (let i = 0; i < msg.to.length; i++) {
  const email = msg.to[i];
  const [contact] = await this.app.storageContactGet([email]);
  msg.to[i] = contact && contact.name ? `${contact.name} <${email}>` : email;
}
```
As written, it looked more complicated than necessary ¯\\\_(ツ)\_/¯

michael-volynets: Yes, agree. Your alternatives look simpler

Issue #1833, opened 2019-07-12
> Issue #1774 Move cleaning threadId 


michael-volynets:
> Issue #1774
> 
> Time Spent: 8-9h
> There were some issues with drafts.
> The `threadId` is cleaned only when the result of draft saving is not found
> 
> ```ts
>         } else if (!this.urlParams.draftId && Api.err.isNotFound(e)) {
>           // not found - creating draft on a thread that does not exist
>           this.urlParams.threadId = ''; // forget there was a threadId
> ```
> 
> I would rather set `threadId` to empty string here:
> 
> ```ts
>       const determined = await this.app.emailProviderDetermineReplyMsgHeaderVariables();
>       if (determined) {
>         ........
>         } else {
>             this.urlParams.threadId = '' ;       
>         }
> ```
> 
> What do you think?

That's ok with me, let's do it as a separate PR after this one is merged.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/1832#issuecomment-510881914_

PR #1837, opened 2019-07-15
Issue 1833 move cleaning thread

michael-volynets:
Issue #1833 
Time Spent: 30m

Went Smoothly


extension/js/common/composer.ts
```diff
@@ -513,10 +513,6 @@ export class Composer {
           // not a draft - updating draft that was since sent as a message (in another window), and is not a draft anymore
           this.urlParams.draftId = ''; // forget there was a draftId - next step will create a new draftId
           await this.draftSave(true); // forceSave=true to not skip
-        } else if (!this.urlParams.draftId && Api.err.isNotFound(e)) {
```
tomholub: You still need this. New messages (not a reply) may also be opened from drafts.

michael-volynets: Okay, got it

michael-volynets:
Okay, got it.

PR #1840, opened 2019-07-15
Issue #1818 added validation for  header

michael-volynets:
Issue #1818 
Time Spent: 2h

Went well

But I don't know why the tests are stopped and there aren't any errors


michael-volynets:
I restarted the tests but there is the same result

test/source/mock/google-api-mock.ts
```diff
@@ -190,9 +190,28 @@ const parseMultipartDataAsMimeMsg = async (multipartData: string): Promise<Parse
...
+        if (message.raw) {
+          const parsedMimeMsg = await Parse.convertBase64ToMimeMsg(message.raw);
+          if (parsedMimeMsg.messageId === inReplyToMessageId) {
+            isMessageExists = true;
+            break;
+          }
+        }
+      }
+      if (!isMessageExists) {
+        throw new HttpClientErr('You are replying to unexisting message id.', 400);
```
tomholub: Please make the error message a bit more verbose (it's for internal debugging, so good to know exactly what's up)
```
Error: suplied In-Reply-To header (${inReplyToMessageId}) does not match any messages present in the mock data for thread ${threadId}
```

test/source/mock/google-api-mock.ts
```diff
@@ -190,9 +190,28 @@ const parseMultipartDataAsMimeMsg = async (multipartData: string): Promise<Parse
...
-  if (threadId && !new Data(acct).getThreads().find(t => t.id === threadId)) {
-    throw new HttpClientErr('The thread you are replying to not found', 404);
+const validateMimeMsg = async (acct: string, mimeMsg: ParsedMail, threadId?: string) => {
+  if (threadId) {
+    const messages = new Data(acct).getMessagesByThread(threadId);
+    if (!messages || !messages.length) {
+      throw new HttpClientErr('The thread you are replying to not found', 404);
+    }
+    const inReplyToMessageId = mimeMsg.headers.get('in-reply-to') ? mimeMsg.headers.get('in-reply-to')!.toString() : '';
+    if (inReplyToMessageId) {
```
tomholub: If there is `threadId` but no `inReplyToMessageId`, it should fail - very loudly - that's the most important thing to test.

test/source/mock/google-api-mock.ts
michael-volynets: Okay, that looks better

tomholub:
I'm not sure what's up with the tests. Looks like a Semaphore bug. Trying to re-run it third time.

test/source/mock/google-api-mock.ts
michael-volynets: So, if there is a `threadId` it's a reply message anyway, yes?

tomholub:
I just thought of two more sanity checks:
1) If subject starts with `Re: ` or `Fwd: `, AND (!threadId || !inReplyToMessageId), then fail
2) If message has threadId OR inReplyToMessageId present, it MUST have either `Re: ` or `Fwd: ` as the beginning of the subject.

This could uncover some potential nasty bugs.

michael-volynets:
Hmm, I've checked how it works now (how it builds subjects).
And currently, it doesn't add `Re: ` to the subject if we are relying on some message

michael-volynets:
Let me check something, I found the code that should add `Re: ` but looks like it doesn't work.
I need to check it

tomholub:
There you go, that's the kind of thing we want to test.

extension/chrome/elements/compose.ts
```diff
@@ -84,7 +84,7 @@ Catch.try(async () => {
     });
     to = reply.to;
     from = reply.from;
-    subject = Google.gmail.findHeader(gmailMsg, 'subject') || '';
+    subject = subject || Google.gmail.findHeader(gmailMsg, 'subject') || '';
```
michael-volynets: I don't know If I fixed it correctly.
What was the reason to re-set subject?

michael-volynets: There was a subject starting with "Re: " but then you change it to the same one but without "Re: "

test/source/mock/google-api-mock.ts
tomholub: Yes, from the perspective of the sender. It will be appended to convo on recipient side.

From the perspective of the recipient, that depends on the email headers.

extension/chrome/elements/compose.ts
tomholub: I'm not sure, I'll have to take a look.

tomholub:
As for the `Re: `. It should be taking the original subject and removing all `Re: ` and `Fwd: ` from the beginning. Then, when it's about to be sent, it should add a `Re: ` or `Fwd: ` in the beginning.

The reason is that if you don't remove it, you'll end up with `Re: Re: Fwd: Re: Subject` which may be technically correct but sucks in my opinion. So I like to strip all of that, and add just one, based on what the user is doing.

I'll take a look why it doesn't work.

extension/chrome/elements/compose.ts
tomholub: I'll try to fix it and push the result. From that you'll see the intention of what is supposed to happen, and if the tests still fail, fix up my fix to work properly.

tomholub: :smiley:

tomholub:
I think the issue was here:
```js
      if (method === 'forward') {
        this.urlParams.subject = 'Fwd: ' + this.urlParams.subject;
        this.S.cached('icon_show_prev_msg').remove();
        await this.appendForwardedMsg(this.messageToReplyOrForward.text);
      } else {
        if (!this.messageToReplyOrForward.headers.from || !this.messageToReplyOrForward.headers.date) {
          this.S.cached('icon_show_prev_msg').hide();
          return;
        }
        const sentDate = new Date(String(this.messageToReplyOrForward.headers.date));
        this.msgExpandingHTMLPart = '<br><br>' + this.generateHTMLRepliedPart(this.messageToReplyOrForward.text, sentDate, this.messageToReplyOrForward.headers.from);
        this.setExpandingTextAfterClick(this.msgExpandingHTMLPart);
      }
```
adding Fwd but not Re.

However I've refactored it a bit, pushing soon

tomholub:
Please take a look, see if it makes sense

michael-volynets:
Okay, now I understood how it works.
You always keep the original subject in `this.urlParams.subject` (until `renderReplyMsgComposeTable` is called) without "Re: " or "Fwd: ".  I thought that the subject that we passed into URL params should be sent but I was incorrect. 
Thanks!


I also spent +3h on investigating it.

I restarted the tests. Looks like the same bug with Semaphore CI

tomholub:
There seems to be a legit failure

image

You should try returning err 404 for threadId `16804894591b3a4b` (intentionally testing missing thread) but still err 400 for all others when missing. Just hard code it in the mock.

tomholub:
```
equest: 400 GET http://localhost:8001/gmail/v1/users/me/threads/16804894591b3a4b?format=full: ok
error: JSHandle@error

    {"status":400,"url":"http://localhost:8001/gmail/v1/users/me/threads/16804894591b3a4b","responseText":"{\"error\":{\"message\":\"MOCK thread not found for flowcrypt.compatibility@gmail.com: 16804894591b3a4b\"}}","statusText":"Bad Request"}
```

tomholub:
There is another test with exactly the same problem:
```
compose[global:compatibility] - reply - old gmail threadId fmt
```
But here the solution is more nuanced. You should try to run the test on live Gmail and see what it does. Then mock should mimic what Gmail does.

michael-volynets:
Okay, I'm going to check it

tomholub:
To run it directly on Gmail, try to add `.only` to the failing test, then try to change:
```diff
- const buildDir = `build/chrome-${(testVariant === 'CONSUMER-LIVE-GMAIL' ? 'CONSUMER' : testVariant).toLowerCase()}`;
+ const buildDir = `build/chrome-consumer`;
```
Just for local testing. Then the test should run live instead of though a mock. I think mock will still be started, but it should not be used.

michael-volynets:
When I'm trying to run tests it throws me this error:

image

On this line of code:
https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/browser/controllable.ts#L144

Maybe I also should change something in `test-secrets.json`?

michael-volynets:
It fails when it tries to fill the password for `test.ci.compose@gmail.com`

michael-volynets:
> There is another test with exactly the same problem:
> 
> ```
> compose[global:compatibility] - reply - old gmail threadId fmt
> ```
> 
> But here the solution is more nuanced. You should try to run the test on live Gmail and see what it does. Then mock should mimic what Gmail does.

I didn't do anything to fix this test, when I fixed previous test this started working

extension/js/common/composer.ts
```diff
@@ -1181,13 +1181,13 @@ export class Composer {
...
 
   private renderReplyMsgComposeTable = async (method: 'forward' | 'reply' = 'reply'): Promise<void> => {
-    this.urlParams.subject = `${(method === 'reply' ? 'Re' : 'Fwd')}: ${this.urlParams.subject}`;
     this.S.cached('prompt').css({ display: 'none' });
     this.S.cached('input_to').val(this.urlParams.to.join(',') + (this.urlParams.to.length ? ',' : '')); // the comma causes the last email to be get evaluated
     await this.renderComposeTable();
     if (this.canReadEmails) {
       const determined = await this.app.emailProviderDetermineReplyMsgHeaderVariables();
       if (determined) {
+        this.urlParams.subject = `${(method === 'reply' ? 'Re' : 'Fwd')}: ${this.urlParams.subject}`;
```
tomholub: This is a nuanced thing, and I'm not super sure what is the right way to do it, but I think maybe putting it above in the original position may be a bit better.

I'm imagining a situation when user is intending to send a reply, but for whatever reason, the right headers cannot be found. In such cases, I would still want subject to show `Re: ` so that the intention is visible that the user meant to reply.

If this ever causes normal replies to be missing headers, we should catch it in the ci tests.

What do you think?

tomholub:
> > There is another test with exactly the same problem:
> > ```
> > compose[global:compatibility] - reply - old gmail threadId fmt
> > ```
> > 
> > 
> > But here the solution is more nuanced. You should try to run the test on live Gmail and see what it does. Then mock should mimic what Gmail does.
> 
> I didn't do anything to fix this test, when I fixed previous test this started working

Thanks for mentioning this, really good to know details like this.

Let's just leave it as is. We can refer to this comment in the future in case we run into the same issue.

extension/js/common/composer.ts
michael-volynets: I'm not sure, 

How can they reply to unexisting `threadId`?
Is there a way to reproduce it using default Gmail Reply box and check how they solved this.

tomholub: 1) Find a conversation to reply to
2) Open a reply box
3) Go to your phone and delete the convo, then delete it from trash
4) wait a bit for the phone to sync changes to Gmail server
5) (back on desktop) send

I don't worry too much about what Gmail does in this scenario, it's too infrequent and not really central to the UX experience etc.

tomholub: Also, we could simply sometimes parse out the thread id wrong. That could happen.

michael-volynets: Got it,

If we can minimize chances of wrong parsing `threadId` then I think it's the right way if we remove `Re: ` prefix.
That's what I would expect from Flowcrypt or Gmail. (I mean I know that something wrong with the thread I'm replying to).

tomholub: Ok then. Fine with me.

Issue #1843, opened 2019-07-15
fix potential GPGTools signed message bug


tomholub:

tomholub:
![image](https://user-images.githubusercontent.com/6306961/61230955-94dcd700-a71a-11e9-97c6-aa1b33206d3b.png)

```
-----BEGIN PGP SIGNATURE-----
Comment: GPGTools - http://gpgtools.org

...
-----END PGP SIGNATURE-----
```

PR #1844, opened 2019-07-15
issue #1843 GPGTools signed compatibility

tomholub:
#1843 
40 mins so far, renders, but verification does not match yet


tomholub:
+30 min, tested with GPGTools, it doesn't verify it either

Issue #963, opened 2017-11-26
option to forget pass phrase anytime - if only in session


tomholub:
should be as a part of the gmail ui (maybe on top)

https://mail.google.com/mail/u/0/#sent/15ffa1ab59b31ea2

tomholub:
Better to show as a part of the pgp_block ui

tomholub:
Maybe some sort of UI that shows eye / crossed eye Android icon

When clicked to forget pass phrase, a message should broadcast to all frames to forget it & reload

tomholub:
css

```css
.cryptup_convo_button:not(:first-child) { margin-left: 5px; }
```

tomholub:
Will do after change of storage mechanism - less work

tomholub:
When pass phrase is remembered in current session, users may want to finish current session so that pass phrase is forgotten.

So there should be a button to do that somewhere (only if the user uses pass phrases in session).

I'm not sure what is the best place to put it. One option is inside, around or under each `pgp_block`. Another option would be on top right, where `see original` button is. Yet another option would be after clicking the FlowCrypt browser action button.

I'm not sure what the button should be called either. `forget pass phrase` sounds weird. Maybe `re-lock encrypted content` or just `lock encrypted content`.

You can tell that pass phrase is in session if you try to get pass phrase from storage `ignoreSession=true`, and there is nothing, and then you get pass phrase with `ignoreSession=false` and you get pass phrase.

Remember, pass phrase must always be brought from the background script, because sessions are different in each frame / background etc. So we always get/set session through background script.

To start, maybe take a screenshot of Gmail and try to find a good place where the button should be and what would it look like, then share the proposals with me, so we can choose the right ux.

tomholub:
Actually, it could be in the form of an open lock icon. When hovered over, it could say `Lock encrypted content again`.

Where should this icon be, I'm not sure - we'll have to experiment as described above.

tomholub:
I had this idea when using settings menu on a mac, where after you enter admin password, there's an open lock icon. You can make it forget the admin password by clicking the lock on the bottom. Eg example here:

![image](https://user-images.githubusercontent.com/6306961/61531810-882deb00-aa5a-11e9-8055-c91410dc1f0f.png)

michael-volynets:
What do you think about the lock icon somewhere here? (In one of those squares)
[...]

PR #1860, opened 2019-07-22
Issue #963 Added btn for finish session, base logic

michael-volynets:
Issue #963 
Time Spent: about 2 days,

There were some problems with `broadcast` and I spent some time investigating how I can do it with all tabs together. 

Result: 
 Gmail -
image

 Inbox -
image


tomholub:
I just learned that GitHub allows to open a draft pr: https://github.blog/2019-02-14-introducing-draft-pull-requests/

extension/js/common/inject.ts
```diff
@@ -54,4 +55,20 @@ export class Injector {
     }
   }
 
+  insertEndSessionBtn = async (acctEmail: string) => {
+    $(this.factory.btnEndPPSession())
+    .insertBefore($('.gb_Zd').children().last())
+    .click(Ui.event.prevent('double', async el => {
+      for(const key of (await Store.keysGet(acctEmail))) {
```
tomholub: I recommend to put this logic into `Store.isPassPhraseCurrentlyInSession`

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -550,4 +552,18 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
   }
 
+  private addEndSessionBtnIfNeeded = async () => {
+    const keys = await Store.keysGet(this.acctEmail);
+    for(const key of keys) {
+      // Check if passpharse in the session
+      if (!(await Store.passphraseGet(this.acctEmail, key.longid, true)) &&
```
tomholub: I recommend to put this logic into `Store.isPassPhraseCurrentlyInSession`

extension/js/common/inject.ts
michael-volynets: Yes, Agree, then there won't be code duplications.

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: Agree, thanks

michael-volynets:
> Sessions expire on their own after some time. I'm not sure after how long. The lock should disappear when that happens.

The Interval function will do that

> Sessions persist after reloading the tab. What if the user closes and reopens gmail?

If they close and reopen it  the session will be ended and they will need to enter PP again

michael-volynets:
And do we need tests for this issue?

tomholub:
Yes, we have a decrypt message with pass phrase in session test.

This test should be appended:
 - at the beginning, there should be no lock in inbox
 - the end, there should be a lock in inbox
 - after clicking the lock, trying to open the message again will ask for pass phrase again

michael-volynets:
done, +2h

extension/js/common/platform/store.ts
```diff
@@ -239,6 +239,19 @@ export class Store {
     return fromSession && !ignoreSession ? fromSession : undefined;
   }
 
+  static getKeysCurrentlyInSession = async (acctEmail: string) => {
```
tomholub: I would have it return a boolean, eg `hasAnyPassPhraseCurrentlyInSession`. That will simplify this method as well as the usage, because it's used that way anyway. Less code is better :)

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -550,4 +557,16 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
   }
 
+  private addOrRemoveEndSessionBtnIfNeeded = async () => {
```
tomholub: This logic is exactly the same as in inbox. You can create a file `js/common/webmail.ts` where we can put shared functionality, for example.

test/source/tests/page_recipe.ts
```diff
@@ -277,16 +277,18 @@ export class InboxPageRecipe extends PageRecipe {
...
     const inboxPage = await browser.newPage(t, Url.extension(`chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId}`));
     await inboxPage.waitAll('iframe');
-    const pgpBlockFrame = await inboxPage.getFrame(['pgp_block.htm']);
+    let pgpBlockFrame = await inboxPage.getFrame(['pgp_block.htm']);
     await pgpBlockFrame.waitAll('@pgp-block-content');
     await pgpBlockFrame.waitForSelTestState('ready');
     if (enterPp) {
-      await pgpBlockFrame.waitAndClick('@action-show-passphrase-dialog', { delay: 1 });
+      await inboxPage.notPresent("@finish-session");
+      pgpBlockFrame.waitAndClick('@action-show-passphrase-dialog', { delay: 1 });
```
tomholub: Missing await here

test/source/tests/page_recipe.ts
```diff
@@ -296,6 +298,17 @@ export class InboxPageRecipe extends PageRecipe {
...
   }
 
+  public static checkFinishingSession = async (t: AvaContext, browser: BrowserHandle, acctEmail: string, threadId: string) => {
+    const inboxPage = await browser.newPage(t, Url.extension(`chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId}`));
+    await inboxPage.waitAll('iframe');
+    await inboxPage.waitAndClick('@finish-session');
+    let pgpBlockFrame = await inboxPage.getFrame(['pgp_block.htm']);
+    await pgpBlockFrame.waitAll('@pgp-block-content');
+    await pgpBlockFrame.waitForSelTestState('ready');
+    pgpBlockFrame.waitAndClick('@action-show-passphrase-dialog', { delay: 1 });
```
tomholub: Missing await here

tomholub: The missing await could sometimes cause confusion. For example, if the command fails, you could lose the stack trace that failed. Then maybe the tests would fail on some other line later, but you would not know why.

Adding await helps not only wait for completion of the `Promise`, but if there is error, helps immediately throw the error up the stack.

extension/js/common/platform/store.ts
michael-volynets: yes but then I work with an array of keys that the method returned:
```TS
      const keysInSession = await Store.getKeysCurrentlyInSession(acctEmail);
      if (keysInSession.length) {
        await Promise.all(keysInSession.map(async k => await Store.passphraseSave('session', acctEmail, k.longid, undefined)));
      }
```

test/source/tests/page_recipe.ts
michael-volynets: Got it, 
Thanks.

extension/js/common/platform/store.ts
tomholub: Sorry, I overlooked that. Got it.

tomholub:
The lock looks not bad. It's not exactly beautiful but doesn't bother me either. It's mostly the shade of orange at fault. I think we can use it as is. Where is the icon from?

image

There is a typo, `Sesion` should be `Session`. But I think we should make it more explicit: `Finish Encrypted Session`.

Finally, when I click it, in Gmail, you don't have to reload the whole tab. You can just reload all encrypted frames, like this:

```js
        $('.' + (window as ContentScriptWindow).reloadable_class).each((i, reloadableEl) => {
          $(reloadableEl).replaceWith($(reloadableEl)[0].outerHTML); // xss-reinsert - inserting code that was already present should not be dangerous
        });
```

I think it will make the experience a little better.

In other places (inbox etc), you can reload the whole tab

Also, in inbox, please give it a `title` like the nearby buttons, so that I know what it does when I hover over it.

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: Yes, it will be better,

Thanks!

michael-volynets:
Yes, I also was thinking about relading only encrypted messages but I didn't know how to do that.

Thanks!

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -56,9 +57,17 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     this.injector = injector;
     this.gmailVariant = gmailVariant;
     this.notifications = notifications;
+    this.webmailCommon = new WebmailCommon(acctEmail, injector);
+  }
+
+  getIntervalFunctions = (): Array<IntervalFunction> => {
+    return [
+      { interval: 1000, handler: this.everything },
+      { interval: 30000, handler: () => this.webmailCommon.addOrRemoveEndSessionBtnIfNeeded() }
```
tomholub: You should keep the same callback style if possible, eg
```js
      { interval: 1000, handler: () => this.everything() },
      { interval: 30000, handler: () => this.webmailCommon.addOrRemoveEndSessionBtnIfNeeded() }
```
or
```js
      { interval: 1000, handler: this.everything },
      { interval: 30000, handler: this.webmailCommon.addOrRemoveEndSessionBtnIfNeeded }
```
It confused me for a second. Unless there is a special reason, in which case maybe a comment would be appropriate

extension/js/common/xss_safe_factory.ts
```diff
@@ -211,6 +211,14 @@ export class XssSafeFactory {
     return `<div class="${this.destroyableCls} reply_message_button"><img title="Secure Reply" src="${this.srcImg('svgs/reply-icon.svg')}" /></div>`;
   }
 
+  btnEndPPSession = () => {
+    return `<div class="finish_session gb_qe gb_oe" data-test="finish-session">
```
tomholub: Also please add a title so that the encrypted inbox icon has it, and double check it doesn't clash in Gmail.

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: yes, thanks, I missed that.

extension/js/common/xss_safe_factory.ts
michael-volynets: I add the title here:
```TS
$(this.factory.btnEndPPSession())
.attr(this.webmailName === 'gmail' ? 'data-tooltip' : 'title', 'End Current Session')
...
```

tomholub:
@michael-volynets where was the icon from? Is it reused from somewhere, or did you draw it? It looks familiar.

michael-volynets:
it's you found it:

>Let's try this one, until I find something more fitting:
https://ionicons.com/ionicons/svg/md-unlock.svg

Issue #1744, opened 2019-06-07
relax use restrictions on public email domains


tomholub:
prerequisite to #1743

tomholub:
I would recommend a static method on `Rules` class, eg:
```js
public static relaxSubscriptionRequirements = (emailAddr: string) {
  return ['gmail.com', 'yahoo.com', 'outlook.com', 'live.com'].includes(emailAddr.split('@')[1] || 'NONE')
}
```
When sending email, you would look at the `from` email, not at email account when evaluating this.

When `from:` is a public domain address, not a private domain, you can allow the user to send large attachments even if they don't have subscription. In fact, you don't even check if they have a subscription if this rule matches their email.

tomholub:
That means I could have email account `human@flowcrypt.com` and an alias `my.personal.email@gmail.com`.

Or an account `my.personal.email@gmail.com` with alias `human@flowcrypt.com`.

Either way, when sending from the @gmail address, I don't have to have a subscription to send large files.

tomholub:
This one will definitely make some people happy :laughing:

PR #1862, opened 2019-07-24
issue #1744 Added emails whitelist that can send large attachments

michael-volynets:
Issue #1744 

Time Spent: 1h 30m

Went smoothly

### DIFF
--- a/test/source/tests/tests/account.ts
+++ b/test/source/tests/tests/account.ts
@@ -60,7 +60,7 @@ export const defineConsumerAcctTests = (testVariant: TestVariant, testWithNewBro
     ava.test.todo('settings > subscribe > expire > compose > footer > subscribe');
 
   } else {
-    ava.test('compose > large file > gmail account (regular user)', testWithSemaphoredGlobalBrowser('compose', async (t, browser) => {
+    ava.test('compose > large file > public domain account (should not prompt to upgrade)', testWithSemaphoredGlobalBrowser('compose', async (t, browser) => {
       const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
       await ComposePageRecipe.fillMsg(composePage, 'human@flowcrypt.com', 'a large file test (gmail account)');
       const fileInput = await composePage.target.$('input[type=file]');
--- a/test/source/tests/tests/account.ts
+++ b/test/source/tests/tests/account.ts
@@ -60,8 +60,8 @@ export const defineConsumerAcctTests = (testVariant: TestVariant, testWithNewBro
     ava.test.todo('settings > subscribe > expire > compose > footer > subscribe');
 
   } else {
-    ava.test('compose > large file > public domain account (should not prompt to upgrade)', testWithSemaphoredGlobalBrowser('compose', async (t, browser) => {
-      const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
+    ava.test('compose > large file > public domain account (should not prompt to upgrade)', testWithSemaphoredGlobalBrowser('compatibility', async (t, browser) => {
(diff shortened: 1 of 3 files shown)


michael-volynets:
Fixing the test

michael-volynets:
@tomholub Can we change `flowcrypt.test.trial@gmail.com` to `flowcrypt.test.trial@flowcrypt.com`?

michael-volynets:
Even would be better to add `flowcrypt.test.trial@flowcrypt.com` as an alias there

tomholub:
I think this involves a backend change (I have an exception there to be able to wipe this account's settings for ci). 

I could try to give it an email alias that is @somecompany

michael-volynets:
Yes, it doesn't matter what custom domain we will use

tomholub:
This may be a bit tricky, I'll see what I can do

tomholub:
You could use `test.ci.compose@org.flowcrypt.com` for the trial test. Then use any `@gmail` account to verify that it lets user send bigger files. It's not perfect but it should allow us to test both scenarios.

tomholub:
That means we can stop using flowcrypt.test.trial@flowcrypt.com completely

tomholub:
Plus you can test the @ gmail account using a mock.

For the org.flowcrypt account, you'll have to use live Gmail.

tomholub:
You can replace:
```js
await FlowCryptApi.hookCiAcctDelete(Config.secrets.ci_dev_account);
```
with:
```js
await FlowCryptApi.hookCiAcctDelete('test.ci.compose@org.flowcrypt.com');
```
So that the trial account gets removed before you test with that email.

michael-volynets:
> You can replace:
> 
> ```js
> await FlowCryptApi.hookCiAcctDelete(Config.secrets.ci_dev_account);
> ```
> 
> with:
> 
> ```js
> await FlowCryptApi.hookCiAcctDelete('test.ci.compose@org.flowcrypt.com');
> ```
> 
> So that the trial account gets removed before you test with that email.

Thanks!
I was already thinking about what I can do with that

tomholub:
The servers are still updating. Some are updated, some not yet (will be done in a few minutes). So if you get an error from backend that it's not a dev account, just try a few minutes later.

test/source/tests/tests/account.ts
```diff
@@ -59,6 +59,15 @@ export const defineConsumerAcctTests = (testVariant: TestVariant, testWithNewBro
 
     ava.test.todo('settings > subscribe > expire > compose > footer > subscribe');
 
-  }
+  } else {
+    ava.test('compose > large file > gmail account (regular user)', testWithSemaphoredGlobalBrowser('compose', async (t, browser) => {
```
tomholub: Looks good, just for clarity let's call it `compose > large file > public domain account (should not prompt to upgrade)`

michael-volynets: Okay, I'm renaming it

tomholub:
Our ci secrets missing pwd again, I think

tomholub:
I updated it, hopefully now it goes through

tomholub:
image

tomholub:
I can just start sending all test failures to your email as well. You will just be getting a bunch of emails that are not related to your work

michael-volynets:
no problem, you can send them to me

michael-volynets:
Sorry, that was my mistake, I was using `compose` account instead of `compatibility`

michael-volynets:
I don't know how I could have missed that.

but the first test (`[standalone] compose > large file > subscribe > trial > attach again`) works on my side.
Is there something wrong with semaphoreci?

tomholub:
Yes, new secrets missing in legacy semaphore setup

tomholub:
Hopefully now it will pass

Issue #1865, opened 2019-07-29
No styles when opening new draft from Drafts folder


michael-volynets:
![image](https://user-images.githubusercontent.com/49104848/62059829-a4851100-b22c-11e9-8bfe-67b1dbda085b.png)

Steps to reproduce:
1. Require PP
2. Open New Secure Compose
3. Create some Draft
4. End PP session if PP is in the session
5. Open draft from drafts folder in Gmail

We can see that the composer asks us to enter PP (`waiting for PP to open draft`) but the composer box is transparent

michael-volynets:
I guess this could have happened after I fixed the issue with attachments about 2 months ago

tomholub:
Interesting. Can give this prompt a white background. Otherwise the style can stay for now, as long as the close button works.

michael-volynets:
When I started working on this issue I wanted to make the compose window looks like this:
![image](https://user-images.githubusercontent.com/49104848/62374070-cbe92000-b543-11e9-980c-51dd93e9a64f.png)

but when I have made this I found out that it broke some other pages, so it can take more time and efforts to fix it that way I wanted.
In the next PR, I'll only change background-color.

Maybe this worths to create another issue to make this composer window looks like in the screenshot when PP isn't entered.

PR #1873, opened 2019-08-02
Issue 1865 added white background

michael-volynets:
Issue #1865 
Time Spent: 5m for adding background color but I'd spent about 2h before for making it looks like what I described in the comment:
https://github.com/FlowCrypt/flowcrypt-browser/issues/1865#issuecomment-517705260


tomholub:
Looks good, thanks

Issue #911, opened 2017-09-06
reply type based on message type: sign -> sign, encrypt -> encrypt


RamonBeast:
When replying to a signed message, by default the "encrypt and send" option is selected, in any case it's not possible (at least in Inbox, correct me if I'm wrong) to reply without signature to a signed message, the outcome is that if a message is addressed to multiple people and only one signs it, everyone using cryptup is forced to add his signature too. I think that this pattern is not entirely correct, I would like to propose a new one:

1) If the message is encrypted the reply button could show the following options in order: reply encrypted, sign and send, reply in clear (this last option can be omitted if too dangerous)

2) If the message is signed: sign and send, reply encrypted, reply in clear (in this case the last option doesn't present any danger)

3) If the message is in clear: no action is taken since cryptup doesn't activate

I think this pattern might work better than the current one, in any case I believe that it should be possible for the user to without signature to a signed message.

tomholub:
I agree, we'll improve it. Thanks for your input.

RamonBeast:
Thank you Tom for your amazing work and always swift response.

tomholub:
I don't know what's a good way to implement this. @michael-volynets you can take a look, maybe you'll have an idea.

tomholub:
However, I only want to implement a simplified version:

If last incoming message was signed, by default when rendering reply box, click the "sign" button for the user. If last incoming message was encrypted, leave as is (default encrypted).

PR #1874, opened 2019-08-04
issue #911 Sign new message if the message user replying to is signed

michael-volynets:
Issue #911 

Time Spent: 1h

When I was working on this issue I found another issue, when we saving signed draft Flowcrypt saves it as encrypted msg (not a signed msg). I was trying to fix it, and I fixed but the duration of draft saving (if signed) increased in two or more times so I removed the changes from this PR.
Maybe we need to create another issue to make an ability to save signed drafts, what do you think?

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1425,7 +1425,7 @@ export class Composer {
       const message = await Mime.process(Buf.fromBase64UrlStr(raw!));
       const readableBlocks = message.blocks
         .filter(b => ['encryptedMsg', 'plainText', 'plainHtml', 'signedMsg'].includes(b.type));
-      const encryptedCount = readableBlocks.filter(b => ['encryptedMsg', 'signedMsg'].includes(b.type)).length;
+      const pgpBlockCount = readableBlocks.filter(b => ['encryptedMsg', 'signedMsg'].includes(b.type)).length;
       const decryptedAndFormatedContent: string[] = [];
       for (const [index, block] of readableBlocks.entries()) {
         const stringContent = String(block.content);
@@ -1436,7 +1436,7 @@ export class Composer {
           const htmlParsed = Xss.htmlSanitizeAndStripAllTags(htmlBlock ? htmlBlock.content.toString() : 'No Content', '\n');
           decryptedAndFormatedContent.push(Xss.htmlUnescape(htmlParsed));
           if (progressCb) {
-            progressCb(60 + (40 / encryptedCount) * (index + 1));
+            progressCb(60 + (40 / pgpBlockCount) * (index + 1));
           }
         } else if (block.type === 'plainHtml') {
           decryptedAndFormatedContent.push(Xss.htmlUnescape(Xss.htmlSanitizeAndStripAllTags(stringContent, '\n')));
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -3,96 +3,37 @@
 'use strict';
 
 import { Catch, UnreportableError } from './platform/catch.js';
-import { Store, Subscription, ContactUpdate, DbContactObjArg } from './platform/store.js';
+import { Store, Subscription } from './platform/store.js';
 import { Lang } from './lang.js';
 import { Value, Str } from './core/common.js';
 import { Att } from './core/att.js';
-import { BrowserMsg, Extension, BrowserWidnow, Bm } from './extension.js';
-import { Pgp, Pwd, FormatError, Contact, KeyInfo, PgpMsg } from './core/pgp.js';
-import { Api, ProgressCb, ChunkedCb, AjaxError } from './api/api.js';
+import { BrowserMsg, Extension, BrowserWidnow } from './extension.js';
+import { Pgp, Pwd, PgpMsg } from './core/pgp.js';
+import { Api } from './api/api.js';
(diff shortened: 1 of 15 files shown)


tomholub:
Good observation (on the speed). Signing is an expensive operation (similar to decrypting). I'm thinking we will not support signed drafts at all for that reason.

RSA signing, decrypting - expensive
RSA encrypting, verifying - (relatively) cheap

extension/js/common/composer.ts
```diff
@@ -1418,12 +1424,12 @@ export class Composer {
         progressCb ? (progress: number) => progressCb(progress * 0.6) : undefined);
       const message = await Mime.process(Buf.fromBase64UrlStr(raw!));
       const readableBlocks = message.blocks
-        .filter(b => b.type === 'encryptedMsg' || b.type === 'plainText' || b.type === 'plainHtml');
-      const encryptedCount = readableBlocks.filter(b => b.type === 'encryptedMsg').length;
+        .filter(b => ['encryptedMsg', 'plainText', 'plainHtml', 'signedMsg'].includes(b.type));
+      const encryptedCount = readableBlocks.filter(b => ['encryptedMsg', 'signedMsg'].includes(b.type)).length;
```
tomholub: Maybe `encryptedCount` could be renamed to `pgpBlockCount` (purpose has changed)

michael-volynets: Yes, agree, 
from now it stores count for both encrypted and signed blocks

michael-volynets:
Okay, thanks,
I forgot about the test,
And maybe, you have some cases when I can be 100% sure that I need to add the tests for the issue? Because sometimes I just don't know whether I need to add tests or not

tomholub:
Here is my rule of thumb when I'm not sure if I should be adding test. I ask myself:

1) Is it manageable to manually test this feature after other code changes to make sure it still works?
2) If it fails in the future, would it be easy to discover?
3) If it fails, is it a big deal?

If answer to 1 or 2 is no, then I probably want to write a test.

But then if answer to 3 is also no, maybe it's ok to skip a test.

In this case, I would not want to have to manually test this after changes, so a test may be appropriate. But then, if it fails, it's not a big deal (in this case). So the answer is subjective.

The reason I thought to make a test in this case is that I want to re-work the "sign and send" UX in the future (to be a button with a down chevron for message type: sign/encrypt/sign+encrypt/plain). I'm afraid this functionality will be easy to overlook when we rework it, so I would like to add a test for that reason.

Overall, now that we use mocks for testing, reliability of the tests should generally not be a problem. So more tests are a good thing, as long as they test something useful.

tomholub:
I just thought, for composer changes specifically, I would lean on the side of adding tests, because composer is quite complex yet quite easy to test independently.

michael-volynets:
Thanks for the explanation, I'll try to follow these rules.

I also have a question, is there a way to get HTMLElement in the tests?
I mean, to test whether the reply box is signed I want to check if there is '.active' class on sign icon:
image

But I can't access this sign icon,
I tried using `.target.$('.icon.action_sign'))` and it returns  me `ElementHandle<Element>` object but I can't access `class` attribute through that object.

Maybe you have an idea of how to access it?

tomholub:
Two approaches

One approach is to try to figure out the class from found element. You could try something like
```js
let signIcon = await page.waitAny('@action-toggle-sign-message');
let classProperty = signIcon.getProperty("class");
let classJson = classProperty.jsonValue();  // evaluate the classJson object, see if you can get the class from there
```
I'm just guessing from https://github.com/GoogleChrome/puppeteer/blob/v1.19.0/docs/api.md#class-elementhandle - if you will take this approach, it would make sense to add it as a static convenience method, maybe on `PageController`.

The other option is to query it including the intended class eg `.icon.action_sign.whatever_intended_class`

It would be better to use `@selectors` that we typically have for testing. You could try to edit this function:
```js
  protected selector = (customSelLanguageQuery: string): string => { // supply browser selector, xpath, @test-id or @test-id(contains this text)
    let m;
    if (this.isXpath(customSelLanguageQuery)) {
      return customSelLanguageQuery;
    } else if (m = customSelLanguageQuery.match(/^@([a-z0-9\-]+)$/)) { // tslint:disable-line:no-conditional-assignment
      return `[data-test="${m[1]}"]`;
    } else if (m = customSelLanguageQuery.match(/^@([a-z0-9\-]+)\(([^()]*)\)$/)) { // tslint:disable-line:no-conditional-assignment
      return `//*[@data-test='${m[1]}' and contains(text(),'${m[2]}')]`;
    } else {
      return customSelLanguageQuery;
    }
  }
```
Maybe add something like:
```js
    } else if (m = customSelLanguageQuery.match(/^@([a-z0-9\-]+)\.[a-z\-_]+$/)) { // tslint:disable-line:no-conditional-assignment
      return `[data-test="${m[1]}"].${m[2]}`;
```
Now our selectors will be a bit smarter and you can select a specific class with it:
```js
let signIcon = await page.waitAll('@action-toggle-sign-message.whatever_intended_class');
```
Which should allow you to achieve what you want.

I haven't tested any of this code.

michael-volynets:
Thanks a lot,

I used the first approach in the PR

tomholub:
@michael-volynets I can merge this when the conflicts are resolved

michael-volynets:
Resolved

Issue #1798, opened 2019-07-04
test that sent message includes previous quote


tomholub:
When #1774 - mock email parsing - is done, the next step would be to make sure that messages sent eg by tests from #1794 indeed include the quoted part.

We can differentiate messages by the subject.

eg, very roughly, skipping many details:

```js
'/msg/send': async (req) => {
  const {text, subject, threadId} = await parseAsEmailReq(req);
  if(subject.includes('include quoted part')) {
    expect(await decrypted(body)).to.contain(data.getLastEmail(threadId).body);
  } else if (subject.includes('... other subject')) {
    // test some other thing
 } else {
    // default just pass
  }
}
```

tomholub:
Maybe this is a little too aggressive in terms of testing, given the functionality. There's probably easier and more impactful tests we could do. If you haven't started on this, I'd move it to sometime later.

michael-volynets:
No, I haven’t started it yet. You can move it

PR #1876, opened 2019-08-05
issue #1798 Added test for testing if the message contains prev quoute

michael-volynets:
Issue #1798
Time Spent: ~4h


tomholub:
Here I'm waiting for the updated secrets file. Thanks!

Issue #1879, opened 2019-08-07
+ add intro button does not disappear when I click "Sign" icon


tomholub:
1) add recipient (no pgp) and message
2) switch to sign
3) Green `+ intro` button is still visible. It should hide when I click sign.

tomholub:
Please add a test for this one

michael-volynets:
Can you please explain to me the logic?
I found the code that should hide `add_intro` button but I see this `if` expression:
```TS
    if (!$('.recipients span').length) {
      this.hideMsgPwdUi(); // <-- This call hides add_intro btn
      this.S.cached('send_btn').removeClass('gray').addClass('green');
    }...
```
If we follow the code it works as it's expected

tomholub:
Here's what it means:
```
    if (!$('.recipients span').length) {
      this.hideMsgPwdUi(); // <-- This call hides add_intro btn
      this.S.cached('send_btn').removeClass('gray').addClass('green');
    }
```
If the user has just deleted the last recipient, there is no need to continue showing UI elements related to "password protected message", which includes the intro. 

Without this code, when I add askjdlkasjdklasjdaklj@gmail.com, then remove it, the compose window will be stuck in "add password" until I add another recipient, which is weird.

You can add a short comment to clarify the code.

Also, this may be an interesting thing to add a test for, as well.

michael-volynets:
Got it, Thanks

PR #1880, opened 2019-08-07
Issue #1879 Hid and intro btn and wrote a test for it

michael-volynets:
Issue #1879 

Time Spent: 40m

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -686,7 +686,7 @@ export class Composer {
     this.S.cached('send_btn_note').text('');
     this.S.cached('send_btn').removeAttr('title');
     const wasPreviouslyVisible = this.S.cached('password_or_pubkey').css('display') === 'table-row';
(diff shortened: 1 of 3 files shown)


extension/js/common/composer.ts
```diff
@@ -686,10 +686,8 @@ export class Composer {
     this.S.cached('send_btn_note').text('');
     this.S.cached('send_btn').removeAttr('title');
     const wasPreviouslyVisible = this.S.cached('password_or_pubkey').css('display') === 'table-row';
-    if (!$('.recipients span').length) {
-      this.hideMsgPwdUi();
-      this.S.cached('send_btn').removeClass('gray').addClass('green');
-    } else if (this.S.cached('icon_sign').is('.active')) {
+    this.hideMsgPwdUi();
```
tomholub: Here you're always hiding it, and then maybe re-showing later, which cannot be a good thing

michael-volynets: Yes, I also wasn't sure about it, 

Thanks for your answer in the issue

Issue #1881, opened 2019-08-07
Refactor `composer-contacts.ts`


michael-volynets:
The main reason for refactoring is to make the code easier to understand and to make the easier implementation of **CC/BCC**

tomholub:
related  #1065

PR #1884, opened 2019-08-08
Issue #1881 Refactored Compooser-contacts.ts

michael-volynets:
Issue #1881 

Time Spent: 6h


michael-volynets:
Here I refactored little bit `composer-contacts.ts`

michael-volynets:
Also, Do you know what `auth_contacts` is?
image

There is a handler but no HTML elements with this class name

tomholub:
We used to have a button at the bottom of search results, that said "Search Google Contacts". It would have this `.auth_contacts` class. It would add permission through oAuth and than we could search directly through google.

Later when I implemented contact search through sent emails, I stupidly removed Google contacts support, thinking that we don't need it. But the current search results are very imperfect. So soon we will be adding the prompt back..

michael-volynets:
Oh, sorry, I was looking in the code before removing it, 
I found only one place where it was used, this is here:
```TS
               if (errsMode === 'gentleRecipientErrs') {
                    gentleErrInvalidEmails += rawRecipientAddrInput;
                    this.composer.debug(`parseRenderRecipients(${errsMode}).6-b (${email}|${rawRecipientAddrInput})`);
                }
```

And then `gentleErrInvalidEmails` was replaced with `input_to` value:
```TS
        this.composer.S.cached('input_to').val(gentleErrInvalidEmails);
```

So I removed it and replaced `input_to` with invalid emails:
```TS
 inputTo.val(validationResult.notValid.join(','));
```

> In certain situations, when I enter a recipient with name eg "tom", which is not a valid email and then hit enter, it should give me an obvious red error (what I call a "harsh" error). I want to see that the entry makes no sense, so tom will be highlighted in a red bubble/chip.

I'm trying to reproduce it on the production version of the extension but I can't see any highlighting,
Could you please check if it works correctly?

tomholub:
You're right, it's broken at the moment.

tomholub:
This is, roughly, what's supposed to happen for `harsh` errors:

image

michael-volynets:
Okay, I got it,

Can I fix it in this PR or should I create another?

tomholub:
Fix it here please, and please add a test too, so that it doesn't get broken again.

tomholub:
You don't 100% have to follow the gentle/harsh way of categorizing it, as long as the result is as desired. In particular, we should look into the one situation where we used "gentleRecipientErrs" and make sure it remains that way after refactoring.

tomholub:
There is another issue on prod:
1) add tom and hit enter
2) it was supposed to become red (which we just discovered that it's broken)
3) hit send, and "tom" will disappear, and you get a message "please add recipient"

If the "tom" was correctly red, the send button would respond differently. It would notice there is a "bad" recipient, and would say "Please correct recipients marked in red".

This would be good to test too.

michael-volynets:
Okay, Thanks,

To make sure:
If a user types an incorrect email address and presses enter - we will add it as the recipient but with a red background, also it should be saved in draft

If a user types an incorrect email address and blurs the input we won't add it to the recipients' list just leave it in the input

If there is at least one recipient with a red background - we can't send a message, we have to show modal

Am I correct?

tomholub:
Roughly correct.

> If a user types an incorrect email address and blurs the input we won't add it to the recipients' list just leave it in the input

I think it should be red in this case too.

Let me take a look exactly when we use gentle errors.

tomholub:
You were right. Lost of focus should result in just leaving it in the input (if it would otherwise be an error. If it's correct email, that's a different story).

```js
    this.S.cached('input_to').blur(Ui.event.handle(async (target, e) => {
      this.debug(`input_to.blur -> parseRenderRecipients start causedBy(${e.relatedTarget ? e.relatedTarget.outerHTML : undefined})`);
      const isRecipientAdded = await this.parseRenderRecipients('gentleRecipientErrs'); // gentle because sometimes blur can happen by accident, it can get annoying (plus affects CI)
      if (isRecipientAdded) {
        this.draftSave(true).catch(Catch.reportErr);
      }
      this.debug(`input_to.blur -> parseRenderRecipients done`);
    }));
```

michael-volynets:
Okay, Great,

Thanks!

tomholub:
I've marked it as work-in-pgoress in the title. Once ready, please remove & tag me again for review. Thanks!

michael-volynets:
Sure,
I will do it

michael-volynets:
I've got a problem,
When I want to save draft with an invalid email address the `Mime.encode` returns me a string of MimeMsg but without `To: ` header, here is a screenshot:
image

I want to save in draft with this email `invalid_email` (it's highlighted with red color now) but when Mime encoded it there isn't `To:` header, 
I have checked `emailjs-mime-builder` repository and found this code:

```JS
    this._headers.forEach(header => {
      var list = []
      if (header.key === 'From' || (!envelope.from && ['Reply-To', 'Sender'].indexOf(header.key) >= 0)) {
        convertAddresses(parseAddresses(header.value), list)
        if (list.length && list[0]) {
          envelope.from = list[0]
        }
      } else if (['To', 'Cc', 'Bcc'].indexOf(header.key) >= 0) {
        convertAddresses(parseAddresses(header.value), envelope.to)
      }
    })
```
They parse the emails themselves and if it's invalid they just don't add it to the Mime Msg.

Maybe, Gmail receives another header with invalid emails for draft saving

I can look for it or we can just leave it without storing in drafts, what do you think?

tomholub:
> Maybe, Gmail receives another header with invalid emails for draft saving
> I can look for it or we can just leave it without storing in drafts, what do you think?

I understand the issue but do not understand the two sentences here. I'm not sure what you meant. Can you elaborate?

Does Google accept a draft without to: header? If they do, then I think the behavior is good, I don't mind invalid recipients disappearing.

michael-volynets:
>I understand the issue but do not understand the two sentences here. I'm not sure what you meant. Can you elaborate?

I meant that Gmail receives `To:` header for valid Emails and, maybe, some another header for invalid emails because they do save draft with invalid emails (I've checked it). If there is another header I need to find it because I don't know the header name yet.

Yes, they accept a draft without `to: ` header, I checked it too.

michael-volynets:
If it's okay that invalid emails won't be saved in the draft then I can leave it as is

tomholub:
Yup, if they're invalid, I'm happy they're gone :+1:

extension/js/common/composer.ts
```diff
@@ -127,6 +127,7 @@ export class Composer {
       await this.initComposeBox();
```
tomholub: I'm not sure this part will do what you expect.

I think, if `await this.initComposeBox();` fails, you may have an exception that gets silently swallowed. Either way, as is, with async resolver inside a Promise, it's not 100% clear what will happen from the code.

My recommendation is to refactor the whole thing as simply an async self-invoked function:

```js
 this.initialized = (async () => {
      await this.initComposeBox();
      await this.initActions();
      await this.checkEmailAliases();
})();
```
This returns a promise, which you can then await elsewhere.

michael-volynets: Totally agree

extension/js/common/composer.ts
```diff
@@ -99,6 +100,10 @@ export class Composer {
   public canReadEmails: boolean;
   public initialized: Promise<void>;
 
+  public get Recipients() {
```
tomholub: What does this do? I'm not familiar with the `get` syntax.

extension/js/common/composer/composer-contacts.ts
```diff
@@ -30,6 +30,10 @@ export class ComposerContacts extends ComposerComponent {
     private myAddrsOnKeyserver: string[] = [];
     private recipientsMissingMyKey: string[] = [];
 
+    public get Recipients() {
```
tomholub: Same question here

extension/js/common/composer/composer-contacts.ts
```diff
@@ -185,60 +196,59 @@ export class ComposerContacts extends ComposerComponent {
...
+        const invalidEmails = validationResult.invalid.filter(em => !!em); // remove empty strings
+        if (force && invalidEmails.length) {
+            recipientsToEvaluate = [...recipientsToEvaluate, ...this.createRecipientsElements(container, invalidEmails, true)];
+            inputTo.val('');
+        } else {
+            inputTo.val(validationResult.invalid.join(','));
+        }
+        this.evaluateRecipients(recipientsToEvaluate).catch(Catch.reportErr);
         this.composer.resizeInputTo();
-        this.composer.debug(`parseRenderRecipients(${errsMode}).9`);
```
tomholub: Please add back the debug calls. You have just refactored a bunch of important code, and if there is a bug that is later discovered by the CI, you will be really thankful that the debug logs are present.

extension/js/common/composer/composer-contacts.ts
```diff
@@ -267,9 +277,9 @@ export class ComposerContacts extends ComposerComponent {
     }
 
     private authContacts = async (acctEmail: string) => {
-        const lastRecipient = $('.recipients span').last();
-        this.composer.S.cached('input_to').val(lastRecipient.text());
-        lastRecipient.last().remove();
+        const lastRecipient = this.addedRecipients[this.addedRecipients.length - 1];
+        this.composer.S.cached('input_to').val(lastRecipient.email);
+        this.removeReceiver(lastRecipient.element);
```
tomholub: please rename to `removeRecipient`

extension/js/common/composer.ts
michael-volynets: It's a getter, I have learned it from C#,

I can write the same thing using the method, something like this:
```TS
public getRecipients() {
  return this.addedRecipients;
}
```

but the way I've written it takes less code

extension/js/common/composer/composer-contacts.ts
michael-volynets: Answered above

extension/js/common/composer/composer-contacts.ts
michael-volynets: Oh, yes, I wanted to add back the debug calls at the end of refactoring but I forgot,
Adding

extension/js/common/composer.ts
tomholub: hmm... no problem with it. But it is not used anywhere else in the code, so it looks alien. If we were to start using it, we should refactor the whole codebase with such code for consistency. But maybe in this case the benefit is very marginal, so I think it's not worth it. If you don't mind, please use the more common syntax.

We will definitely one day do a big code style / consistency refactor, and can consider it at that point.

michael-volynets: Okay, I agree :+1:

extension/js/common/composer.ts
```diff
@@ -118,11 +119,23 @@ export class Composer {
...
+    this.initialized = (async () => {
       await this.initComposeBox();
       await this.initActions();
       await this.checkEmailAliases();
-    });
+    })();
+  }
+
+  updateFooterIcon = (include?: boolean) => {
+    if (typeof include === 'undefined') { // decide if pubkey should be included
```
tomholub: should say "decide if footer should be included"

tomholub:
Because the git diff will be much much smaller when you leave the methods in original order.

michael-volynets:
Okay, got it,
I will undo the commit with methods rearranging

tomholub:
I wish git was smarter about this and showed me that it's exactly the same method, just moved. Or show me that the method also has a 1 line change or something. But it is what it is :) ...

Issue #1883, opened 2019-08-07
sometimes public key save button has white background


tomholub:
@michael-volynets 

this is a fairly serious usability bug because it seems like there is no button, and this kind of thing would not be noticed by automated tests.

![image](https://user-images.githubusercontent.com/6306961/62660032-7f934b00-b95c-11e9-984d-d7a51a24ff91.png)

After highlighting:

![image](https://user-images.githubusercontent.com/6306961/62659853-0d226b00-b95c-11e9-9991-a657b63d42c5.png)

```
-----BEGIN PGP PUBLIC KEY BLOCK-----
Version: FlowCrypt 6.9.3 Gmail Encryption
Comment: Seamlessly send and receive encrypted email

xsFNBF1LPPYBEAC6RQetgLc+EtjqDhvthakUH2/eynsEchRb3WkwylkBX5nF
zOC/ZcT/yJCZxQV8e5VgFLRoUpsfu4TJCZYO3xJqb1JTQKJ6Ss75DG/aPnOZ
/EaqhCyKUZ52FFpEhly5xb6xUF7kLLeMqCkiZPQKoP/J0LZPjOrZ3vIJSw/U
raVGCxgJymZIGuI4LJZmHbwMTjyFaIIeMy0UogxnfQJo7k6+JfVwbojJOwMd
5nOShKQrfqhuibg+NBba7GTnRcO7od/rDxWq6Lx29namqnH9Dc1vlCDYitKx
Ki7XCJeRqdri7v8V8SQZXQjpJvXdHEWNahzYGMASjR+Thg4ueOqwBQWql0zZ
FcOR0cGO3oqyjldxw4GY5kYCRAGQgOLzQ/tah1VGsOf/GM+W/Od2fhP8EsH5
atrsBh3pGbF/nf/YgtZVTsF/Rad9T5jz5CR0rSjkTCyrYTX3YDEIVQ2iAie1
x1XXNsdwNMmxvhT8xSWN5KHbpvG5D+UjHmYYQ7S2ExqR1QBMCYULdKdD+pcy
KqxKosEdi3h16nrw+pIKackXtxwPBvVXmRD8Rn++0oCMARfwqKcmvDpyPU8x
tA+FJQO3nBBO5KhlJdxijW37YmolUf93YQBoO6/tRdkwUHXseAU3BDMd/F0z
[...]

PR #1886, opened 2019-08-08
Issue #1883 Leave else case empty instead of 'inherit'

michael-volynets:
Issue #1883 
Time Spent: 20m

Yes, sorry, that was my bad,
I don't know why but I thought that `background-color: inherit;` inherits the previous color but it doesn't, it inherits something else (maybe I will investigate it later)


michael-volynets:
Now, It looks like this:
image

tomholub:
Much better, thanks for taking care of it late at night

tomholub:
Maybe inherit inherits from the parent. Which is white

tomholub:
I removed the screen because it contained user email. Best is to black out any emails present in screen shots

michael-volynets:
No problem, I'm trying to be available anytime :)

Yes, sure, I will black emails next time

PR #1917, opened 2019-08-12
Use Logical OR instead of Bitwise OR

davalapar:


tomholub:
I was actually imagining a bitwise OR may be more appropriate. It would be weird if `SWIFT` and `TS` were constants that just have `true`/`false`. More likely you would expect them to be some form of identifiers, possibly binary-spaced. OpenPGP protocol does this a lot, for example in the `key flags` which is one byte describing a bunch of capabilities. For example number 8 (100 in binary) may mean encryption, 4 (010) may mean signing, etc. To put them together, you use bitwise OR. So I may need to #wontfix this PR :+1:

PR #1922, opened 2019-08-13
Close the 'New Secure Message' on Esc keydown

### LINKED ISSUE #1908: Escape should perform save & close action
Closing Secure Compose Window should be possible with a keyboard. The common practice for closing popups is to use `Esc`. Gmail, for example, is using that.

At the moment, nothing is happening with Secure Compose Window when pressing `Esc`.

limonte:
Closes #1908

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -903,7 +903,7 @@ export class Composer {
     if (Catch.browser().name === 'firefox') { // the padding cause issues in firefox where user cannot click on the message password
       this.S.cached('input_text').css({ 'padding-top': 0, 'padding-bottom': 0 });
     }
-    this.S.cached('body').keydown(Ui.escape(() => !this.composeWindowIsMinimized && $('.close_new_message').click()));
+    this.S.cached('body').keydown(Ui.escape(() => !this.composeWindowIsMinimized && !this.urlParams.isReplyBox && $('.close_new_message').click()));
     this.S.cached('body').keypress(Ui.ctrlEnter(() => !this.composeWindowIsMinimized && this.extractProcessSendMsg()));
     this.S.cached('send_btn').click(Ui.event.prevent('double', () => this.extractProcessSendMsg()));
     this.S.cached('send_btn').keypress(Ui.enter(() => this.extractProcessSendMsg()));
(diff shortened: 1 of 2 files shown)


limonte:
@tomholub I also noticed that Gmail has another handy Esc handler: in case the popup is minimized, Esc will restore it to normal/maximize state. Should I align with Gmal in this case?

extension/js/common/composer.ts
```diff
@@ -903,6 +903,7 @@ export class Composer {
     if (Catch.browser().name === 'firefox') { // the padding cause issues in firefox where user cannot click on the message password
       this.S.cached('input_text').css({ 'padding-top': 0, 'padding-bottom': 0 });
     }
+    this.S.cached('body').keydown(Ui.escape(() => !this.composeWindowIsMinimized && this.app.closeMsg()));
```
tomholub: My recommendation here is to manually trigger the event like so: `$('.close_new_message').click()` and let the handler worry about it, for two reasons:
1) there are additional checks in that handler, and this way all paths will go through the same "chokepoint". See the handler:
```js
      $('.close_new_message').click(Ui.event.handle(async () => {
        if (!this.isSendMessageInProgress || await Ui.modal.confirm('A message is currently being sent. Closing the compose window may abort sending the message.\nAbort sending?')) {
          this.app.closeMsg();
        }
      }, this.getErrHandlers(`close message`)));
```
2) It should not trigger for message replies. I believe, doing what I'm suggesting, it won't. I'm guessing, though have not tested it, that your method may (or it may close another open new message, although you are focused onto a reply frame? That could be good or bad, but either way somewhat unexpected).

Please try the above suggestion, and then test that it does not affect a reply window, and test that if you have a reply window focused, it should not (I think?) affect other open compose windows.

tomholub:
> @tomholub I also noticed that Gmail has another handy Esc handler: in case the popup is minimized, Esc will restore it to normal/maximized state. Should I align the Esc behavior with Gmal?

That's really weird, I would maybe hold off on this one unless you feel strongly in favor of it.

limonte:
> That's really weird, I would maybe hold off on this one unless you feel strongly in favor of it.

This behavior was a bit of surprise for me as well, so I don't feel strongly about it.

extension/js/common/composer.ts
limonte: > My recommendation here is to manually trigger the event like so: $('.close_new_message').click() and let the handler worry about it

Makes sense, done!

> It should not trigger for message replies

It was closing by Esc, so I added the additional check in https://github.com/FlowCrypt/flowcrypt-browser/pull/1922/commits/5346285d668d6b49d590cc9478b700dcd84b933a

>  test that if you have a reply window focused, it should not (I think?) affect other open compose windows.

Tested, works nicely :+1:

PR #1926, opened 2019-08-13
Remove setting padding-top and padding-bottom to 0 in Firefox

### LINKED ISSUE #1902: Top padding is missing for the mail body
Spacing between the horizontal line and email body is missing:

image

This is a tiny detail, but my brain is reacting to it as "something is broken here" :smile:

limonte:
Closes #1902 

I wasn't able to find the original PR adding this workaround, probably because it was added long ago and not needed anymore.

```js
// the padding cause issues in firefox where user cannot click on the message password
```

I guess "the message password" is this one:

image

I manually tested it with changes from this PR, works nicely (Firefox 68)

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -900,9 +900,6 @@ export class Composer {
   private renderComposeTable = async () => {
     this.debugFocusEvents('input_text', 'send_btn', 'input_to', 'input_subject');
     this.S.cached('compose_table').css('display', 'table');
-    if (Catch.browser().name === 'firefox') { // the padding cause issues in firefox where user cannot click on the message password
-      this.S.cached('input_text').css({ 'padding-top': 0, 'padding-bottom': 0 });
-    }
     this.S.cached('body').keydown(Ui.escape(() => !this.composeWindowIsMinimized && !this.urlParams.isReplyBox && $('.close_new_message').click()));
     this.S.cached('body').keypress(Ui.ctrlEnter(() => !this.composeWindowIsMinimized && this.extractProcessSendMsg()));
     this.S.cached('send_btn').click(Ui.event.prevent('double', () => this.extractProcessSendMsg()));
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -668,7 +668,8 @@ export class Composer {
         this.refBodyHeight = this.S.cached('body').height() || 605;
       }
       const attListHeight = $("#att_list").height() || 0;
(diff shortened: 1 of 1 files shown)


tomholub:
Thanks. I think it had to do with reply, possibly, when you then write more text and cause the frame to expand with the text size. I'll also double check before merging

limonte:
Hm, @tomholub please postpone the review of this PR, I've found the regression produced by it:

image

Let me check what's causing it.

tomholub:
Yup, that's where your new margins went :laughing: 

This was probably the reason I was somewhat content with the ugly fix. If you could figure this out, that would be good.

Please note that window sizing on Firefox uses completely different logic compared to Chrome. On Chrome, we let the browser do it. In Firefox, we do it manually...

```js
  /**
   * On Firefox, we have to manage textbox height manually. Only applies to composing new messages
   * (else ff will keep expanding body element beyond frame view)
   * A decade old firefox bug is the culprit: https://bugzilla.mozilla.org/show_bug.cgi?id=202081
   *
   * @param updateRefBodyHeight - set to true to take a new snapshot of intended html body height
   */
  public setInputTextHeightManuallyIfNeeded = (updateRefBodyHeight: boolean = false) => {
    if (!this.urlParams.isReplyBox && Catch.browser().name === 'firefox') {
      this.S.cached('input_text').css('height', '0');
      let cellHeightExceptText = 0;
      for (const cell of this.S.cached('all_cells_except_text')) {
        cellHeightExceptText += $(cell).is(':visible') ? ($(cell).parent('tr').height() || 0) + 1 : 0; // add a 1px border height for each table row
      }
      if (updateRefBodyHeight || !this.refBodyHeight) {
        this.refBodyHeight = this.S.cached('body').height() || 605;
      }
      const attListHeight = $("#att_list").height() || 0;
      this.S.cached('input_text').css('height', this.refBodyHeight - cellHeightExceptText - attListHeight);
    }
  }
```

tomholub:
I think I have the solution, you just subtract the padding from it.

```diff
- this.S.cached('input_text').css('height', this.refBodyHeight - cellHeightExceptText - attListHeight);
+ const inputTextVerticalPadding = 2 * ????;
+ this.S.cached('input_text').css('height', this.refBodyHeight - cellHeightExceptText - attListHeight - inputTextVerticalPadding);
```

Sorry that we're doing the browser's job here, I didn't find a better way to work around it.

limonte:
Should be good now, thank you for hints @tomholub! Please go ahead with the review.

PR #1932, opened 2019-08-14
Focus iframe before focusing #input_to (Firefox needs that)

### LINKED ISSUE #1910: Click the "Secure Compose" button should autofocus the 'Add Recepient' field
At the moment, it's required either to use mouse or press Tab multiple times to focus into the Secure Compose Window after clicking the "Secure Compose"

The 'Add Recepient' field should be auto-focused after clicking the "Secure Compose" button to smooth the UX by eliminating the additional unnecessary click/keystrokes.

limonte:
Closes #1910 

Used this SO answer as a hint: https://stackoverflow.com/a/1784925


limonte:
> Very good, please call it FocusFrame instead of FocusNewMessage because it may be useful for any frame (eg pass phrase frame, reply frame in some situations, or whatever frames we have in the future).

Done.

PR #1933, opened 2019-08-14
Height of the minimized iframe should be equal header's height

### LINKED ISSUE #1929: Minimized state looks cut off
The minimized state looks a bit odd because of the vertical alignment of its text:

image

The height of minimized popup is 36px

image

When the height of header is 40px, so minimized is cut by 4px

I believe the height of minimized state should be equal to header's height and be 40px

limonte:
Closes #1929

@tomholub WDYT about adding

```css
transition: height 0.1s ease-in-out
```

to `#new_message`? To me personally, the height animation makes an impression better.


tomholub:
Overall I usually go easy on animations (I disabled them in SweetAlert2 too). But let's give it a try, I think it may make sense here to make it clear where the minimized msg went.

Issue #1065, opened 2018-04-25
bcc and cc


chuacw:
Is there no BCC, and CC available when sending emails?

tomholub:
Not yet. I have a list of users who are interested in this, I'll leave this open to track the progress. Thanks.

tomholub:
ref https://mail.google.com/mail/u/0/#sent/FMfcgxwBTjvMwPmdNCmSVkvHtRwndRph

tomholub:
This is a relatively UI-heavy change. The result should work very similarly to what our Android app does with cc/bcc.

tomholub:
This should also support drag and drop between to:, cc: and bcc:, else it's hard to use.

chuacw:
That's true. The program model and user model needs to match here.

tomholub:
This is a relatively large change, and will need thorough tests added.

michael-volynets:
@tomholub Can you please share the screenshot of Flowcrypt Android App where I can see BCC/CC?

tomholub:
I recommend to download the app and test it out (do you have an android phone?) so that you can mimic the whole interaction. There is a down chevron on the right side next to recipient, which you can click to view cc and bcc options.

tomholub:
![image](https://user-images.githubusercontent.com/6306961/61883309-b479b980-aee9-11e9-983d-64a2505ed0ff.png)

Can otherwise follow Gmail

michael-volynets:
In this issue, I will also refactor `composer-contacts.ts` to make it easier to understand code and to make the logic with `CC/BCC` similar to `to`, (perhaps I will have to delete `parserenderRecipients` and rewrite or split it)

tomholub:
Of course. Is it possible to refactor it in a separate PR with no functionality change? I know it will be more work for you, but it will be much easier for me to review, and therefore less likely that a mistake will slip through.

michael-volynets:
Yes, sure,

Okay, I will try to refactor it in separate PR

tomholub:
Great. Let's do the refactor first, then you'll add new functionality after that.

michael-volynets:
@tomholub, will we allow users to add the same email address to the to/cc/bcc if the email was already added in another input?
I checked how Gmail does it and they allow to do it but they even allow to add the same email addresses in one input.

In my opinion, it doesn't make sense to allow doing it but there maybe there are some situations that I don't know where it will be useful

tomholub:
[...]

PR #1939, opened 2019-08-14
Issue #1065 Add cc/bcc logic

michael-volynets:
Issue #1065 

Here I added the JS logic to send CC/BCC messages,

Things to do:

- [x] Collapse inputs
- [x] Add CC/BCC buttons in the top-right corner
- [x] Drag & Drop
- [x] Save CC/BCC recipients in Draft
- [x] Fix&Write tests
- [x] Fix CSS in reply window


extension/js/common/composer/interfaces/composer-types.ts
```diff
@@ -2,12 +2,18 @@
 
 'use strict';
 
-export type RecipientElement = {
-  email: string,
-  element: HTMLElement
+export type SendingType = 'to' | 'cc' | 'bcc';
+
+export interface BaseRecipient {
```
tomholub: Why not `Recipients = {to: string[], cc?: string[], bcc?: string[]}` ? That will simplify your code

michael-volynets: Sorry, but I don't exactly understand how it will simplify the code.

Could you please explain to me a little bit more what do you mean?

tomholub: You will be able to remove this whole chunk of code:
```js
    const to: string[] = [];
    const cc: string[] = [];
    const bcc: string[] = [];
    for (const recipient of recipients) { // Using JS loop to not dublicate .filter().map()
      switch (recipient.sendingType) {
        case 'to':
          to.push(recipient.email);
          break;
        case 'cc':
          cc.push(recipient.email);
          break;
        case 'bcc':
          bcc.push(recipient.email);
          break;
      }
    }
```
createMsgObj API will become more obvious, from:
```js
await Google.createMsgObj(acctEmail, acctEmail, [{ email: acctEmail, sendingType: 'to' }], subj, { 'text/html': emailMsg }, emailAtts);
```
to:
```js
await Google.createMsgObj(acctEmail, acctEmail, { to: [acctEmail] }, subj, { 'text/html': emailMsg }, emailAtts);
```

michael-volynets: Okay, got it,
If it's the only reason so I can just change the code this way:
```TS
    const result: Recipients = { to: [], cc: [], bcc: [] };
    // tslint:disable-next-line: standard-loops
    recipients.forEach(r => result[r.sendingType].push(r.email));
```

Because there are some places where it's better to use `BaseRecipient` or `RecipientElement` like here:

if we change method signature from this:
```TS
  public static createMsgObj = async (
    acctEmail: string, from: string = '', recipientElements: RecipientElement[] = [], subject: string = '', by: SendableMsgBody, atts?: Att[], threadRef?: string
  ): Promise<SendableMsg> 
```
to that:
```TS
  public static createMsgObj = async (
    acctEmail: string, from: string = '', recipients: Recipients, subject: string = '', by: SendableMsgBody, atts?: Att[], threadRef?: string
  ): Promise<SendableMsg> 
```
Then we need to join (`Array.join`) all `to:`, `cc:` and `bcc:` values to make this validation working. (I guess I will need to do these joins in other places too).
```TS
    if (!recipientElements.length) {
      throw new Error('The To: field is empty. Please add recipients and try again');
    }
    const invalidEmails = recipientElements.filter(r => !Str.isEmailValid(r.email));
    if (invalidEmails.length) {
      throw new Error(`The To: field contains invalid emails: ${invalidEmails.join(', ')}\n\nPlease check recipients and try again.`);
    }
```

So, I would like to leave it as is but change that large `for` cycle to the small one I have written above.

tomholub: I'll take a look later today

tomholub: In `if (!recipientElements.length) {` you should actually do `recipients.to.length`. Email without `To` header is invalid (to my knowledge). So you would, as it is now, have to rewrite that with a filter anyway.

Maybe you meant `Array.concat` ? You can do `Array.prototype.concat(...Object.values(recipients))`. If there are more than two places that need this, you could put it as ComposerRecipients (or whatever is the class name) method, etiher normal or static, eg `asArray(recipients: Recipients): string[]` or something like that.

I didn't examine the code in deep detail so I may be missing something. I just feel like `(acctEmail, acctEmail, [{ email: acctEmail, sendingType: 'to' }]` is a bit awkward as an API.

michael-volynets: Okay, got it, 
Yes, it makes sense, I will rewrite it according to what you just wrote.

michael-volynets: It should be easier to use  :+1:

tomholub: That's always nice :)

michael-volynets:
@tomholub I want to test it on Firefox Browser but I get this error when I run the command `run_firefox`:
image

I checked if there is the Firefox profile and everything is okay,
I didn't have that issue before.

Does web-ext work well on your side?

tomholub:
Yes, I use it all the time. I would try to rename the profile folder, then try to re-run it. It may create another profile in its place, and run just fine (maybe?).

tomholub:
On linux

michael-volynets:
Okay, I will check it,
It worked before, I don't know what's happened

Thanks.

michael-volynets:
@tomholub I think you can start reviewing it,
The functionality is almost done, I listed things that aren't done in the PR header.

tomholub:
Thanks! I'll take a look

extension/js/common/composer/interfaces/composer-types.ts
michael-volynets: It isn't resolved yet

I will resolve it when I finish all other things

michael-volynets:
@tomholub I have a question,

When I'm replying to a message can I somehow extract CC & BCC headers? 
I found that we parse `To` Header from the HTML page and I can extract CC/BCC the same way but they are in the popover:
image
and I find little difficult to parse it and maybe parsing is not the best way for doing that.

Maybe you know some other ways to extract that data?

extension/js/common/composer/composer-contacts.ts
```diff
@@ -623,7 +623,7 @@ export class ComposerContacts extends ComposerComponent {
       container.find('r_loader').remove();
     }
     const restHTML = `<span class="rest"><span id="rest_number"></span> others</span>`;
-    container.html(restHTML);
+    container.html(restHTML); // xss-direct
```
tomholub: Yup, correct

michael-volynets:
Also, I've got an issue with tests,

I added `data-test='collapsed'` attribute to the element that hides inputs.
```HTML
          <div class="collapsed" data-test="collapsed">
            <span class="placeholder">Recipients</span>
            <div class="email_preview"></div>
          </div>
``` 
And when I want to click on it, it seems to not be invoked (I added some logs in the methods to check and I couldn't see them when the test was running).
This code `await composePageOrFrame.click('@collapsed');` should open inputs so I can type the recipients' emails but it doesn't work. Do you know why it may not work?

tomholub:
You should call it in line with other ones, like `action-expand-cc-bcc-fields`

But that's not causing the issue. I'm not sure what is.

michael-volynets:
Okay, I will rename it and try to investigate why that happens

tomholub:
You could try to `waitAny` it, then get the element and do `await el.click()`, as a way to debug it.

You could also execute javascript on the frame, with `$(".yourclass").click()` in it as a test, through puppeteer.

tomholub:
Other than the send_from issue, please also align cc/bcc to the right side to have the same distance to the right as the input field has from the left:

image

Currently it's not clear how was the distance from the right side chosen :)

image

Please also make it lowercase: `cc`, `bcc`.

tomholub:
> @tomholub I have a question,
> 
> When I'm replying to a message can I somehow extract CC & BCC headers?
> I found that we parse `To` Header from the HTML page and I can extract CC/BCC the same way but they are in the popover:
> image
> and I find little difficult to parse it and maybe parsing is not the best way to doing that.
> 
> Maybe you know some other ways to extract that data?

Sorry, I missed that. Don't parse it from dom. Get it from Gmail API. We are already getting this info from Gmail, I'll find where.

tomholub:
image

Here. Just change:
```diff
-if (!isReplyBox || (threadId && threadId !== threadMsgId && to.length && subject)) {
+if (!isReplyBox || !threadId) {
   return; // either not a reply box, or reply box & has all needed params
 }
```
That means we will be ignoring the `to` and `subject` in url, and always fetching metadata from Gmail if it's a reply.

Later we will get rid of these header url parameters altogether (except for threadId), but in another PR (this PR is already getting big).

In the `gmailMsg` object you will find the needed headers.

michael-volynets:
Thanks for the previous answer.

Also, Is there any reason why you added this CSS class:
```CSS
table#compose td.input.show_send_from {
    height: 84px;
}
```
That code adds these paddings:
image

But without that paddings it looks like this:
image

In my opinion, it looks better without those paddings and I want to remove that CSS code, what do you think?

tomholub:
The goal of the css, I think, was to make all of these 3 cells the same height:
 - recipients
 - sender
 - subject

I tried to measure it and it's already a bit broken (prod):

image

Which explains the broken padding, visually (supposed to be the same distance):

image

If that was made the same distance, it will all work:

image

For the one that you proposed, I'm lacking any sense of consistency, don't know what to make of it:

image

So from my point of view, as it is now in production is slightly broken, your "before" screenshot is somewhat more broken, and your proposal also :) I'm not a UI designer (as you can tell) but I really have a thing for UIs with apparent visual consistency (especially padding / sizing / positioning).

michael-volynets:
Got the goal of that code, Thanks!

I will fix the CSS to make the paddings the same in all 3 inputs (Recipients&From&Subject).

tomholub:
If you will be editting it, here is I think the best form of consistency, which is a bit more compact:

image

Remember to make sure it's consistent regardless is "from" is showing or not.

michael-volynets:
Looks great, 

Got it

michael-volynets:
@tomholub Could you please review it again?

I have done all your requested changes,

Thanks

tomholub:
Some user:

image

tomholub:
Sorry, I somehow missed the message above. I'll take another look

tomholub:
I have to block merging for now for lack of visual consistency :laughing: ... sorry I'm quite anal about these things :roll_eyes: 

image means good, image means off by 1 pixel (should still be fixed), image means too far off

1) Outer border padding. All of the red lines are supposed to be the same distance, but they all seem to differ.

image

2) vertically it seems to need a bit more space. When I measured it, it's just 1px, but my eye is maybe really missing that pixel :)

image

3) cc and bcc have inconsistent margin from the other elements, vertically

image

extension/chrome/elements/reply_pubkey_mismatch.ts
```diff
@@ -103,10 +104,15 @@ Catch.try(async () => {
     return { 'In-Reply-To': '', 'References': '' };
   };
 
+  // TODO: Test this method before merging
   $('#send_btn').off().click(Ui.event.prevent('double', async target => {
     $(target).text('sending..');
     const body = { 'text/plain': $('#input_text').get(0).innerText };
-    const message = await Google.createMsgObj(acctEmail, from, to, subject, body, [att], threadId);
+    const recipients: BaseRecipient[] = to.map(email => {
+      const recipient: BaseRecipient = { email, sendingType: 'to' };
```
tomholub: Please use Recipients instead of BaseRecipient[]

extension/chrome/settings/modules/backup.ts
```diff
@@ -218,7 +218,7 @@ Catch.try(async () => {
   const doBackupOnEmailProvider = async (acctEmail: string, armoredKey: string) => {
     const emailMsg = String(await $.get({ url: '/chrome/emails/email_intro.template.htm', dataType: 'html' }));
     const emailAtts = [asBackupFile(acctEmail, armoredKey)];
-    const msg = await Google.createMsgObj(acctEmail, acctEmail, [acctEmail], GMAIL_RECOVERY_EMAIL_SUBJECTS[0], { 'text/html': emailMsg }, emailAtts);
+    const msg = await Google.createMsgObj(acctEmail, acctEmail, [{ email: acctEmail, sendingType: 'to' }], GMAIL_RECOVERY_EMAIL_SUBJECTS[0], { 'text/html': emailMsg }, emailAtts);
```
tomholub: Please use Recipients here

extension/js/common/api/email_provider_api.ts
```diff
@@ -9,28 +9,32 @@ import { Value, Dict, Str } from '../core/common.js';
...
 export type ProviderContactsQuery = { substring: string };
-export type SendableMsg = { headers: Dict<string>; from: string; to: string[]; subject: string; body: SendableMsgBody; atts: Att[]; thread?: string; };
+export type SendableMsg = { headers: Dict<string>; from: string; recipients: Recipients; subject: string; body: SendableMsgBody; atts: Att[]; thread?: string; };
 type LastMsgHeaders = { lmSender: string | undefined, lmRecipients: string[], lmReplyTo: string | undefined };
 
 export class EmailProviderApi extends Api {
 
   public static createMsgObj = async (
-    acctEmail: string, from: string = '', to: string[] = [], subject: string = '', by: SendableMsgBody, atts?: Att[], threadRef?: string
+    acctEmail: string, from: string = '', recipientElements: BaseRecipient[] = [], subject: string = '', by: SendableMsgBody, atts?: Att[], threadRef?: string
```
tomholub: Please use Recipients here

extension/js/common/composer.ts
```diff
@@ -970,7 +978,7 @@ export class Composer {
         document.getElementById('input_text')!.focus(); // #input_text is in the template
         // Firefox will not always respond to initial automatic $input_text.blur()
         // Recipients may be left unrendered, as standard text, with a trailing comma
-        await this.composerContacts.parseRenderRecipients(this.S.cached('input_to_container')); // this will force firefox to render them on load
+        this.composerContacts.parseRenderRecipients(this.S.cached('input_to')).catch(Catch.reportErr); // this will force firefox to render them on load
```
tomholub: Any reason not to await this? Are there external API calls?

tomholub:
By the way I like this :+1:

image

It wasn't part of the issue, but I think it increases usability. I'll go ahead and adjust the design a bit.

tomholub:
Another issue: it seems to create broken html?
```html
<span class="email_address has_pgp" "="">
```

tomholub:
Sometimes I click away from the recipients and it does not result in showing recipient preview. I'll make a video.

tomholub:
When I clicked away from input, the clicked back, it stopped responding to input / stopped adding email bubbles! This one is quite serious, if you can reproduce it, please also add a test for this before fixing it.

image

I keep clicking back and forth, pressing tab, etc, the last two emails do not get processed.

tomholub:
It's definitely on the right track, but has many quirks that need fixing - see all above.

extension/js/common/composer.ts
michael-volynets: Sorry, I think I added it accidentally because it could block the thread.

I tested it along with `await` and it works the same way, thanks!

michael-volynets:
https://github.com/FlowCrypt/flowcrypt-browser/pull/1939#issuecomment-524644262

Thanks, I that was my typo mistake

michael-volynets:
https://github.com/FlowCrypt/flowcrypt-browser/pull/1939#issuecomment-524644479
Fixed,

Now when you will focus RecipientElement and then blur it the inputs will collapse

michael-volynets:
> When I clicked away from input, the clicked back, it stopped responding to input / stopped adding email bubbles! This one is quite serious, if you can reproduce it, please also add a test for this before fixing it.
> 
> image
> 
> I keep clicking back and forth, pressing tab, etc, the last two emails do not get processed.
> 
> Tried hitting enter, and it processed them both as one email, even though there is a space between them:
> 
> image

Sorry, but we didn't have that functionality before, we can only add 2 or more emails at one time when we write them by comma
e.g. `test@test.com,test2@test.com` then they will be processed.

But if you noticed it I will add it, it won't take a lot of time to write that.

michael-volynets:
Little question, I found this code: 
```TS
  private getReplyParams = (convoRootEl: JQueryEl): FactoryReplyParams => {
    const headers = Google.determineReplyCorrespondents(this.acctEmail, this.addresses, {
      lmSender: this.domGetMsgSender(convoRootEl),
      lmRecipients: this.domGetMsgRecipients(convoRootEl),
      lmReplyTo: undefined, // cannot get this info from DOM, will be later added from API
    });
....
  }
```
In `gmail_element_replacer.ts` file.
Can I get rid of this? I mean I want to replace it with Gmail API call to fetch the to/cc/bcc data, will it be okay?

michael-volynets:
> Little question, I found this code:
> 
> ```ts
>   private getReplyParams = (convoRootEl: JQueryEl): FactoryReplyParams => {
>     const headers = Google.determineReplyCorrespondents(this.acctEmail, this.addresses, {
>       lmSender: this.domGetMsgSender(convoRootEl),
>       lmRecipients: this.domGetMsgRecipients(convoRootEl),
>       lmReplyTo: undefined, // cannot get this info from DOM, will be later added from API
>     });
> ....
>   }
> ```
> 
> In `gmail_element_replacer.ts` file.
> Can I get rid of this? I mean I will only use Gmail API call (in composer.ts) to fetch the to/cc/bcc data, will it be okay?

Resolved

michael-volynets:
@tomholub, It's ready for review,

One thing I have done without asking you is I removed this arrow in the top-right corner:
image

because it brought me some problems with CSS and I didn't want to spend a lot of time on that.
Now it immediately opens emails aliases in the reply box (the same way like you open a compose window from the button in Gmail sidebar). If it isn't okay for you (if you want to leave that arrow) so I can somehow try to fix that.

tomholub:
> Now it immediately opens emails aliases in the reply box (the same way like you open a compose window from the button in Gmail sidebar). If it isn't okay for you (if you want to leave that arrow) so I can somehow try to fix that.

It was done this way because reply box has much more limited vertical space. I'll try it.

extension/js/common/composer/composer-draft.ts
```diff
@@ -67,12 +67,6 @@ export class ComposerDraft extends ComposerComponent {
         await this.abortAndRenderReplyMsgComposeTableIfIsReplyBox('!armored');
         return false;
       }
-      if (String(parsedMsg.headers.subject)) {
```
tomholub: Removing this introduced a bug, subjects are not loaded when I load a draft.

extension/js/common/composer/composer-draft.ts
```diff
@@ -113,7 +107,13 @@ export class ComposerDraft extends ComposerComponent {
         }
         const subject = String(this.composer.S.cached('input_subject').val() || this.urlParams.subject || 'FlowCrypt draft');
         const to = this.composer.getRecipients().map(r => r.email); // else google complains https://github.com/FlowCrypt/flowcrypt-browser/issues/1370
-        const mimeMsg = await Mime.encode(body, { To: to.join(','), From: this.composer.getSender(), Subject: subject }, []);
+        const recipients: Recipients = { to: [], cc: [], bcc: [] };
+        // tslint:disable-next-line: standard-loops
+        this.composer.getRecipients().forEach(r => recipients[r.sendingType]!.push(r.email));
```
tomholub: standard for/of loop can be used here

tomholub: That's why we have a linter for it! :)

extension/js/common/composer/composer-draft.ts
```diff
@@ -173,38 +173,30 @@ export class ComposerDraft extends ComposerComponent {
     }
   }
 
-  private async decryptAndRenderDraft(encryptedArmoredDraft: string, headers: { from?: string; to: string[] }): Promise<boolean> {
+  private async decryptAndRenderDraft(encryptedArmoredDraft: string, headers: { subject?: string, from?: string; to: string[], cc: string[], bcc: string[] }): Promise<boolean> {
     const passphrase = await this.app.storagePassphraseGet();
     if (typeof passphrase !== 'undefined') {
       const result = await PgpMsg.decrypt({ kisWithPp: await Store.keysGetAllWithPp(this.urlParams.acctEmail), encryptedData: Buf.fromUtfStr(encryptedArmoredDraft) });
       if (result.success) {
+        if (headers.subject) {
```
tomholub: This looks like it should work, but for whatever reason, when tested, subject is not added. Not sure why

extension/js/common/core/mime.ts
```diff
@@ -111,20 +114,16 @@ export class Mime {
...
-    let headerFrom;
-    if (Array.isArray(parsedMimeMsg.headers.from) && parsedMimeMsg.headers.from[0] && parsedMimeMsg.headers.from[0].address) {
-      headerFrom = parsedMimeMsg.headers.from[0].address;
-    }
-    if (Array.isArray(parsedMimeMsg.headers.to)) {
-      for (const to of parsedMimeMsg.headers.to) {
-        if (to.address) {
-          headerTo.push(String(to.address));
+  private static headerGetAddress = (parsedMimeMsg: MimeContent, headersNames: Array<SendingType | 'from'>) => {
+    const result: { from: string[], to: string[], cc: string[], bcc: string[] } = { from: [], to: [], cc: [], bcc: [] };
```
tomholub: Why would from be an array?

extension/js/common/core/mime.ts
```diff
@@ -111,20 +114,16 @@ export class Mime {
...
+    for (const hdrName of headersNames) {
+      if (Array.isArray(parsedMimeMsg.headers[hdrName])) {
```
tomholub: Hopefully `from` should never be an array? So not sure if this would work?

test/source/mock.ts
```diff
@@ -12,7 +12,7 @@ export const mock = async (logger: (line: string) => void) => {
       return; // missing mock data, not yet used
     }
     const filename = `${email.replace(/[^a-z0-9]+/g, '')}.json`;
-    const url = `https://github.com/FlowCrypt/flowcrypt-bin/blob/master/gmail-mock-data/${filename}?raw=true`;
+    const url = `https://github.com/michael-volynets/flowcrypt-bin/blob/update-mock-data/gmail-mock-data/${filename}?raw=true`; // UNDO: before commit
```
tomholub: Note for me - remember to fix

extension/js/common/core/mime.ts
michael-volynets: I agree, it would be better to have `from` like `string | undefined`.

I wanted to have all headers the same type so I can work with them easier

PR #1940, opened 2019-08-14
Issue 1924 contacts keyboard navigation

### LINKED ISSUE #1924: Recipients dropdown isn't keyboard friendly
When the recipients dropdown is shown, it should be possible to:

- navigate it by pressing `Arrow up` and `Arrow Down`
- select the recipient by pressing `Enter`
- close the dropdown by pressing `Esc`

### LINKED ISSUE #1934: When I press `ESC` button while I'm searching contacts the whole compose window closes
When I press the `ESC` button if I'm searching for some contact I expect that the search box will close but not the whole compose window.

I checked how Gmail does it and they close search box firstly and on the second `ESC` press they close the whole window

limonte:
Closes #1924 
Closes #1934

Also re-enabled `Esc` for closing the 'New Secure Message' window (rollback of https://github.com/FlowCrypt/flowcrypt-browser/commit/1ba6be58ec109c015daf3e996df91f953381aa8f)

### DIFF
--- a/extension/js/common/composer/composer-contacts.ts
+++ b/extension/js/common/composer/composer-contacts.ts
@@ -108,6 +108,7 @@ export class ComposerContacts extends ComposerComponent {
       currentActive.click(); // select contact
       return true;
     } else if (e.key === 'ArrowUp') {
+      e.preventDefault();
       let prev = currentActive.prev();
       if (!prev.length) {
         prev = this.composer.S.cached('contacts').find('ul li.select_contact').last();
@@ -116,6 +117,7 @@ export class ComposerContacts extends ComposerComponent {
       prev.addClass('active');
       return true;
     } else if (e.key === 'ArrowDown') {
+      e.preventDefault();
       let next = currentActive.next();
       if (!next.length) {
         next = this.composer.S.cached('contacts').find('ul li.select_contact').first();
--- a/extension/js/common/composer/composer-contacts.ts
+++ b/extension/js/common/composer/composer-contacts.ts
@@ -90,15 +90,22 @@ export class ComposerContacts extends ComposerComponent {
    */
   private recipientInputKeydownHandler = (e: JQuery.Event<HTMLElement, null>): boolean => {
     const currentActive = this.composer.S.cached('contacts').find('ul li.select_contact.active');
-    if (e.key === 'Enter') {
-      currentActive.click();
-      return true;
-    } else if (e.key === 'Escape') {
+    if (e.key === 'Escape') {
       if (this.composer.S.cached('contacts').is(':visible')) {
         e.stopPropagation();
         this.hideContacts();
         this.composer.S.cached('input_to').focus();
       }
+      return true;
+    } else if (!currentActive.length) {
+      return false; // all following code operates on selected currentActive element
+    } else if (e.key === 'Tab') {
+      e.preventDefault(); // don't switch inputs
+      e.stopPropagation(); // don't switch inputs
+      currentActive.click(); // select contact
+      return true;
+    } else if (e.key === 'Enter') {
+      currentActive.click(); // select contact
       return true;
     } else if (e.key === 'ArrowUp') {
       let prev = currentActive.prev();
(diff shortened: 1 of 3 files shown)


extension/js/common/composer/composer-contacts.ts
```diff
@@ -38,13 +38,51 @@ export class ComposerContacts extends ComposerComponent {
   }
 
   initActions(): void {
-    this.composer.S.cached('input_to').on('keyup', Ui.event.prevent('veryslowspree', () => this.searchContacts()));
+    let preventSearchContacts = false
+    this.composer.S.cached('input_to').on('keyup', Ui.event.prevent('veryslowspree', () => {
+      if (!preventSearchContacts) {
+        this.searchContacts()
```
limonte: @tomholub what would be the right way to resolve the `Promises must be handled appropriately` error from the code quality check?

extension/js/common/composer/composer-contacts.ts
```diff
@@ -38,13 +38,51 @@ export class ComposerContacts extends ComposerComponent {
...
 
   initActions(): void {
-    this.composer.S.cached('input_to').on('keyup', Ui.event.prevent('veryslowspree', () => this.searchContacts()));
+    let preventSearchContacts = false;
+    this.composer.S.cached('input_to').on('keyup', Ui.event.prevent('veryslowspree', () => {
+      if (!preventSearchContacts) {
+        this.searchContacts();
+      }
+    }));
     this.composer.S.cached('input_to').on('keydown blur', Ui.event.handle(async (target, e) => {
```
tomholub: Please move this whole function body into a private method on the class, eg `recipientInputKeydownHandler`, with a comment that it does keyboard navigation in search results etc.

Then you could still leave the `blur` handler directly in here, that one is not too long.

I have a thing for functions/methods that can ideally fit on my screen vertically :)

extension/js/common/composer/composer-contacts.ts
tomholub: Most likely you're missing an `await` somewhere. Possibly in front of `this.searchContacts()`, which means you'd also put `async` in front of `() => {`.

If you don't await a promise, any promise rejections will be ignored / errors silently swallowed, which is what it's trying to keep you from doing.

extension/js/common/composer/composer-contacts.ts
limonte: Done :+1:

extension/js/common/composer/composer-contacts.ts
limonte: Done, thanks! :+1:

limonte:
> Looks good! I'm thinking, it could also select the recipient with tab. I'll try to add support for that as I test it out.

Gmail works that way, so yes, I agee with `Tab` acting like `Enter`, thanks for taking care of this :+1:

tomholub:
Finally I get to break your ux code instead of you breaking mine :)

https://vimeo.com/353863120 - choose HD so you can see it properly

I think the hover handler is making other competing elements active. I would make sure, on hover, that active status is removed from all other elements, which should fix this.

tomholub:
While @limonte is addressing this, @michael-volynets would you please do a code review?

limonte:
> Finally I get to break your ux code instead of you breaking mine :)
> 
> https://vimeo.com/353863120 - choose HD so you can see it properly
> 
> I think the hover handler is making other competing elements active. I would make sure, on hover, that active status is removed from all other elements, which should fix this.

Good catch, done :+1:

Also, I added `cursor: pointer` fro dropdown items to indicate for users that those are clickable.

limonte:
> Everything looks good and it's much better to work with contacts this way. Thanks!
> But I found a minor thing that we could fix:
> 
> When you press 'ArrowDown' the cursor position is set to the end of the text:
> image
> 
> When you press 'ArrowUp' the cursor position is set to the beginning of the text:
> image
> 
> I think would be better if it always was at the end of the text.

Good catch, thanks @michael-volynets! Done :+1:

PR #1949, opened 2019-08-15
Add new icon for 'public key' button in footer

### LINKED ISSUE #1904: The "Public key" icon looks off
image

This icon looks odd and interrupting the consistency of UI.

Something like the "key" icon would be expected here by me:

image

limonte:
Closes #1904

The icon is taken from https://icons8.com/icon/11169/certificate, rezized and manually fixed to be sharp in that small size (20x20px)

Result:

image

@tomholub could you please apply the same "blue" filter to `footer-icon-pubkey-standard.png` and regenerate `footer-icon-pubkey-active.png`?


tomholub:
I have no clue how I did the blue tbh, most likely I got the RGB value of the blue colour, then used the "bucket fill" icon in an image editor such as gimp to change the gray to a single shade blue. 

If the gray is multi-shade, then I apply a blue filter of some sort to get it close enough to intended blue, then use the bucket with the "correct blue" to fill the main lines. 

Not a designer though (you can tell by looking at the pk icon which I made ... )

limonte:
I made two active icons by using Gimp's `Color -> Colorize...` tool with these values:

image

tomholub:
Perfect, that's useful to know. We have an internal design repo, I'll post it there.

tomholub:
It's good, better than before. I think, for the blue highlight, we should be using a "filled up icon" version of the icon, in blue. That would make it more obvious. I'll make that another issue, for sometime later.

PR #1953, opened 2019-08-16
Include #input_from to taborder

### LINKED ISSUE #1925: evaluate removing tabindex="-1" on #input_from
From http://webaim.org/techniques/keyboard/tabindex

> A tabindex="-1" value removes the element from the default navigation flow (i.e., a user cannot tab to it), but it allows it to receive programmatic focus, meaning focus can be set to it from a link or with scripting.** This can be very useful for elements that should not be tabbed to, but that may need to have focus set to them.

As a user, I'd like to be able to access `#input_from` from a keyboard. Is there any particular reason to exclude it from the taborder?

limonte:
Closes #1925 

> When there is just one email in input_from, the from-select-option was hidden but I think it would still trap focus when tabbing. So you would need to double-tab to get from recipients to body. It wasn't smooth. Please test if it's still the case, and if it is, see what could be done about it.

Tested, no longer the case, because there's no `<select>` when there is just one email:

image


limonte:
Also, I removed `outline: none;` from `<select>`

[There's a reason for that ugly outline!](https://stackoverflow.com/a/21758143)

tomholub:
Arguably, the one on ff looks a lot better.

PR #1954, opened 2019-08-16
Align focused state of action buttons with their hover state

### LINKED ISSUE #1906: Controls in Secure Compose Window should be tab-friendly and indicate the focused state
Some people prefer not to touch their mouse/trackpad and use a keyboard for navigating fields, buttons and other controls:

Peek 2019-08-12 12-19

In Secure Compose Window I can't always see which element is currently focused, all these buttons aren't accessible with a keyboard:

image

limonte:
Closes #1906

Replaced action `<div>`s with `<button>`s with appropriate `tabindex`es so they are focusable and clickable with keyboard (Enter or Space).

What I wasn't able to achieve is to trigger file upload dialog with the keyboard. By hitting Space or Enter the button is getting clicked, not the `<input>` inside it. @tomholub Do you maybe have any hint which could help here?

### DIFF
--- a/extension/chrome/elements/compose.htm
+++ b/extension/chrome/elements/compose.htm
@@ -95,16 +95,16 @@
         <td class="bottom">
           <div id="send_btn" class="button green" data-test="action-send" tabindex="6"><i class=""></i><span>Encrypt and
               Send</span></div>
-          <div class="icon action_sign" data-test="action-switch-to-sign" tabindex="-1"
-            title="Sign this message instead of encrypting it"><i alt="sign"></i></div>
-          <div class="icon attach" id="fineuploader_button" tabindex="-1" title="Attach a file"><img
-              src="/img/svgs/paperclip-icon.svg" class="attach-file" alt="Attach File"></div>
+          <button class="icon action_sign" data-test="action-switch-to-sign" tabindex="7"
+            title="Sign this message instead of encrypting it"><i alt="sign"></i></button>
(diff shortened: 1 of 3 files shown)


tomholub:
Sorry, I don't have many hints for that. We use fine-upploader which does some magic with a checkbox, I think. I forgot what exactly was it doing tbh. We could address that sometime in the future, it's possible we will one day deprecate fine-upploader.

limonte:
> it's possible we will one day deprecate fine-upploader.

It is deprecated (archived) already. So, it makes sense to use something else in the future.

tomholub:
Yes. But it works well for now, so we'll see. Eventually.

tomholub:
Testing this, I found another issue: tab all the way to the `footer`, then enter, then escape - whole window goes away - not good.

limonte:
> Testing this, I found another issue: tab all the way to the footer, then enter, then escape - whole window goes away - not good.

I'll address that in scope of this PR :+1:

tomholub:
Ah, ok, I'll close both #1955 and #1906 when this merges. Thanks!

limonte:
> Testing this, I found another issue: tab all the way to the footer, then enter, then escape - whole window goes away - not good.

whole window is not going away now, but Esc isn't closing the footer modal, even though `closeOnEsc` is `true` by default. I believe the reason for that is iframe inside featherlight and the `keydown` events aren't bubbling to featherlight handlers.

limonte:
@tomholub should we change #1955 to "Esc isn't closing footer modal"?

tomholub:
Can

PR #1956, opened 2019-08-16
Issue 1907 focus trap

### LINKED ISSUE #1907: Secure Compose Window should trap the focus
When inside Secure Compose Window, the focus should be "trapped" in the same way as native Gmail popup is doing. Here's the demo of "trapped" focus http://davidtheclark.github.io/focus-trap/demo/

This is important in order to make keyboard-ninjas and disabled people happy with FlowCrypt.

limonte:
Closes #1907


tomholub:
To understand, the goal of trapping focus is that you can tab around the compose window in circles without focusing things outside the iframe?

limonte:
> To understand, the goal of trapping focus is that you can tab around the compose window in circles without focusing things outside the iframe?

Yes.

limonte:
The PR isn't ready for review yet. I'll request your review once it's ready.

PR #1959, opened 2019-08-17
Implement clicking the 'Attach a file' button with a keyboard

### LINKED ISSUE #1958: The 'Attach a file' button in not clickable with the keyboard
Nothing is happening when hitting Enter or Space while `.icon.attach` is focused.

limonte:
Fixes #1958

I made the `<input>` inside of `.icon.attach` focusable, that seems to be the only way to make it clickable with the keyboard.

Also, moved the `title` attr to `<input>`. Previously, browsers were showing the default input title instead of "Attach a file":

image

### DIFF
--- a/extension/chrome/elements/compose.htm
+++ b/extension/chrome/elements/compose.htm
@@ -97,8 +97,7 @@
               Send</span></div>
           <button class="icon action_sign" data-test="action-switch-to-sign" tabindex="7"
             title="Sign this message instead of encrypting it"><i alt="sign"></i></button>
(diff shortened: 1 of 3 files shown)


tomholub:
Best is to avoid force push when possible, just tag on another commit

limonte:
> Best is to avoid force push when possible, just tag on another commit

In that case there will be multiple "Fix code quality errors" commits, but I'm fine with that :+1:

tomholub:
Yeah, just fix them in a few commits. I don't mind that, since I squash all merges into one commit anyway.

Issue #1972, opened 2019-08-23
double check and test export-secret-subkeys GNU extension


tomholub:
after #1965 + add a test with key from https://github.com/openpgpjs/openpgpjs/issues/251#issuecomment-466794922

tomholub:
+ update to OpenPGP.js v4.6.0
+ gpg-dummy key tested in flowcrypt-mobile-core

PR #1973, opened 2019-08-23
Issue 1972 gpg dummy fix and openpgpjs upgrade

tomholub:
#1972 3-4 hours


extension/js/common/core/pgp.ts
```diff
@@ -15,7 +15,7 @@ import { FcAttLinkData } from './att.js';
 import { Buf } from './buf.js';
 import { Xss } from '../platform/xss.js';
 
-const openpgp = requireOpenpgp();
+export const openpgp = requireOpenpgp();
```
tomholub: Should make sure we use this throughout the project

Issue #1914, opened 2019-08-12
unexpectedly seeing feedback screen after subscription?


limonte:
As a result of successful payment, this form was shown:

![image](https://user-images.githubusercontent.com/6059356/62866087-b2509100-bd18-11e9-8738-46d7715b622e.png)

I'm not sure why exactly the feedback form was shown to me after payment, I was expecting something like 'We received your payment, thank you' message.

tomholub:
Very weird, that's not where you're supposed to end up after a successful payment, at all.

tomholub:
@michael-volynets please could you take a minute sometime take a look at the code and see why would this happen? If you cannot find anything, I'll try to see if I can also reproduce it too, and if yes, try to investigate it myself.

tomholub:
I will make another issue for the missing initial subscription email.

michael-volynets:
Looking into it

michael-volynets:
I've subscribed to Flowcrypt and I didn't get that form. The notification was sown and the modal dialog was closed.
I also checked the code and it's written correctly. Any code paths don't drive us to that form.

@limonte Had you been subscribed to the `trial` plan before you subscribed to  `Pro` plan?

limonte:
> @limonte Had you been subscribed to the trial plan before you subscribed to Pro plan?

Yes.

michael-volynets:
I just tried this way and I have the same result as before.
I don't really know what could provoke that behavior

tomholub:
Thank you @michael-volynets for looking into it. Reading through the discussion, I have a guess.

@limonte went to the settings through a help icon here: 

![image](https://user-images.githubusercontent.com/6306961/62959917-180c4e00-bde9-11e9-8e90-1a6174ceec34.png)

Then he ended up on url: moz-extension://39553fca-76d4-4791-bf00-b4dfede6fd45/chrome/settings/index.htm?acctEmail=help.cryptup.org%40gmail.com&page=%2Fchrome%2Fsettings%2Fmodules%2Fhelp.htm

Which would show the help page, which he closed. But the url is left unchanged.

Then he subscribed right away, and at the end of the subscription process, there is probably a `window.reload()`. Guess which url it reloaded & guess which page will come up again?

If my guessing is correct, the solution would be to, when user closes a modal, make sure there is no `page=` url param on the host page. I think this can be done by adjusting url history through js, without reloading the page.

[...]

PR #1984, opened 2019-09-03
Issue #1914,  remove 'page' query parameter when closing the modal di…

michael-volynets:
…alog

Issue #1914 
Time Spent: 1h 30m


tomholub:
I like the use of URLSearchParams, wasn't aware of it

Issue #1981, opened 2019-09-02
This OpenPGP key is not usable: add option to view the pubkey


tomholub:
![image](https://user-images.githubusercontent.com/6306961/64131478-5e394980-cdb8-11e9-80e2-74d6f13fc07d.png)

There should still be option to view the full pubkey

PR #1986, opened 2019-09-03
Issue #1981, Allowed users to see details of not usable pub key

michael-volynets:
Issue #1981 

Time Spent: ~1h 30m

### DIFF
--- a/extension/chrome/elements/pgp_pubkey.ts
+++ b/extension/chrome/elements/pgp_pubkey.ts
@@ -83,7 +83,7 @@ Catch.try(async () => {
           const pubToEmail = (pubkey: OpenPGP.key.Key) => Str.parseEmail(pubkey.users[0].userId ? pubkey.users[0].userId!.userid : '').email;
           Xss.sanitizeAppend('.add_contact', Xss.escape(' for ' + pubs.map(pubToEmail).filter(e => !!e).join(', ')));
         }
-        Xss.sanitizePrepend('#pgp_block.pgp_pubkey .result', `<span>This message includes a Public Key for <span class= "email">${emailText}</span>.</span>`);
+        Xss.sanitizePrepend('#pgp_block.pgp_pubkey .result', `<span>This message includes a Public Key for <span class= "email">${Xss.escape(emailText)}</span>.</span>`);
         $('.pubkey').addClass('good');
         setBtnText().catch(Catch.reportErr);
       }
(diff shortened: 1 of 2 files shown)


extension/chrome/elements/pgp_pubkey.ts
```diff
@@ -67,18 +67,25 @@ Catch.try(async () => {
...
           }
         } else {
-          $('.email').text('more than one person');
+          emailText = 'more than one person';
           $('.input_email').css({ display: 'none' });
           const pubToEmail = (pubkey: OpenPGP.key.Key) => Str.parseEmail(pubkey.users[0].userId ? pubkey.users[0].userId!.userid : '').email;
           Xss.sanitizeAppend('.add_contact', Xss.escape(' for ' + pubs.map(pubToEmail).filter(e => !!e).join(', ')));
         }
+        $('#pgp_block.pgp_pubkey .result')
+          .prepend(`<span>This message includes a Public Key for <span class= "email">${emailText}</span>.</span>`); // xss-direct
```
tomholub: Xss vulnerability.

tomholub: Always use the `sanitize` methods unless you have a reason not to. Don't be complacent.

michael-volynets: Got It,

Corrected

extension/chrome/elements/pgp_pubkey.ts
```diff
@@ -67,18 +67,24 @@ Catch.try(async () => {
...
-            $('.email').text(email);
           }
         } else {
-          $('.email').text('more than one person');
+          emailText = 'more than one person';
           $('.input_email').css({ display: 'none' });
           const pubToEmail = (pubkey: OpenPGP.key.Key) => Str.parseEmail(pubkey.users[0].userId ? pubkey.users[0].userId!.userid : '').email;
           Xss.sanitizeAppend('.add_contact', Xss.escape(' for ' + pubs.map(pubToEmail).filter(e => !!e).join(', ')));
         }
+        Xss.sanitizePrepend('#pgp_block.pgp_pubkey .result', `<span>This message includes a Public Key for <span class= "email">${emailText}</span>.</span>`);
```
tomholub: Better, but please don't fully rely on sanitizing only, because it may also fail. If it's a simple case like this, it is appropriate to xss escape the email before running it through sanitization.

Security is always applied in layers, like an onion. If one layers fails, which it may anytime, there should be another layer that ends up saving your (user's) butt.

michael-volynets: Okay, Got It,
I will sanitize any text that we get from the user :)

Issue #1741, opened 2019-06-06
Change auto-generated text when we forward encrypted message


michael-volynets:
Now we have this generated content if we want to forward someone a message:
```
Some text here.

Forwarded message:

> Some text there.
```

I think would be better if we give more information to the user about the message (From Who, Time, Subject, To), the same way as Gmail does.

```
Some text here.

---------- Forwarded message ---------
From: User Name <User Email>
Date: mm/dd/yyyy
Subject: Re: Subject Line here
To: <User Email>

> Some text there.
```

tomholub:
Make it do whatever the reply does today. But maybe it's already fixed?

PR #1987, opened 2019-09-03
Issue #1741, Changed text and corrected quoting algorithm

michael-volynets:
Issue #1741 
Time Spent: 1h 15m


michael-volynets:
Do you mean to add exactly the same text as when we're replying to some message?

tomholub:
Yes, I don't see any reason not to :) When you forward a message, you want to know what the other person said and when.. and we already have a mechanism for that. I'd reuse the same.

michael-volynets:
Okay, sorry then, I didn't catch it from the first time.

michael-volynets:
Done

Issue #1928, opened 2019-08-13
small layout issue contact page settings


tomholub:
Around the name input, misaligned 

![image](https://user-images.githubusercontent.com/6306961/62969617-37ad7180-bdfd-11e9-86da-bef1c2218f92.png)

PR #1989, opened 2019-09-04
issue #1928, Corrected margins of placeholder

michael-volynets:
Issue #1928,
Time Spent: 30m

Result:
image

I fixed alignment of `Full Name:` ​​placeholder but actually I'd like to get rid of it because we don't have any similar inputs in the extension (I haven't seen that large placeholder) and I think our users can intuitively understand that there should be their name so I think without a placeholder it would look better.


tomholub:
One day we have to completely redo that form, but it's a low priority for now. Thanks!

Issue #1977, opened 2019-09-02
don't render private keys in pubkey import


michael-volynets:
Screenshot:
![image](https://user-images.githubusercontent.com/49104848/64116521-38517c00-cd9b-11e9-869c-65b179e26b22.png)

tomholub:
We should not even be displaying this at all. In the public key import screen, when we encounter private keys, it should probably show text `Found Private Key ${longid}. To import Private Keys, see Additional Settings -> My Keys.`

You can do that by replacing the block with a "text" block that includes this sentence in `content` (I think).

michael-volynets:
I just researched that we are able to put there everything we want :laughing: (except plain/text :))
I tried to put there PGP message and it decrypted it.

I will add validation for the PGP keys.

PR #1991, opened 2019-09-04
Issue #1977, Added validation and error messages

michael-volynets:
Issue #1977,
Time Spent: 1h

### DIFF
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -143,7 +143,7 @@ Catch.try(async () => {
               container.append(replacedHtmlSafe); // xss-safe-factory
             }
           } else {
-            container.append(`<div class="bad">Skipping found '${Pgp.friendlyMsgBlockTypeName(block.type)}'</div>`); // xss-direct
+            Xss.sanitizeAppend(container, `<div class="bad">Skipping found '${Pgp.friendlyMsgBlockTypeName(block.type)}'</div>`)
           }
         }
         container.css('display', 'block');
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -111,31 +111,43 @@ Catch.try(async () => {
   const actionProcessBulkImportTextInput = async () => {
     try {
       const value = Str.normalize(String($('#bulk_import .input_pubkey').val())).trim();
+      if (!value) {
+        await Ui.modal.warning('Please paste public key(s).');
+        return;
+      }
       const normalizedLongid = KeyImportUi.normalizeLongId(value);
-      let key: string;
+      let pub: string;
       if (normalizedLongid) {
         const data = await Keyserver.lookupLongid(acctEmail, normalizedLongid);
         if (data.pubkey) {
-          key = data.pubkey;
+          pub = data.pubkey;
         } else {
           await Ui.modal.warning('Could not find any Public Key in our public records that matches this fingerprint or longid');
           return;
         }
       } else {
-        key = value;
+        pub = value;
       }
-      if (Pgp.key.isPossiblePublic(key)) {
-        const replacedHtmlSafe = XssSafeFactory.replaceRenderableMsgBlocks(factory, key);
-        if (replacedHtmlSafe && replacedHtmlSafe !== value) {
-          $('#bulk_import #processed').html(replacedHtmlSafe).css('display', 'block'); // xss-safe-factory
-          $('#bulk_import .input_pubkey, #bulk_import .action_process, #file_import #fineuploader_button').css('display', 'none');
-        } else {
-          await Ui.modal.warning('Could not find any new public keys');
+      let { blocks } = Pgp.armor.detectBlocks(pub);
+      blocks = blocks.filter((b, i) => blocks.findIndex(f => f.content === b.content) === i); // remove duplicates
+      if (!blocks.length) {
+        await Ui.modal.warning('Could not find any new public keys.');
+      } else if (blocks.length === 1 && blocks[0].type === 'plainText') { // Show modal because users could make a mistake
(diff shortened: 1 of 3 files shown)


extension/chrome/settings/modules/contacts.ts
```diff
@@ -109,26 +110,32 @@ Catch.try(async () => {
 
   const actionProcessBulkImportTextInput = async () => {
     try {
-      const value = String($('#bulk_import .input_pubkey').val());
+      const value = Str.normalize(String($('#bulk_import .input_pubkey').val())).trim();
       const normalizedLongid = KeyImportUi.normalizeLongId(value);
-      let pubkey: string;
+      let key: string;
```
tomholub: I like to call pubkeys explicitly either `pub` or `pubkey` instead of just `key`, for easier security reviews.

Private keys could be called `prv` for example.

Basically nothing should be called just `key` because that's ambiguous, but I know there are many such places that we have to refactor.

extension/chrome/settings/modules/contacts.ts
```diff
@@ -109,26 +110,32 @@ Catch.try(async () => {
...
         }
       } else {
-        pubkey = value;
+        key = value;
       }
-      const replacedHtmlSafe = XssSafeFactory.replaceRenderableMsgBlocks(factory, pubkey);
-      if (replacedHtmlSafe && replacedHtmlSafe !== value) {
-        $('#bulk_import #processed').html(replacedHtmlSafe).css('display', 'block'); // xss-safe-factory
-        $('#bulk_import .input_pubkey, #bulk_import .action_process, #file_import #fineuploader_button').css('display', 'none');
+      if (Pgp.key.isPossiblePublic(key)) {
```
tomholub: You don't have to add these `isPossible` methods. You can instead just get the `MsgBlock[]`, and then filter the blocks, replacing all that is not a public key with a note `Skipping found XXX`, where XXX is the name of the block you don't like.

That way you still render the public keys, and there will be notes around them that the other things were skipped (not modals)

extension/chrome/settings/modules/contacts.ts
michael-volynets: Yes, maybe it's an incorrect name of the variable,

I wanted to name it just `key` because we don't know whether the key is public or private or even some other string. We don't know what users can enter :)

I will rename it.

michael-volynets:
+ ~45m

extension/chrome/settings/modules/contacts.ts
```diff
@@ -109,26 +110,44 @@ Catch.try(async () => {
...
         }
       } else {
-        pubkey = value;
+        pub = value;
       }
-      const replacedHtmlSafe = XssSafeFactory.replaceRenderableMsgBlocks(factory, pubkey);
-      if (replacedHtmlSafe && replacedHtmlSafe !== value) {
-        $('#bulk_import #processed').html(replacedHtmlSafe).css('display', 'block'); // xss-safe-factory
+      let { blocks } = Pgp.armor.detectBlocks(pub);
+      blocks = blocks.filter((b, i) => blocks.findIndex(f => f.content === b.content) === i); // remove duplicates
```
tomholub: I didn't think of removing duplicates. That's a good idea!

extension/chrome/settings/modules/contacts.ts
```diff
@@ -109,26 +110,44 @@ Catch.try(async () => {
...
+      if (!blocks.length) {
+        await Ui.modal.warning('Could not find any new public keys.');
+      } else if (blocks.length === 1 && blocks[0].type === 'plainText') { // Show modal because users could make a mistake
```
tomholub: Very good

extension/js/common/core/pgp.ts
```diff
@@ -289,6 +289,46 @@ export class Pgp {
     },
   };
 
+  public static friendlyMsgBlockTypeName = (type: MsgBlockType) => {
```
tomholub: Ok, why not. Could be implemented as a dictionary lookup (less code to write), but doesn't matter either way.

extension/chrome/settings/modules/contacts.ts
```diff
@@ -109,26 +110,44 @@ Catch.try(async () => {
...
+      } else { // Render Results
+        const container = $('#bulk_import #processed');
+        for (const block of blocks) {
+          if (block.type === 'publicKey') {
+            const replacedHtmlSafe = XssSafeFactory.replaceRenderableMsgBlocks(factory, String(block.content));
+            if (replacedHtmlSafe && replacedHtmlSafe !== value) {
+              container.append(replacedHtmlSafe); // xss-safe-factory
+            }
+          } else {
+            container.append(`<div class="bad">Skipping found '${Pgp.friendlyMsgBlockTypeName(block.type)}'</div>`); // xss-direct
```
tomholub: Just use sanitize append. I know it's safe here, and I know it's less efficient. But I will have less things to scrutinize during security reviews, and maybe it will be cheaper and simipler to do external security reviews, because they will also have fewer things to review. Thanks!

extension/js/common/core/pgp.ts
michael-volynets: I also was thinking about dictionary. I don't remember exactly why I choose method implementation of it.

I will replace it with the dictionary in the next commit. It will be better

tomholub: But to be honest, it's not really necessary to have a friendly name for these. It's a bit of a fringe case, so I don't mind showing `signedMsg` as a part of an error, that only happens on one screen.

extension/chrome/settings/modules/contacts.ts
michael-volynets: Thanks

extension/chrome/settings/modules/contacts.ts
michael-volynets: Thanks

Issue #1980, opened 2019-09-02
test that subject is rendered when loading from a draft


tomholub:
![image](https://user-images.githubusercontent.com/6306961/64127112-ee1ec980-cd9f-11e9-9a5e-a7ed230eca01.png)

PR #1992, opened 2019-09-04
Issue #1980, added drafts to samples, corrected generating of samples…

michael-volynets:
…, wrote two tests

Issue #1980,
Time Spent: 4h 30m

Please also merge a PR in the repo with mocks and re-run tests

### DIFF
--- a/test/source/mock.ts
+++ b/test/source/mock.ts
@@ -12,7 +12,7 @@ export const mock = async (logger: (line: string) => void) => {
       return; // missing mock data, not yet used
     }
     const filename = `${email.replace(/[^a-z0-9]+/g, '')}.json`;
-    const url = `https://github.com/FlowCrypt/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
+    const url = `https://github.com/michael-volynets/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
     const filepath = `./test/samples/${filename}`;
     if (!existsSync(filepath)) {
       const { body, statusCode } = await request.get({ url, encoding: null }); // tslint:disable-line:no-null-keyword
--- a/test/source/mock.ts
+++ b/test/source/mock.ts
@@ -12,7 +12,7 @@ export const mock = async (logger: (line: string) => void) => {
       return; // missing mock data, not yet used
     }
     const filename = `${email.replace(/[^a-z0-9]+/g, '')}.json`;
-    const url = `https://github.com/michael-volynets/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
+    const url = `https://github.com/FlowCrypt/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
     const filepath = `./test/samples/${filename}`;
     if (!existsSync(filepath)) {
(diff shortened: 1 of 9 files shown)


tomholub:
There's one problem here - I always tend to delete drafts from the drafts folder, because they keep accumulating.

So I'm almost certain one day I'll delete them and then we will be missing them.

So I'm thinking if, for drafts, a different solution could be used.. maybe we could hard code them into the data class, but just for that account.. what do you think?

michael-volynets:
Agree, if we accidentally delete a draft that will be a problem.
But on the other hand, it will be harder to add a new test draft message if drafts are hardcoded.

I marked drafts that are used in tests with a label `tests/draft` to decrease chances of deleting them :)

So, if you want I can put them in another `.json` file and parse them separately.

tomholub:
I'll take a look

test/source/mock.ts
```diff
@@ -12,7 +12,7 @@ export const mock = async (logger: (line: string) => void) => {
       return; // missing mock data, not yet used
     }
     const filename = `${email.replace(/[^a-z0-9]+/g, '')}.json`;
-    const url = `https://github.com/FlowCrypt/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
+    const url = `https://github.com/michael-volynets/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
```
michael-volynets: Undo before merge

tomholub: You can undo it now, I have merged the -bin repo

michael-volynets: Okay, thanks!

test/source/tests/tests/compose.ts
```diff
@@ -316,6 +316,28 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
...
+    ava.test('compose[global:compatibility] - loading drafts - reply', testWithNewBrowser(async (t, browser) => {
+      await BrowserRecipe.setUpCommonAcct(t, browser, 'compatibility');
+      const appendUrl = 'isReplyBox=___cu_true___&threadId=16cfa9001baaac0a&skipClickPrompt=___cu_false___&ignoreDraft=___cu_false___&threadMsgId=16cfa9001baaac0a';
+      const initialScript = () => {
+        chrome.storage.local.set({ 'cryptup_flowcryptcompatibilitygmailcom_drafts_reply': { '16cfa9001baaac0a': 'r-1543309186581841785' } });
+      };
+      const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility', { appendUrl, hasReplyPrompt: true, skipClickPropt: true, initialScript });
+      await composePage.click('@action-expand-cc-bcc-fields');
+      await isRecipientElementsExists(composePage, { to: true, cc: false });
+      expect(await composePage.read('@input-body')).not.to.be.empty;
```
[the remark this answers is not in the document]
tomholub: I have to unresolve this one - please add some specific text to look for (eg just one line).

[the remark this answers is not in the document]
michael-volynets: Sorry, I missed that

Issue #2001, opened 2019-09-05
Got an error when asking sender to update a key


michael-volynets:
![image](https://user-images.githubusercontent.com/49104848/64348884-a3e75380-cffe-11e9-9dc7-6c909831e58b.png)

tomholub:
Please include a test for this one

michael-volynets:
Ok

michael-volynets:
Issue fixed, Writing a test

Issue #1999, opened 2019-09-05
public key block should not be center aligned


tomholub:
![image](https://user-images.githubusercontent.com/6306961/64337765-89e24c80-cfcf-11e9-9cc5-74c9880c3651.png)

michael-volynets:
Fixed in #2009

PR #2009, opened 2019-09-05
Issue 2001 error when asking sender for key

michael-volynets:
Issue #2001 and #1999
Time Spent: 1h 30m

### DIFF
--- a/test/source/tests/tests/compose.ts
+++ b/test/source/tests/tests/compose.ts
@@ -307,7 +307,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
       await ComposePageRecipe.sendAndClose(composePage);
     }));
 
-    ava.test('compose[global:compose] - standalone - CC&BCC test reply', testWithSemaphoredGlobalBrowser('compatibility', async (t, browser) => {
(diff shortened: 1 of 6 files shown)


test/source/tests/tests/compose.ts
```diff
@@ -339,6 +339,21 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
...
+      const replyMismatchPage = await ComposePageRecipe.openReplyKeyMismatch(t, browser, 'compatibility', appendUrl);
+      await Util.sleep(3);
+      const emailsPreview = await replyMismatchPage.waitAny('@email-preview');
+      const recipients = await emailsPreview.$$('span');
+      expect(recipients.length).to.equal(1);
+      const recipientEmail = await (await recipients[0].getProperty('textContent')).jsonValue() as string;
+      expect(recipientEmail).to.equal('flowcrypt.compatibility@gmail.com');
+      const text = await replyMismatchPage.read('@input-body');
+      expect(text).to.include('I was not able to read your encrypted message because it was encrypted for a wrong key.');
+      expect(await replyMismatchPage.isElementPresent('@attachment')).to.be.true;
```
tomholub: Please also try to send the message out, to see that there's no error (through mock).

michael-volynets: Ok

michael-volynets: Thanks

Issue #1869, opened 2019-07-29
load expired key from attester instead of dropping it


tomholub:
When #1763 / #415 lands, we can improve it further.

Currently when a key is loaded from Attester and it's expired, it will be dropped and it will look like recipient doesn't use encryption at all:

![image](https://user-images.githubusercontent.com/6306961/62078351-f5503600-b23b-11e9-836e-8598d486f4bc.png)

Instead, it should load the public key anyway, which should cause it to show in orange.

Then we should add test for this.

Email `expired.on.attester@domain.com`

tomholub:
We had a customer bumping exactly into this issue, so they'll appreciate if we can fix it.

tomholub:
Two customers, actually.

PR #2010, opened 2019-09-05
Issue #1869, not to skip expired keys

michael-volynets:
Issue #1869,

Time Spent: 10m


extension/chrome/elements/compose.ts
```diff
@@ -154,7 +154,7 @@ Catch.try(async () => {
             if (!parsed.keys[0]) {
               Catch.log('Dropping found but incompatible public key', { for: email, err: parsed.err ? ' * ' + parsed.err.join('\n * ') : undefined });
               lookupResult.pubkey = null; // tslint:disable-line:no-null-keyword
-            } else if (! await parsed.keys[0].getEncryptionKey()) {
+            } else if (! await parsed.keys[0].getEncryptionKey() && !Pgp.key.usableButExpired(parsed.keys[0])) { // Not to skip expired keys
```
tomholub: Should be `! await Pgp.keyusableButExpired`

Please also add a test. Thanks!

tomholub: Actually try this:
```js
} else if (! await Pgp.key.usable(...) && ! await Pgp.key.expired(...)) {
```
It should be more accurate.

tomholub: Actually - please not just test that it becomes orange, but try to send a message out. There is supposed to be a modal dialog asking the user to confirm to send the message even though key is expired - please test that it shows the dialog too, and then that the message gets sent properly.

michael-volynets: Sorry, but we can't send emails if there is at least on expired email.
Do I have to allow it?

michael-volynets: I wrote the tests, but I've got an issue:
When sending the message with an expired pub. Key it shows the modal that says there is expired key then I click `Ok` and it tries to send the message but openpgp.js returns me this:
image

Do you know what this means?

Also, I found two ways I can get an error when sending an email with an expired recipient:
1) It's the way where I have `confirm` modal dialog
2) This one when I see error modal immediately:
```TS
    if (usableTimeFrom > usableTimeUntil) { // used public keys have no intersection of usable dates
      await Ui.modal.error('The public key of one of your recipients has been expired for too long.\n\nPlease ask the recipient to send you an updated Public Key.');
      throw new ComposerResetBtnTrigger();
    }
```

I guess I need to test this way too. But I don't know how to get the key that will fit this expression.

tomholub: I think I'll take a look at this, it's getting somewhat complex.

michael-volynets: Ok, I pushed the changes

test/source/tests/tests/compose.ts
```diff
@@ -385,20 +400,16 @@ const baseQuotingTest = async (composePage: Controllable, textToInclude: string)
...
-const isRecipientElementsExists = async (controllable: ControllablePage, exists: { to?: string[], cc?: string[] }) => {
-  const containerTo = await controllable.waitAny('@container-to', { visible: false });
-  const containerCc = await controllable.waitAny('@container-cc', { visible: false });
-  const recipientsTo = await containerTo.$$('.recipients span');
-  const recipientsCc = await containerCc.$$('.recipients span');
-  const checkIfRecipientsContains = async (emails: string[] | undefined, recipientElements: ElementHandle<Element>[]) => {
-    if (!emails || !emails.length) {
-      return;
-    }
+const expectRecipientElements = async (controllable: ControllablePage, expected: { to?: string[], cc?: string[], bcc?: string[] }) => {
```
tomholub: For whatever reason, after merging, this kept giving me trouble (tests failing).

I was not exactly able to understand how this function works, so that I refactored it in a way I can understand. But it still doesn't work. I think I made it a little stricter - if you leave out one of to/cc/bcc in the expected emails, but the DOM does have it, it would fail.

Anyway, I'm off to sleep now. When you have a chance, please take a look at this and see why is it failing.

michael-volynets: Done,
I corrected this function.

There were some issues with finding a container with recipients in it and also I removed `.not` when excepting `.length` property of two arrays

Issue #1721, opened 2019-05-25
fix message compose when my own key expired


tomholub:
currently it will say that recipient's email is expired.

It should instead say that my own key expired, and offer to renew it.

tomholub:
Sometimes people import keys that may expire in the future.

Then they use the extension happily for a while, and then their key expires.

Then they start seeing "One of your recipient uses expired key" once they hit send. But it's actually their own key at fault. Because every outgoing message also encrypts for their own key, they'll see it for every recipient they try to send a message to.

To test this, we could allow importing of private keys that are expired. With a warning (modal):

> You are importing a key that is expired. You can still import it to read messages from the past, but you will not be able to send new messages using this key. You can add more keys in the settings later.
>
> Proceed with expired key?
>
> [OK] [Cancel]

Then add a test for the setup screen with an expired key.

Then fix the issue when sending out encrypted messages on such account (to make it clear it's the user's own key that's expired). It should say (error modal):

> This message could not be encrypted because your own Private Key is expired.
>
> You can extend expiration of this key in other OpenPGP software (such as gnupg), then re-import updated key in FlowCrypt Settings -> Additional Settings -> Public Key -> Show Private -> update key.
>
> [OK]

(the OK closes the modal, but will not send the message, so user will be looking at the Compose Window)

Then, add an automated test for doing that (updating the key in settinsg). Finally, the test should re-try sending a message, and it should go through after the retry.

tomholub:
Please also check if there is a test for sending encrypted messages when RECIPIENT public keys are expired. If we are not testing it today, please add a test for that too.

michael-volynets:
Just for a note:
I created a key for my Gmail Account that will expire tomorrow at `2021-08-12 ~15:00:00`

tomholub:
Did you use gnupg?

michael-volynets:
No, I used `gpg --gen-key` and then edited expiration date for the key in Linux console

tomholub:
That is gnupg, also known as gpg :)

michael-volynets:
Okay, thanks, I didn't know that :)

PR #2015, opened 2019-09-09
Issue #1721 sending msg with own expired key

tomholub:
I made a PR so that I can comment on the code below

#1721 #2016


extension/chrome/elements/compose.ts
```diff
@@ -190,17 +191,17 @@ Catch.try(async () => {
       BrowserMsg.send.closeNewMessage(parentTabId);
     }
   };
-  const collectAllAvailablePublicKeys = async (acctEmail: string, recipients: string[]) => {
+  const collectAllAvailablePublicKeys = async (acctEmail: string, recipients: string[]): Promise<{ armoredPubkeys: PubkeyResult[], emailsWithoutPubkeys: string[] }> => {
     const contacts = await storageContactGet(recipients);
     const { public: senderArmoredPubkey } = await storageGetKey(acctEmail);
-    const armoredPubkeys = [senderArmoredPubkey];
+    const armoredPubkeys = [{ pubkey: senderArmoredPubkey, email: acctEmail }];
```
tomholub: Could use `{pubkey: string, isMine: boolean}` - see below

extension/js/common/composer.ts
```diff
@@ -545,18 +545,27 @@ export class Composer {
...
   }
 
-  private encryptMsgAsOfDateIfSomeAreExpired = async (armoredPubkeys: string[]): Promise<Date | undefined> => {
+  private encryptMsgAsOfDateIfSomeAreExpired = async (acctEmail: string, armoredPubkeys: PubkeyResult[]): Promise<Date | undefined> => {
     // todo - disallow in certain situations
     const usableUntil: number[] = [];
     const usableFrom: number[] = [];
-    for (const armoredPubkey of armoredPubkeys) {
-      const { keys: [pub] } = await openpgp.key.readArmored(armoredPubkey);
+    const myKey = armoredPubkeys.find(a => a.email === acctEmail);
```
tomholub: Here I would go by `armoredPubkeys.find(ap => ap.isMine)` - else I feel it would end up too complex when we take into account our own email aliases, and allowing to use different pubkey depending on alias.. if all you're trying to figure out is if the key is user's or not, I think it would be simpler to be explicit.

extension/js/common/composer.ts
```diff
@@ -545,18 +545,27 @@ export class Composer {
     });
   }
 
-  private encryptMsgAsOfDateIfSomeAreExpired = async (armoredPubkeys: string[]): Promise<Date | undefined> => {
+  private encryptMsgAsOfDateIfSomeAreExpired = async (acctEmail: string, armoredPubkeys: PubkeyResult[]): Promise<Date | undefined> => {
```
tomholub: I think I have edited this method since, please pull from master

michael-volynets: Yes, I've already merged it

tomholub:
@michael-volynets please check the tests. There was an `.only` left. When I removed it, many tests fail.

tomholub:
Also my previous comments were not addressed at all.

michael-volynets:
Yes, I'm working on them now and I have fixed the issue with the tests

tomholub:
Got it, thanks

michael-volynets:
Total time spent: 10h 30m

tomholub:
Thanks, ready for review? (sorry - this time I created the PR)

michael-volynets:
Yes, Sure,
You can review it

extension/chrome/settings/setup.ts
```diff
@@ -488,6 +488,9 @@ Catch.try(async () => {
         return await Ui.modal.warning(e.message);
       } else if (e instanceof KeyCanBeFixed) {
         return await renderCompatibilityFixBlockAndFinalizeSetup(e.encrypted, options);
+      } else if (e instanceof ResetForm) {
```
tomholub: I know you followed the behavior of what we do in compose form here, with the reset exception. While that's what I did in composer, I'm not very proud of it, and don't really like this pattern. Anyway, in this case, there is no (large) harm in leaving the inputs as they are. I recommend to remove this if branch.

I will add related comments in other places.

extension/js/common/browser.ts
```diff
@@ -463,9 +463,14 @@ export class Ui {
         },
       });
     },
-    error: async (text: string): Promise<void> => {
+    error: async (text: string, isHTML: boolean = false): Promise<void> => {
+      if (isHTML) {
```
tomholub: Ternary operator is ok when assigning value or returning, eg
```js
text = isHTML ? Xss.htmlSanitize(text) : Xss.escape(text).replace(/\n/g, '<br>');
```
I only hate ternary operator when it's compounded (hard to read) or when it's not returning any value, eg:
```js
condition ? doBlah() : doBleh();
```
is not good.

But in this case, no prob.

tomholub: Thanks for the proper escaping!

extension/js/common/composer.ts
```diff
@@ -548,17 +548,29 @@ export class Composer {
...
    * Try to find an intersection of time that public keys of all recipients were usable (if user confirms this with a modal)
    */
-  private encryptMsgAsOfDateIfSomeAreExpiredAndUserConfirmedModal = async (armoredPubkeys: string[]): Promise<Date | undefined> => {
+  private encryptMsgAsOfDateIfSomeAreExpiredAndUserConfirmedModal = async (armoredPubkeys: PubkeyResult[]): Promise<Date | undefined> => {
     const usableUntil: number[] = [];
     const usableFrom: number[] = [];
-    for (const armoredPubkey of armoredPubkeys) {
-      const { keys: [pub] } = await openpgp.key.readArmored(armoredPubkey);
+    const myKeys = armoredPubkeys.filter(ap => ap.isMine);
+    for (const armoredPubkey of armoredPubkeys.filter(k => !myKeys.includes(k))) {
```
tomholub: Actually here please cycle through all keys.

I know you reasoned correctly here, you want to either worry about recipients, or about sender's key, and separate concerns.

But because of the complexity of signatures, their creation time, their expiration time, etc, it's actually important to look at **intersection of all keys**, including our own.

It should not cause trouble with the rest of the logic, because you are testing sender's key first. So if it's actually the sender's key causing trouble, user will always know about it either way.

Maybe what you did will produce the same behavior. I'm not sure. But I'd feel a bit better if it looked at all keys here.

tomholub: That means simply `for (const armoredPubkey of armoredPubkeys) {`

test/source/tests/page_recipe.ts
```diff
@@ -5,10 +5,23 @@ import { expect } from 'chai';
...
 import { FlowCryptApi } from './api';
-import { EvaluateFn } from 'puppeteer';
+import { EvaluateFn, ElementHandle } from 'puppeteer';
 
-class PageRecipe {
+export class PageRecipe {
+  public static getElementPropertyJson = async (elem: ElementHandle<Element>, property: string) => await (await elem.getProperty(property)).jsonValue() as string;
 
+  public static waitForModalAndRespond = async (controllable: Controllable, name: 'confirm' | 'error', { contentToCheck, clickOn }:
+    { contentToCheck?: string, clickOn?: 'confirm' | 'cancel' }) => {
```
tomholub: Nice, thanks!

test/source/tests/page_recipe.ts
```diff
@@ -338,10 +351,13 @@ export class InboxPageRecipe extends PageRecipe {
 export class ComposePageRecipe extends PageRecipe {
 
   public static openStandalone = async (
-    t: AvaContext, browser: BrowserHandle, group: CommonBrowserGroup, options:
+    t: AvaContext, browser: BrowserHandle, group: CommonBrowserGroup | string, options:
       { appendUrl?: string, hasReplyPrompt?: boolean, skipClickPropt?: boolean, initialScript?: EvaluateFn } = {}
   ): Promise<ControllablePage> => {
-    const email = (group === 'compose') ? 'test.ci.compose%40org.flowcrypt.com' : 'flowcrypt.compatibility%40gmail.com';
+    if (['compatibility', 'compose'].includes(group)) { // Most common accounts
```
tomholub: Actually here it would probably be cleaner to do one by one..
```
if(group === compose) {
  email = 'test.ci.compose@org.flowcrypt.com';
} else if (group === compatibility) {
  email = ...
} else {
  throw 
}
```
Or, like before, you could define a dictionary somewhere `ccctEmailByGroup = {[group: string]: string}`, then use that here. Maybe even without the throw, because the types should protect you.

extension/chrome/settings/setup.ts
michael-volynets: I think would be better to clean the form,
I can try to implement the same but in another way. What do you think?

tomholub: I think it's a fringe case, I don't mind not cleaning the form at all.

If the user is using an expired key, they are almost 100% a technical user. I don't mind to give them a bit of inconvenience having to press `CTRL+A` and then `CTRL+V` again with an updated key. They'll probably have to update the key in gnupg which is a lot more complicated.

Not worth complicating our codebase for this. However, if this was functionality that is expected to be used by normal clueless users, that would be a different story.

extension/js/common/browser.ts
michael-volynets: Sorry, you know that I sometimes can forget using ternary operators.
Thanks for noticing!

extension/chrome/settings/setup.ts
michael-volynets: Okay, Got it

extension/js/common/composer.ts
michael-volynets: Yes, agree, Thanks!

extension/chrome/settings/setup.ts
michael-volynets: I just realized that I need to either replace this `ResetForm` with UserAlert or create a new one because I need to throw an exception here to stop validation here:
```TS
      } else if (await Pgp.key.usableButExpired(k)) {
        // Maybe would be better to give user 3 abilities:
        // 1) Confirm importing expired key
        // 2) Fix expired key
        // 3) Cancel
        const isConfirmed = await Ui.modal.confirm('You are importing a key that is expired. You can still import it to read messages from the past, ' +
          'but you will not be able to send new messages using this key. You can add more keys in the settings later.\n\nProceed with expired key?');
        if (!isConfirmed) {
          throw new ResetForm(); // <- here
        }
...
```

so maybe would be better to leave it as is or I can recode it but I don't really think that it's worth it

tomholub: Good point, you can `throw new UserAlert(` then, like a few lines below it does. At least it will be a bit more consistent, and the same logic can handle both situations.

tomholub: The UserAlert would then say: `You chose to not import expired key.\n\nPlease import another key, or edit the expired key in another OpenPGP software to extend key validity.`

michael-volynets: Okay!

michael-volynets:
+45m for tests correction

michael-volynets:
@tomholub, It's ready for Review

michael-volynets:
18hrs in Jira

Issue #1493, opened 2019-01-12
simplify compose reply


tomholub:
- do not try to parse subject, from, to, etc from the DOM
- work exclusively with threadId and threadMsgId, then always fetch the remaining metadata later from API
- when cannot fetch, show retry button
- when cannot get threadId, render error showing that it failed to get threadId for a reply (users can write us)

tomholub:
This one may mostly be done now, please take a look

michael-volynets:
As I see now, we don't use 'threadMsgId' variable in `compose.ts`.
I think would be better to use this variable because maybe in the future we will have the ability to reply not only to the last message.
So I'll return this variable to the code. If there isn't `threadMsgId` and there is `threadId` then I'll set the last message-id of the thread as `threadMsgId` ;

michael-volynets:
Also, do we need to get `isReplyBox` variable from the `URLParams`?
I think we don't, we could say that it's the reply box depending on whether we have threadMsgId or not.
What do you think?

tomholub:
Please don't add threadMsgId back, because it's useless now. We'll add it back when it's actually useful.

At which point we will probably call it `replyToMsgId` for clarity.

tomholub:
> Also, do we need to get `isReplyBox` variable from the `URLParams`?
> I think we don't, we could say that it's the reply box depending on whether we have threadMsgId or not.
> What do you think?

I guess you're right. You could remove it.

michael-volynets:
I need to change these 2 tests:
![image](https://user-images.githubusercontent.com/49104848/64735847-0d310000-d4f2-11e9-8b1c-ab2c48c3cafa.png)

I will test here if the retry button is present in the tests, Is it okay? Or would it be better to change a little bit API mock so when I load the message the first time the mock returns me `NotFound` and when I try to get it the second time it will return me a normal message, what do you think?

tomholub:
> I will test here if the retry button is present in the tests, Is it okay?

sounds ok

> Or would it be better to change a little bit API mock so when I load the message the first time the mock returns me NotFound and when I try to get it the second time it will return me a normal message, what do you think?

That sounds like too much magic

PR #2020, opened 2019-09-11
[3h+] issue #1493 - rm reply headers from url

michael-volynets:
2h 30m so far

### DIFF
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -31,8 +31,7 @@ Catch.try(async () => {
 
   const ksLookupsByEmail: { [key: string]: PubkeySearchResult | Contact } = {};
 
-  const uncheckedUrlParams = Env.urlParams(['acctEmail', 'parentTabId', 'draftId', 'placement', 'frameId', 'threadId', 'threadMsgId',
-    'skipClickPrompt', 'ignoreDraft', 'debug']);
+  const uncheckedUrlParams = Env.urlParams(['acctEmail', 'parentTabId', 'draftId', 'placement', 'frameId', 'threadId', 'skipClickPrompt', 'ignoreDraft', 'debug']);
   const acctEmail = Assert.urlParamRequire.string(uncheckedUrlParams, 'acctEmail');
   const parentTabId = Assert.urlParamRequire.string(uncheckedUrlParams, 'parentTabId');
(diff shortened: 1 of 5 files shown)


michael-volynets:
+ 20m

extension/chrome/elements/compose.ts
```diff
@@ -64,43 +57,47 @@ Catch.try(async () => {
...
+  let to: string[] = [];
+  let cc: string[] = [];
+  let bcc: string[] = [];
+  if (threadId) {
+    const fetchSuccess = await (async () => {
+      Xss.sanitizePrepend('#new_message', Ui.e('div', { id: 'loader', html: 'Loading secure reply box..' + Ui.spinner('green') }));
+      try {
+        const thread = await Google.gmail.threadGet(acctEmail, threadId, 'metadata');
+        const gmailMsg = await Google.gmail.msgGet(acctEmail, thread.messages[thread.messages.length - 1].id, 'metadata');
+        const reply = Google.determineReplyCorrespondents(acctEmail, storage.addresses || [], gmailMsg);
```
tomholub: If I'm reading this correctly, you could simplify as:
```js
const {from, to, cc, bcc} = Google.determineReplyCorrespondents(acctEmail, storage.addresses || [], gmailMsg);
```

tomholub:
You can ignore the codacy check

extension/chrome/elements/compose.ts
michael-volynets: I couldn't make it exactly the same way you wrote but I have made it a little better

michael-volynets:
> I feel like there should be some more code in webmail_elements_replacer (or similar) to remove? Like parsing recipients and subject from the DOM, and adding that to url params, etc.

I removed parsing `subject` form that file.
I think I've mostly done it in some of previous PRs because I can't see parsing of a message in `gmail_element_replacement`, it parses only ThreadMSgId, and threadId now.

tomholub:
> > I feel like there should be some more code in webmail_elements_replacer (or similar) to remove? Like parsing recipients and subject from the DOM, and adding that to url params, etc.
> 
> I removed parsing `subject` form that file.
> I think I've mostly done it in some of previous PRs because I can't see parsing of a message in `gmail_element_replacement`, it parses only ThreadMSgId, and threadId now.

Sounds good, as long as we no longer pass subject, to, ..., ... in the url either? And the XssSafeFactory was cleaned up of this as well?

Issue #1911, opened 2019-08-12
Alias name is being ignored


limonte:
Let's say a user has the primary email address `john@gmail.com` and name `John Doe`.
When the email is sent, a recipient sees `John Doe <john@gmail.com>`.

Now, the same user can have an email alias `alice@gmail.com` and name for that alias `Alice Doe`.

By using Gmail, when the user is using the alias to send an email, a recipient will see `Alice Doe <alice@gmail.com>`.

But, if the user is using FlowCrypt and the same alias, a recipient will see `John Doe <alice@gmail.com>`.

The name of the original email address is used, not alias which might lead to confusions on the recipient side, as a recipient will see name unrelated to email address.

tomholub:
Very good discovery

tomholub:
I have not assigned you (yet) because this maybe digs a bit deeper into the internals of the plugin. We'll see if there's time for it later, or if we end up fixing this ourselves.

tomholub:
@michael-volynets please change how alias addresses are saved, from a string array into a dictionary.

You can pull all of the information below from `sendAs`.

You can just forget/start ignoring the `addresses: string[]` field, and instead add `sendAs` field in `acctStorage`.

For example: 
```
sendAs: {
  "joe@gmail.com": { isPrimary: true, name: "Joe Doe", footer: "--\nJoe Doe" },
  "joe@your-local-baker.com": { isPrimary: false, name: "Joe Doe (baker)", footer: "--\nJoe Doe, your personal baker" },
}
```
Then when you need just a list of addresses, you would get it from storage and do `const addresses = Object.keys(storage.sendAs)`.

For now, please ignore the footer. That will be another issue.

tomholub:
And of course, make sure the right name is used based on the alias.

Fallback on the account name on Gmail that we store today in case the alias name is empty.

PR #2023, opened 2019-09-13
Issue #1911, Changed addresses to 'sendAs'

michael-volynets:
Issue #1911,
Time Spent: 8h so far

I changed `addresses: string[]` to `sendAs: Dict<{ name: string, isDefault: boolean, isPrimary: boolean }>`

The only thing to do is  the fix issue with alias names


tomholub:
Please also include the footer/signature in our storage too

tomholub:
from `sendAs`, as described in the issue

michael-volynets:
Done, + 1h

tomholub:
Correction - please don't use the stored footer from sendAs yet. That will be for another issue, because it will involve more decisions (we will have to probably completely remove the `footer` setting icon, use it regardless if free or paid, etc).

Just store it, but don't query it in any way yet.

tomholub:
Unless you wan to include it in this issue.. and remove footer.htm, and everything else related to footers, and just always include it if present.

tomholub:
But if you do that, you may need to include footer in the quoted content to let users edit it.. it's not so simple. I'd prefer to do that in another PR. Better keep PRs small for my sanity.

michael-volynets:
Okay, let's do that in another PR,

So do I need to store email footer in both `SendAs` and `email_footer` but get it only from `email_footer`, yes?

tomholub:
Please only save footers from `sendAs` into `sendAs` property.

Leave everything else as it was before, so that footers are completely unaffected from user perspective - for now.

michael-volynets:
That's what I did in this PR, I changed from `email_footer` to `sendAs[emailAlias].footer`

extension/chrome/elements/shared/footer.ts
```diff
@@ -13,28 +13,34 @@ Catch.try(async () => {
 
   Ui.event.protect();
 
-  const uncheckedUrlParams = Env.urlParams(['acctEmail', 'parentTabId', 'grandparentTabId']); // placement: compose||settings
+  const uncheckedUrlParams = Env.urlParams(['acctEmail', 'parentTabId', 'grandparentTabId', 'emailAlias']); // placement: compose||settings
```
tomholub: please revert

michael-volynets:
Okay, I didn't get what you mean exactly,

Thanks for the explanation and comments

michael-volynets:
I removed the last commit with Footer changes and added only one property called `footer`.
I don't do anything with this property

extension/chrome/settings/modules/keyserver.ts
```diff
@@ -113,17 +113,19 @@ Catch.try(async () => {
     }));
 
     $('.action_remove_alias').click(Ui.event.prevent('double', async self => {
```
tomholub: I think this functionality no longer makes sense. If the user wants to remove an alias, they can do it in Gmail and it will get refreshed next time by pulling from sendAs API. So I would remove the button & this handler.

extension/chrome/settings/modules/keyserver.ts
```diff
@@ -113,17 +113,19 @@ Catch.try(async () => {
...
       window.location.reload();
     }));
 
     const contentEl = Xss.sanitizeAppend('#content', '<div class="line"><a href="#" class="action_fetch_aliases">Missing email address? Refresh list</a></div>');
     contentEl.find('.action_fetch_aliases').click(Ui.event.prevent('parallel', async (self, done) => {
       Xss.sanitizeRender(self, Ui.spinner('green'));
       try {
-        const addresses = await Settings.fetchAcctAliasesFromGmail(acctEmail);
-        await Store.setAcct(acctEmail, { addresses: Value.arr.unique(addresses.concat(acctEmail)) });
+        await Settings.refreshAcctAliases(acctEmail);
```
tomholub: Since we're at it, you can remove this button and also the handler. Please use the same logic as in compose - load sendAs asynchronously on every page load, and if it changed, offer to reload the page.

Fewer buttons to confuse users. Thanks!

extension/chrome/settings/setup.ts
```diff
@@ -97,14 +97,14 @@ Catch.try(async () => {
       if (!GoogleAuth.hasReadScope(storage.google_token_scopes || [])) {
         $('.auth_denied_warning').css('display', 'block');
       }
-      if (typeof storage.addresses === 'undefined') {
+      if (typeof storage.sendAs === 'undefined') {
         if (GoogleAuth.hasReadScope(storage.google_token_scopes || [])) {
```
tomholub: This `if` check no longer makes sense (sorry, I overlooked this last time).

We needed to check if we had read permission because we used to guess aliases based on sent emails, and read permission was needed to examine sent emails.

We now use proper api for this, so this check is not needed.

extension/chrome/settings/setup.ts
```diff
@@ -97,14 +97,14 @@ Catch.try(async () => {
...
-          Settings.fetchAcctAliasesFromGmail(acctEmail).then(saveAndFillSubmitOption).catch(Catch.reportErr);
+          Settings.fetchAcctAliasesFromGmail(acctEmail, true).then(saveAndFillSubmitOption).catch(Catch.reportErr);
         } else { // cannot read emails, don't fetch alternative addresses
```
tomholub: Which means you can remove this else branch completely.

extension/js/common/composer.ts
```diff
@@ -1060,9 +1062,9 @@ export class Composer {
     this.composeWindowIsMinimized = !this.composeWindowIsMinimized;
   }
 
-  private renderSenderAliasesOptions() {
-    const addresses = this.app.storageGetAddresses();
-    if (addresses.length > 1) {
+  private renderSenderAliasesOptions(sendAs: Dict<SendAsAlias>) {
```
tomholub: Yes, passing them as argument is better :+1:

extension/js/common/composer.ts
```diff
@@ -1072,7 +1074,10 @@ export class Composer {
       Xss.sanitizeAppend(inputAddrContainer, selectElHtml);
       inputAddrContainer.find('#input_from_settings').click(Ui.event.handle(() => this.app.renderSendingAddrDialog(), this.getErrHandlers(`open sending address dialog`)));
       const fmtOpt = (addr: string) => `<option value="${Xss.escape(addr)}" ${this.getSender() === addr ? 'selected' : ''}>${Xss.escape(addr)}</option>`;
-      Xss.sanitizeAppend(inputAddrContainer.find('#input_from'), addresses.map(fmtOpt).join('')).change(() => this.composerContacts.updatePubkeyIcon());
+      emailAliases = emailAliases.sort((a, b) => {
+        return (sendAs[a].isDefault === sendAs[b].isDefault) ? 0 : sendAs[a].isDefault ? -1 : 1;
```
tomholub: This looks good. Maybe we can completely remove the "choose order of aliases" screen, since they can do this in Gmail settings?

extension/js/common/settings.ts
```diff
@@ -20,11 +20,18 @@ declare const zxcvbn: Function; // tslint:disable-line:ban-types
...
     const response = await Google.gmail.fetchAcctAliases(acctEmail);
-    return response.sendAs
-      .filter(alias => alias.isDefault || alias.verificationStatus === 'accepted')
-      .map(alias => alias.sendAsEmail);
+    const validAliases = response.sendAs.filter(alias => (alias.isDefault as boolean) || alias.verificationStatus === 'accepted');
+    const result: Dict<SendAsAlias> = {};
+    for (const alias of validAliases) {
+      result[alias.sendAsEmail] = { name: alias.displayName, isPrimary: false };
+    }
+    if (includeAcctAddress) {
```
tomholub: I imagine this could always be true, unless you have usecases in mind that I have overlooked

tomholub: But this would overwrite the "footer" property from Gmail, so maybe always false is better? Just skip this completely?

tomholub:
You can ignore the codacy check

extension/chrome/elements/compose.ts
```diff
@@ -89,7 +89,7 @@ Catch.try(async () => {
     if (gmailMsg.threadId) {
       threadId = gmailMsg.threadId;
     }
-    const reply = Google.determineReplyCorrespondents(acctEmail, storage.addresses || [], gmailMsg);
+    const reply = Google.determineReplyCorrespondents(acctEmail, Object.keys(storage.sendAs || {}), gmailMsg);
```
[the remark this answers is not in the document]
michael-volynets: Done, I used the second option.

extension/chrome/settings/modules/keyserver.ts
michael-volynets: Agree, removed

extension/chrome/settings/setup.ts
michael-volynets: Got it, removed

extension/chrome/settings/modules/keyserver.ts
michael-volynets: Yes, Will do

extension/chrome/elements/sending_address.ts
```diff
@@ -20,33 +19,35 @@ Catch.try(async () => {
...
 
     Xss.sanitizeRender(container, addresses.map(emailAddrToHtmlRadio).join(''));
-    container.find('input').first().prop('checked', true);
     container.find('input').click(Ui.event.handle(async target => {
       const chosenSendingAddr = String($(target).val());
-      if (chosenSendingAddr !== addresses[0]) {
-        const orderedAddrs = Value.arr.unique([chosenSendingAddr].concat(storage.addresses || []));
-        await Store.setAcct(acctEmail, { addresses: orderedAddrs });
-        window.location.reload();
+      if (storage.sendAs && !storage.sendAs[chosenSendingAddr].isDefault) {
```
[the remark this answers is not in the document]
michael-volynets: Yes, it will, I'll fix it.

Thanks!

extension/js/common/composer.ts
michael-volynets: In my opinion, it would be better to leave it as is. The way users can change email alias using FlowCrypt is much easier. When I tried to change my alias first time in Gmail I spent too much time to find the tab with email aliases I was looking for.
And maybe, some users want to have different email aliases in Gmail and in FlowCrpypt, so I would recommend you to leave this screen.

extension/chrome/elements/compose.ts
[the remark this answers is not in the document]
michael-volynets: I just realized that Gmail always returns email aliases including our primary email address, so maybe we don't need to make this ternary operator at all and make sendAs not to be 'Dict<T> | undefined' but just  'Dict<T>'?

extension/js/common/composer.ts
tomholub: There is some truth to it, but presumably these people also use normal unencrypted email, so hopefully they already have gone through the pain of configuring it.

I think the best possible scenario is that the user would never have to configure anything in FlowCrypt and it would just work as expected. That's not realistic, but this update would push us a bit closer in that direction. 

One worry is that we may end up needing this for Microsoft accounts.. but regardless I would rather have it removed if it's not necessary. 

FlowCrypt is becoming a bit of an effort to maintain, so less code is better.

extension/chrome/elements/compose.ts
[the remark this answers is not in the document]
tomholub: It will unfortunately probably still be | undefined because it's coming from the storage.. who know if and when it was set. Not sure.

extension/js/common/composer.ts
michael-volynets: Okay, that makes sense,

Then I will remove it

extension/chrome/elements/compose.ts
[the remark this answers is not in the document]
michael-volynets: Oh, yes, I forgot that something can happen with the storage.
Thanks

extension/js/common/settings.ts
michael-volynets: I removed it at all,
When we have a footer it will be better to not overwrite existing `sendAs` aliases and just update properties `isPrimary` and 'isDefault'

tomholub: I would overwrite everything, why not?

tomholub: I'm not sure we are talking about the same thing. When you receive new data from sendAs endpoint, is there any reason not to update our sendAs storage completely?

tomholub: There is no need to complicate it, I think. Every time we receive sendAs from the API, we completely update sendAs in storage.

There will be no way to set aliases in FlowCrypt otherwise.

In another PR, there will be no way to set footers either and sendAs will be used.

I don't see any need for any logic more complicated then that, but it's possible you see something I don't (because you are the one implementing it)

michael-volynets: > I would overwrite everything, why not?

If I remember correctly we are going to have a 'footer' property in `sendAs` object so if we overwrite everything we will lose our footer.

tomholub: We will use footers from Gmail also, in another PR.

Until then, we are using footers from 'footer' storage. But later we'll ignore the footer storage and just use sendAs, and there will be no way to change footers in FlowCrypt.

michael-volynets: So, if I understood everything correctly you want me to update the `sendAs` property in our storage right away in `fetchAcctAliasesFromGmail`? (I will rename it if yes)

michael-volynets: Sorry, I just got what you meant,
Done.

tomholub: Sure, you could do that, if you want to have a consistent place to grab new results and update storage.

Or it could do saving only, eg:
```
async updateStorageSendAs(apiResult): Promise<void> {
   // 
}
```
Which may be slightly better decoupled. Either way is good.

michael-volynets: we already have this method `refreshAcctAliases`, it was returning fetched aliases from Gmail before and now it returns `boolean` that indicates whether the 'sendAs' was Changed or not.

tomholub: That sounds ok, we could just use that, then pull from local storage.

tomholub:
Removing code is my favorite part of development :laughing: :+1:

image

extension/js/common/settings.ts
```diff
@@ -109,11 +113,14 @@ export class Settings {
...
 
-  static refreshAcctAliases = async (acctEmail: string) => {
-    const addresses = await Settings.fetchAcctAliasesFromGmail(acctEmail);
-    const all = Value.arr.unique(addresses.concat(acctEmail));
-    await Store.setAcct(acctEmail, { addresses: all });
-    return all;
+  static refreshAcctAliases = async (acctEmail: string): Promise<boolean> => {
+    const fetchedSendAs = await Settings.fetchAcctAliasesFromGmail(acctEmail);
+    const { sendAs: storedAliases } = (await Store.getAcct(acctEmail, ['sendAs']));
+    if (!storedAliases || !compareObjects(fetchedSendAs, storedAliases)) {
```
tomholub: Too complicated, I think unnecessarily.

The only reason we actually need to refresh a window is if there is an alias missing. That means you can just compare:

```js
!storedAliases || Object.keys(fetchedSendAs).sort().join(',') !== Object.keys(storedAliases).sort().join(',')
```

Then you update the stored object either way, but only return true if list of aliases changed.

extension/js/common/platform/util.ts
```diff
@@ -32,3 +32,50 @@ export const moveElementInArray = <T>(arr: Array<T>, oldIndex: number, newIndex:
   arr.splice(newIndex, 0, arr.splice(oldIndex, 1)[0]);
   return arr;
 };
+// tslint:disable:no-unsafe-any
+export const compareObjects = (object1: { [key: string]: any }, object2: { [key: string]: any }) => {
```
tomholub: Related to another comment, this whole section can be removed.

Also, this file should only contain platform-specific code. The kind of code that would be written differently for each platform (browser vs node).

It says it right above:
```js
/**
 * Functions which must be written differently to run in NodeJS versus in web browsers.
 *
 * If the code would be the same on both platforms, it does not belong here (or anywhere in platform/ directory)
 */
```

extension/js/common/settings.ts
michael-volynets: Unfortunately, I also was thinking about it but the reason I had to create this `compareObjects` function is that we can't change default alias using Flowcrpyt now so I need to check not only if alias emails changed but if `isDefault` property is changed too. 

The first idea I came up was to replace `Object.keys`  with `JSON.stringify` but that didn't work because Gmail returned **the same** object with the same properties but in another order so the result of `JSON.stringify` was different. Then I decided to create this function called `compareObjects`.

extension/js/common/platform/util.ts
michael-volynets: Please take a look at my comment above and if I can leave this method in Flowcrypt I can move it to another file.

extension/js/common/settings.ts
michael-volynets: I also agree that this method (`compareObjects`) is too large and complicated but I guess that just checking those two fields (`key` and `isDefault`) would be a little weird and incorrect. So I wrote this general solution.

michael-volynets: Yes, this is much easier and readable, I thought that this was would be little incorrect but if you wrote this I'm going to replace my code with yours. 
As I said, I wanted to make more general solution but when you show me your example I realized that it seems like it doesn't really make sense

Thanks, I will change it

tomholub: Your solution was a prudent one, to offer best user experience. But it came out as so much work (and so much code) that the tradeoff may not be completely worth it in this case, considering it gets updated next time user reopens the window.

tomholub:
I think LGTM got confused after I reverted the merge

tomholub:
@michael-volynets please have a look at the test failures after merging, plus the issue I mentioned above.

michael-volynets:
Sorry, but I can't exactly understand what you mean by `switching from the previous system to the new one`.
Could you please give me more details about that?
Thanks

michael-volynets:
Sorry, but I can't exactly understand what you mean by `switching from the previous system to the new one`.
Could you please give me more details about that?
Thanks

tomholub:
Previous versions of FlowCrypt used storage.addresses. The new one uses storage.sendAs.

When we release the new version and it updates for the user, the compose dialog will show "Your aliases have changed" (as above), which is not true.

This happens because the new version is only comparing to `sendAs` which is `undefined` the first time the extension gets updated.

Reproduce:
1) set up account using master branch (on account that has more than one alias)
2) switch to this branch
3) rebuild and open compose window

The solution may be to, when comparing `sendAs` for changes, if `sendAs` is `undefined`, we compare the contents with what is in `storage.addresses` instead. (this would be temporary migration code that we can eventually remove).

michael-volynets:
Thanks for the explanation, I got what you meant,

Yes, I think it's a great idea to make this temporary migration, 
I will do it,
When this PR is merged I'll create an issue to remove that temporary code after some time.

tomholub:
This looks like a good solution. I'll test it.

tomholub:
18hrs from Jira after fixes

Issue #1963, opened 2019-08-17
rework compose button / sign / encrypt / plain with a chevron


tomholub:
Instead of current situation with send btn and sign btn:
```
[ send encrypted message ]  [.!]
```
We should have an up or down chevron in the button to choose message type:
```
[ Send encrypted message |⌄]
```

When I click it, it will give me several lines that I can click and add tick to them, or click again and remove tick:
 - Encrypt message
 - Sign message

So I can send encrypted message, signed message, signed+encrypted message, or plain, depending which ones I choose.

That will also allow use cases like this:

> When I write someone I don't usually know beforehand whether they're going to have FlowCrypt set up or not. So when I start drafting an encrypted message and see that they don't, I usually don't want to go with a one-time password (since that expires and is just generally a bit of a hassle for the other person for ordinary day-to-day communication). So I would like an option where you just draft the email, if the other person has set up Flowcrypt the message is sent encrypted, and if they don't, you can send it unencrypted and add a little gentle message at the bottom about what FlowCrypt is and why they might want to use it.

tomholub:
Very roughly (but not so ugly :) ):

![image](https://user-images.githubusercontent.com/6306961/64734723-b5d16680-d4d6-11e9-99a6-9ba7cd113ed7.png)

Possible button texts:
 - Sign and Send
 - Encrypt and Send
 - Encrypt, Sign and Send
 - Send plain (not encrypted)

You can use checkboxes from our docs: https://flowcrypt.com/docs/technical/openpgp-compatibility.html

Please make the pop-up options flexible enough that we could add a third or fourth option later (the other options would not change button text, but may change some other settings).

michael-volynets:
Your variant looks nice, I like it :+1:

michael-volynets:
I think we don't need to show a red cross icon, would be better to show just green tick icon next to the selected item, what do you think?

tomholub:
You can try it with just the green ticks, and we'll see one I test it. Thanks

PR #2028, opened 2019-09-19
[15h] Issue 1963 add sending type popover

michael-volynets:
Issue #1963,
Time Spent: 15h

Result:
image

1) I think would be better to change  Plain Message Icon (I just picked some from svgs folder)
2) I haven't done logic for sending `signed + encrypted` and `plain` message. Will it be okay if I add the code in this PR or do I need to make another one?

(Haven't checked on Firefox yet)

I'm going to fix tests but you can start reviewing it


tomholub:
> I haven't done logic for sending signed + encrypted and plain message. Will it be okay if I add the code in this PR or do I need to make another one?

In this one I think, else it would be broken when I merge it. But you could also leave it to me to finish this logic, either way is ok.

michael-volynets:
> > I haven't done logic for sending signed + encrypted and plain message. Will it be okay if I add the code in this PR or do I need to make another one?
> 
> In this one I think, else it would be broken when I merge it. But you could also leave it to me to finish this logic, either way is ok.

Okay, I will do that logic in this PR

michael-volynets:
And I think it would be great to add sending tests for each option in the popover, wouldn't it?

Also, To send Encrypted + Signed message I need to Encrypt the text then sign it or vice versa?

tomholub:
Signing + encryption is done in one step. Simply provide both the public keys (to encrypt for) and private key (to sign with) to the `PgpMsg.encrypt` function.

tomholub:
```
static encrypt: PgpMsgMethod.Encrypt = async ({ pubkeys, signingPrv, pwd, data, filename, armor, date }) => {
```

michael-volynets:
Got it, thanks!

extension/js/common/composer/interfaces/composer-types.ts
```diff
@@ -61,3 +61,6 @@ export type ComposerUrlParams = {
 };
 
 export type PubkeyResult = { pubkey: string, email: string, isMine: boolean };
+
+export type ComposerPopoverItem = { type: 'selectEncryptionType', HTMLContent: string, data: EncryptionType, iconPath?: string };
```
tomholub: It seems the `type: 'selectEncryptionType'` is not currently used. It's best to not introduce code "in case it's useful in the future" because who knows, the future may not be what we expect :) Unless I overlooked some usage of this.

extension/js/common/composer.ts
```diff
@@ -306,12 +311,86 @@ export class Composer {
...
+    $('.sending-container').removeClass('popover-opened');
+    this.showHidePwdOrPubkeyContainerAndColorSendBtn();
+  }
+
+  private openOrClosePopover = (event: JQuery.Event<HTMLElement, null>) => {
+    event.stopPropagation();
+    const sendingContainer = $('.sending-container');
+    sendingContainer.toggleClass('popover-opened');
+    if (sendingContainer.hasClass('popover-opened')) {
+      $('body').click((event) => {
```
tomholub: Should use `Ui.event.handle`

extension/js/common/composer/interfaces/composer-types.ts
michael-volynets: Thanks, I added it because I thought it's useful in the future but I agree that the future may not be what we expect  :)

I will remove it

michael-volynets:
> Also, maybe we can consider creating a separate composer/composer-send-btn.ts file where we could put some of this logic? Anything that concerns how the send button behaves.

Yes, that would be great I think but I would recommend you to do it in another PR. And maybe move there all logic related to sending messages.

tomholub:
Ok. Another PR for that is fine with me, one day.

michael-volynets:
Please don't merge it yet, I found some CSS bugs I need to fix

tomholub:
Ready or should I still wait until further notice?

michael-volynets:
Sorry, yes, it's ready for review

tomholub:
Thanks! I'll take a look tomorrow

michael-volynets:
> you could make "Encrypt and send" have a dark gray color of lock, then "Encrypt, sign and send" could have a green color lock (instead of green sign icon)

Maybe that orange color for plain text?

tomholub:
> Maybe that orange color for plain text?

Orange would work, you can use the same as the expired key orange

michael-volynets:
What do you think about this?
image

I think I need to change the color of tick icon, make it a little bit more dark green because it's too light I think.

michael-volynets:
Also, I think I can find a more consistent icon for plain text because this orange one isn't really good for my opinion.

tomholub:
Why not the red lock as I was suggesting above? Or even a red crossed lock?

Or maybe give the plain text just no icon.

I still think the sign-and-send should have a gray sign icon.

Agree the tick should have consistent/darker color.

tomholub:
You could give the "Send plain" a Gmail icon 😆 ... really. To reinforce that Gmail is by default not end-to-end encrypted.

tomholub:
We should have a Gmail icon you can use from the Connect to Gmail button.

tomholub:
plus tests

michael-volynets:
> You could give the "Send plain" a Gmail icon laughing ... really. To reinforce that Gmail is by default not end-to-end encrypted.

I agree, that should be good

michael-volynets:
> I still think the sign-and-send should have a gray sign icon (even if it looks less visually consistent, but seems clearer)

So then maybe make green-sign icon next to `Encrypt, Sign And Send`

michael-volynets:
The result:
image

tomholub:
No please don't use the green sign icon. The encryption is more important then the signing. I would rather have two green locks, one sign icon and one gmail icon.

tomholub:
I can finalize it, there are some other minor css fixes I want to do.

michael-volynets:
Okay, Thanks,

Then you can remove the green icon at all, it was used only in that place.

tomholub:
Ok, will do

tomholub:
@michael-volynets how about the tests? They are still failing on something.

michael-volynets:
Wow, does this only happen in the tests? I haven't seen this before

michael-volynets:
> @michael-volynets how about the tests? They are still failing on something.

Sorry, I didn't notice that, thanks! I thought it's LGTM failing

michael-volynets:
Fixed the popover in the reply box, looking into tests

michael-volynets:
Tests fixed too :+1:

tomholub:
39hrs from jira after tests / fixes

Issue #879, opened 2017-07-08
offer to search contacts with Google api


tomholub:
#1621

ref https://mail.google.com/mail/u/0/#inbox/160add19970fd443

tomholub:
On enterprise builds, this permission should be included during setup.

tomholub:
Currently, we search contacts on our local database + based on sent email.

Searching based on sent email is good because we don't have to ask for another permission (contacts). But it's also bad because the results are imperfect. Often times, users are not able to find the contacts they are looking for, and they are writing me about this all the time.

Google has some APIs for this:
https://developers.google.com/contacts/v3/ - the older XML version
https://developers.google.com/people - similar JSON API

Scope:
https://www.googleapis.com/auth/contacts.readonly - it looks like it's not a restricted scope

The functionality should be as follows, depending if user has added the contacts oauth scope or not (which will NOT be added by default during setup for consumer builds, so that we don't scare users):

#### Default behavior (when app does not have contacts read scope enabled)

1) search in local contacts like today
2) search in sent gmail like today
3) render results
4) on the bottom rendered results, show `Enable Google Contact Search` (clickable)

When user clicks it, `authPopup` will show up where there should also be contacts permission included. User should go through the popup, similar to when we handle `authPopupNeeded`, except this time there should also be the contacts permission.

Once done, the app should be able to check if we have contacts permission from storage (similar to canReadEmails or whatever we called the email read permission).

If there is this permission, it will behave like described below.

#### Google contact search enabled - behavior

1) search in local contacts like today
2) search contacts from Google API
3) render results

michael-volynets:
Okay, everything described pretty well, Thanks,

I've got one question below:

When I search for contacts from Google API and there are some contacts that have PGP installed I think it would be great to render them with green lock and etc.. So maybe when I fetch  contacts from Gmail  then I need to find out which of them have PGP installed through our Attester and then I add all of the fetched contacts into Local DB.

What do you think about it?

tomholub:
That sounds like a good idea. Let's add that in a separate PR, to make it easier for me to review.

michael-volynets:
So, in the PR for this issue, I just load contacts and doesn't matter whether they have PGP or not, I will render them with gray color, yes?

tomholub:
Yes, just like we do today. But remember that we are rendering results from from the DB as well (always search local db first).

[...]

PR #2052, opened 2019-09-30
Issue 879 search contacts google api

michael-volynets:
Issue #879,
Time Spent: 16h


extension/chrome/elements/compose.ts
```diff
@@ -47,7 +47,10 @@ Catch.try(async () => {
 
   const storage = await Store.getAcct(acctEmail, ['google_token_scopes', 'addresses', 'sendAs', 'addresses_keyserver', 'email_footer', 'email_provider',
     'hide_message_password', 'drafts_reply']);
-  const canReadEmail = GoogleAuth.hasReadScope(storage.google_token_scopes || []);
+  const scopes = {
+    canReadEmails: GoogleAuth.hasReadScope(storage.google_token_scopes || []),
+    canSearchContacts: (storage.google_token_scopes || []).includes(GoogleAuth.OAUTH.scopes.contactsReadOnly)
```
tomholub: You could do `GoogleAuth.hasReadContactsScope` for consistency

extension/js/common/api/google.ts
```diff
@@ -554,6 +554,7 @@ export class GoogleAuth {
       profile: 'https://www.googleapis.com/auth/userinfo.profile', // needed so that `name` is present in `id_token`, which is required for key-server auth when in use
       compose: 'https://www.googleapis.com/auth/gmail.compose',
       modify: 'https://www.googleapis.com/auth/gmail.modify',
+      contactsReadOnly: 'https://www.googleapis.com/auth/contacts.readonly',
```
tomholub: For our purposes maybe we can call it `readContacts`

extension/js/common/api/google.ts
```diff
@@ -554,6 +554,7 @@ export class GoogleAuth {
...
       contacts: 'https://www.google.com/m8/feeds/',
```
tomholub: If what you do ends up working as expected, the original `contacts` scope will be outdated, so you can remove it.

extension/js/common/composer/composer-contacts.ts
```diff
@@ -269,7 +272,7 @@ export class ComposerContacts extends ComposerComponent {
     }
   }
 
-  private renderSearchRes = (input: JQuery<HTMLElement>, contacts: Contact[], query: ProviderContactsQuery) => {
+  private renderSearchRes = (input: JQuery<HTMLElement>, contacts: Contact[], query: ProviderContactsQuery, research: () => Promise<void>) => {
```
tomholub: minor nitpick, `research` sounds like academic activity, maybe let's name it `reSearch`

tomholub: But I'm thinking, would it be possible to trigger re-search without passing a callback? It looks a bit awkward. maybe if you could pass `reSearchDbOnly` into `renderSearchRes`, then you could re-call `searchContacts` directly inside `renderSearchRes` without having to pass a callback?

michael-volynets: Yes, Agree :)
I'm working on it now

michael-volynets: I think I can get rid of it at all

extension/js/common/api/google.ts
michael-volynets: Okay, will do, thanks!

michael-volynets:
Sorry, got it, I think I somehow missed that

michael-volynets:
Done, + 40m

test/source/tests/tests/compose.ts
```diff
@@ -409,6 +409,19 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
...
+      await BrowserRecipe.setUpCommonAcct(t, browser, 'compose');
+      const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
+      await composePage.click('@action-expand-cc-bcc-fields');
+      await composePage.type('@input-to', 'contact');
+      let contacts = await composePage.waitAny('@container-contacts');
+      expect(await PageRecipe.getElementPropertyJson((await contacts.$('ul li:first-child'))!, 'textContent')).to.eq('No Contacts Found');
+      const oauthPopup = await browser.newPageTriggeredBy(t, () => composePage.click('@action-auth-with-contacts-scope'), 'test.ci.compose@org.flowcrypt.com');
+      await OauthPageRecipe.google(t, oauthPopup, 'test.ci.compose@org.flowcrypt.com', 'approve');
+      contacts = await composePage.waitAny('@container-contacts');
+      expect(await PageRecipe.getElementPropertyJson((await contacts.$('ul li:first-child'))!, 'textContent')).to.eq('contact.test@flowcrypt.com');
```
tomholub: Here, it would be good to also test that after we reload the compose page, it remembers the newly allowed scope, and will right away just load the contacts.

PR #2053, opened 2019-10-01
Issue #2045 make 'Send' and 'Show send options' buttons focusable

### LINKED ISSUE #2045: The Send Options button is not keyboard-friendly
It's not possible to access the Send Options button with a keyboard (using `Tab`)

limonte:
Fixes #2045 

I converted `<div>`s to `<button>`s. It's always a good idea to use `<button>` tags for buttons, by doing that we're getting all a11y features for free (it's built-in in browsers), e.g. this line was removed:

```js
this.S.cached('send_btn').keypress(Ui.enter(() => this.extractProcessSendMsg()));
```

`<button>`s are getting clicked automatically with `Enter` or `Spacebar`, no need in adding additional keyboard handlers.

---

Also, I removed the `border-radius: 2px` from `#send_btn` and `#toggle_send_options` in order to avoid messing around with `border-top/bottom-left/right-radius` properties. @tomholub please let me know if sharp buttons are agains design styleguide.

### DIFF
--- a/extension/chrome/elements/compose.htm
+++ b/extension/chrome/elements/compose.htm
@@ -120,11 +120,12 @@
           <div class="sending-container">
             <div class="sending-options-container">
             </div>
-            <div id="send_btn" class="button green" tabindex="6">
-              <span class="btn-text" data-test="action-send"><i></i>Encrypt and Send</span>
-              <div class="action-show-options-popover" data-test="action-show-options-popover">
-                <div class="icon-chevron"></div>
-              </div>
+            <button id="send_btn" class="button green" tabindex="6">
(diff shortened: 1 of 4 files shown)


limonte:
> The button is no longer vertically centered in the window footer (it's too low)?

Hm, I can't reproduce this. Could you please specify OS and browser?

limonte:
> What does the screenshot look like on your end (on this branch - remember have to re-build)

For me it was exactly the same this morning. Something was wrong with my eyes yesterday afternoon :)

Paddings are now fixed, also fixed switching between green and gray state for the `toggle_send_options` button, take a look.

PR #2054, opened 2019-10-01
Remove outline: none on #input_from

limonte:
This is probably one of the most popular mistakes of front-end developers, there's even the website for it :) http://outlinenone.com/

@michael-volynets @tomholub please do not use `outline: none` without adding some other way of indicating the focused state, e.g. adding `border` to `:focus` state.

Without outline, it's not possible to see the focused state of the element and UI becomes unusable for keyboard users (hackers, people with inabilities to use mouse/trackpoint, etc.).


tomholub:
Thanks! I believe you have already done this one time, and I'm embarrassed if this is another merging error.

PR #2057, opened 2019-10-01
Make the 'Recepients' field accessible with keyboard

### LINKED ISSUE #2044: The Recipients field is not keyboard-friendly
It's not possible to access the Recipients field with a keyboard (using `Tab`)

limonte:
Closes #2044

Also, in scope of this PR I made `cc` and `bcc` buttons accessible with keyboard (they are at the end of taborder)

As discussed per email conversation I changed some lines in JS side to use IDs instead of tag names or classnames:

- `recipients_placeholder: '#recipients_placeholder'`
- `cc: '#cc'`
- `bcc: '#bcc'`

@michael-volynets @tomholub this practice is bad and should be avoided in future:

```js
this.composer.S.cached('email_copy_actions').find('span')
```

The reason why it's bad is that JS should not rely on tag names at all. In this PR I was changing `<span>`s to `<button>`s and was not expecting to break something on the JS side by doing that.

Copy-pasting the good practice suggested by @tomholub here:

> - tags only have html structure significance, and not used anywhere else
> - css file will only contain references to classes, not ids nor tags
> - both classes and ids may be used in JS, with preference to ids when there is just one thing to refer to
> - only data-test attrs should be used for tests, where possible (we already do this)

### DIFF
--- a/extension/chrome/elements/compose.htm
+++ b/extension/chrome/elements/compose.htm
@@ -47,7 +47,7 @@
       </tr>
       <tr>
         <td class="input recipients-inputs">
-          <div id="input_addresses_container" tabindex="0" style="display: none;">
+          <div id="input_addresses_container" class="invisible">
             <!-- to -->
             <div id="input-container-to" data-sending-type="to" data-test="container-to">
               <span class="recipients recipients-to"></span>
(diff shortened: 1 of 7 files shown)


tomholub:
To be honest, I would not mind to leave cc/bcc fields not accessible with keyboard. I think it should be a rare enough functionality that I don't feel bad if it's not accessible. But of course, since it's done, it doesn't hurt.

PR #2058, opened 2019-10-02
Issue #2056 Focus input_text asyncronously when replying

### LINKED ISSUE #2056: does not always auto-focus on text when replying - chrome
To reproduce, on Chrome:

1) open a conversation
2) click the reply frame area
3) start typing. Nothing gets typed, message body not focused

But interestingly, when you right-click the secure reply gray area and choose "reload frame", then click the reply frame area, it will get focused! ..

In either case, it works well on Firefox.

limonte:
Closes #2056 

To be honest, I do not understand what exaclty is the reason syncronous `.focus()` is not working in Chrome. It doesn't seem that something else is stealing focus from `input_text`.

The fix is a little ugly becase of so-called "magic" number `100`, I personally don't like magic numbers and try to avoid them when possible. @tomholub please let me know if it's too ugly to be in the codebase.

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -1090,8 +1090,7 @@ export class Composer {
     }
     if (this.urlParams.isReplyBox) {
       if (this.urlParams.to.length) {
-        // Chrome needs async focus (#2056)
-        setTimeout(() => {
+        Catch.setHandledTimeout(() => { // Chrome needs async focus: https://github.com/FlowCrypt/flowcrypt-browser/issues/2056
           this.S.cached('input_text').focus();
           document.getElementById('input_text')!.focus(); // #input_text is in the template
         }, 100);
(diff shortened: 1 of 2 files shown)


limonte:
I'm not sure why the code quality check is failing: https://flowcrypt.semaphoreci.com/jobs/dd6f10a3-1c67-4fbb-bf32-7c41c8c3b5b7

@tomholub could you give me a hint what's wrong with the code?

extension/js/common/composer.ts
```diff
@@ -1090,8 +1090,11 @@ export class Composer {
     }
     if (this.urlParams.isReplyBox) {
       if (this.urlParams.to.length) {
-        this.S.cached('input_text').focus();
-        document.getElementById('input_text')!.focus(); // #input_text is in the template
+        // Chrome needs async focus (#2056)
+        setTimeout(() => {
+          this.S.cached('input_text').focus();
+          document.getElementById('input_text')!.focus(); // #input_text is in the template
+        }, 100);
```
tomholub: It's complaining that if there was an exception thrown inside the anonymous arrow function, the exception would not get reported: a silently dropped error.

We have methods such as `Catch.setHandledTimeout` or something similar. I'll update and merge it.

The err message is not the best.

Issue #1789, opened 2019-07-01
make sure wrong msg checksum will show err


tomholub:
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/WhctKHTCDdzrhZvpxRGDFDVFZFqRsdXXDsDrJwpgXvFGWpbnmnpBJTpcdbLlHSGGHfsFMVL

tomholub:
```
https://github.com/openpgpjs/openpgpjs/issues/924
```

tomholub:
This has been fixed with an OpenPGP.js update, and it now correctly shows error:

![image](https://user-images.githubusercontent.com/6306961/65723788-a3425a00-e09e-11e9-8e0e-68a840139661.png)

It would be good to add a test to one of our decrypt tests, to test for the string:

`Ascii armor integrity check on message failed`

PR #2065, opened 2019-10-03
Issue #1789, added a test to check failure

michael-volynets:
Issue #1789,
Time Spent: 30m


test/source/tests/tests/decrypt.ts
```diff
@@ -183,6 +183,14 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithNewBrowser:
       await pubFrame.close();
     }));
 
+    ava.default('[standalone] - wrong message - checksum throws error', testWithNewBrowser(async (t, browser) => {
```
tomholub: here you could test with shared browser, because the test is not changing any settings. It will be faster.

Issue #1512, opened 2019-01-27
do not show private key - either copy to clipboard or save to file


tomholub:

tomholub:
copy to clipboard:
```
[Node was not found]
removeFake@moz-extension://39553fca-76d4-4791-bf00-b4dfede6fd45/lib/clipboard.js:239:19
selectFake/this.fakeHandlerCallback@moz-extension://39553fca-76d4-4791-bf00-b4dfede6fd45/lib/clipboard.js:203:26
catch.js:78:2
FlowCrypt ERROR: Please report errors above to human@flowcrypt.com. We fix errors VERY promptly.
```

tomholub:
![image](https://user-images.githubusercontent.com/6306961/51802380-70ab3d80-2241-11e9-965f-72fb1aa50204.png)

tomholub:
When saving to file the copy is weird, should not talk about decryption
```
File is ready.
Right-click the link and select Save Link As
DECRYPTED FILE
```

michael-volynets:
Okay, will do,

I've got an idea for the future:
Maybe show public and private keys on one page?
It will look similar to this:

![image](https://user-images.githubusercontent.com/49104848/66094912-5975da00-e59e-11e9-966a-b81b84e347a1.png)

What do you think?

tomholub:
Actually, that's a good idea. Less confusing, but I will re-arrange your design a bit, first.

PR #2066, opened 2019-10-03
Issue #1512, added ability to download/copy prv, and removed ability …

michael-volynets:
…to see it

Issue #1512,
Time Spent: 1h

There may be some duplications or similar code, that's because I was following original code
I think this file needs some refactoring.

### DIFF
--- a/extension/chrome/settings/modules/my_key.ts
+++ b/extension/chrome/settings/modules/my_key.ts
@@ -51,10 +51,16 @@ Catch.try(async () => {
   $('.show_when_showing_public').css('display', '');
   $('.show_when_showing_private').css('display', 'none');
 
-  $('.action_download_pubkey, .action_download_prv').click(Ui.event.prevent('double', (e) => {
-    const keyStr = e.className.includes('pubkey') ? primaryKi.public : primaryKi.private;
-    const keyAtt = new Att({ data: Buf.fromUtfStr(keyStr), type: 'application/pgp-keys', name: `0x${primaryKi.longid}.private.asc` });
-    Browser.saveToDownloads(keyAtt, Catch.browser().name === 'firefox' ? $('body') : undefined);
+  $('.action_download_pubkey').click(Ui.event.prevent('double', () => {
+    const name = `0x${primaryKi.longid}.asc`;
+    const pubKeyAtt = new Att({ data: Buf.fromUtfStr(primaryKi.public), type: 'application/pgp-keys', name });
+    Browser.saveToDownloads(pubKeyAtt, Catch.browser().name === 'firefox' ? $('body') : undefined);
+  }));
+
+  $('.action_download_prv').click(Ui.event.prevent('double', () => {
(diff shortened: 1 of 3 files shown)


extension/chrome/settings/modules/my_key.ts
```diff
@@ -50,27 +51,27 @@ Catch.try(async () => {
   $('.show_when_showing_public').css('display', '');
   $('.show_when_showing_private').css('display', 'none');
 
-  $('.action_download_pubkey').click(Ui.event.prevent('double', () => {
-    Browser.saveToDownloads(Att.keyinfoAsPubkeyAtt(primaryKi), Catch.browser().name === 'firefox' ? $('body') : undefined);
+  $('.action_download_pubkey, .action_download_prv').click(Ui.event.prevent('double', (e) => {
+    const keyStr = e.className.includes('pubkey') ? primaryKi.public : primaryKi.private;
```
tomholub: I really appreciate that you made it into one handler, but I'll refactor it as two handlers. Sorry - this is a rare case that's very security sensitive. I've caused a vulnerability before by wrongly using private key instead of pubkey before, so I'd rather have this separate, to make it harder to introduce a vulnerability when someone edits it in the future.

extension/chrome/settings/modules/my_key.ts
```diff
@@ -50,27 +51,27 @@ Catch.try(async () => {
...
       $('.action_show_other_type').text('show private key').removeClass('good').addClass('bad');
       $('.key_type').text('Public Key Info');
       $('.show_when_showing_public').css('display', '');
       $('.show_when_showing_private').css('display', 'none');
     }
   }));
 
-  const clipboardOpts = { text: () => prv.toPublic().armor() };
-  new ClipboardJS('.action_copy_pubkey', clipboardOpts); // tslint:disable-line:no-unused-expression no-unsafe-any
+  const clipboardOpts = { text: () => $('.action_show_other_type').text().toLowerCase() === 'show private key' ? primaryKi.public : primaryKi.private };
```
tomholub: For now ok but a bit similar comment to above - I guess this should be refactored in https://github.com/FlowCrypt/flowcrypt-browser/issues/2070

extension/chrome/settings/modules/my_key.ts
```diff
@@ -50,27 +51,27 @@ Catch.try(async () => {
   $('.show_when_showing_public').css('display', '');
   $('.show_when_showing_private').css('display', 'none');
 
-  $('.action_download_pubkey').click(Ui.event.prevent('double', () => {
-    Browser.saveToDownloads(Att.keyinfoAsPubkeyAtt(primaryKi), Catch.browser().name === 'firefox' ? $('body') : undefined);
+  $('.action_download_pubkey, .action_download_prv').click(Ui.event.prevent('double', (e) => {
+    const keyStr = e.className.includes('pubkey') ? primaryKi.public : primaryKi.private;
+    const keyAtt = new Att({ data: Buf.fromUtfStr(keyStr), type: 'application/pgp-keys', name: `0x${primaryKi.longid}.private.asc` });
```
tomholub: And this is actually an error, it would always say `.private.asc` at the end, even for public key, which may lead to confusion.

PR #2069, opened 2019-10-03
Issue #1957, Add toasts when clicking PK icon

### LINKED ISSUE #1957: show such toast when clicking PK icon
It would be great if we could show a simple toast after certain actions.

image

This would be in particular useful in the secure compose window. Eg when I click (or tab-enter) the public key icon, it could say `Attaching your Public Key to the message` or `Removing your Public Key from the message`. Same for signing. This could be the first usecase, but there may be other usecases where this could be super useful.

Toast should disappear within 2 seconds.

limonte:
Closes #1957


tomholub:
I'll look at the integration tests

tomholub:
Also, it would trigger toast messages even in situations when user would not expect it (like adding a recipient, removing recipient, etc). I'm updating it to only toast when user specifically clicked the PK button.

Issue #1517, opened 2019-02-04
Chrome focus after sending encrypted email


circld:
first off, huge thank you for an awesome extension! great idea and excellent execution. kudos.

**issue**
after replying to an encrypted message, the focus appears to interfere with Gmail shortcuts from registering/functioning. (this also impacts other extensions for driving using the keyboard, e.g., Vimium).

while this is a relatively minor issue (hitting tab a couple times and then attempting to shortcuts is a workaround), it is rather disruptive for power users used to driving Gmail almost entirely by keyboard.

**steps to reproduce**
1. enable Gmail shortcuts
2. focus on an encrypted conversation
3. use the `n` key to navigate down to the latest message in the conversation
4. press 'r' to reply to the message
5. type a message, press send (since `cmd-enter` is [not available as a shortcut](https://github.com/FlowCrypt/flowcrypt-browser/issues/183) yet, `tab` then `enter`)

at this point, the focus appears to be screwed up and no Gmail shortcuts work.

**system details**
OS: OSX
Browser: Version 72.0.3626.81 (Official Build) (64-bit)

tomholub:
That's interesting. We are open to contributions if someone wants to tackle this. We may also get to this ourselves sometime. Sorry to screw up your workflow!

tomholub:
If you are familiar with browser javascript, see if doing a vanilla JS equivalent to `$('body').focus()` in the dev console makes Gmail pick up on your shortcuts again. That would be good to know.

circld:
looks like upon sending a reply, the focus is on the body of the reply text:

<img width="542" alt="image" src="https://user-images.githubusercontent.com/5423134/52715467-27a5f980-2f62-11e9-8f03-9d0270b25a87.png">

running `> document.body.focus()` or `> $('body').focus()` doesn't seem to change the `document.activeElement`, nor restore shortcuts.

anything else you'd like me to try?

tomholub:
But were you attempting to focus on the body in the context if the inner iframe, or the Gmail page as a whole? Something to check.

Another thing to try: pick a Gmail element that is always there, that does not have any clickable buttons or links in it. Maybe the `Last account activity` on the bottom would be a good candidate. Try to issue a click programmatically on that element from the console. Would that restore shortcuts?

(if not, would clicking on it with your mouse fix the shortcuts?

tomholub:
Please verify this is still an issue, I have a feeling we may have fixed it.

[...]

PR #2084, opened 2019-10-04
Issue #1517, Focus body after sending message to make Gmail shortcuts work

limonte:
@tomholub I accidentally pressed the "Create PR" button instead of creating a draft :man_facepalming: 

I need you help here. This change should fix the issue, but for now it doesn't. It seems that this listener isn't being triggered https://github.com/FlowCrypt/flowcrypt-browser/pull/2084/files#diff-668908dbb47e93967b144b8ef0658932R109 and I can't understand why. Could you give me any hint on this?

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -789,7 +789,7 @@ export class Composer {
     BrowserMsg.send.notificationShow(this.urlParams.parentTabId, {
       notification: `Your ${isSigned ? 'signed' : 'encrypted'} ${this.urlParams.isReplyBox ? 'reply' : 'message'} has been sent.`
     });
-    BrowserMsg.send.focusBody(this.urlParams.parentTabId) // Bring focus back to body so Gmails shortcuts will work
+    BrowserMsg.send.focusBody(this.urlParams.parentTabId); // Bring focus back to body so Gmails shortcuts will work
     await this.composerDraft.draftDelete();
     this.isSendMessageInProgress = false;
     if (this.urlParams.isReplyBox) {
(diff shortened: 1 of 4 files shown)


tomholub:
One thing to note is that these BrowserMsg are asynchronous, so it may take a few more event loop cycles for it to actually do something.

Have you confirmed that the body of the handler runs? With a console.log/alert or something?

A side note, you could probably just re-use `focus_frame` since it does the same thing, but maybe have a special keyword like `root` or a null, so that it knows to focus the main frame. Then you don't need to be adding another browser msg type.

tomholub:
A few more hints:
1) this may be the first time you have edited a content script. When you edit content scripts (unlike editing extension frames), you have to actually reload the whole plugin in `chrome://extensions` or in `about:addons`, else the old code will still run (without any handler). Same for background scripts.
2) make sure the handler is called at all with a log/alert
3) make sure the other focus_frame handler is NOT run from elsewhere, overriding your focus
4) eventually make them both a single handler

Let me know if this helps

limonte:
> this may be the first time you have edited a content script. When you edit content scripts (unlike editing extension frames), you have to actually reload the whole plugin in chrome://extensions or in about:addons, else the old code will still run (without any handler). Same for background scripts.

This was helpful, thank you! The PR is ready for review now. I decided to keep the `focus_body` listener because now it's actually different from `focus_frame` and also IMO `focusBody` it's better for code readability than `focusFrame(null)`.

tomholub:
It works!!! :tada:

Issue #2076, opened 2019-10-03
replying to signed message should produce signed message


tomholub:
1) open a signed message
2) click reply
3) button will wrongly say "Encrypt and Send" (supposed to be Sign and Send)
4) I'm not able to change the button to sign and send. I CAN change it to `Encrypt, Sign and Send`, but not to `Sign and Send`

Please make a test for this first, and make sure the test is failing. Then produce a fix in another commit. That way we know this won't break in the future.

michael-volynets:
I'll actually extend this one
```TS
ava.default.only('compose[global:compatibility] - reply - signed message'
```
instead of adding new

tomholub:
Definitely

Issue #2031, opened 2019-09-21
move send btn logic into a separate composer file


tomholub:
> Also, maybe we can consider creating a separate composer/composer-send-btn.ts file where we could put some of this logic? Anything that concerns how the send button behaves.

Yes, that would be great I think but I would recommend you to do it in another PR. And maybe move there all logic related to sending messages.

_Originally posted by @michael-volynets in https://github.com/FlowCrypt/flowcrypt-browser/pull/2028#issuecomment-533811233_

PR #2091, opened 2019-10-07
Issue 2076 replying to signed msg

michael-volynets:
Issue #2076,
Time Spent: 1h 30m (1h for a test, 30m for fixing)

I know this is not the best way of fixing it and I'm going to refactor it in #2031, 
I want to change `  private popoverItems?: ComposerPopoverItem[];` from array to dictionary (it would simplify send btn code) but this thing isn't for this PR so I decided to do it today later in another PR.

### DIFF
--- a/extension/js/common/composer.ts
+++ b/extension/js/common/composer.ts
@@ -312,7 +312,7 @@ export class Composer {
       await this.composerContacts.setEmailsPreview(this.getRecipients());
     }
     this.initComposerPopover();
-    this.waitUntilRecipientsLoadedAndSetTestStateReady().catch(Catch.reportErr);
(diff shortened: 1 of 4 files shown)


extension/js/common/composer.ts
```diff
@@ -1286,6 +1283,11 @@ export class Composer {
     }
   }
 
+  private waitUntilRecipientsLoadedAndSetTestStateReady = async () => {
```
tomholub: I'll go ahead and rename this to `loadRecipientsThenSetTestStateReady`

michael-volynets: Ok, Thanks!

PR #2093, opened 2019-10-08
issue #2092 option to only build static assets

### LINKED ISSUE #2092: option to only build static assets
Roger, thanks!

While working on this, I got a question. While working on `htm` files, it should be possible to quickly copy them to the build folders without rebuilding the whole plugin. Right now, the development process is slow for me because in order to see changes in `htm` files, I'm rebuilding the whole plugin which takes time and also kinda distracts focus.

@tomholub is there any way to copy static files (`htm`, `css`, etc.) without rebuilding the whole plugin?

_Originally posted by @limonte in https://github.com/FlowCrypt/flowcrypt-browser/issues/398#issuecomment-538880863_

tomholub:
close #2092

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,8 +4,14 @@ OUTDIR=./build/generic-extension-wip
 SRCDIR=./extension
 
 set -euxo pipefail
-
 shopt -s globstar
+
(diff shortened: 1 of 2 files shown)


tomholub:
@limonte you'll be able to use `npm run-script build-assets-only` or configure it as a task in your IDE.

PR #2097, opened 2019-10-09
Set Up: enforce users to write down their passphrase

### LINKED ISSUE #1579: visual cue to write pass phrase down on paper
I like how keybase does this. We keep asking users to note their pass phrase down, but it would be really good to reinforce it with a graphic that looks like a paper with pass phrase written on it.

image

limonte:
Closes #1579 

image

### DIFF
--- a/extension/chrome/settings/setup.ts
+++ b/extension/chrome/settings/setup.ts
@@ -521,8 +521,8 @@ Catch.try(async () => {
   });
 
   const isCreatePrivateFormInputCorrect = async () => {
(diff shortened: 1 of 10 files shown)


limonte:
While working on this feature I found two issues in SweetAlert2 - https://github.com/sweetalert2/sweetalert2/pull/1766 and https://github.com/sweetalert2/sweetalert2/pull/1767

Fixes in SweetAlert2 are released, I bumped the lib in separated commits.

@tomholub The PR is ready for review now.

PR #2105, opened 2019-10-11
Fix SweetAlert2's font-size for Bootstrap 3

limonte:
We're using Bootstrap 3 (`bootstrap.min.css`) in `settings/index.htm` and `settings/setup.htm`.

SweetAlert2 has a small well-known issue with Bootstrap 3. Because Bootstrap 3 sets `font-size: 10px` to `html`, modals look small:

image

So I added the solution from https://github.com/sweetalert2/sweetalert2/issues/834#issuecomment-359237178 to `bootstrap.min.css`. 

Adding changes to third-party libraries is usually a bad practice, but in this particular case I think it's better because:

- Probably we will not upgrade Bootstrap at all, but rather get rid of it
- If we upgrade Bootstrap, that upgrade will be to Bootstrap 4 and Bootstrap 4 doesn't have that issue

In both cases (getting rid of bootstrap or upgrading to 4) we want the change from this PR to be removed as well. In order not to forget to remove it, it's better to have it in `bootstrap.min.css` rather than for example in `settings.css`.

With this change modals look normal:

image


limonte:
@tomholub apologies for not catching this issue in my previous PR #2104

Issue #984, opened 2018-01-17
forwarded messages should include files/attachments


tomholub:
- should work with detached attachments
 - should also work with pgp/mime

tomholub:
ref https://mail.google.com/mail/u/0/#inbox/160fed6e9ffb3fa4

tomholub:
This one could be workable (but second priority).

PR #2116, opened 2019-10-16
[9h] Issue #984, Decrypt and add attachments from previous message

michael-volynets:
Issue #984,
Time Spent: 5h

Now when we forward a message the attachments will be added to the forwarded message too,
Originally I was thinking to send encrypted attachments without decryption and encryption again but then I realized that we will have a `key mismatch` issue so I had to decrypt and encrypt them with primary user private key.

### DIFF
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -138,7 +138,7 @@ Catch.try(async () => {
         const container = $('#bulk_import #processed');
         for (const block of blocks) {
           if (block.type === 'publicKey') {
-            const replacedHtmlSafe = XssSafeFactory.replaceRenderableMsgBlocks(factory, String(block.content));
+            const replacedHtmlSafe = XssSafeFactory.replaceRenderableMsgBlocks(factory, block.content.toString());
             if (replacedHtmlSafe && replacedHtmlSafe !== value) {
               container.append(replacedHtmlSafe); // xss-safe-factory
             }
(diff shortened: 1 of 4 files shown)


tomholub:
Yes, you cannot just pass them along as is, because the recipient may differ. They indeed have to be decrypted and re-encrypted.

michael-volynets:
Great improvement, actually I didn't think about it, Thanks!

michael-volynets:
So, now when replying on some message the attachments won't be decrypted but I think to avoid loading attachments at all we need to change getting message from `raw` format to `full` but to do this we have to change almost whole `getAndDecryptMessage` method so I would recommend to move it to another issue.

michael-volynets:
+30m

tomholub:
It works for detached attachments, but not for pgp/mime type of attachments.

Try this message: https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#label/tests%2Fattachment/WhctKHTCKmjFMNVCrvMcxSVtFQHKSBTflcZqSlQqTZSsWjNgQwXQjMNPHblkSqRscsRghmQ

tomholub:
Another issue is that when I click `Forward`, the recipients field should be empty, and in focus. I'll make another issue for this.

michael-volynets:
let me check, Thanks

michael-volynets:
+2h 30m

michael-volynets:
Lookin into tests

michael-volynets:
+ 30m

Issue #2118, opened 2019-10-17
Forward should show empty recipients + recipients in focus


tomholub:
Another issue is that when I click `Forward`, the recipients field should be empty, and in focus. I'll make another issue for this.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/2116#issuecomment-543297500_

There is code somewhere that will make recipients in focus if it's a new message, so you could add an OR there to handle if this is a forwarded message.

This must be a relatively recent regression. Please add a test.

michael-volynets:
`There is code somewhere that will make recipients in focus if it's a new message, so you could add an OR there to handle if this is a forwarded message.` I couldn't find this code so I added focus in `renderReplyMsgComposeTable` function

PR #2127, opened 2019-10-21
Fixed bug when forwarding, fixed another issue with cc/bcc buttons, a…

michael-volynets:
…dded test

Issue #2118,
Time Spent: 2h 30m,

Fixed issue when forwarding messages and during working on this one I found another issue.
Another issue is when you clicked on cc/bcc button and hide inputs the cc/bcc button will never show, it's also fixed in this PR.
Added a simple test.


michael-volynets:
Looking into tests

michael-volynets:
+40m for fixing tests

Issue #2039, opened 2019-09-26
remove all footer settings, use sendAs 


tomholub:
The tricky part will be: please add footer on top of the quoted part.

 - if this is a reply, it will be appended on top of the quote
 - if this is a new message, it will be the quote

In either case, it will be hidden under the `...`

In either case, our existing mechanisms (when storing drafts, loading drafts, etc) should pick it up so that the footer should not double up.

tomholub:
Allow this for everybody, subscribed or not.

michael-volynets:
```
if this is a reply, it will be appended on top of the quote
if this is a new message, it will be the quote
```
To clarify,
Now when we send a message we don't encrypt footer but if we have footer under `...` we will have to encrypt it in the whole message. Am I correct?

tomholub:
Yes, you would encrypt the footer, too. Thanks for asking!

PR #2134, opened 2019-10-22
[10h] Issue 2039 changing footer

michael-volynets:
Issue #2039,
Time Spent: 8h

It's ready but I want to refactor the code before it's merged. If you want you can test this feature, whatever you decide will be the right decision.
I will do some minor changes and fix tests if they fail.


michael-volynets:
I'm also thinking about adding tests for this issue.
What do you think about it?

michael-volynets:
+30m spent on Refactoring

tomholub:
Ok, you could add tests, but for this one I think it would be Third Priority milestone

michael-volynets:
Ok, I will create an Issue

extension/chrome/elements/compose.ts
```diff
@@ -68,7 +68,19 @@ Catch.try(async () => {
       }
       return result;
     };
-    return storage.sendAs || (storage.addresses && arrayToSendAs(storage.addresses));
+    const result = storage.sendAs || (storage.addresses && arrayToSendAs(storage.addresses));
+    if (result && storage.email_footer) {
```
tomholub: I would ignore whatever was set in `email_footer`, because going forward, footers should just be set from Gmail.

If user has no footer set on Gmail, then there is no footer.

extension/chrome/elements/compose.ts
```diff
@@ -380,13 +387,21 @@ Catch.try(async () => {
...
     collectAllAvailablePublicKeys
-  }, processedUrlParams, await Store.subscription());
+  }, processedUrlParams);
 
   BrowserMsg.addListener('close_dialog', async () => {
     $('.featherlight.featherlight-iframe').remove();
   });
-  BrowserMsg.addListener('set_footer', async ({ footer }: Bm.SetFooter) => {
-    storage.email_footer = footer;
+  BrowserMsg.addListener('set_footer', async ({ email, footer }: Bm.SetFooter) => {
```
tomholub: There should not be any need for a set_footer BrowserMsg. It was used because there was a separate footer.htm frame, which is no longer needed. Since there is no possibility to set a different footer than what is on Gmail, this should be removed.

extension/chrome/elements/shared/footer.ts
```diff
@@ -4,7 +4,7 @@
 'use strict';
```
tomholub: Please remove footer.ts completely, will only get footers from Gmail

extension/js/common/composer.ts
```diff
@@ -210,8 +205,9 @@ export class Composer {
     this.S.cached('input_text').get(0).onpaste = this.inputTextPasteHtmlAsText;
     this.S.cached('icon_footer').click(Ui.event.handle(target => {
```
tomholub: There probably doesn't need to be a footer icon anymore, either. User can change their footer on Gmail, or edit it by clicking the three dots.

extension/js/common/xss_safe_factory.ts
```diff
@@ -77,8 +77,8 @@ export class XssSafeFactory {
     return this.frameSrc(this.extUrl('chrome/elements/add_pubkey.htm'), { emails, placement });
   }
 
-  srcAddFooterDialog = (placement: Placement, grandparentTabId: string) => {
-    return this.frameSrc(this.extUrl('chrome/elements/shared/footer.htm'), { placement, grandparentTabId });
+  srcAddFooterDialog = (placement: Placement, grandparentTabId: string, emailAlias?: string) => {
```
tomholub: Could probably be removed

michael-volynets:
Thanks for the review, 
Will do

extension/chrome/elements/compose.ts
michael-volynets: Agree, removed

extension/chrome/elements/shared/footer.ts
michael-volynets: Removed

extension/js/common/xss_safe_factory.ts
michael-volynets: Yes, we don't need it anymore, removed

extension/chrome/elements/compose.ts
michael-volynets: Agree, removed

michael-volynets:
Done, +1h
(Total 9h 30m)

extension/js/common/platform/xss.ts
```diff
@@ -64,8 +68,8 @@ export class Xss {
     return cleanHtml;
   }
 
-  public static htmlSanitizeAndStripAllTags = (dirtyHtml: string, outputNl: string): string => {
-    let html = Xss.htmlSanitizeKeepBasicTags(dirtyHtml);
+  public static htmlSanitizeAndStripAllTags = (dirtyHtml: string, outputNl: string, removeImgs: boolean = false): string => {
```
tomholub: Can you please share more info about how `removeImgs` is used, when and why? This would probably make for a good comment

tomholub: Can you please share more info about how removeImgs is used, when and why? This would probably make for a good comment

michael-volynets: It's used because when the signature contains an image FlowCrypt won't render the image in the composer, it will instead render show image text, so I decided to remove Images from signature at all.
I used that here:
```TS
  private createFooterHTML = (footer: string) => {
    return `<br>-<br><br>${Xss.htmlSanitizeAndStripAllTags(footer, '<br>', true)}`;
  }
```

maybe when we have a WYSIWYG editor in FlowCrypt we could show images.

tomholub: Understood. I don't like footer images anyway, we don't have to support them.

tomholub:
We should do a more gracious form of updating sendAs in composer, since now it will also notice footer changes, and it's too obstructing to always ask user to "reload the window".

So that when sendAs changes, detected by the following method:

```js
  static refreshAcctAliases = async (acctEmail: string): Promise<boolean> => {
    const fetchedSendAs = await Settings.fetchAcctAliasesFromGmail(acctEmail);
    const { sendAs: storedAliases, addresses: oldStoredAddresses } = (await Store.getAcct(acctEmail, ['sendAs', 'addresses']));
    await Store.setAcct(acctEmail, { sendAs: fetchedSendAs });
    if (!storedAliases) { // Aliases changed (it was previously undefined)
      if (oldStoredAddresses) { // Temporary solution
        return Object.keys(fetchedSendAs).sort().join(',') !== oldStoredAddresses.sort().join(',');
      }
      return true;
    }
    if (Settings.getDefaultEmailAlias(fetchedSendAs) !== Settings.getDefaultEmailAlias(storedAliases)) { // Changed (default email alias was changed)
      return true;
    }
    if (Object.keys(fetchedSendAs).sort().join(',') !== Object.keys(storedAliases).sort().join(',')) { // Changed (added/removed email alias)
      return true;
    }
    if (Object.keys(fetchedSendAs).filter(email => fetchedSendAs[email].footer).map(email => fetchedSendAs[email].footer).sort().join(',') !==
      Object.keys(storedAliases).filter(email => storedAliases[email].footer).map(email => storedAliases[email].footer).sort().join(',')) {
      return true;
    }
    return false; // Nothing changed
  }
```

It should transparently update sender options and also the footer information for use in the same secure window immediately.

Issue #1452, opened 2019-01-01
better message signing test - try to verify


tomholub:

tomholub:
Not only test that we can send a signed message, but also check that the message can be verified.

Et the very least the mock api should check this.

PR #2135, opened 2019-10-23
[2h] Issue #1452 test signed messages

michael-volynets:
Issue #1452 
Time Spent: 1h 40m

When I finished writing the test and I ran it, unfortunately, it didn't pass, I didn't know why because the test was written correctly. The problem was because of `email-mime-codec` `foldLines` method, it added a new line symbol and the message changed.

Could you please explain to me why do we need this method?

### DIFF
--- a/test/source/mock/strategies/send-message-strategy.ts
+++ b/test/source/mock/strategies/send-message-strategy.ts
@@ -13,7 +13,6 @@ class SignedMessageTestStrategy implements ITestMsgStrategy {
     const decrypted = await PgpMsg.decrypt({ kisWithPp: keyInfo!, encryptedData: Buf.fromUtfStr(mimeMsg.text) });
     if (decrypted.success && decrypted.signature) {
       const content = decrypted.content.toUtfStr();
-      console.log(content);
       if (!content.includes(this.expectedText)) {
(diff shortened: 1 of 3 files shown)


tomholub:
It's descried here, if we don't fold them ourselves before signing, Gmail will fold them for us after signing, resulting in faulty signatures (text changed after signing).
```js
            // Folding the lines or GMAIL WILL RAPE THE TEXT, regardless of what encoding is used
            // https://mathiasbynens.be/notes/gmail-plain-text applies to API as well
            // resulting in.. wait for it.. signatures that don't match
            // if you are reading this and have ideas about better solutions which:
            //  - don't involve text/html ( Enigmail refuses to fix: https://sourceforge.net/p/enigmail/bugs/218/ - Patrick Brunschwig - 2017-02-12 )
            //  - don't require text to be sent as an attachment
            //  - don't require all other clients to support PGP/MIME
            // then please const me know. Eagerly waiting! In the meanwhile..
            plaintext = (window as unknown as BrowserWidnow)['emailjs-mime-codec'].foldLines(plaintext, 76, true); // tslint:disable-line:no-unsafe-any
```
You could just test a shorter message that doesn't get folded.

tomholub:
No prob, that's why we have a review process

Issue #1606, opened 2019-03-26
MIME is rendered as body on encrypted e-mails with attachments but no body


lkslawek:
When I'm viewing an e-mail that has no body but only attachment, FlowCrypt shows MIME as message body. The attachment is still available for download.

I've sent the test e-mail from Apple Mail 12.2 (3445.102.3) with GPGMail 3.0.1.

The decrypted `encrypted.asc` constructed by this client is:

```
Content-Type: multipart/mixed;
	boundary="Apple-Mail=_2E6442E0-11D3-4D77-B3C0-93356DA62015"


--Apple-Mail=_2E6442E0-11D3-4D77-B3C0-93356DA62015
Content-Disposition: attachment;
	filename=attachment.txt
Content-Type: text/plain;
	x-unix-mode=0644;
	name="attachment.txt"
Content-Transfer-Encoding: 7bit

An attachment.

--Apple-Mail=_2E6442E0-11D3-4D77-B3C0-93356DA62015--
```

And this above is exactly what gets displayed in Gmail:

<img width="429" alt="Screenshot 2019-03-26 at 16 31 39" src="https://user-images.githubusercontent.com/3786106/55010557-ad4aa980-4fe4-11e9-8e54-3f9317cfe1b8.png">

I've tried it also with Thunderbird + Enigmail pair, but it includes a body part, even if empty.

Also above is how Apple Mail is sending unencrypted e-mails with no character in body, that is, only with attachment part in MIME. I do not know whether the body part is required or not formally, so please decide yourselves if this is a bug for fixing or not.

Observed on Firefox 65.0.1 and FlowCrypt 6.7.0.

tomholub:
Great bug (compatibility) report, thank you. 

Could you please have such email sent to flowcrypt.compatibility@gmail.com

Pubkey at FlowCrypt.com/lookup

Subject: Apple Mail attachment-only PGP/mime with empty body

File content: please add link to this GitHub issue

Thanks again!

lkslawek:
Sure. Actually, I've sent there two e-mails, because accidentally I've discovered just know, that in the same case, when the e-mail is only signed and not encrypted, FlowCrypt shows error "Could not open this message with FlowCrypt" in place of the body. The attachment is available for download also in this case.

So one encrypted and not signed, and one signed and not encrypted.

tomholub:
Awesome, that helps!

tomholub:
The received test emails are at 

[...]

PR #2138, opened 2019-10-24
[3h] Issue #1606, corrected Mime.decode method

michael-volynets:
Issue #1606,
Time Spent: 2h 30m (Almost all time I was investigating why it isn't working),

I'm not sure it's the right solution but I tried to make it work as Thunderbird does.
I tried to open this message in Thunderbird and they render the attachment and the text. So I did the same. Now, FlowCrypt decrypts the message correctly.

If you approve this solution I'll start writing the tests.

### DIFF
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -378,8 +378,10 @@ Catch.try(async () => {
         isHtml = true;
       } else if (typeof decoded.text !== 'undefined') {
         decryptedContent = decoded.text;
+      } else {
+        decryptedContent = '';
       }
-      if (decoded.subject) {
+      if (decoded.subject && isEncrypted) {
         decryptedContent = getEncryptedSubjectText(decoded.subject, isHtml) + decryptedContent;
       }
       for (const att of decoded.atts) {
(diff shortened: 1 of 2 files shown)


tomholub:
`!Mime.getNodeFilename(node)` is there so that it would not confuse attachments (which have a filename) with `text/html` or `text/plain` body. For example, you could attach a html file, and it would wrongly assume it's the body (or txt file, like above and like Thunderbird does it, which is wrong).

tomholub:
Corrected, now it looks as follows:

image

There is some duplication of the attachment which is ugly but does not bother me too much.

extension/chrome/elements/pgp_block.ts
```diff
@@ -378,8 +378,10 @@ Catch.try(async () => {
         isHtml = true;
       } else if (typeof decoded.text !== 'undefined') {
         decryptedContent = decoded.text;
+      } else {
+        decryptedContent = '';
       }
-      if (decoded.subject) {
+      if (decoded.subject && isEncrypted) {
```
tomholub: Here I have fixed unrelated issue, it does not make sense to show `Encrypted Subject` if this was a signed message.

extension/chrome/elements/pgp_block.ts
```diff
@@ -378,8 +378,10 @@ Catch.try(async () => {
         isHtml = true;
       } else if (typeof decoded.text !== 'undefined') {
         decryptedContent = decoded.text;
+      } else {
+        decryptedContent = '';
```
tomholub: This takes care of setting the (empty), so that the whole mime message does not get wrongly rendered as abefore.

Issue #920, opened 2017-09-29
Reply in conversation to not last element


axkibe:
Am I doing something wrong or is it missing. I cannot reply to a conversation that is not the last message. In "original" view, I have a reply button on every message, but in crypt-view I only can reply to the last message of conversation.

tomholub:
For now, that's indeed a limitation. The thing is, CryptUp doesn't really quote the message contents anyway, so in a conversation, even if you were allowed to reply to a middle message, the recipient would probably not see a difference (as long as they are using conversation view).

Once we start quoting messages, then it will make sense to allow this.

axkibe:
Well I had to copy/paste the email adresses by hand, since the last reply I got was a personal reply to a conversation that included a bunch of people.

I don't know if each just quoting or other email headers involved, my reply that should go to the group should have nothing to do with the personal reply I received.

PS: This is also something that didn't work, dragging the emails into the new composition field. Or to copy/paste the email boxes, I had to expant them and manually copy/paste each email address.

tomholub:
There are email headers involved, at this moment we're always referring to the last message in conversation as you'd expect.

tomholub:
depends on https://github.com/FlowCrypt/flowcrypt-browser/issues/873 which in turns depends on https://github.com/FlowCrypt/flowcrypt-browser/issues/92

tomholub:
Marking actionable, third priority - the blocking issues have been fixed.

Have to re-work the logic to know which exact message we are replying to, instead of always assuming it's the last message in conversation.

Issue #2128, opened 2019-10-21
merge `reply_pubkey_mismatch.htm` with `compose.htm`


michael-volynets:
I think that we need to totally remove `reply_pubkey_mismatch.htm` and use `compose.htm` instead of that. 
The reason is when I made some new features for `compose.htm` I also have to check if everything works well `reply_pubkey_mismatch.htm`.
Like in issue #2119.

tomholub:
I'm fine with this, but will only merge the PR if it's obvious that it's an improvement to code clarity - as long as `composer.js` or `compose.htm` doesn't get too complicated as a result.

tomholub:
I'm refactoring `compose.ts` in #2284 and will cover this too

Issue #1735, opened 2019-05-31
Standard reply box gets replaced when you want to send a nonenrypted email.


michael-volynets:
**editted by Tom with latest issue description to avoid confusion**
----------------------

Existing behavior to preserve:
- if last message is encrypted, and I click encrypted reply, render encrypted reply box
- if last message is plain, and I click plain reply, render plain reply
- if last message is plain, and I click encrypted reply, render encrypted reply

Current unwanted behavior:
 - if last message is encrypted, and I **click plain reply icon on top right**, it still shows an encrypted reply box

Instead, we should find a way to NOT replace the plain reply box that shows **after explicitly clicking on plain reply** and instead augument it with a very bright red box above it, which says: `The last message was encrypted, but you are composing a reply without encryption. [switch to encrypted reply]`

In a separate issue, we should have a way to switch from within the secure reply to replying without encryption.

That should be less invasive and make everyone happy.

tomholub:
This is historically intended behavior, but should improve it.

I will add some notes here over the weekend.

tomholub:
- add info about reply buttons in general

tomholub:
#920 is somewhat affecting it, once that one lands, I will re-visit this issue

tomholub:
I'm not entirely sure what is the right approach here.

Users will habitually be clicking the original reply button and think that the reply will be encrypted, but it won't.

At the same time, more mindful users will sometimes want to actually reply without encryption, but we don't allow them atm.

Best would be if it's possible to show a warning, something like:

> The message you are replying to is encrypted. Do you really want to reply without encryption?

But I'm really not sure.

tomholub:
Moving to discussion, curious about your input

tomholub:
Thought about this some more.

Existing behavior to preserve:
- if last message is encrypted, and I click encrypted reply, render encrypted reply box
- if last message is plain, and I click plain reply, render plain reply
- if last message is plain, and I click encrypted reply, render encrypted reply

Current unwanted behavior:
 - if last message is encrypted, and I **click plain reply icon on top right**, it still shows an encrypted reply box

Instead, we should find a way to NOT replace the plain reply box that shows **after explicitly clicking on plain reply** and instead augument it with a very bright red box above it, which says: `The last message was encrypted, but you are composing a reply without encryption. [switch to encrypted reply]`

In a separate issue, we should have a way to switch from within the secure reply to replying without encryption.

That should be less invasive and make everyone happy.

[...]

PR #2150, opened 2019-10-29
#920, Reply To not last element

michael-volynets:
Issue #920,
Time Spent: 20h

Currently, I haven't implemented drafts when replying to not last element, I created another issue (#2151)
It wasn't really a hard issue but most of the time I spent on testing because when I changed something in files like `gmail_element_replacer.ts` I should have tested everything works correctly
Tomorrow I want to go thought the changed files again because I might miss something. So you can start reviewing it but please don't merge it.
Thanks.


extension/chrome/elements/compose.ts
```diff
@@ -31,17 +31,20 @@ Catch.try(async () => {
...
   const frameId = Assert.urlParamRequire.string(uncheckedUrlParams, 'frameId');
   const skipClickPrompt = uncheckedUrlParams.skipClickPrompt === true;
   const ignoreDraft = uncheckedUrlParams.ignoreDraft === true;
+  const removeAfterClose = uncheckedUrlParams.removeAfterClose === true;
   const placement = Assert.urlParamRequire.oneof(uncheckedUrlParams, 'placement', ['settings', 'gmail', undefined]);
   const disableDraftSaving = false;
   const debug = uncheckedUrlParams.debug === true;
   let draftId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'draftId') || '';
   let threadId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadId') || '';
+  const threadMsgId = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'threadMsgId') || '';
```
tomholub: We should call this `replyToMsgId` for clarity, since that's how we're using it now

tomholub: That means maybe we don't need `threadId`? Since we have to make an api call anyway? (to retrieve message info - in the same api call we can retrieve `threadId` instead of getting it through url params?)

extension/chrome/elements/compose.ts
```diff
@@ -74,12 +77,11 @@ Catch.try(async () => {
...
   }
   const replyParams: { from: string, subject: string, to: string[], cc: string[], bcc: string[] } = { from: '', subject: '', to: [], cc: [], bcc: [] };
-  if (threadId) {
+  if (threadMsgId) {
     const fetchSuccess = await (async () => {
       Xss.sanitizePrepend('#new_message', Ui.e('div', { id: 'loader', html: 'Loading secure reply box..' + Ui.spinner('green') }));
       try {
-        const thread = await Google.gmail.threadGet(acctEmail, threadId, 'metadata');
-        const gmailMsg = await Google.gmail.msgGet(acctEmail, thread.messages[thread.messages.length - 1].id, 'metadata');
+        const gmailMsg = await Google.gmail.msgGet(acctEmail, threadMsgId, 'metadata');
```
tomholub: We saved one API roundtrip, that's good.

extension/js/common/composer.ts
```diff
@@ -412,22 +412,19 @@ export class Composer {
...
-        this.urlParams.subject = `${(method === 'reply' ? 'Re' : 'Fwd')}: ${this.urlParams.subject}`;
-        this.composerSendBtn.additionalMsgHeaders['In-Reply-To'] = determined.headers['In-Reply-To'];
-        this.composerSendBtn.additionalMsgHeaders.References = determined.headers.References;
-        if (!this.urlParams.draftId) { // if there is a draft, don't attempt to pull quoted content. It's assumed to be already present in the draft
-          (async () => { // not awaited because can take a long time & blocks rendering
-            const footer = this.getFooter();
-            await this.composerQuote.addTripleDotQuoteExpandBtn(determined.lastMsgId, method, footer);
-            if (this.composerQuote.messageToReplyOrForward && this.composerQuote.messageToReplyOrForward.isSigned) {
+      this.urlParams.subject = `${(method === 'reply' ? 'Re' : 'Fwd')}: ${this.urlParams.subject}`;
+      if (!this.urlParams.draftId) { // if there is a draft, don't attempt to pull quoted content. It's assumed to be already present in the draft
```
tomholub: Currently, when sending from a draft, what will happen?

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -82,22 +82,24 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   setReplyBoxEditable = async () => {
     const replyContainerIframe = $('.reply_message_iframe_container > iframe').last();
     if (replyContainerIframe.length) {
-      $(replyContainerIframe).replaceWith(this.factory.embeddedReply(this.getReplyParams(this.getGonvoRootEl(replyContainerIframe[0])), true)); // xss-safe-value
+      $(replyContainerIframe).replaceWith(this.factory.embeddedReply(this.getLastMsgReplyParams(this.getGonvoRootEl(replyContainerIframe[0])), true)); // xss-safe-value
```
tomholub: Try this:
1. open a conversation where the first message was encrypted, and second was plain, and third was plain
2. make open the middle message, but not the first
3. click to reply to the middle message (plain) and wait for standard reply box to render
4. now click to open the first (encrypted) message in the conversation. element replacer should replace the reply box
5. I suspect it will replace the reply box thinking that you are replying to the last message in conversation? Even though user is replying to the middle one.

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -130,28 +132,23 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
-        }
+      // only replace with FlowCrypt reply button if does not have any buttons replaced yet, and only replace the last one
+      for (const elem of convoReplyBtnsToReplaceArr) {
+        $(elem).addClass('inserted');
+        const element = $(this.factory.btnReply()).insertBefore($(elem).children().last());  // xss-safe-factory
+        const messageContainer = $(elem.closest('.h7') as HTMLElement);
+        element.click(Ui.event.prevent('double', Catch.try(async () => {
+          if (messageContainer.is(':last-child')) {
+            if (isEncrypted) {
+              await this.setReplyBoxEditable();
```
tomholub: note - I have to look at implications of this

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -130,28 +132,23 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
-        // hide original gmail reply buttons, except for the last one, but ONLY IF this is an encrypted conversation
-        // the reason is that currently encrypted replies only support replying to last message. If some other reply button
-        // was clicked, and FlowCrypt replaced the reply box with an encrypted one mid-message, it would create a false imagination
-        // for the user that they are replying mid-convo, when in fact it would end up replying to the last message anyway
-        // todo - revisit this (and potentially remove) when we support encrypted replies mid-convo
-        for (const replyBtn of convoReplyBtnsToReplaceArr) { // all other non-last reply buttons to be hidden
-          $(replyBtn).addClass('replaced').text(''); // hide all except last
-        }
+      // only replace with FlowCrypt reply button if does not have any buttons replaced yet, and only replace the last one
+      for (const elem of convoReplyBtnsToReplaceArr) {
```
tomholub: note - I have to give this a second look

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -130,28 +132,23 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
-          element.click(Ui.event.prevent('double', Catch.try(() => this.replaceStandardReplyBox(true, true)))); // xss-safe-factory
-        }
-      }
-      if (isEncrypted) {
-        // hide original gmail reply buttons, except for the last one, but ONLY IF this is an encrypted conversation
-        // the reason is that currently encrypted replies only support replying to last message. If some other reply button
-        // was clicked, and FlowCrypt replaced the reply box with an encrypted one mid-message, it would create a false imagination
-        // for the user that they are replying mid-convo, when in fact it would end up replying to the last message anyway
-        // todo - revisit this (and potentially remove) when we support encrypted replies mid-convo
-        for (const replyBtn of convoReplyBtnsToReplaceArr) { // all other non-last reply buttons to be hidden
```
tomholub: Good, related issue: #1735

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -437,13 +434,29 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     return $(anyInnerElement).closest('div.if, td.Bu').first();
   }
 
-  private replaceStandardReplyBox = async (editable: boolean = false, force: boolean = false) => {
+  private insertEncryptedReplyBox = async (messageContainer: JQuery<HTMLElement>) => {
+    if (!messageContainer.hasClass('h7')) {
+      throw new Error("Incorrect message container");
```
tomholub: Would this be user facing error? How would it be handled?

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -437,13 +434,29 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
+    }
+    const convoRoot = $(this.sel.convoRoot);
```
tomholub: Note - need to give this section a second look

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -130,28 +132,23 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
-      if (isEncrypted) {
-        // hide original gmail reply buttons, except for the last one, but ONLY IF this is an encrypted conversation
-        // the reason is that currently encrypted replies only support replying to last message. If some other reply button
-        // was clicked, and FlowCrypt replaced the reply box with an encrypted one mid-message, it would create a false imagination
-        // for the user that they are replying mid-convo, when in fact it would end up replying to the last message anyway
-        // todo - revisit this (and potentially remove) when we support encrypted replies mid-convo
-        for (const replyBtn of convoReplyBtnsToReplaceArr) { // all other non-last reply buttons to be hidden
-          $(replyBtn).addClass('replaced').text(''); // hide all except last
-        }
+      // only replace with FlowCrypt reply button if does not have any buttons replaced yet, and only replace the last one
```
tomholub: "only replace the last one" is an outdated comment

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -130,28 +132,23 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     const visibleReplyBtns = $('td.acX:visible');
     if (visibleReplyBtns.not('.replaced, .inserted').length) { // last reply button in convo gets replaced
       const convoReplyBtnsToReplace = visibleReplyBtns.not('.replaced, .inserted');
-      const hasVisibleReplacements = visibleReplyBtns.filter('.replaced, .inserted').length > 0;
       const convoReplyBtnsToReplaceArr = convoReplyBtnsToReplace.get();
```
tomholub: `convoReplyBtnsToReplaceArr` should be called `convoReplyBtnsArr` (we are not replacing them anymore)

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -130,28 +132,23 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
-        // for the user that they are replying mid-convo, when in fact it would end up replying to the last message anyway
-        // todo - revisit this (and potentially remove) when we support encrypted replies mid-convo
-        for (const replyBtn of convoReplyBtnsToReplaceArr) { // all other non-last reply buttons to be hidden
-          $(replyBtn).addClass('replaced').text(''); // hide all except last
-        }
+      // only replace with FlowCrypt reply button if does not have any buttons replaced yet, and only replace the last one
+      for (const elem of convoReplyBtnsToReplaceArr) {
+        $(elem).addClass('inserted');
+        const element = $(this.factory.btnReply()).insertBefore($(elem).children().last());  // xss-safe-factory
+        const messageContainer = $(elem.closest('.h7') as HTMLElement);
```
tomholub: I imagine this could be a memory leak, because `messageContainer` will be used at unknown time in the future inside `element.click`, and the content scripts are similar to SPAs (will never have a chance to de-allocate memory in this case).

We actually already have a few memory leaks in the content script that customers were complaining about.

Is it possible to compute/ find `messageContainer` inside the handler, based on which element was clicked?

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -437,13 +434,29 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     return $(anyInnerElement).closest('div.if, td.Bu').first();
   }
 
-  private replaceStandardReplyBox = async (editable: boolean = false, force: boolean = false) => {
+  private insertEncryptedReplyBox = async (messageContainer: JQuery<HTMLElement>) => {
```
tomholub: It seems this method does not have to be async

extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -130,28 +132,23 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
+        const messageContainer = $(elem.closest('.h7') as HTMLElement);
+        element.click(Ui.event.prevent('double', Catch.try(async () => {
+          if (messageContainer.is(':last-child')) {
+            if (isEncrypted) {
+              await this.setReplyBoxEditable();
+            } else {
+              await this.replaceStandardReplyBox(undefined, true, true);
+            }
+          } else {
+            await this.insertEncryptedReplyBox(messageContainer);
```
tomholub: Can probably remove the await

extension/chrome/elements/compose.ts
michael-volynets: Let me check, if we don't need it anymore I will remove it

michael-volynets: Look what I found here:
```TS
  reinsertReplyBox = (subject: string, threadId: string) => {
    const params: FactoryReplyParams = { subject, sendAs: this.sendAs, replyMsgId: threadId };
    $('.reply_message_iframe_container:visible').last().append(this.factory.embeddedReply(params, false, true)); // xss-safe-value
  }
```
Is it correct to pass threadId into replyMsgId; I'm not sure

tomholub: Sort of, the first message in the thread has the same `msgId` as the `threadId`. But we should not rely on that, and properly fix it.

extension/js/common/composer.ts
michael-volynets: Everything is good, the message was sent in the same thread

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: yes, it is possible, will do

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: The result:
image

It replaces everything correctly but the standard reply box got replaced with `Draft Skipped`
I think would be better to leave it as a standard reply box or remove it. What do you think?

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: Yes, thanks

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: Yes, thanks

tomholub:
Thanks @michael-volynets , is it ready for review?

michael-volynets:
There is an unresolved review item, please take a look on that.

extension/js/content_scripts/webmail/gmail_element_replacer.ts
michael-volynets: @tomholub

tomholub: @michael-volynets I think it would be indeed appropriate to have it render a standard reply box, whatever Gmail would/should do, in that situation, or whatever you'd expect as user.

michael-volynets:
@tomholub, I'm thinking to replace the reply box (after the last message) depend on the last message type.
If the last message was encrypted then replace it otherwise leave a standard reply box.
But it may be a little complicated, I'm not sure if we need this. What do you think?

tomholub:
> If the last message was encrypted then replace it otherwise leave a standard reply box.

Yes, that's how it should ideally work. Right now, it will replace the last message box if ANY previous message was encrypted, which is annoying, and I've had several users uninstall for that reason.

tomholub:
We could cover that in another PR if you think existing behavior did not get worse then before. Or you can add it to this PR. Up to you.

michael-volynets:
Okay, I will add it in this PR

michael-volynets:
Added, +1h

extension/chrome/elements/reply_pubkey_mismatch.ts
```diff
@@ -120,9 +120,9 @@ Catch.try(async () => {
   const determineReplyHeaders = async () => {
     const thread = await Google.gmail.threadGet(acctEmail, threadId, 'full');
     if (thread.messages && thread.messages.length > 0) {
-      const threadMsgIdLast = Google.gmail.findHeader(thread.messages[thread.messages.length - 1], 'Message-ID') || '';
+      const replyMsgIdLast = Google.gmail.findHeader(thread.messages[thread.messages.length - 1], 'Message-ID') || '';
```
tomholub: Here needs improving, I can fix it.

Since we don't always just reply to last message, but to a specific message, we should take care to construct the headers using the message we are replying to, not the last in thread.

tomholub: Actually, this is just reply_pubkey_mismatch which is not a problem. You will probably refactor this away in #2128

extension/js/common/composer.ts
```diff
@@ -414,22 +416,19 @@ export class Composer {
...
+        (async () => { // not awaited because can take a long time & blocks rendering
+          const footer = this.getFooter();
+          await this.composerQuote.addTripleDotQuoteExpandBtn(this.urlParams.replyMsgId, method, footer);
+          if (this.composerQuote.messageToReplyOrForward) {
+            this.composerSendBtn.additionalMsgHeaders['In-Reply-To'] = this.composerQuote.messageToReplyOrForward.headers['in-reply-to'];
```
tomholub: For a reply to be properly constructed, mail servers need to know which email address you are replying to. That means `In-Reply-To` should contain the id of the message you are replying to.

extension/js/common/composer.ts
```diff
@@ -414,22 +416,19 @@ export class Composer {
...
+            this.composerSendBtn.additionalMsgHeaders.References = this.composerQuote.messageToReplyOrForward.headers.references;
```
tomholub: Here also a mistake, you can compare with previous implementation which does 
```js
'References': threadMsgRefsLast + ' ' + replyMsgIdLast
```
When unsure, you can check the spec: https://tools.ietf.org/html/rfc4021#section-2.1.10

> The message identifier(s) of other message(s) to which the current
      message may be related.  In RFC 2822, the definition was changed
      to say that this header field contains a list of all Message-IDs
      of messages in the preceding reply chain.  Defined as standard by
      RFC 822.

PR #2168, opened 2019-11-06
Issue 2165 take icon_show_prev_msg's height into account in setInputTextHeightManuallyIfNeeded()

### LINKED ISSUE #2165: compose window cropped bottom on firefox
image

Only happens when the email has more than one alias. Maybe forgot to resize window after adding the alias selector.

Not tested on chrome

limonte:
Closes #2165 

This small node was forgotten in `setInputTextHeightManuallyIfNeeded()`:

image


limonte:
@tomholub since 7.0.7, `icon_show_prev_msg` will always be there for users with footer and the New Secure window is always overflowing, consider releasing 7.0.1 with this fix as the initial UX for Firefox users isn't prefect now.

tomholub:
Agree. Thanks!

PR #2173, opened 2019-11-10
Improve CSS formatting, add stylelint

limonte:
In this PR I dit nothing but changed the formatting to CSS files (`stylelint --fix`). Aside from formatting there are some other issues that were hard-to-spot before:

image

Stylelint is the default tool for linting styles (ESLint for CSS), it's well-maintained and there are plenty of useful plugins e.g. https://github.com/sweetalert2/stylelint-config/blob/master/index.js

@tomholub let me know what do you think about this change in coding style. I personally prefer it because it's much easier to read, spot issues and Git diffs are way more useful when each CSS property is on its own line.


tomholub:
Cool with me. I was actually going to eventually use scss instead of css, which would sure help with the formatting. For the time being, I'm good with reformatting it.

tomholub:
By the way, one of the tests didn't run because it insists on `issue-0000-some-name` branch naming convention.

Issue #2148, opened 2019-10-25
choose encrypt, sign + send as default


tomholub:
Not yet actionable, should do sometime end of November

PR #2178, opened 2019-11-13
Issue #2148, encryptAndSend by default, prettifier removed unnecesarr…

michael-volynets:
…y tab in composer-send-btn
Issue #2148 
Time Spent: 15m


tomholub:
@michael-volynets please have a look at the conflics - thanks!

PR #2182, opened 2019-11-15
Issue #1074 test file backup during setup

### LINKED ISSUE #1074: add automated test for create key manually -> download
- [x] fix it (got fixed on recent versions)
- [ ] add test (on chrome)
- [ ] also test settings -> backup -> as file

michael-volynets:
close #1074 
Time Spent: 1h

There was a problem. When I check `backup as file` checkbox and submit the form puppeteer throws an error with a message that the page is closed but when I check `if (settingsPage.page.isClosed())` it returns me `false` so I decided to leave the checkbox unchecked and just check if the checkbox is present.


tomholub:
Here I still have to verify what exactly it's testing, that's why I didn't merge it yet

test/source/tests/page_recipe.ts
```diff
@@ -69,7 +69,8 @@ export class SetupPageRecipe extends PageRecipe {
     } else if (backup === 'email') {
       throw new Error('tests.setup_manual_create options.backup=email not implemented');
     } else if (backup === 'file') {
-      throw new Error('tests.setup_manual_create options.backup=file not implemented');
+      await settingsPage.waitAny('@action-backup-download'); // Only check if there is a checkbox.
+      return;
```
tomholub: This doesn't actually proceed with the backup - the test is not useful.

We should test scenarios that user would do - set up the extension (completely) using this backup method.

tomholub:
> There was a problem. When I check backup as file checkbox and submit the form puppeteer throws an error with a message that the page is closed but when I check if (settingsPage.page.isClosed()) it returns me false so I decided to leave the checkbox unchecked and just check if the checkbox is present.

I now see you mentioned this. I'll have a look.

Issue #2184, opened 2019-11-19
refactor composer file


tomholub:
in prep for  #1620

PR #2185, opened 2019-11-19
refactor composer + popover

tomholub:
#2184


tomholub:
I wanted to add another set of options and it's starting to get very tall.

image

I will refactor the popover options as toggles (encrypt, sign) which reduces the amount of items to a half.

tomholub:
Also likely fixed rarely reported issues with popover positioning

image

PR #2186, opened 2019-11-19
refactor formatters + composer + experimental pgp/mime

### LINKED ISSUE #921: PGP/Mime - detached signatures
I recently got a complaint that one receiver of my signed message (signed only not encrypted) was confused by all the Mumbo Jumbo.

So I guess FlowCrypt uses Inline PGP as default? Can I change this somewhere to do PGP/MIME instead?

### LINKED ISSUE #1522: option to send outgoing messages as PGP/mime html
#1621
- This probably means we have to use PGP/MIME
- This also means we have to have a way for people to switch back to plain text. The UI should be similar to Gmail switch to plain text
- When someone switches to plain text, it will stay that way on their account, until switched back
- for now no actual html formatting. That's tracked as a separate issue

### LINKED ISSUE #1620: option to send outgoing message as pgp/mime
#1621

The settings of using rich text or poor text should be sticky.

related #921

### LINKED ISSUE #2184: refactor composer file

tomholub:
close #2184 
close #1620 (experimental), without signatures
close #921
close #1522
approx 14 hours


tomholub:
Still todo:
 - [ ] add all err throwing and handling from composer-send-btn into composer-errs
 - [ ] create ComposerRender
 - [ ] rename composer modules

tomholub:
- [ ] make `NewMsgData` include `SendableMsgBody` ?

Issue #2189, opened 2019-11-21
Signature is adding itself to message each time draft is being closed and opened again


limonte:
### Steps to reproduce:

1. Click the "Secure Compose" button
2. Select a recipient
3. Wait for the "Saved" notification in the footer
4. Close the "New Secure Message" window
5. Go to Drafts, open the most recent draft message

### Actual result

The signature is doubled, the message contains it and the footer (three dots) contains it.

### Expected result

The message is empty, the signature is hidden in the footer (three dots)

---

Latest Firefox on Linux

tomholub:
I believe the intention was that when loading message from a draft, footer would not be appended again. It's assumed it's already present in the draft.

PR #2221, opened 2019-11-27
[4h] Issue #2189, fixed bug when footer doubled

michael-volynets:
Issue #2189,
Time Spent: 3h

I also fixed an issue when `TripleDots` button was shown on the reply box with a draft and the button was also rendered (but actually it shouldn't have been rendered), and when you click on it nothing happens.

I haven't found the issue related to this `Tripedots` button bug on Github (it might not be reported).

### DIFF
--- a/extension/js/common/composer/composer-quote.ts
+++ b/extension/js/common/composer/composer-quote.ts
@@ -47,7 +47,7 @@ export class ComposerQuote extends ComposerComponent {
       this.composer.S.cached('icon_show_prev_msg').removeClass('progress');
     }
     if (!this.messageToReplyOrForward && msgId) {
-      this.composer.S.cached('icon_show_prev_msg').click(Ui.event.handle(async el => {
+      this.composer.S.cached('icon_show_prev_msg').click(Ui.event.handle(async () => {
         this.composer.S.cached('icon_show_prev_msg').unbind('click');
         await this.addTripleDotQuoteExpandBtn(msgId, method);
         if (this.messageToReplyOrForward) {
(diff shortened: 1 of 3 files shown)


michael-volynets:
Yes, I also was thinking like you and actually my first implementation was kind of you wrote in your comment.
But I checked issue #2189 and @limonte described that the expected result is:
```
The message is empty, the signature is hidden in the footer (three dots)
```
And I also think that it's better from the UI side so I implemented it.

But both ways are acceptable so if you want I can remove it and it will make code simpler :)

tomholub:
Sorry, I'm a bit confused now. Did the simpler fix not achieve that? I thought that it was already saving draft including the footer. That means simply not adding another footer when it was loaded for the draft should achieve this result:

> The message is empty, the signature is hidden in the footer (three dots)

Because it was already saved as part of the draft. If I'm wrong, please let me know.

michael-volynets:
Yes, there is a little oversight of what you are saying.
When we save draft we save the footer as well.
So when we do steps that described in the issue the result will be a draft with a recipient and with the message body containing your footer. I mean when we load draft we don't append the footer. The code you listed:
```TS
if(footer && !this.composer.draft.msgWasLoadedFromDraft) {
      this.footerHTML = this.createFooterHTML(footer);
      safePreviousMsg  += this.footerHTML;
}
```
Adds your footer to the expanding part of the message (not to the message body).
So we anyway have to remove footer from the message body and add it to the expanding part.

Your code says that we don't need to show `threedots` button because we already got the footer in the draft message body.

michael-volynets:
+20m

Issue #2219, opened 2019-11-25
end-to-end test password encrypted attachments


tomholub:
We should end-to-end test password encrypted attachments, all the way to opening it on the website

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/2218#issuecomment-558380628_

PR #2222, opened 2019-11-27
[13h?] Issue #2219, e2e test of password encrypted attachments

michael-volynets:
…word message

Issue #2219,
Time Spent: 5h

I stuck with this issue a little and I have some problems such as:
1) Do we have an account with a paid subscription? Because I need to send an encrypted by password message that won't expire. This one I'm using in this test will expire in a month.
2) I think the variable `approveAppLogin` might not be needed. I wanted to use `action: "close"` to skip approving login into the app (Google Login), I would have had to write some code like this:
```TS
if (action === 'approve') {
 await oauthPage.waitAll('#submit_approve_access'); 
}
```
but I assume that I could break some other tests.
3) I didn't find a way to download files using Puppeteer. There is an opened issue in their GitHub Repository (https://github.com/puppeteer/puppeteer/issues/299) about this. 
I tried to find another way like when I added this line of code at the end of the test to click on the button to download a file:
```TS
...
      await flowcryptPage.waitAndClick('.action_decrypt');
      expect(await flowcryptPage.read('.pgp_block')).to.include('Test Message');
      expect(await flowcryptPage.read('.attachment')).to.include('test-img.jpg.pgp');
      await flowcryptPage.waitAndClick('.attachment'); // <-- this one
```
The page closes or crashes (you could test it by yourself), so I could test it like: if page is closed after click then everything is good but when I call `flowcryptPage.page.isClosed()` it somehow returns me `false`. And I don't really know how to test if attachments are downloaded correctly.

I will also send you an updated secrets file.

@tomholub Could you please help me? Thanks!

### DIFF
--- a/extension/js/common/composer/formatters/encrypted-mail-msg-formatter.ts
+++ b/extension/js/common/composer/formatters/encrypted-mail-msg-formatter.ts
@@ -187,7 +187,6 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter implements Mail
     const storage = await Store.getAcct(this.composer.urlParams.acctEmail, ['outgoing_language']);
     const lang = storage.outgoing_language || 'EN';
     const msgUrl = `${this.FC_WEB_URL}/${short}`;
-    $('#new_message').attr('data-test-flowcrypt-msg-url', msgUrl);
     const a = `<a href="${Xss.escape(msgUrl)}" style="padding: 2px 6px; background: #2199e8; color: #fff; display: inline-block; text-decoration: none;">
                     ${Lang.compose.openMsg[lang]}
                    </a>`;
--- a/extension/js/common/composer/formatters/encrypted-mail-msg-formatter.ts
+++ b/extension/js/common/composer/formatters/encrypted-mail-msg-formatter.ts
@@ -187,6 +187,7 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter implements Mail
     const storage = await Store.getAcct(this.composer.urlParams.acctEmail, ['outgoing_language']);
     const lang = storage.outgoing_language || 'EN';
     const msgUrl = `${this.FC_WEB_URL}/${short}`;
+    $('#new_message').attr('data-test-flowcrypt-msg-url', msgUrl);
     const a = `<a href="${Xss.escape(msgUrl)}" style="padding: 2px 6px; background: #2199e8; color: #fff; display: inline-block; text-decoration: none;">
                     ${Lang.compose.openMsg[lang]}
                    </a>`;
--- a/test/source/mock/google-api-mock.ts
+++ b/test/source/mock/google-api-mock.ts
@@ -5,7 +5,7 @@
 import { Api, HttpClientErr, Status } from './api';
 import { IncomingMessage } from 'http';
 import { OauthMock } from './oauth';
-import { Data, GmailMsg } from './data';
+import { Data } from './data';
 import { ParsedMail } from "mailparser";
 import * as http from 'http';
 import Parse, { ParseMsgResult } from '../util/parse';
@@ -69,7 +69,7 @@ export const startGoogleApiMock = async (logger: (line: string) => void) => {
               { gd$email: [{ address: 'contact.test@flowcrypt.com', primary: "true" }] }
             ]
           }
-        }
+        };
       }
       throw new HttpClientErr(`Method not implemented for ${req.url}: ${req.method}`);
     },
@@ -157,7 +157,9 @@ export const startGoogleApiMock = async (logger: (line: string) => void) => {
               throw e;
             }
           }
-          return { id: 'mockfakesend', labelIds: ['SENT'], threadId: parseResult.threadId };
+          // added msg survives this particular `data` object
+          const id = new Data(acct).storeSentMessage(parseResult.mimeMsg, parseResult.threadId);
+          return { id, labelIds: ['SENT'], threadId: parseResult.threadId };
         }
       }
       throw new HttpClientErr(`Method not implemented for ${req.url}: ${req.method}`);
@@ -271,4 +273,3 @@ const validateMimeMsg = async (acct: string, mimeMsg: ParsedMail, threadId?: str
     throw new HttpClientErr('You can\'t send a message from unexisting email address(es)');
   }
 };
-
--- a/test/source/mock/google-api-mock.ts
+++ b/test/source/mock/google-api-mock.ts
@@ -157,9 +157,7 @@ export const startGoogleApiMock = async (logger: (line: string) => void) => {
               throw e;
             }
           }
-          // added msg survives this particular `data` object
-          const id = new Data(acct).storeSentMessage(parseResult.mimeMsg, parseResult.threadId);
-          return { id, labelIds: ['SENT'], threadId: parseResult.threadId };
+          return { id: 'fakesendid', labelIds: ['SENT'], threadId: parseResult.threadId };
         }
       }
       throw new HttpClientErr(`Method not implemented for ${req.url}: ${req.method}`);
--- a/test/source/mock/google-api-mock.ts
+++ b/test/source/mock/google-api-mock.ts
@@ -5,7 +5,7 @@
 import { Api, HttpClientErr, Status } from './api';
 import { IncomingMessage } from 'http';
 import { OauthMock } from './oauth';
-import { Data } from './data';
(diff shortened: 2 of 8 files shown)


tomholub:
> 1. Do we have an account with a paid subscription? Because I need to send an encrypted by password message that won't expire. This one I'm using in this test will expire in a month.

I don't think it's needed, you can send a message, verify it that it works well, then forget it. You don't need to keep testing the same message from before, and you shouldn't need subscription to do it.

The goal is to test the end-to-end experience from sending all the way to downloading in one go, to see that it works.

> 2. I think the variable approveAppLogin might not be needed. I wanted to use action: "close" to skip approving login into the app (Google Login), I would have had to write some code like this:

I don't think it's necessary to log in at all.

> 3) I didn't find a way to download files using Puppeteer. There is an opened issue in their GitHub Repository (puppeteer/puppeteer#299) about this.

If Puppeteer cannot do this, we can instead find a way to add downloadable URL as a meta-tag onto the donwload button on the web. Then you parse out the URL, and you just do `Requests.get(url)` in Node. Then you try to decrypt that to verify it was encrypted properly, like we do with the other mock messages.

You should be able to do all of this when using mock Gmail api, I hope.

test/source/tests/tests/gmail.ts
```diff
@@ -96,6 +96,19 @@ export const defineGmailTests = (testVariant: TestVariant, testWithNewBrowser: T
       expect(content).to.contain('STONE NEED REMAIN SLIDE DEPOSIT BRICK');
     }));
 
+    ava.default('mail.google.com[global:no.pgp] - process to flowcrypt website and decypt the message', testWithNewBrowser(async (t, browser) => {
+      const gmailPage = await browser.newPage(t, 'https://accounts.google.com/AccountChooser?service=mail&continue=https://mail.google.com/mail/');
```
tomholub: You don't actually need to log in as the recipient. Use a mock. In the sent mock email, there will be URL link to the website. Open that link with Puppeteer.

michael-volynets: Got it, thanks!

michael-volynets:
```
If Puppeteer cannot do this, we can instead find a way to add downloadable URL as a meta-tag onto the donwload button on the web. Then you parse out the URL, and you just do Requests.get(url) in Node. Then you try to decrypt that to verify it was encrypted properly, like we do with the other mock messages.
```
Yes, that would be great to add the meta-tag.
I also changed the test so we use only our mock now. This should be better. You can go through it again.

michael-volynets:
+1h

test/source/tests/tests/compose.ts
```diff
@@ -461,6 +461,27 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
...
+      await composePage.waitForSelTestState('closed', 60);
+      const [elemWithFlowcryptURL] = await composePage.page.$x('.//body/@data-test-flowcrypt-msg-url');
+      const url = await PageRecipe.getElementPropertyJson(elemWithFlowcryptURL!, 'value');
+      await composePage.close();
+      const flowcryptPage = await browser.newPage(t, url);
+      await flowcryptPage.waitAndType('.decrypt_answer', pwd);
+      await flowcryptPage.waitAndClick('.action_decrypt');
+      expect(await flowcryptPage.read('.pgp_block')).to.include(subject);
+      expect(await flowcryptPage.read('.pgp_block')).to.include('The best footer ever!'); // test if footer is present
+      expect(await flowcryptPage.read('.attachment')).to.include('small.png.pgp');
```
tomholub: Yes, this is closer.

Instead of parsing the URL using `const [elemWithFlowcryptURL] = await composePage.page.$x('.//body/@data-test-flowcrypt-msg-url');`, best would be if the you could do something like `const sentMsgFromMock = mockApi.findSentMessageBySubject(subject)`.

Then you parse out the URL from the mock message. I'm not exactly sure if this is possible, but it would be better.

Then finally, there should be a way to download the attachment (you need it to provide a data-test-donwload-url="https://..." in the file button on the web). I'll give you access to the web repo to add that.

tomholub: You already have access to https://github.com/FlowCrypt/flowcrypt-web/

There is no TypeScript, no build system. Just edit the js/html.

michael-volynets: Okay, It will be better, 
I will implement it. I think I have an idea of how to do that

test/source/mock/google-api-mock.ts
```diff
@@ -5,7 +5,7 @@
 import { Api, HttpClientErr, Status } from './api';
 import { IncomingMessage } from 'http';
 import { OauthMock } from './oauth';
-import { Data } from './data';
+import { Data, GmailMsg } from './data';
```
tomholub: Seems to be unnecessary

extension/js/common/composer/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -187,6 +187,7 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter implements Mail
     const storage = await Store.getAcct(this.composer.urlParams.acctEmail, ['outgoing_language']);
     const lang = storage.outgoing_language || 'EN';
     const msgUrl = `${this.FC_WEB_URL}/${short}`;
+    $('#new_message').attr('data-test-flowcrypt-msg-url', msgUrl);
```
tomholub: Should be removed, can now be parsed from sent mail

test/source/mock/strategies/send-message-strategy.ts
```diff
@@ -4,6 +4,26 @@ import { HttpClientErr } from '../api.js';
 import { PgpMsg, Pgp } from "../../core/pgp.js";
 import { Buf } from '../../core/buf.js';
 import { Config } from '../../util/index.js';
+import { GmailMsg, Data } from '../data.js';
+
+class AddMessageToStorageTestStrategy implements ITestMsgStrategy {
```
tomholub: It's not so clear that this should be a test strategy, it should rather just be a method on `Data`? What is it testing?

tomholub: Now I see how it's used, this should be renamed to `PwdEncryptedMessageStrategy` then.

test/source/mock/strategies/send-message-strategy.ts
```diff
@@ -4,6 +4,26 @@ import { HttpClientErr } from '../api.js';
...
+    new Data(mimeMsg.from.value[0].address).storeMessage(gmailMsg);
+  }
+
+  createGmailMsg = (parsedMail: ParsedMail): GmailMsg => {
+    return {
+      id: '',
+      historyId: '',
+      payload: {
+        headers: [{ name: 'Subject', value: parsedMail.subject }],
+        body: { data: parsedMail.text, attachmentId: '', size: 0 }
```
tomholub: `size` should be `parsedMail.text.length`

test/source/util/types.ts
```diff
@@ -0,0 +1,4 @@
+
+export type Await<T> = T extends {
+    then(onfulfilled?: (value: infer U) => unknown): unknown;
```
tomholub: Where is this type used?

test/source/tests/tests/compose.ts
```diff
@@ -461,6 +464,32 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
       await ComposePageRecipe.sendAndClose(composePage);
     }));
 
+    ava.default.only('compose[global:compatibility] - standalone - send pwd encrypted msg & check on flowcrypt site', testWithSemaphoredGlobalBrowser('compatibility', async (t, browser) => {
+      const res = await request.get({ url: 'https://s3.eu-central-1.amazonaws.com/flowcrypt/attachment/ac96108791bd9f6c59c33b963b7d947ce43a2595.pgp' });
```
tomholub: I'll have a look at this

tomholub:
The reason it didn't work:
1) need to pass `encoding: null` into `request.get` so that it doesn't try to utf-decode the binary data (that's probably why you were getting "not an openpgp message")
2) need to hash the pwd with `Pgp.hash.challengeAnswer` before decrypting with it (Sorry I forgot to mention this - you'd notice it both in the composer code as well as in pgp_block.ts).

michael-volynets:
Thanks, Tom.
Now I know it.
Actually, I spent about 19 hrs on it.
I will change it.

PR #2232, opened 2019-12-02
Test toggle minimized state by clicking header

### LINKED ISSUE #2047: add test for clicking secure message header
should minimize / maximize based on state

limonte:
Closes #2047

### DIFF
--- a/test/source/tests/tests/compose.ts
+++ b/test/source/tests/tests/compose.ts
@@ -25,6 +25,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
       await inboxPage.waitAll('iframe');
       const composeFrame = await inboxPage.getFrame(['compose.htm']);
       const initialComposeFrameHeight = await inboxPage.getOuterHeight('iframe');
+      composeFrame.waitAll('#section_header');
       const composeFrameHeaderHeight = await composeFrame.getOuterHeight('#section_header');
       // mimimize compose frame
       await composeFrame.waitAndClick('@header-title');
--- a/test/source/tests/tests/compose.ts
+++ b/test/source/tests/tests/compose.ts
@@ -19,7 +19,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
 
   if (testVariant !== 'CONSUMER-LIVE-GMAIL') {
 
-    ava.default.only('compose[global:compatibility] - toggle minimized state by clicking compose window header', testWithNewBrowser(async (t, browser) => {
+    ava.default('compose[global:compatibility] - toggle minimized state by clicking compose window header', testWithNewBrowser(async (t, browser) => {
       await BrowserRecipe.setUpCommonAcct(t, browser, 'compatibility');
       const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=flowcrypt.compatibility@gmail.com`));
       await inboxPage.waitAndClick('@secure-compose');
--- a/test/source/tests/tests/compose.ts
+++ b/test/source/tests/tests/compose.ts
@@ -25,7 +25,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
       await inboxPage.waitAll('iframe');
       const composeFrame = await inboxPage.getFrame(['compose.htm']);
       const initialComposeFrameHeight = await inboxPage.getOuterHeight('iframe');
-      composeFrame.waitAll('#section_header');
+      await composeFrame.waitAll('#section_header');
       const composeFrameHeaderHeight = await composeFrame.getOuterHeight('#section_header');
       // mimimize compose frame
       await composeFrame.waitAndClick('@header-title');
--- a/test/source/tests/tests/compose.ts
+++ b/test/source/tests/tests/compose.ts
@@ -19,8 +19,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
 
   if (testVariant !== 'CONSUMER-LIVE-GMAIL') {
 
-    ava.default('compose[global:compatibility] - toggle minimized state by clicking compose window header', testWithNewBrowser(async (t, browser) => {
-      await BrowserRecipe.setUpCommonAcct(t, browser, 'compatibility');
+    ava.default('compose[global:compatibility] - toggle minimized state by clicking compose window header', testWithSemaphoredGlobalBrowser('compatibility', async (t, browser) => {
       const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=flowcrypt.compatibility@gmail.com`));
       await inboxPage.waitAndClick('@secure-compose');
       await inboxPage.waitAll('iframe');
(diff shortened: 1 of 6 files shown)


limonte:
@tomholub I can't see the output with test results for SemaphoreCI, when I click "Details"

image

this screen is shown

image

I can't see what went wrong in this build

tomholub:
I have added you to the project - can you please try again, making sure you are logged in?

image

You should, after clicking on the failed test "chrome-consumer-live-gmail" and then clicking the line that shows "npm run-script ..." see the following unroll:

image

It was expecting a `compose.htm` frame but it was not there.

A more thorough fail debug output has been sent to my email.

I'll have it send you emails too, but it will send you tons of spam when everybody else runs their CI :( I recommend to create a filter to have them all go straight to trash in Gmail. Then when you need to look one up, search the id in Gmail, in this case `cdlqmc2e5nc` and it will show:

image

Clicking it will help you find it.

Open the attached file in Firefox and it should give you enough info to find out what went wrong in detail.

This means it retried 3 times, always failed:

image

Open one of the attempts:

image

Voila, the compose window is in fact there.

Almost certainly, you were testing its presence before it had time to show up.
Before you try to grab the frame, try to add `await gmailPage.waitForAll('iframe')` or something similar (take a look at other code that does something similar).

Anyway, don't test this functionality on Live Gmail. Live Gmail is expensive, slow and flakey.

Instead test this in FlowCrypt encrypted inbox, which you can run under mock tests:

image

Where you can test minimizing, maximizing the compose window just fine, with less hassle.

Do this by moving the code maybe to `test/source/tests/tests/compose.ts` where you'll notice `if (testVariant !== 'CONSUMER-LIVE-GMAIL') {` meaning these are NOT live gmail tests (a good thing).

limonte:
Thank you for the thorough comment @tomholub, that was super helpful!

The PR is ready for review now.

test/source/tests/tests/compose.ts
```diff
@@ -16,6 +16,22 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
 
   if (testVariant !== 'CONSUMER-LIVE-GMAIL') {
 
+    ava.default.only('compose[global:compatibility] - toggle minimized state by clicking compose window header', testWithNewBrowser(async (t, browser) => {
```
tomholub: Common oversight - please don't leave `.only` checked in - would make our CI skip all tests except yours

limonte: :man_facepalming: oops, sorry!

test/source/tests/tests/compose.ts
```diff
@@ -16,6 +16,22 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
 
   if (testVariant !== 'CONSUMER-LIVE-GMAIL') {
 
+    ava.default('compose[global:compatibility] - toggle minimized state by clicking compose window header', testWithNewBrowser(async (t, browser) => {
```
tomholub: Here `testWithNewBrowser` is super slow because it has to spin up a new browser, log into an account, set up the extension for each test.

When you are not changing any account settings, better use `testWithSemaphoredGlobalBrowser`. This will run the test on a recycled browser that already has an account set up, and then release the browser to be used for other similar tests - much faster.

You can choose which account you need for the test, eg `compatibility`.

test/source/tests/tests/compose.ts
```diff
@@ -16,6 +16,22 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
...
+      await BrowserRecipe.setUpCommonAcct(t, browser, 'compatibility');
```
tomholub: which means you will skip this step, extension will be already set up

tomholub:
Tests failing:

image

By the way you don't have to be too precise with the heights. I'd be quite happy with testing for <100 pixels for one state and >200 pixels for another, as an example.

limonte:
> Tests failing:
> 
> image
> 
> By the way you don't have to be too precise with the heights. I'd be quite happy with testing for <100 pixels for one state and >200 pixels for another, as an example.

I ran the test a couple of times and wasn't able to reproduce this case.

Currently another test is failing:
```
compose[global:flowcrypt.test.key.new.manual@gmail.com] - standalone - own key expired
```

tomholub:
> Tests failing:
> 
> image
> 
> By the way you don't have to be too precise with the heights. I'd be quite happy with testing for <100 pixels for one state and >200 pixels for another, as an example.

One reason this could happen without `.only` but not with `.only` is because when we run all tests, a bunch of them run in parallel and the machine gets really busy. Certain browser actions may take longer to complete (or not be as immediate). Maybe you are not awaiting an element to show up at some point, and when we run tests in parallel, the element does not have time to show up before you pull it's height (or some such scenario).

tomholub:
> Currently another test is failing:
> compose[global:flowcrypt.test.key.new.manual@gmail.com] - standalone - own key expired

Pulled from master, that should fix this particular flakey test.

tomholub:
> I ran the test a couple of times and wasn't able to reproduce this case.

One reason this could happen without .only but not with .only is because when we run all tests, a bunch of them run in parallel and the machine gets really busy. Certain browser actions may take longer to complete (or not be as immediate). Maybe you are not awaiting an element to show up at some point, and when we run tests in parallel, the element does not have time to show up before you pull it's height (or some such scenario).

limonte:
> > Currently another test is failing:
> > compose[global:[flowcrypt.test.key.new.manual@gmail.com](mailto:flowcrypt.test.key.new.manual@gmail.com)] - standalone - own key expired
> 
> Pulled from master, that should fix this particular flakey test.

rebased, thanks :+1:

> Maybe you are not awaiting an element to show up at some point, 

Added `waitAll('#section_header')` before measuring it, let's see.

BTW: can I somehow re-run semaphore tests?

tomholub:
> BTW: can I somehow re-run semaphore tests? e.g. I want to run them to ensure there's no flakiness.

Yes, if they run on Semaphore 2.0 (mock tests) then using these buttons, click "rerun":

https://flowcrypt.semaphoreci.com/workflows/afce8c01-dd7e-47f2-8d66-8378cdd2974d

image

Or if it runs Semaphore 1.0 (live gmail tests) then you'll click the branch name, then "rebuild last revision":

https://semaphoreci.com/flowcrypt/flowcrypt-browser/branches/issue-2047-test-clicking-secure-message-header/builds/9

image

image

But generally speaking, if it builds and I give it a LGTM, we merge it, and if it turns out to be flakey, we address it as the flakiness surfaces over time. So you don't need to manually re-run it a few times just to test if it's flakey or not.

test/source/tests/tests/compose.ts
```diff
@@ -25,6 +25,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithNewBrowser:
       await inboxPage.waitAll('iframe');
       const composeFrame = await inboxPage.getFrame(['compose.htm']);
       const initialComposeFrameHeight = await inboxPage.getOuterHeight('iframe');
+      composeFrame.waitAll('#section_header');
```
tomholub: you have to `await` this or it will have no effect

Basically everything that has to do with the browser has to be awaited

tomholub:
(if I'm lazy and want to re-run all tests, I just add or remove a newline in readme..)

test/source/tests/tests/compose.ts
limonte: oops, stupid mistake :man_facepalming: 

In ESLint there's the https://eslint.org/docs/rules/require-await rule which can be helpful for such cases.

Offtopic: did you think about using ESLint for `flowcrypt-browser`. `tslint` has been deprecated in favor of ESLint: https://medium.com/palantir/tslint-in-2019-1a144c2317a9

ESLint is truly amazing tool for keeping TS/JS in good and consistent shape and preventing silly human mistakes like this one with missing `await`.

PR #2241, opened 2019-12-04
ESLint base config

### LINKED ISSUE #2240: enable eslint - for now only rule require-await
In ESLint there's the https://eslint.org/docs/rules/require-await rule which can be helpful for such cases.

Offtopic: did you think about using ESLint for `flowcrypt-browser`? `tslint` has been deprecated in favor of ESLint + `@typescript-eslint`: https://medium.com/palantir/tslint-in-2019-1a144c2317a9

ESLint is truly amazing tool for keeping TS/JS in good and consistent shape and preventing silly human mistakes like this one with missing `await`.

Take a look on the config we're using for SweetAlert2: https://github.com/sweetalert2/eslint-config/blob/master/index.js

_Originally posted by @limonte in https://github.com/FlowCrypt/flowcrypt-browser/pull/2232_

limonte:
Closes #2240

The basic configuration is in this PR, uncomment `plugin:@typescript-eslint/recommended` and comment `plugin:@typescript-eslint/base` to see what kinds of issues can be caught. To be precise, these are more coding style opinionated fixes rather than issues.

Unfortunately, it's not possible to enforce `await` for statements returning promises, here's the great explanation why: https://github.com/typescript-eslint/typescript-eslint/issues/1203#issuecomment-553725876

Also, the `eslint-plugin-no-only-tests` plugin isn't working for us currently, reported an issue: https://github.com/levibuzolic/eslint-plugin-no-only-tests/issues/14


limonte:
I also added the `semi` rule because as you can see it's the issue for my code as in my project I'm not using semicolons and I'm forgetting them all the time :)

I also added the `eslint-plugin-no-only-tests` plugin to catch `.only` earlier and without humans.

limonte:
I added the testing commit with `.only(` to verify that CI should fail.

limonte:
All is well

image

image

extension/js/common/core/mime.ts
```diff
@@ -177,7 +177,7 @@ export class Mime {
   }
 
   public static decode = (mimeMsg: Uint8Array): Promise<MimeContent> => {
-    return new Promise(async resolve => {
+    return new Promise(resolve => {
```
tomholub: good catch, that must have been swallowing errors silently!

Issue #2231, opened 2019-12-02
use a view-framework for simpler settings modules


tomholub:
in `settings/modules/`:
 - add_key.ts
 - auth_denied.ts
 - compatibility.ts
 - contact_page.ts
 - debug_api.ts
 - my_key_update.ts
 - my_key_user_ids.ts
 - my_key.ts
 - security.ts
 - test_passphrase.ts

I've skipped the more complex ones which I'll have a look at, and maybe refactor further.

michael-volynets:
I will create a PR with one refactored file.
If I've done everything correctly I will continue working on others. Okay?
Or I can even split every file to separate PR so it will be easier for review.

PR #2242, opened 2019-12-04
Issue #2231, refactored add_key.ts file

michael-volynets:
Issue #2231,
Time Spent: 40m

I also found a bug:
When we call `BrowserMsg.send.reload(this.parentTabId, { advanced: true });`
It doesn't refresh a page immediately, it refreshed the page in about 4-5 sec after clicking. I tried to investigate it but the code has been written well. I don't know why it behaves this way yet. Need more investigation.

### DIFF
--- a/extension/chrome/settings/modules/add_key.ts
+++ b/extension/chrome/settings/modules/add_key.ts
@@ -39,10 +39,10 @@ View.run(class AddKeyView extends View {
   }
 
   setHandlers() {
-    $('.action_add_private_key').click(this.setHandlerPrevent('double', el => this.addPrivateKeyHandler(el)));
+    $('.action_add_private_key').click(this.setHandlerPrevent('double', this.addPrivateKeyHandler));
     $('#input_passphrase').keydown(this.setHandler((el, ev) => {
       if (ev.which === 13) {
-        this.addPrivateKeyHandler(el).catch(Catch.reportErr);
+        $('.action_add_private_key').click();
       }
     }));
   }
--- a/extension/chrome/settings/modules/add_key.ts
+++ b/extension/chrome/settings/modules/add_key.ts
@@ -14,63 +14,83 @@ import { Assert } from '../../../js/common/assert.js';
 import { KeyImportUi, UserAlert, KeyCanBeFixed } from '../../../js/common/ui/key_import_ui.js';
 import { initPassphraseToggle } from '../../../js/common/ui/passphrase_ui.js';
 import { Xss } from '../../../js/common/platform/xss.js';
+import { View } from '../../../js/common/view.js';
 
-Catch.try(async () => {
+View.run(class AddKeyView extends View {
 
-  const uncheckedUrlParams = Url.parse(['acctEmail', 'parentTabId']);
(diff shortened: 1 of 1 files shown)


michael-volynets:
looking into tests

michael-volynets:
After re-running everything is okay

tomholub:
> I also found a bug:
> When we call `BrowserMsg.send.reload(this.parentTabId, { advanced: true });`
> It doesn't refresh a page immediately, it refreshed the page in about 4-5 sec after clicking. I tried to investigate it but the code has been written well. I don't know why it behaves this way yet. Need more investigation.

That sounds like the bug you had last time, which manifested on your device but not on mine. Unsure.

michael-volynets:
Could you try to reproduce it by just adding a new key?
If the page refreshes immediately than maybe it's only from my side

extension/chrome/settings/modules/add_key.ts
```diff
@@ -14,63 +14,83 @@ import { Assert } from '../../../js/common/assert.js';
...
+    await this.loadAndRenderKeyBackupsOrRenderError();
+    $('.source_selector').css('display', 'block');
+    $('#spinner_container').text('');
+  }
 
-  const keyinfos = await Store.keysGet(acctEmail);
-  const privateKeysLongIds = keyinfos.map(ki => ki.longid);
-  let keyBackups;
+  setHandlers() {
+    $('.action_add_private_key').click(Ui.event.prevent('double', this.addPrivateKeyHandler));
```
tomholub: 1) Here please use `this.setHandlerPrenvent` instead.
2) I'm almost certain that this must have broken the functionality of `addPrivateKeyHandler` because `this` would be the HTMLElement, not this class. That means our tests are not sufficient to notice the brokenness. Please manually test it to see if it's broken or not, and assuming it is, please design a test that specifically fails when this functionality is utilized, and which should get fixed when you switch to `this.setHandlerPrevent`

michael-volynets: Oh, sorry, I didn't notice that there is this type of handlers.
Okay, will do

michael-volynets:
```
I'm almost certain that this must have broken the functionality of addPrivateKeyHandler because this would be the HTMLElement, not this class. That means our tests are not sufficient to notice the brokenness. Please manually test it to see if it's broken or not, and assuming it is, please design a test that specifically fails when this functionality is utilized, and which should get fixed when you switch to this.setHandlerPrevent
```
I think I checked it before but I will double-check it

michael-volynets: Yes, It worked because I have declared `addPrivateKeyHandler` function like a lamda expression. But I will change the code a little so the `addPrivateKeyHandler` function will be like a method.

tomholub: Ok, that's an interesting solution. I don't like that we would have to think about which methods are lambdas and which are just regular methods (I like consistency). But I do like that you wrote functioning code :)

I will also go ahead and remove the (el) => this... and just directly reference this.xxx in the callback, to make it same as the other files.

Thanks!

PR #2257, opened 2019-12-05
eslint all ts files recursively

limonte:
In #2241 I made a mistake, `**/*` isn't working recursively in bash.

Now all TS files are getting checked by `eslint`, there are some issues. @tomholub I'm not sure if these issues should be fixed or ignored, PTAL:

image


tomholub:
no-async-promise-executor looks promising, meaning it should probably be fixed, because I think it would drop errors silently. I'll take a look in more detail, some of these may be tricky to fix.

tomholub:
`no-async-promise-executor` is a really good linter :+1: I really hate when errors fall through for any reason, and we had a few of these in the codebase.

PR #2264, opened 2019-12-05
Issue #2231, refactored security file

michael-volynets:
Issue #2231,
Time Spent: 50m

### DIFF
--- a/extension/chrome/settings/modules/security.ts
+++ b/extension/chrome/settings/modules/security.ts
@@ -129,5 +129,6 @@ View.run(class SecurityView extends View {
         return true;
       }
     }
+    return false;
   }
 });
--- a/extension/chrome/settings/modules/security.ts
+++ b/extension/chrome/settings/modules/security.ts
@@ -16,6 +16,7 @@ import { View } from '../../../js/common/view.js';
 import { KeyInfo } from '../../../js/common/core/pgp.js';
 
 View.run(class SecurityView extends View {
+
   private readonly acctEmail: string;
   private readonly parentTabId: string;
   private primaryKi: KeyInfo | undefined;
@@ -36,7 +37,7 @@ View.run(class SecurityView extends View {
     const storage = await Store.getAcct(this.acctEmail, ['hide_message_password', 'outgoing_language']);
     $('#hide_message_password').prop('checked', storage.hide_message_password === true);
     $('.password_message_language').val(storage.outgoing_language || 'EN');
-    await this.renderPPOptionsIfStoredInSession();
+    await this.renderPassPhraseOptionsIfStoredPermanently();
     await this.loadAndRenderPwdEncryptedMsgSettings();
   }
 
@@ -47,7 +48,7 @@ View.run(class SecurityView extends View {
     $('.password_message_language').change(this.setHandler(() => this.onMsgLanguageUserChange()));
   }
 
-  private async renderPPOptionsIfStoredInSession() {
+  private async renderPassPhraseOptionsIfStoredPermanently() {
     const keys = await Store.keysGet(this.acctEmail);
     if (await this.isAnyPassPhraseStoredPermanently(keys)) {
       $('.forget_passphrase').css('display', '');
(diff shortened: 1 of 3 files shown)


extension/chrome/settings/modules/security.ts
```diff
@@ -7,122 +7,135 @@ import { Store } from '../../../js/common/platform/store.js';
...
-  const parentTabId = Assert.urlParamRequire.string(uncheckedUrlParams, 'parentTabId');
-  const embedded = uncheckedUrlParams.embedded === true;
-
-  await initPassphraseToggle(['passphrase_entry']);
+  async render() {
+    await initPassphraseToggle(['passphrase_entry']);
+    [this.primaryKi] = await Store.keysGet(this.acctEmail, ['primary']);
+    Assert.abortAndRenderErrorIfKeyinfoEmpty(this.primaryKi, false);
+    if (!this.primaryKi) {
+      return; // added do_throw=false above + manually exiting here because security.htm can indeed be commonly rendered on setup page before setting acct up
```
tomholub: Actually this is no longer true, we don't render security.htm before setting up acct anymore. You can remove the `false` above, and remove this if/return

extension/chrome/settings/modules/security.ts
```diff
@@ -7,122 +7,135 @@ import { Store } from '../../../js/common/platform/store.js';
...
+    }
+    if (this.embedded) {
```
tomholub: We no longer support embedded security form, so we can remove this, and also remove all references to this.embedded

extension/chrome/settings/modules/security.ts
```diff
@@ -7,122 +7,135 @@ import { Store } from '../../../js/common/platform/store.js';
...
-  const authInfo = await Store.authInfo(acctEmail);
+  private async renderPPOptionsIfStoredInSession() {
+    let storedInSession = false;
+    const keys = await Store.keysGet(this.acctEmail);
+    for (const key of keys) {
+      if (await Store.passphraseGet(this.acctEmail, key.longid, true)) {
+        storedInSession = true;
+        break;
+      }
+    }
```
tomholub: Actually the current naming is not correct, if I'm seeing correctly, we are testing if pass phrase is stored in permanent local storage, not in session.

This is a good opportunity to move these lines into a separate method, called more precisely `isAnyPassPhraseStoredPermanently`. then below you do `if(await this.isAnyPassPhraseStoredPermanently())`

extension/chrome/settings/modules/security.ts
```diff
@@ -7,122 +7,135 @@ import { Store } from '../../../js/common/platform/store.js';
...
+          $('.confirm_passphrase_requirement_change').click();
+        }
+      }));
+    }
+  }
 
-  if (embedded) {
-    $('.change_passhrase_container, .title_container').css('display', 'none');
-    $('.line').css('padding', '7px 0');
+  private async renderSubscriptionDetails() {
```
tomholub: this could be renamed `loadAndRenderPwdEncryptedMsgSettings`

extension/chrome/settings/modules/security.ts
michael-volynets: Okay, got it,
Done

extension/chrome/settings/modules/security.ts
michael-volynets: Thanks,
Done

extension/chrome/settings/modules/security.ts
michael-volynets: Yes, this is better,
Done

extension/chrome/settings/modules/security.ts
michael-volynets: Okay, but if I create a method like `isAnyPassPhraseStoredPermanently(keys)`, will it be okay? This way I won't call `await Store.keysGet(this.acctEmail);` function twice.
Or Store.keysGet(this.acctEmail);` isn't expensive operation?

tomholub: `isAnyPassPhraseStoredPermanently(keys)` is better

tomholub:
I wanted to refactor it further but won't be able to get to this in the next two days, just a note to myself

PR #2289, opened 2019-12-08
issue 1870 frames doubled

### LINKED ISSUE #1870: sometimes pgp frames double up
This is an intermittent error. Sometimes the content scripts will render two iframes instead of one.

For example, a message will show up twice (two pgp_block iframes) or it will show double the amount of attachments (10 instead of 5).

We replace everything once a second. I think this will happen when some stage of the replacing takes more than a second, before the DOM is updated to indicate that the content script is replacing it, but after a second has passed for another replace to take place.

image

Tests will also fail because they expect one block, but two show:

image

image

You can probably catch this error easily when you temporarily lower replace interval down to 100 or 50 ms, then try to open a few messages. I think sooner or later you should encounter it. If not, try even faster interval.

tomholub:
close #1870 1 hr

### DIFF
--- a/extension/js/content_scripts/webmail/gmail_element_replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail_element_replacer.ts
@@ -192,10 +192,17 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
   }
 
+  /**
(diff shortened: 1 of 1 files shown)


limonte:
Aha, this was the right place to fix it. Good to know! Thanks for the well-explained comment!

Issue #1438, opened 2018-12-30
Refactor Api.err and Pgp.* methods into separate classes


tomholub:
...

tomholub:
Please don't do one gigantic PR, maybe make PRs file by file if possible. Then we can have several PRs related to one issue.

michael-volynets:
Okay, will do,
There are more than 20 files, I think it's better to include 3-4 files in PR if files are small and make some few PRs with large files.

tomholub:
We will follow your style here.

What I particularly want to refactor are two things:

#### property methods - example `settings.ts`

```js
// current
export class Settings {
  static fetchAcctAliasesFromGmail = async (acctEmail: string): Promise<...> => {
    // ...
  }
}
// refactored
export class Settings {
  static async fetchAcctAliasesFromGmail(acctEmail: string): Promise<...> {
    // ...
  }
}
```

#### object literals with methods on them - example `pgp.ts`

There are various properties with groups of methods, eg `Pgp.key`, `Pgp.armor` etc. These should become separate classes such as `PgpKey`, `PgpArmor`, `PgpHash`. These classes should then go to separate files eg `pgp-key.ts`, `pgp-armor.ts` all in the same folder, where now `pgp.ts` is.

```js
// current
export class Pgp {
  public static key = {
    create: async (...): Promise<...> => {
      // ...
    },
  };
}
// refactored
export class PgpKey {
  static async create(...): Promise<...> {
    // ...
  }
}
```

tomholub:
Updated regexes to find these

`\n\s+(public|private)?\s*(static)?\s+[a-zA-Z0-9]+ = (async )?\([^)]*\)[^\n]+=> \{`
595 results

`\s+(static)?\s+[a-zA-Z0-9]+: (async )?\([^)]+\)[^\n]+=> \{`
98 results

tomholub:
I'll try to get a stab at the bigger group with a regex-replace

michael-volynets:
Got it, Thanks.
Will do

tomholub:
Remaining to refactor:
- [ ] Api.err as ApiErr
- [ ] Pgp.armor as PgpArmor
- [ ] Pgp.hash as PgpHash
- [ ] Pgp.key as PgpKey
- [ ] Pgp.armor as PgpArmor
- [ ] Pgp.armor as PgpArmor
- [ ] Pgp.password as PgpPwd
[...]

PR #2300, opened 2019-12-09
issue #1438 refactor with regex

tomholub:
#1438


tomholub:
@limonte this rule just saved me

image

PR #2303, opened 2019-12-09
"typescript.tsdk": "node_modules/typescript/lib"

limonte:
Connected to #2297

VSCode is still shipped with TS 3.6, in order to tell it to use TS 3.7 this setting should be there.

Alternatively, we can remove `.vscode/settings.json` from the repo at all. Let me know @tomholub wdyt

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,1 +1,3 @@
-{}
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}


tomholub:
We have it in the workspace settings - just restart your workspace after `npm install` and it should pick up on it.

image

limonte:
Note: `.vscode/settings.json` is generated automatically after `Select TS version -> Use workspace version`.

limonte:
> We have it in the workspace settings - just restart your workspace after npm install and it should pick up on it.

Thanks, will try and let you know.

limonte:
Yup, restarting the workspace helped :+1:

Issue #2239, opened 2019-12-03
use a view-framework for simpler frame elements 


tomholub:
in `extension/chrome/elements`:
 - `add_pubkey.ts`
 - `backup.ts`
 - `passphrase.ts`
 - `pgp_pubkey.ts`
 - `subscribe.ts`

I've skipped the more complex ones which I'll have a look at, and maybe refactor further.

michael-volynets:
I will be working on it the same way as I was working on #2231 (one file - one PR)

michael-volynets:
Note: `Backup` file has been already refactored

michael-volynets:
No, `Backup` element isn't updated, I was looking at `backup` module file

PR #2304, opened 2019-12-09
Refactored pgp_pubkey.ts file

michael-volynets:
Issue #2239
Time Spent: 1h


tomholub:
Refactor `public_key.ts` to expect several several pubkey with several emails

 - [ ] expect possibly several public keys contained, render them all
 - [ ] expect that each public key possibly contains several emails, render each
 - [ ] each pubkey should have `import for all emails` button as well (if > 1 email)
 - [ ] on top should be `import all public keys` button (if > 1 public key)

PR #2315, opened 2019-12-09
issue 2314 fix progress percentage

### LINKED ISSUE #2314: fix progress percentage in send btn

tomholub:
close #2314


tomholub:
we could make a test for this - we have a test that sends larger file, it could check that `%` shows at some point in the `send_btn`.

PR #2321, opened 2019-12-10
Make Setup process accessible with keyboard

### LINKED ISSUE #2096: FlowCrypt Set Up isn't accessible with keyboard
There are `<div>`s which look like buttons, but they're not buttons and therefore inaccessible for keyboard users, e.g.

image

As a general rule, everything which behaves like a button should be `<button>`

limonte:
Fixes #2096

### DIFF
--- a/extension/chrome/elements/add_pubkey.htm
+++ b/extension/chrome/elements/add_pubkey.htm
@@ -38,9 +38,9 @@
       <textarea class="pubkey" placeholder="ASCII Armored Public Key"></textarea>
     </div>
     <div class="line">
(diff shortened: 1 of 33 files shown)


limonte:
There are 24 more `<div class="button ...>` outside of the `settings` folder, do you want me to fix them in the scope of this PR or to handle them separately?

tomholub:
> @tomholub There are 24 more `<div class="button ...>` outside of the `settings` folder, do you want me to fix them in the scope of this PR or to handle them separately?

You can go ahead and fix them all, but have a look at a few of them to make sure that the visible styling doesn't break. Thanks!

image

Issue #2324, opened 2019-12-10
remove archaic openInSettings compose test method


tomholub:
Causes trouble needlessly

![image](https://user-images.githubusercontent.com/6306961/70554810-baf58e80-1b75-11ea-8c74-7920f76c3823.png)

PR #2325, opened 2019-12-10
issue 2324 remove tests that use new message in settings footer

tomholub:
#2324 30 min


tomholub:
Should also remove the button itself, and all compose dialog settings that relate to placement=settings (rendering add_pubkey dialog, rendering passphrase dialog, closing).

PR #2332, opened 2019-12-10
Issue #1438, refactoring Pgp.armor

michael-volynets:
Don't review it yet, I should fix the issue with the errors in the tests and go through the code


extension/js/content_scripts/webmail/gmail_element_replacer.ts
```diff
@@ -20,6 +19,7 @@ import { Gmail } from '../../common/api/email_provider/gmail/gmail.js';
 import { Ui } from '../../common/browser/ui.js';
 import { Browser } from '../../common/browser/browser.js';
 import { BrowserMsg } from '../../common/browser/browser-msg.js';
+import { PgpArmor } from '../../../../test/source/core/pgp/armor.js';
```
tomholub: This is the reason. You have imported from `test` folder, but you should be importing from `../../common/core` folder.

michael-volynets: I got it
Thanks, I think it would take me much more time if you didn't help me.

michael-volynets:
Got it, thanks,
going to change the file name

michael-volynets:
Total Time spent: ~1h 40m

PR #2347, opened 2019-12-11
Add eslint max-len rule

limonte:
This is related to #2321 where I made `max-len` violations.

Now, `eslint` is taking care of `max-len`.


limonte:
> You should also install tslint, so that you get instant feedback on the other tslint rules. We also wrote some rules of our own.

I did it, but VSCode isn't highlighting tslint issue for me. Will double-check :+1:

tomholub:
I'm using TSLint 1.2.3 extension from Microsoft

Writing
```
      const x = {
        hello: 1,
        there: 2
      };
```
Should produce

image

I'm also using ErrorLens, because some of the TSLint rules are quite brutal, to make the errors less annoying while I'm still writing code.

limonte:
Thanks! btw you can remove the `stylelint` extention, as `stylelint-plus` is the replacement. Also, don't forget to install the `eslint` extension!

tomholub:
Great, thanks :)

tomholub:
Somehow I'm still not getting eslint to show anything in my IDE, even after the extension was installed. Is there anything else to configure?

limonte:
Yes, by default eslint plugin for VSCode validates only JS:

image

This PR should help: https://github.com/FlowCrypt/flowcrypt-browser/pull/2351

PR #2360, opened 2019-12-12
Issue #1438, refactored pgp password

michael-volynets:
Issue #1438,
Time Spent: 20m

### DIFF
--- a/extension/js/common/core/pgp-password.ts
+++ b/extension/js/common/core/pgp-password.ts
@@ -18,9 +18,31 @@ interface PasswordStrengthResult {
 }
 
 export class PgpPwd {
+  private static CRACK_GUESSES_PER_SECOND = 10000 * 2 * 4000;
+  private static CRACK_TIME_WORDS_PWD = [ // the requirements for a one-time password are less strict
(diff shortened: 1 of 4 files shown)


extension/js/common/core/pgp-password.ts
```diff
@@ -0,0 +1,40 @@
...
+    pass: boolean;
+  };
+  seconds: number;
+  time: string;
+}
+
+export class PgpPwd {
+  static estimateStrength = (zxcvbnResultGuesses: number, type: 'passphrase' | 'pwd' = 'passphrase'): PasswordStrengthResult => {
+    const timeToCrack = zxcvbnResultGuesses / Pgp.CRACK_GUESSES_PER_SECOND;
+    for (const word of type === 'pwd' ? Pgp.CRACK_TIME_WORDS_PWD : Pgp.CRACK_TIME_WORDS_PASS_PHRASE) {
```
tomholub: please move `CRACK*` props to this class

tomholub: Then you can probably make them private

michael-volynets: I didn't notice that,
Thanks

PR #2366, opened 2019-12-13
Add recommended extensions .vscode/extensions.json

limonte:
I made a small Wiki page on how to get started with VSCode: https://github.com/FlowCrypt/flowcrypt-browser/wiki/Getting-started-in-VSCode

This PR will ensure that when a new developer starts working on the project, he/she will be prompted to install recommended extensions:

image

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,15 @@
+{
+	// See https://go.microsoft.com/fwlink/?LinkId=827846 to learn about workspace recommendations.
+	// Extension identifier format: ${publisher}.${name}. Example: vscode.csharp
+
+	// List of extensions which should be recommended for users of this workspace.
+	"recommendations": [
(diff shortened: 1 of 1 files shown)


tomholub:
That's useful, that way if we add an extension requirement, we can easily prompt each other too.

limonte:
Hey @tomholub 

Would you mind if I remove this line

```js
    "editor.fontFamily": "'Hack', 'Droid Sans Mono', 'monospace', monospace, 'Droid Sans Fallback'",
```

from `flowcrypt-browser.code-workspace`?

I believe, the font prefenrences shouldn't be in the project settings, but rather in user settings (`.config/Code/User/settings.json`)

tomholub:
Correct, it should be in user settings. Every user should set their own. I've set this before there were any other people on the project.

PR #2372, opened 2019-12-13
Issue 1438 refactor pgp internal

michael-volynets:
Issue #1438, refactored a part of `Pgp.internal`.
Another part I will refactor when I move `PgpMsg` class to another file.

### DIFF
--- a/extension/js/common/core/pgp-key.ts
+++ b/extension/js/common/core/pgp-key.ts
@@ -203,6 +203,17 @@ export class PgpKey {
     return await PgpKey.longid(await PgpKey.fingerprint(keyOrFingerprintOrBytes));
   }
 
+  static longids = async (keyIds: OpenPGP.Keyid[]) => {
+    const longids: string[] = [];
+    for (const id of keyIds) {
+      const longid = await PgpKey.longid(id.bytes);
+      if (longid) {
+        longids.push(longid);
+      }
+    }
+    return longids;
+  }
+
   static usable = async (armored: string) => { // is pubkey usable for encrytion?
     if (!PgpKey.fingerprint(armored)) {
       return false;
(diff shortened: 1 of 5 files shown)


extension/js/common/core/pgp-key.ts
```diff
@@ -308,4 +319,9 @@ export class PgpKey {
     }
     throw new Error('No valid signature found in key');
   }
+
+  static cryptoKeyOptionalMatchingKeyid = (key: OpenPGP.key.Key, forMsgKeyids: OpenPGP.Keyid[] | undefined) => {
```
tomholub: Actually, despite the name, this one should go into `PgpMsg` because I believe it will be used there privately.

extension/js/common/core/pgp.ts
```diff
@@ -182,22 +182,10 @@ export class Pgp {
   }
 
   public static internal = {
-    msgBlockObj: (type: MsgBlockType, content: string | Buf, missingEnd = false): MsgBlock => ({ type, content, complete: !missingEnd }),
-    msgBlockDecryptErrObj: (encryptedContent: string | Buf, decryptErr: DecryptError): MsgBlock => ({ type: 'decryptErr', content: encryptedContent, decryptErr, complete: true }),
     msgBlockAttObj: (type: MsgBlockType, content: string, attMeta: AttMeta): MsgBlock => ({ type, content, complete: true, attMeta }),
```
tomholub: One day I may move these into a separate class, but I have to think about it first.

MsgBlockFactory or something like that. But if I did that, there would be a whole bunch of other things I'd have to refactor (move things from `PgpArmor` and `Mime`) and I'm not 100% sure it makes sense yet.

I'll make an issue.

extension/js/common/core/pgp-key.ts
michael-volynets: Okay, I will check it

Issue #337, opened 2017-01-11
pgp message contained inside signed message - trouble


tomholub:
nvimp convo 14e6c09f8dede152 from MC

tomholub:
repro - copy armored encrypted message into a new secure compose and paste that into secure compose, then switch to sign. Send it. Can it be verified?

I believe this may have been fixed since, in OpenPGP.js

schmorrison:
@tomholub  I am getting this problem, the encrypted message block has a single dash and whitespace character before the opening and closing on the encrypted message block:

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

- -----BEGIN PGP MESSAGE-----
Version: FlowCrypt 6.6.7 Gmail Encryption
Comment: Seamlessly send and receive encrypted email

wcFMA/bY5+sl2uFqAQ//RkCDk8rXoD1lv4trb4lgoWQX/HtrhOmoDSmLXs0F
fJz9b8WoCmWOLOsv5Qp2Xa88uQMvC9ByWx75Gi9fPcbq8TW4ul+tbHgTQ4Vr
wkols3CABbatFNHvmwvyvGXMrrvKjrSfu7DoImR0roq2WUzB3uNaP66XF5gs
XrGDXlcfFgQeZkDdTjvdnuOSo4O3p1QCHQdKa1dS3ezBm+fMjjjmKA0sqJWW
GvEsE/SKlrlQiRHlCxwiKmaGxLHJt7fR9/xbpOeNLmBsbtfOyEArBDyT2r2Z
cFj5p/aQKhAAD303PFt/7vf+8bnWfobyvjKtm54qhaZX2nCdtwStz7nB34lF
0MaHHh2zyRoqOdt9ZUTOUtdRsWtCp3EX9Z5Q3hQjDc8C/bgdRDXyu8hk+fAs
Yfx82tcw+gRm18NXV1Ger8Kucx6iusbNlx8fvNB9rYkGbPgwy0SwcKTHXwv9
zjYbttW5JfJLasPQ6NyRdQXmFAglnwquHcm+1gBpaJWPcMM7a/ThVQmhNJv/
iCefCys/3W3viFUYjI2A4eL93ELW3eaEbXHscj/2A6iivdOfV/+Z8Z3rC47T
j1WTR36+K7gaZSWXnp2jCzY1mzBJoUcy7iD2Q233/Oj7z0oxBuV9UHWWZG+l
[...]

PR #2384, opened 2019-12-16
Issue 337 pgp message contained inside signed

michael-volynets:
Issue #337 
Time Spent: 2h

Please merge new mock data first and the rerun tests.
Thanks

### DIFF
--- a/test/source/tests/tests/decrypt.ts
+++ b/test/source/tests/tests/decrypt.ts
@@ -191,6 +191,13 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithNewBrowser:
       await InboxPageRecipe.checkDecryptMsg(t, browser, { acctEmail, threadId, expectedContent });
     }));
 
+    ava.default('[standalone] - inbox - encrypted message inside signed', testWithSemaphoredBrowser('compatibility', async (t, browser) => {
+      const inboxPage = await browser.newPage(t, 'chrome/settings/inbox/inbox.htm?acctEmail=flowcrypt.compatibility%40gmail.com&threadId=16f0bfce331ca2fd');
+      const pgpBlock = await inboxPage.getFrame(['pgp_block.htm']);
+      const content = await pgpBlock.read('#pgp_block');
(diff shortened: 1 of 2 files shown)


test/source/tests/tests/decrypt.ts
```diff
@@ -191,6 +191,13 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithNewBrowser:
       await InboxPageRecipe.checkDecryptMsg(t, browser, { acctEmail, threadId, expectedContent });
     }));
 
+    ava.default('[standalone] - inbox - encrypted message inside signed', testWithSemaphoredBrowser('compatibility', async (t, browser) => {
+      const inboxPage = await browser.newPage(t, 'chrome/settings/inbox/inbox.htm?acctEmail=flowcrypt.compatibility%40gmail.com&threadId=16f0bfce331ca2fd');
+      const pgpBlock = await inboxPage.getFrame(['pgp_block.htm']);
```
tomholub: You have to first await the frame above, look at other usages of `getFrame`

tomholub: That will make such tests a little lest flaky, I think

test/source/tests/tests/decrypt.ts
```diff
@@ -191,6 +191,13 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithNewBrowser:
...
+      const content = await pgpBlock.read('#pgp_block');
```
tomholub: You have to also await test state "ready"

tomholub: image

test/source/tests/tests/decrypt.ts
michael-volynets: Agree, Thanks!

test/source/tests/tests/decrypt.ts
michael-volynets: I don't know how I couldn't have noticed this. The test passed when I ran it locally.
Thanks.

Issue #2048, opened 2019-09-30
proactively check if new recipients have PGP as they are being loaded


tomholub:
When I search for contacts from Google API and there are some contacts that have PGP installed I think it would be great to render them with green lock and etc.. So maybe when I fetch  contacts from Gmail  then I need to find out which of them have PGP installed through our Attester and then I add all of the fetched contacts into Local DB.

What do you think about it?

_Originally posted by @michael-volynets in https://github.com/FlowCrypt/flowcrypt-browser/issues/879#issuecomment-536582662_

tomholub:
Attester will be fine, can proceed with this.

PR #2385, opened 2019-12-16
Issue #2048, check using attester and add to DB

michael-volynets:
Issue #2048,
Time Spent: 1h 20m (30m dev, ~50 testing)

It was easier than I expected but I had to test it well so I spent almost one hour testing it.
I've tested it pretty well and didn't find any bugs.

### DIFF
--- a/extension/chrome/elements/composer/composer-recipients.ts
+++ b/extension/chrome/elements/composer/composer-recipients.ts
@@ -22,6 +22,8 @@ import { ApiErr } from '../../../js/common/api/error/api-error.js';
 import { PgpKey, Contact } from '../../../js/common/core/pgp-key.js';
 
 export class ComposerRecipients extends ComposerComponent {
+  private readonly failedLookupEmails: string[] = [];
+
   private addedRecipients: RecipientElement[] = [];
   private BTN_LOADING = 'Loading..';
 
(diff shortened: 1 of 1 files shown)


michael-volynets:
And it also works pretty fast. Each request takes about 60ms so if we need to look up 8 email addresses it will take only 480ms which is good.

tomholub:
> And it also works pretty fast. Each request takes about 60ms so if we need to look up 8 email addresses it will take only 480ms which is good.

It depends - emails who don't have public key recorded on Attester can take a few seconds per request. Eg try to search `startupsoft`. I haven't tested it yet, how does it behave in such situation?

michael-volynets:
Yes, it's loading too long (about 6-7 sec). I will try to fix it. I will make parallel requests.
Thanks for the found.

tomholub:
> Yes, it's loading too long (about 6-7 sec). I will try to fix it. I will make parallel requests.
> Thanks for the found.

Rather than making parallel requests (and waiting for them to all finish before rendering?) it may be better to just fire off `lookupPubkeyFromDbOrKeyserverAndUpdateDbIfneeded` asynchronously, and don't worry about re-rendering (because as the user types the name or email, it will probably get re-rendered anyway).

In that case, we may need to keep a in-memory cache of emails we looked up that already returned no results, so that we don't keep asking attester about the same missing pubkeys in the same session.

michael-volynets:
I just tested parallel requests and it can take up to 2-3 seconds, I don't think it's too long.
If we will do the way you purposed in the last message the lock (next to the found contact's email) may blink from gray to green once. 

But I'm going to check this how it will work actually.

michael-volynets:
Everything works well if a user types slowly but if I either type fast or copy the email address the compose window won't render it at all.
Maybe it would be better to leave the awaited parallel way? There would be less code, minimum delays and everything should work well.

Or I can spend some more time to fix the bug but as I said I prefer awaited parallel way.

tomholub:
I'll test it. Thanks

tomholub:
But still, we should not keep requesting the same email again and again as I search.

If I have 10 contacts that don't have a pubkey: one@domain.co, two@domain.co, ...

And I very slowly type: `d`, `o`, `m`, `a`, `i`, `n`

Then we have just made 10 * 6 = 60 requests, when it should have ideally been 10 requests (assuming all these contacts already showed when I pressed `d`)

tomholub:
The solution works well when the searched contacts actually do end up having public keys. When they don't, it will make a lot of expensive calls (several seconds each) that could potentially be skipped.

michael-volynets:
I agree, done

PR #2388, opened 2019-12-17
Remove editor.formatOnSave and editor.formatOnPaste

limonte:
Since recently I'm using the `flowcrypt-browser.code-workspace` configuration (previously I was doing `code ~/www/flowcrypt-browser` which isn't taking `.code-workspace` file into account) 

I don't know how `editor.formatOnSave` and `editor.formatOnPaste` weren't bothering anyone before me :)

Here's what's happening:

With the clean (no changes) repo, I open `compose.htm`, hit <kbd>Ctrl</kbd>+<kbd>S</kbd> and the file is changed automatically:

This is happening because of `html.format.wrapLineLength` which is `120` by default. This should be happening for everybody who is using `flowcrypt-browser.code-workspace`.

So these wrapping changes are getting into other changes which isn't desired.

### DIFF
--- a/flowcrypt-browser.code-workspace
+++ b/flowcrypt-browser.code-workspace
@@ -6,8 +6,6 @@
     "typescript.preferences.importModuleSpecifier": "relative",
     "typescript.preferences.quoteStyle": "single",
     "tslint.configFile": "conf/tslint.yaml",
-    "editor.formatOnSave": true,
-    "editor.formatOnPaste": true,
     "eslint.autoFixOnSave": true,
     "eslint.validate": [
       "javascript",


tomholub:
It was bothering me too, but only in html. It's desired in typescript. Is there a way to only disable it for html?

limonte:
> It was bothering me too, but only in html. It's desired in typescript. Is there a way to only disable it for html?

For TS, we have the `max-len` rule enabled: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/.eslintrc.js#L24

It's not automatically fixable though, but I wouldn't want robots to wrap long TS lines anyways, they aren't smart enough (yet).

tomholub:
I don't really care about long lines (ts or html) auto-format. In fact I don't care about any auto-formatting of html.

Anything we can do to keep auto-format on paste/save in ts, without doing unwanted things to html?

limonte:
> Anything we can do to keep auto-format on paste/save in ts, without doing unwanted things to html?

Yup, `"html.format.enable": false,` will do. I'll open another PR.

PR #2394, opened 2019-12-18
Issue 2365 refactoring attachment download view

### LINKED ISSUE #2365: refactor AttachmentDownloadView and InboxView
refactor AttachmentDownloadView and InboxView

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/2364#issuecomment-565146330_

michael-volynets:
Closes #2365,
Time Spent: ~1h 40m

Refactored, I've already reviewed it by myself so you can start reviewing it.


extension/chrome/elements/attachment.ts
```diff
@@ -15,69 +15,96 @@ import { Url } from '../../js/common/core/common.js';
...
-      Xss.sanitizeRender('body.attachment', `Error downloading file - no internet. ${Ui.retryLink()}`);
-    } else {
-      Catch.reportErr(e);
-      Xss.sanitizeRender('body.attachment', `Error downloading file - ${String(e)}. ${Ui.retryLink()}`);
-    }
-  };
+  }
 
-  const getUrlFileSize = (origUrl: string): Promise<number | undefined> => new Promise(resolve => {
+  private getUrlFileSize = (origUrl: string): Promise<number | undefined> => new Promise(resolve => {
```
tomholub: note for me - refactor async / await explicitly

tomholub:
There seems to be a bug (already present before the refactor, I think) that makes it not really show download percentage.

It shows `100%` from the beginning to end, and then it's done.

image


You can test with https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#label/tests%2Fattachment/WhctKHTCDdzqhQwQwQwWwqqKpmXSZPzCrvMpRZmsTjFpcQVSktdMQxRWKLFcbjqrBtvtxLV

To test it easier (manually), you can slow down your browser's internet connection in network tab.

Let's also make a test for this. You'll need to mock up the `gmail/message/.../attachment/...` api. Just add an endpint `gmail/message/specificmsgid/attachment/?` (or similar) where you hard-code the message id, and leave the attachment id as a resource-id parameter. Then test that the attachment frame loads properly, and that you can click on it and it will render the "ready" result.

After that, to test the download percentage, you may need to also slow down how fast the mock serves the mock attachment files, by sending the data over the wire chunk by chunk and adding a delay between each chunk. I could help with this part.

michael-volynets:
Don't you mind if I create another issue to make the test?
I will fix this bug in this PR but I will write the test in another one.

michael-volynets:
I'm started thinking that we need to remove this percentage bar at all.
The reason is that we can't know exactly what the file size is because when we download a file we get a gzipped response (`content-encoding: gzip` in response headers). When we get this type of response we will always have `lengthComputable` set to `false` in Chrome browser that means that `event.total` will be always 0.
The solution would be to send another header that indicates the actual size of the file (e.g. `X-Content-Length`) but I tried to find some info about it, maybe we could request this header from `googlapis` but I haven't found anything.

I also found that there is a `content-length` header I thought I found the solution but then I realized that `content-length` indicates the compressed length of the file and `event.total` is decompressed length so in real example we could have `content-length` = 15 000 bytes and `event.total` = 20 000 bytes so percentages will be about 130% which is not good :/

So as a solution I would recommend showing just a loader without any percentage. We may somewhere `Loading...` label and that's it.

tomholub:
> I'm started thinking that we need to remove this percentage bar at all.
> The reason is that we can't know exactly what the file size is because when we download a file we get a gzipped response (`content-encoding: gzip` in response headers). When we get this type of response we will always have `lengthComputable` set to `false` in Chrome browser that means that `event.total` will be always 0.
> The solution would be to send another header that indicates the actual size of the file (e.g. `X-Content-Length`) but I tried to find some info about it, maybe we could request this header from `googlapis` but I haven't found anything.
> 
> I also found that there is a `content-length` header I thought I found the solution but then I realized that `content-length` indicates the compressed length of the file and `event.total` is decompressed length so in real example we could have `content-length` = 15 000 bytes and `event.total` = 20 000 bytes so percentages will be about 130% which is not good :/
> 
> So as a solution I would recommend showing just a loader without any percentage. We may put somewhere `Loading...` label and that's it.

I think last time I did something like `Math.min((downloadedByteLen / 1.35) / byteLen, 1.00) * 100` or something like that. The logic should still be somewhere to be seen.

It worked fine, and tended to give good indication about when things will be done.

Try to change browser speed to "slow 3g" and download one of the attachments. You don't know if you should go grab a coffee or a lunch, it just keeps circling for what seems like forever. There has to be some progress indication, even if not precise.

tomholub:
It's right here:
```
progressEl.text(`${Math.floor(((received * 0.75) / size) * 100)}%`);
```

tomholub:
Unless something changed in the way they compress it (that they didn't before). But we should really investigate this. It used to work nicely.

michael-volynets:
> I think last time I did something like Math.min((downloadedByteLen / 1.35) / byteLen, 1.00) * 100 or something like that. The logic should still be somewhere to be seen.

Now, `byteLen` is always '0' in Chrome. let me check how it behaves in Firefox. It could behave differently way because I read that the issue is present only in Chrome

michael-volynets:
@tomholub  yes, I was right, on Firefox everything works perfectly so we can at least Firefox as is.
But we have to something with Chrome. You know, we can even simulate downloading, I mean it will change percentage but it will be stuck on about 75% until the file is fully downloaded. I don't think it's a good practice but we don't have any option. On Chrome, we can do either this or don't show the progress bar at all.

tomholub:
> @tomholub yes, I was right, on Firefox everything works perfectly so we can at least leave Firefox as is.
> But we have to do something with Chrome. You know, we can even simulate downloading, I mean it will change percentage but it will be stuck on about 75% until the file is fully downloaded. I don't think it's a good practice but we don't have any option. On Chrome, we can do either this or don't show the progress bar at all.

I don't really understand - why is the length available in Firefox, but not chrome? What is the difference?

Issue #1842, opened 2019-07-15
allow generating revocation cert in settings


tomholub:
> I took issue with FlowCrypt for two things that could not be provided through the browser extension.
> 2) You could not generate revocation certs.

tomholub:
When I go to FlowCrypt Settings -> Additional Settings -> (my key), I should see a `revocation cert` option:

![image](https://user-images.githubusercontent.com/6306961/71090316-a5e2b600-219a-11ea-9384-cdb59c60e5a1.png)

When clicked, it will do a confirmation modal:

> Revocation cert is used when you want to revoke your Public Key (meaning you are asking others to stop using it). You can save it do your hard drive, and use it later in case you ever need it.
> 
> Would you like to generate and save a revocation cert now?

When I confirm it, a new revocation cert should be generated and downloaded.

PR #2396, opened 2019-12-19
Issue #1842, Added ability to download revocation certificate

michael-volynets:
Issue #1842 
Time Spent: 4h

I thought it would be easier :)
I faced some issues that slowed me a little.
1. I had to update typings. First I decided to copy/pasted definitions from here `https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/openpgp/ts3.2/index.d.ts`  but that didn't work so I added that definitions that we only need.
2. I don't know why but `openpgp.key.Key.revoke` returns not `string` but `Stream`. I checked the original method to understand what's going on there. I checked it and didn't find out why so I had to somehow deal with that, and I corrected typings a little, so `openpgp.key.Key.revoke` may return either `string`, `Stream<string>` or `undefined`.
3. And the last thing I had to do is to handle a situation when a user doesn't have stored passphrase. Actually, I didn't have any trouble implementing this step.

### DIFF
--- a/extension/chrome/settings/modules/my_key.ts
+++ b/extension/chrome/settings/modules/my_key.ts
@@ -2,7 +2,6 @@
 
 'use strict';
 
-import { Catch } from '../../../js/common/platform/catch.js';
 import { Store } from '../../../js/common/platform/store.js';
 import { Att } from '../../../js/common/core/att.js';
 import { Browser } from '../../../js/common/browser/browser.js';
@@ -21,11 +20,12 @@ import { Ui } from '../../../js/common/browser/ui.js';
 declare const ClipboardJS: any;
 
 View.run(class MyKeyView extends View {
+
   private readonly acctEmail: string;
   private readonly longid: string;
   private readonly myKeyUserIdsUrl: string;
   private readonly myKeyUpdateUrl: string;
-  private primaryKi!: KeyInfo;
+  private keyInfo!: KeyInfo;
 
   constructor() {
     super();
@@ -37,12 +37,12 @@ View.run(class MyKeyView extends View {
   }
 
   render = async () => {
-    [this.primaryKi] = await Store.keysGet(this.acctEmail, [this.longid]);
-    Assert.abortAndRenderErrorIfKeyinfoEmpty(this.primaryKi);
-    const prv = await PgpKey.read(this.primaryKi.private);
+    [this.keyInfo] = await Store.keysGet(this.acctEmail, [this.longid]);
+    Assert.abortAndRenderErrorIfKeyinfoEmpty(this.keyInfo);
+    const prv = await PgpKey.read(this.keyInfo.private);
     $('.action_view_user_ids').attr('href', this.myKeyUserIdsUrl);
     $('.action_view_update').attr('href', this.myKeyUpdateUrl);
-    $('.key_words').text(this.primaryKi.keywords);
+    $('.key_words').text(this.keyInfo.keywords);
     $('.email').text(this.acctEmail);
     $('.key_fingerprint').text(await PgpKey.fingerprint(prv, 'spaced') || '(unknown fingerprint)');
     await this.setPubkeyContainer();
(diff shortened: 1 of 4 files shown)


extension/chrome/settings/modules/my_key.ts
```diff
@@ -68,13 +79,35 @@ View.run(class MyKeyView extends View {
     }
   }
 
+  private downloadRevocationCert = async (enteredPP?: string) => {
```
tomholub: I'm thinking that the code could really simplify we used `elements/passphrase.htm` module instead.

Or maybe if we could make Sweetalert prompt where user enters some info, then we could use that for pass phrases which is much simpler.

But for now we can go with this.

extension/chrome/settings/modules/my_key.ts
```diff
@@ -44,20 +46,29 @@ View.run(class MyKeyView extends View {
...
+    await initPassphraseToggle(['input_passphrase']);
   }
 
   setHandlers = () => {
-    $('.action_download_pubkey').click(this.setHandlerPrevent('double', () => this.downloadPubKeyHandler()));
-    $('.action_download_prv').click(this.setHandlerPrevent('double', () => this.downloadPrvKeyHandler()));
-    const clipboardOpts = { text: (trigger: HTMLElement) => trigger.className.includes('action_copy_pubkey') ? this.primaryKi!.public : this.primaryKi!.private };
+    $('.action_download_revocation_cert').click(this.setHandlerPrevent('double', () => this.downloadRevocationCert()));
+    $('.action_continue_download').click(this.setHandlerPrevent('double', () => this.downloadRevocationCert(String($('#input_passphrase').val()))));
+    $('#input_passphrase').on('keydown', this.setHandler((el, ev) => {
```
tomholub: Can use `this.setEnterHandlerThatClicks`

extension/chrome/settings/modules/my_key.ts
```diff
@@ -44,20 +46,29 @@ View.run(class MyKeyView extends View {
...
+      if (ev.which === 13) {
+        $('.action_continue_download').click();
+      }
+    }));
+    $('.action_cancel_download_cert').click(this.setHandler(() => { $('.enter_pp').hide(); }));
+    $('.action_download_pubkey').click(this.setHandlerPrevent('double', this.downloadPubKeyHandler));
```
tomholub: should use `() => this...` for handlers, for consistency.

extension/js/common/core/types/openpgp.d.ts
```diff
@@ -850,6 +875,11 @@ declare namespace OpenPGP {
       getKeyId(): Keyid;
     }
 
+    export interface revoke_reasonForRevocation {
```
tomholub: No need to export it

extension/js/common/core/types/openpgp.d.ts
```diff
@@ -818,6 +841,8 @@ declare namespace OpenPGP {
       update(key: Key): void;
       verifyPrimaryKey(): Promise<enums.keyStatus>;
       isRevoked(): Promise<boolean>;
+      revoke(easonForRevocation: revoke_reasonForRevocation, date?: Date): Promise<Key>;
```
tomholub: typo `eason`

extension/js/common/core/pgp-key.ts
```diff
@@ -354,4 +354,15 @@ export class PgpKey {
     throw new Error('No valid signature found in key');
   }
 
+  static revoke = async (key: OpenPGP.key.Key) => {
```
tomholub: Will add signature for clarity

extension/chrome/settings/modules/my_key.ts
```diff
@@ -68,13 +79,35 @@ View.run(class MyKeyView extends View {
...
+          return;
+        }
+      } else {
+        $('.enter_pp').show();
+        return;
+      }
+    }
+    $('.enter_pp').hide();
+    $('#input_passphrase').val('');
+    const revokedArmored = await PgpKey.revoke(prv);
```
tomholub: Should test if not undefined, and alert instead

extension/js/common/core/types/openpgp.d.ts
```diff
@@ -704,6 +704,29 @@ declare namespace OpenPGP {
       signature = 6,
     }
 
+    enum reasonForRevocation {
+      /**
```
tomholub: will compact the comments a bit

extension/js/common/core/types/openpgp.d.ts
```diff
@@ -704,6 +704,29 @@ declare namespace OpenPGP {
...
+       * No reason specified (key revocations or cert revocations)
+       */
+      no_reason = 0,
+      /**
+       * Key is superseded (key revocations)
+       */
+      key_superseded = 1,
+      /**
+       * Key material has been comPromise<any>d (key revocations)
```
tomholub: typo

tomholub:
Todo - still investigate output format (maybe should say revocation)

michael-volynets:
Thanks!

PR #2401, opened 2019-12-19
Make sure that cursor is visible when typing

### LINKED ISSUE #2400: The cursor is not staying in the visible part of the editor
The same as https://github.com/neilj/Squire/issues/374

limonte:
Fixes #2400

### DIFF
--- a/extension/chrome/elements/composer/composer-input.ts
+++ b/extension/chrome/elements/composer/composer-input.ts
@@ -23,6 +23,7 @@ export class ComposerInput extends ComposerComponent {
     this.handlePasteImages();
     this.initShortcuts();
     this.resizeReplyBox();
+    this.scrollIntoView();
   }
 
   removeRichTextFormatting = () => {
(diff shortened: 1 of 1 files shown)


extension/chrome/elements/composer/composer-input.ts
```diff
@@ -137,13 +138,30 @@ export class ComposerInput extends ComposerComponent {
   }
 
   private resizeReplyBox = () => {
-    this.squire.addEventListener('input', () => {
+    this.squire.addEventListener('cursor', () => {
       if (this.composer.view.isReplyBox) {
         this.composer.size.resizeComposeBox();
```
tomholub: Possible concern - in case `cursor` event spams with each keypress, this would probably cause higher CPU usage to evaluate if it should resize or not on each keypress (I've had that problem before). I'll have a look to confirm.

tomholub: It does spam but could not observe any difference in cpu usage (70% while typing, either way).

PR #2421, opened 2019-12-20
Do not autoformat links for plain text messages

### LINKED ISSUE #2409: pasted links should not be clickable when not richText
1) don't select rich text
2) paste a plain link from address bar
3) hit enter

image

expected:
image

limonte:
Fixes #2409

Unfortunately, it doesn't seem possible to fix the issue without changing `squire-raw.js`. In order not to override changes during the upgrade, I added the comment. @tomholub please let me know if you have some other suggestions on how to prevent accidental overriding changes in `lib` files.

Should we maybe create the internal fork of `squire` (as private repo) to mirror the changes we're doing on it?

### DIFF
--- a/extension/lib/squire-raw.js
+++ b/extension/lib/squire-raw.js
@@ -1,5 +1,9 @@
 /* Copyright © 2011-2015 by Neil Jenkins. MIT Licensed. */
 
+// ! CUSTOM CHANGES, please apply the same changes when upgrading the library !
+//
+// 1. handleEnter(): Wrap addLinks( range.startContainer, root, self ) into if (self._config.richText) { }
+
 ( function ( doc, undefined ) {
 
 "use strict";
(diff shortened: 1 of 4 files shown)


extension/lib/squire-raw.js
```diff
@@ -1434,7 +1438,9 @@ var handleEnter = function ( self, shiftKey, range ) {
     // Remove any zws so we don't think there's content in an empty
     // block.
     self._recordUndoState( range );
-    addLinks( range.startContainer, root, self );
+    if (self._config.richText) {
```
tomholub: Squire will still respond to spacebar, so we'd need to do something similar in `space`. 

But preferably we'd do it in a form of a PR to Squire directly: https://github.com/neilj/Squire/issues/375

limonte: Yup, sounds like a proper solution :+1:

tomholub: I went ahead and forked it at https://github.com/FlowCrypt/Squire - you can make a PR there, then use the modified Squire build here in this PR. Later you can make PR from the new fork to their original repo, too.

limonte: Great!

> ERROR: Permission to FlowCrypt/Squire.git denied to limonte.

Please add me as a collaborator to that repo, thanks!

tomholub: https://github.com/FlowCrypt/Squire/invitations done

limonte: @tomholub I updated the `squire-raw.js` accordingly to `FlowCrypt/Squire`, the PR is ready for review.

PR #2422, opened 2019-12-20
Made a test to check if percents is shown

### LINKED ISSUE #2316: ci test to make sure progress % shows in send btn when sending file
we could make a test for this - we have a test that sends larger file, it could check that `%` shows at some point in the `send_btn`.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/2315#issuecomment-563447871_

michael-volynets:
close #2316 
Time Spent: 2h

The only thing that went wrong is I had to test it using Live mock, that's why I spent much more time on it.

### DIFF
--- a/test/source/tests/tests/account.ts
+++ b/test/source/tests/tests/account.ts
@@ -45,7 +45,9 @@ export const defineConsumerAcctTests = (testVariant: TestVariant, testWithNewBro
       fileInput = await composePage.target.$('input[type=file]');
       await fileInput!.uploadFile('test/samples/large.jpg');
       await Util.sleep(2); // give it a little time to make tests less brittle
-      await ComposePageRecipe.sendAndClose(composePage);
+      await composePage.waitAndClick('@action-send', { delay: 1 });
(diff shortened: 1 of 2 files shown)


test/source/tests/tests/account.ts
```diff
@@ -45,7 +45,9 @@ export const defineConsumerAcctTests = (testVariant: TestVariant, testWithNewBro
       fileInput = await composePage.target.$('input[type=file]');
       await fileInput!.uploadFile('test/samples/large.jpg');
       await Util.sleep(2); // give it a little time to make tests less brittle
-      await ComposePageRecipe.sendAndClose(composePage);
+      await composePage.waitAndClick('@action-send', { delay: 1 });
+      await composePage.page.waitForFunction(() => document.querySelector('[data-test="action-send"]')!.textContent!.includes('%'));
```
tomholub: This could end up being flaky - it would be better if we had a function like await `await frame.waitForSelectorWithText(`@selector`, 'some text')`.

Let's see how it goes.

PR #2423, opened 2019-12-21
Editor overflow

### LINKED ISSUE #2420: Compose window gets overflowed when pasting or typing lots of text, can't see borders
Steps to reproduce:
1. Open Compose window
2. Paste several times a large amount of text or start typing only new lines
3. You will see that the top and bottom parts of the composer is getting hidden;

Expected result:
The text is shown between the borders. Composer should have original height.

image

limonte:
Fixes #2420

### DIFF
--- a/extension/css/cryptup.css
+++ b/extension/css/cryptup.css
@@ -1842,6 +1842,7 @@ table#compose td.text div.text_container {
   display: flex;
   flex-direction: column;
   height: 100%;
+  max-height: -webkit-fill-available; /* https://github.com/FlowCrypt/flowcrypt-browser/issues/2420 */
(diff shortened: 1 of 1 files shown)


tomholub:
It appears that the failures are legit - tests will fail if certain elements are expected to be in view but are not.

Look up `tr1me2t0jz` in your email to find more debug info / screenshots about the failure.

limonte:
> It appears that the failures are legit - tests will fail if certain elements are expected to be in view but are not.
> 
> Look up `tr1me2t0jz` in your email to find more debug info / screenshots about the failure.

Wow, that's great! I wasn't expecting the change to affect anything. Good that tests are covering us!

Issue #2416, opened 2019-12-20
"Send btn" is blocked when canceling PP modal


michael-volynets:
Steps to reproduce:
1. Forget all Passphrases stored in a browser
2. Open & fill compose window
3. Click `Send` button.
4. You should see `Enter Passphrase` modal dialog, click `later` 
5. The `send` button gets blocked and you can't send the message anymore.

Screenshot:
![image](https://user-images.githubusercontent.com/49104848/71278264-c14aee00-235f-11ea-91f2-f981102136e6.png)

Expected Result:
Send button isn't blocked when clicking `later` button in the modal winow

michael-volynets:
The same when you are trying to download an encrypted attachment.
When you click on the `Download` button after that you cancel modal dialog the `Download` button won't be clickable anymore.

tomholub:
You could reset the send btn state after you send the `BrowserMsg` to open pp dialog. Then you could programmatically `.click()` it when pass phrase becomes available.

michael-volynets:
Actually, the issue is somewhere deeper, because I found a listener for resetting the button in `composer-storage.ts` file.
```TS
  initActions = () => {
    BrowserMsg.addListener('passphrase_entry', async ({ entered }: Bm.PassphraseEntry) => {
      if (!entered) {
        clearInterval(this.passphraseInterval);
        this.composer.sendBtn.resetSendBtn();
      }
    });
  }
```

tomholub:
I see

michael-volynets:
Composer window is fixed. 
I started fixing AttachmentView.
I will also check other elements/pages where this modal dialog is used, I think this issue may occur in other places.

PR #2424, opened 2019-12-21
Issue 2416 sendbtn is blocked when canceling pp modal

michael-volynets:
Issue #2416 
Time Spent: 1h 50m
I fixed those bugs and also added a new thing called `PromiseCencellationToken`, I should have added it because I found some kind of memory leak. In `attachment.ts` and `backup.ts` when we called `Store.waitUntilPassphraseChanged` and clicked `Later` button the cycle that checks if PP changed was working which is bad because how many times I opened modal dialog and clicked `Later` - how many times the function was running continuously.
This `PromiseCencellationToken` fixes that problem.

### DIFF
--- a/extension/chrome/elements/attachment.ts
+++ b/extension/chrome/elements/attachment.ts
@@ -41,7 +41,7 @@ View.run(class AttachmentDownloadView extends View {
 
   constructor() {
     super();
-    const uncheckedUrlParams = Url.parse(['acctEmail', 'msgId', 'attId', 'name', 'type', 'size', 'url', 'parentTabId', 'content', 'decrypted', 'frameId', 'isEncrypted', 'tabId']);
+    const uncheckedUrlParams = Url.parse(['acctEmail', 'msgId', 'attId', 'name', 'type', 'size', 'url', 'parentTabId', 'content', 'decrypted', 'frameId', 'isEncrypted']);
     this.acctEmail = Assert.urlParamRequire.string(uncheckedUrlParams, 'acctEmail');
     this.parentTabId = Assert.urlParamRequire.string(uncheckedUrlParams, 'parentTabId');
     this.frameId = Assert.urlParamRequire.string(uncheckedUrlParams, 'frameId');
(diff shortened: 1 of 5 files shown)


extension/chrome/elements/attachment.ts
```diff
@@ -36,10 +36,12 @@ View.run(class AttachmentDownloadView extends View {
   private originalButtonHTML: string | undefined;
   private canClickOnAtt: boolean = false;
   private downloadInProgress = false;
+  private cancellationToken: PromiseCancellationToken = { cancel: false };
+  private tabId!: string;
 
   constructor() {
     super();
-    const uncheckedUrlParams = Url.parse(['acctEmail', 'msgId', 'attId', 'name', 'type', 'size', 'url', 'parentTabId', 'content', 'decrypted', 'frameId', 'isEncrypted']);
+    const uncheckedUrlParams = Url.parse(['acctEmail', 'msgId', 'attId', 'name', 'type', 'size', 'url', 'parentTabId', 'content', 'decrypted', 'frameId', 'isEncrypted', 'tabId']);
```
tomholub: Extra unused `tabId`

extension/chrome/elements/attachment.ts
```diff
@@ -204,7 +214,14 @@ View.run(class AttachmentDownloadView extends View {
       Browser.saveToDownloads(new Att({ name: result.filename, type: this.att.type, data: result.content }), $('body'));
     } else if (result.error.type === DecryptErrTypes.needPassphrase) {
       BrowserMsg.send.passphraseDialog(this.parentTabId, { type: 'attachment', longids: result.longids.needPassphrase });
-      await Store.waitUntilPassphraseChanged(this.acctEmail, result.longids.needPassphrase);
+      try {
+        await Store.waitUntilPassphraseChanged(this.acctEmail, result.longids.needPassphrase, 1000, this.cancellationToken);
```
tomholub: This could just return true or false, no need to throw

PR #2429, opened 2019-12-21
Issue 2427 issue when footer doubled

### LINKED ISSUE #2427: The footer (signature) is duplicated when replying
This issue is happening in Chrome only:

michael-volynets:
close #2427,
Time Spent: 1h
I couldn't test it because it happens only when you changed your footer in settings and then open a reply box. But I corrected the test so now it will check if there is no footer before you click on `triple dot` button


tomholub:
@michael-volynets it's failing your own test

image

I would recommend to create a method called `composePage.waitForContent(selector, content)` that samples the text (using our other standard methods) every 10ms, like you did originally.

tomholub:
> @michael-volynets it's failing your own test
> 
> image
> 
> I would recommend to create a method called `composePage.waitForContent(selector, content)` that samples the text (using our other standard methods) every 10ms, like you did originally.

I'll fix it, it blocks other branches too

PR #2435, opened 2019-12-23
Issue 2075 japanese charset

### LINKED ISSUE #2075: Japanese charset ISO-2022-JP garbled
> (uninstall feedback) Japanese (charset="ISO-2022-JP") e-mail message is garbling with this extension.

Asked user to send a sample, worst case we could probably create one.

michael-volynets:
close #2075,
Time Spent: 5h

I tested other Cyrillic/Japanese encodings and the result is that:
All Cyrillic encodings work well.
But `EUC_JP` and `Shift_JIS` Japanese encodings aren't working at all. I tested if `TextDecoder` decodes them the same way as `ISO-2022-JP` and the result is that it does. I haven't implemented that in this PR because maybe it would be better to watch how `TextDecoder` will work with `ISO-2022-JP` in release and if everything is okay we could implement `EUC_JP`, `Shift_JIS` charsets the same way.


test/source/platform/util.ts
```diff
@@ -15,3 +16,8 @@ export const base64encode = (binary: string): string => {
 export const base64decode = (b64tr: string): string => {
   return Buffer.from(b64tr, 'base64').toString('binary');
 };
+
+export const iso2022jpToUtf = (content: Buf) => {
+  const decoder = new TextDecoder();
```
tomholub: This code runs under Node, and wouldn't work anyway. Better to put in `throw new Error('iso2022jpToUtf not implemented on node')`

extension/js/common/platform/util.ts
```diff
@@ -32,3 +34,8 @@ export const moveElementInArray = <T>(arr: Array<T>, oldIndex: number, newIndex:
   arr.splice(newIndex, 0, arr.splice(oldIndex, 1)[0]);
   return arr;
 };
+
+export const iso2022jpToUtf = (content: Buf) => {
+  const decoder = new TextDecoder();
+  return decoder.decode(content);
```
tomholub: Is it guessing the encoding? I think it would be better to tell it which encoding to decode.

michael-volynets: Yes, it's guessing now,
I will take a look at how it will behave
Thanks!

test/source/platform/util.ts
michael-volynets: Thanks

michael-volynets:
>  Please add a test for this, using the tests .js file.

Did you mean `tests.json` file?

michael-volynets:
I just realized that the solution doesn't work with `text/html` encrypted messages. 
I think the reason is that `TextDecoder` think that the HTML is just UTF-8 message and even doesn't try to find encoded characters in `ISO-2022-JP` and decode them.
I will investigate it a little but for now, I have no idea what we could do with that.

michael-volynets:
Also, please merge my PR in `flowcrypt-bin` repository.
Thanks

michael-volynets:
+ 2 hours

test/tests.json
```diff
@@ -1,5 +1,12 @@
 {
   "messages": [
+    {
+      "name": "[enigmail] encrypted iso-2022-jp, plain text",
+      "content": [
+        "ゾし逸現飲"
+      ],
+      "params": "?frameId=none&msgId=16f431a0b9056562&senderEmail=michael.flowcrypt2%40gmail.com&isOutgoing=___cu_false___&acctEmail=flowcrypt.compatibility%40gmail.com&parentTabId=0"
```
tomholub: You can include the message directly using `message=`, so that it's not loaded from the mock. That way it will not rely on mock data.

extension/js/common/platform/util.ts
```diff
@@ -32,3 +34,11 @@ export const moveElementInArray = <T>(arr: Array<T>, oldIndex: number, newIndex:
   arr.splice(newIndex, 0, arr.splice(oldIndex, 1)[0]);
   return arr;
 };
+
+export const iso2022jpToUtf = (content: Buf) => {
+  if (!TextDecoder) {
+    throw new Error('iso2022jpToUtf not implemented on node.');
+  }
+  const decoder = new TextDecoder('utf-8');
```
tomholub: Shouldn't it say `iso-2022-jp`?

test/source/mock.ts
```diff
@@ -13,7 +13,7 @@ export const mock = async (logger: (line: string) => void) => {
       return; // missing mock data, not yet used
     }
     const filename = `${email.replace(/[^a-z0-9]+/g, '')}.json`;
-    const url = `https://github.com/FlowCrypt/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
+    const url = `https://github.com/michael-volynets/flowcrypt-bin/raw/master/gmail-mock-data/${filename}`;
```
tomholub: Will not be needed, can include message direclty

extension/js/common/platform/util.ts
```diff
@@ -32,3 +34,11 @@ export const moveElementInArray = <T>(arr: Array<T>, oldIndex: number, newIndex:
   arr.splice(newIndex, 0, arr.splice(oldIndex, 1)[0]);
   return arr;
 };
+
+export const iso2022jpToUtf = (content: Buf) => {
+  if (!TextDecoder) {
```
tomholub: You have edited `extension/js/common/platform/util.ts` which runs in the browser. My earlier comment was about `test/source/platform/util.ts` which runs in node.

That's why they are under `platform` folder, the files are **different** on each platform.

Therefore, the error message is appropriate for the Node implementation (where it should be left unimplemented and not throw).

For browser, you would say `Your browser does is not supported (missing TextDecoder)`.

test/source/mock.ts
michael-volynets: Sorry, I forgot to remove that.

extension/js/common/platform/util.ts
michael-volynets: Oh, sorry, I thought that they are the same files. 
I will change that.
Thanks

extension/js/common/platform/util.ts
michael-volynets: Yes, it should. I originally thought like you but that code was giving me a broken result.
I just researched it a little more and it's working now but I don't understand why.
Let me explain to you what I did:
Decoding result is broken:
```TS
const buf = Buf.fromUint8(node.content);
const decoder = new TextDecoder('iso-2022-jp');
return decoder.decode(content);
```
Decoding result is good:
```TS
const buf = Buf.fromRawBytesStr(node.rawContent!)
const decoder = new TextDecoder('iso-2022-jp');
return decoder.decode(content);
```

The difference is how I get `buf`. In the first example, I'm getting using `Buf.fromUint8(node.content);`, in the second `Buf.fromRawBytesStr(node.rawContent!)`.
But I don't understand why this affects decoding, I expected the same results.

test/tests.json
michael-volynets: Thanks, I didn't notice that.

extension/js/common/platform/util.ts
tomholub: `rawContent` is my own addition to mime-decoder.js, exactly because I had issues with `content`.

tomholub:
Thanks, happy to have this resolved.

Issue #2358, opened 2019-12-12
eslint rules


michael-volynets:
We could add `Prettier` (https://github.com/prettier/prettier-vscode) VSCode extension to make it easier to format our files.
But first, we should Agree with what style of code we should use.

As I see now, we are using this code style:
```TS
  // We use indent 2 spaces

  method() { // before every method/lambda expression should be new line
    // inside a method we don't use any new lines
    const a = 0;
    if (<condition>) { // when using 'if', 'for', 'while' operators we always use these brackets '{', '}'
      <code>
    }
  }
```

And also we should add everything that described in `tslint.yaml`

tomholub:
Yes, with 1 empty line around methods

```
class Class {
// 1 empty line
  method = () =>  {
  }  
// 1 empty line
  method = () =>  {
  }  
// 1 empty line
}
```
enforce if/for/when brackets - yes
no blank lines in methods - yes
enforce semicolon

For the future:
alphabetic ordering of imports

tomholub:
ordering of methods:
 - constructor should always be the first method
 - public methods should follow
 - then protected methods
 - private methods should only last

michael-volynets:
Yes, and also I think we should have the same with properties:
1. static properties
2. private static properties.
3. public properties.
4. protected properties
5. private properties
6. constructor
7. and the methods as you described above

michael-volynets:
I just noticed something about sorting using `Prettier`.
They can't do this because it sometimes can be unsafe. They described it in this issue: (https://github.com/prettier/prettier/issues/5412)

I think I could find another extension that does what we need. But we will have to fully test the extension to make sure it won't break anything.

tomholub:
We don't need an auto-fix for this, just a linter. eslint can probably do it.

michael-volynets:
Got it, yes, I hope it does

tomholub:
For now un-assigning both of us

tomholub:
Once it gets assigned to somebody, the goal will be to add a linter that covers at some (small part) of what was mentioned above, along with a refactor that makes the code conform to the new rule.

Best to do it one by one (1 new linter rule per pr) so that it's easier to review for bugs.

tomholub:
@michael-volynets what did you use to autofix the import order and method order? When I auto-import, it always seems to put it in the wrong place, and since it doesn't get autofixed, it can get quite frustrating to adjust manually.

tomholub:
[...]

PR #2438, opened 2019-12-23
Issue 2358 file formatting enforce explicit visibility

michael-volynets:
Depends on #2437 
Issue #2358
Time Spent:

I think it's good practice to always write modifiers because:
1) In different programming languages, accessibility by default may be different, for example, in C# it's `private` but in `Typescript` it's public
2) We sometimes write them and sometimes don't. 
Maybe it's a little bit more of code but I think it worths that.  

I also have done that using `Auto-fix` so I have to go through the code firstly

### DIFF
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -69,7 +69,7 @@ export class ComposeView extends View {
     openpgp.initWorker({ path: '/lib/openpgp.worker.js' });
   }
 
-  render = async () => {
+  public render = async () => {
     this.storage = await Store.getAcct(this.acctEmail, ['google_token_scopes', 'addresses', 'sendAs', 'email_provider',
       'hide_message_password', 'drafts_reply']);
     this.tabId = await BrowserMsg.requiredTabId();
@@ -87,7 +87,7 @@ export class ComposeView extends View {
     this.composer = new Composer(this);
   }
 
-  setHandlers = () => {
+  public setHandlers = () => {
     // all handled in Composer
   }
 
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -2,21 +2,22 @@
 
 'use strict';
 
-import { Store, AccountStoreExtension, Scopes, AccountStore } from '../../js/common/platform/store.js';
-import { Ui } from '../../js/common/browser/ui.js';
-import { Composer } from './composer/composer.js';
-import { BrowserMsg } from '../../js/common/browser/browser-msg.js';
-import { openpgp } from '../../js/common/core/pgp.js';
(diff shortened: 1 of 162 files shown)


extension/chrome/elements/composer/composer-errs.ts
```diff
@@ -25,7 +25,7 @@ export class ComposerErrs extends ComposerComponent {
 
   private debugId = Str.sloppyRandom();
 
-  initActions = () => {
+  public initActions = () => {
```
michael-volynets: @tomholub, maybe, it would be good to make `initActions` optional method?
What do you think?

extension/chrome/elements/pgp_block.ts
```diff
@@ -78,7 +78,7 @@ export class PgpBlockView extends View {
     }
   }
 
-  setHandlers = () => {
+  public setHandlers = () => {
```
michael-volynets: @tomholub, the same with `setHandlers`

extension/chrome/popups/default.ts
```diff
@@ -44,7 +44,7 @@ View.run(class DefaultPopupView extends View {
     }
   }
 
-  setHandlers = () => {
+  public setHandlers = () => {
```
michael-volynets: one more

extension/chrome/elements/composer/composer-atts.ts
```diff
@@ -20,11 +20,11 @@ export class ComposerAtts extends ComposerComponent {
     this.attach = new AttUI(() => this.getMaxAttSizeAndOversizeNotice());
   }
 
-  initActions = () => {
+  public initActions = () => {
```
michael-volynets: +1

extension/chrome/elements/compose.ts
```diff
@@ -87,7 +87,7 @@ export class ComposeView extends View {
     this.composer = new Composer(this);
   }
 
-  setHandlers = () => {
+  public setHandlers = () => {
```
michael-volynets: +1

michael-volynets:
LGTM

extension/chrome/elements/composer/composer-errs.ts
tomholub: I'd keep it required, there should be relatively few that don't actually have it declared, and I think it's good to have a place where the handlers would go if needed.

tomholub: But I'll probably eventually refactor it like view modules one day

tomholub:
For the empty handler methods, I'd leave them there for consistency, but thanks for asking

extension/chrome/elements/composer/composer-errs.ts
michael-volynets: Got it, thanks!

PR #2451, opened 2020-01-02
Issue #2442, refactored 'getNodeContentAsUtfStr' method

### LINKED ISSUE #2442: ISO-2022-JP TextDecoder html
> I just realized that the solution doesn't work with `text/html` encrypted messages. 
> I think the reason is that `TextDecoder` think that the HTML is just UTF-8 message and even doesn't try to find encoded characters in `ISO-2022-JP` and decode them.
> I will investigate it a little but for now, I have no idea what we could do with that now.

_Originally posted by @michael-volynets in https://github.com/FlowCrypt/flowcrypt-browser/pull/2435#issuecomment-569098604_

----------

I think this this cold be because you originally did not pass encoding when initializing `TextDecoder`, so it's left guessing. If you pass it in, maybe it will work for html too.

michael-volynets:
Closes #2442,
Time Spent: 3h 30m

The solution was pretty easy. 
Most of the time I spent on exploring/testing how I can fix the issue.


extension/js/common/core/mime.ts
```diff
@@ -343,19 +343,20 @@ export class Mime {
   }
 
   private static getNodeContentAsUtfStr = (node: MimeParserNode): string => {
+    let resultBuf = Buf.fromRawBytesStr(node.rawContent!);
```
tomholub: The solution looks good but it appears you are decoding the result twice in some situations, which is not necessary. Once here, and once again if any of the IFs fit. I would follow the original pattern to avoid unnecessary double decoding.

michael-volynets: Yes, thanks,

I will change code in another way so it won't double decode that.

michael-volynets:
Okay.
I also was thinking about a test but forgot to ask you.

michael-volynets:
+1h

PR #2475, opened 2020-01-09
Issue upload integrity check #1838

michael-volynets:
Issue #1838 
Here I was working on integrity checking. 
I was trying to fix it on our front-end, but I suppose there is should be back-end work.

It's not finished and some names may not be true, example:
`md5Encode`, it should have returned an `MD5` value but it returns base64 of the `MD5` value.


tomholub:
closing as stale, somebody will have a second look later

PR #2445, opened 2019-12-28
Issue 2398 make tests inline images

michael-volynets:
Issue: #2398
Time Spent: about 10hrs

It takes much more than I expected,
At the moment, I only finished writing test for checking if images aren't present in a footer, and if images render correctly when saving/rendering draft.
The best way to test drafts I found is to try to create a draft with an image and render that draft in one test so we can make sure that everything is working. But while working on it I faced a lot of issues.
The first one was that it didn't work (I mean I couldn't have seen any rich text elements (e.g. link, images)) so I fixed that in my first commit.
The next issue is I have to somehow store these newly created drafts in our `mock-data.ts` file, it gave me some troubles to map the data from the request to the `GmailMsg` / `DraftMsg`.
The third issue I had to deal with is there is no way to paste `image/*` using a clipboard in puppeteer, so I added `<input type="hidden" />` only in `debug` when I put a base64 test of image and squire adds that image to the textbox.

When I finish writing tests for encrypted/signed messages with images I will make this PR ready for review

PR #2477, opened 2020-01-10
Remove .vscode/extensions.json

limonte:
We have the same + `sort-imports` in https://github.com/FlowCrypt/flowcrypt-browser/blob/master/flowcrypt-browser.code-workspace#L36 (added in https://github.com/FlowCrypt/flowcrypt-browser/pull/2445/files#diff-b1f11e9665acbca9af0b6c5a772bce38)


tomholub:
For whatever reason I could not find it there. But now I do see it.

PR #2483, opened 2020-01-16
Add selector for encryption key type to Set Up process

### LINKED ISSUE #2159: ecc keys by default
This would mostly relieve the need for performance improvements like #1976 

Should come with an obvious way to choose another type of key

Options (first is default)
 - `curve25512` (shown to user as ECC `Curve25519`)
 - `rsa2048` (RSA 2048bit)
 - `rsa4096` (RSA 4096bit)

Then this string can be put into `Pgp.key.generate`

limonte:
Closes #2159 

@tomholub please let me know if I made some dummy mistakes in my PR, this is the first time I'm working with encryption keys in FlowCrypt.

Also, I'm not sure how shoudl I test this functionality, any hints would be very welcome.

Thanks in advance!


extension/chrome/settings/setup.htm
```diff
@@ -217,6 +228,17 @@ <h1></h1>
             </label>
           </div>
 
+          <div class="line left">
```
tomholub: When importing a key, you are working with a key that was already created before. Therefore there is no reason to ask the user what type of key they want - we're not creating anything.

extension/js/common/core/pgp-key.ts
```diff
@@ -36,6 +36,8 @@ export interface PrvKeyInfo {
   parsed?: OpenPGP.key.Key;     // only for internal use in this file
 }
 
+export type KeyType = 'curve25519' | 'rsa2048' | 'rsa4096';
```
tomholub: I would call this `KeyAlgo` for clarity.

extension/chrome/settings/setup/setup-import-key.ts
```diff
@@ -21,6 +22,7 @@ export class SetupImportKeyModule {
       return;
     }
     const options: SetupOptions = {
+      key_type: $('#step_2a_manual_enter .key_type').val() as KeyType,
```
tomholub: Not needed when importing, not generating a key

extension/chrome/settings/setup/setup-recover-key.ts
```diff
@@ -60,6 +60,7 @@ export class SetupRecoverKeyModule {
       const options: SetupOptions = {
         submit_main: false, // todo - reevaluate submitting when recovering
         submit_all: false,
+        key_type: 'rsa4096', // TODO(@limonte): ask Tom is this right?
```
tomholub: Same here, importing key from a backup - not creating any.

extension/chrome/settings/setup/setup-create-key.ts
```diff
@@ -66,7 +67,7 @@ export class SetupCreateKeyModule {
     await Settings.forbidAndRefreshPageIfCannot('CREATE_KEYS', this.view.rules!);
     const { full_name } = await Store.getAcct(this.view.acctEmail, ['full_name']);
     try {
-      const key = await PgpKey.create([{ name: full_name || '', email: this.view.acctEmail }], 'rsa4096', options.passphrase); // todo - add all addresses?
+      const key = await PgpKey.create([{ name: full_name || '', email: this.view.acctEmail }], options.key_type, options.passphrase); // todo - add all addresses?
```
tomholub: Since it's only needed for creating keys, you could probably grab the value from DOM either in this function, or add KeyAlgo as a second argument to this method, instead of including it in `SetupOptions`. If you think it's indeed better to keep it in SetupOptions, then you'd have to make it optional (needed only when creating).

The cleanest would probably be to change method sig to `public createSaveKeyPair = async (options: SetupOptions, keyAlgo: KeyAlgo)`

extension/chrome/settings/setup.htm
limonte: Yup, that was the dummy mistake I was suspecting. Thanks!

extension/js/common/core/pgp-key.ts
limonte: Done :+1:

extension/chrome/settings/setup/setup-import-key.ts
limonte: Done :+:1

extension/chrome/settings/setup/setup-recover-key.ts
limonte: Done :+1:

extension/chrome/settings/setup/setup-create-key.ts
limonte: The cleanest also seems to be the most simple and readable one, done :+1:

limonte:
> I don't think we'll be making automated tests for this particular feature, although technically we could.
> 
> To test it manually, you can have a look at the created key in settings and paste it into `Experimental -> Test OpenPGP Compatibility`. It should tell you the key algorithms there.

Tested manually, all is well :+1: 

> (20) PK 0 > SK 0 > Algo: [-] ecdh

tomholub:
@limonte 

It's picking up on the sort-imports fails. Assuming you have removed that extension. Then we should probably remove it from the tests, too. I'll take a look.

image

tomholub:
When people leave it on default, the keygen should be a lot faster compared to RSA, too.

PR #2488, opened 2020-01-17
Await the result of Message.prototype.sign()

### LINKED ISSUE #2487: TypeError: msgOrVerResults.verify is not a function
When trying to use OpenPGP key compatibility test with a private key, I'm getting this issue:

> (17) PK 0 > Sign/Verify test: [-] sign msg ok|verify failed: FlowCrypt had trouble verifying this message **(TypeError: msgOrVerResults.verify is not a function)**  

_Originally posted by @limonte in https://github.com/FlowCrypt/flowcrypt-browser/issues/1578#issuecomment-575663401_

limonte:
Closes #2487

`Message.prototype.sign()` is returning a Promise:

image

### DIFF
--- a/extension/chrome/settings/modules/compatibility.ts
+++ b/extension/chrome/settings/modules/compatibility.ts
@@ -172,7 +172,7 @@ View.run(class CompatibilityView extends View {
       if (!key.isFullyDecrypted()) {
         return 'skiped, not fully decrypted';
       }
-      const signedMessage = openpgp.message.fromText(this.encryptionText).sign([key]);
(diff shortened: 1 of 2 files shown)


limonte:
@tomholub this was a quick fix so I pushed it bypassing the queue, but if we also need a test for this, please create another issue so it can be prioritized among other tasks.

tomholub:
> @tomholub this was a quick fix so I pushed it bypassing the queue, but if we also need a test for this, please create another issue so it can be prioritized among other tasks.

That makes sense, you should have flexibility to apply small fixes liberally.

tomholub:
Test won't be needed

PR #2492, opened 2020-01-19
Test your private key compatibility

### LINKED ISSUE #1578: there should be a way to test compatibillity during setup
One way could be to add a footer to SweetAlert2 error modals on import setup page (and settings -> import page). The footer could say `Test your private key compatibility`, a link which should lead to chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/settings/modules/compatability.htm or ff alternative

should only apply to the types of errors that seem like a compatibility issue

limonte:
Closes #1578 

@tomholub two questions

1. From the issue: "should only apply to the types of errors that seem like a compatibility issue". I added a footer to all modals. Please let me know which modals are not for "compatibility issues", I'll remove the footer from them.

2. Could you confirm that an automated test is needed for this functionality and if yes, then which steps would you like to be included in that test? 

Thanks!

### DIFF
--- a/extension/chrome/settings/modules/my_key_update.ts
+++ b/extension/chrome/settings/modules/my_key_update.ts
@@ -69,11 +69,11 @@ View.run(class MyKeyUpdateView extends View {
     if (typeof uddatedKey === 'undefined') {
       await Ui.modal.warning(Lang.setup.keyFormattedWell(this.prvHeaders.begin, String(this.prvHeaders.end)), Ui.testCompatibilityLink);
     } else if (uddatedKey.isPublic()) {
-      await Ui.modal.warning('This was a public key. Please insert a private key instead. It\'s a block of text starting with "' + this.prvHeaders.begin + '"', Ui.testCompatibilityLink);
+      await Ui.modal.warning('This was a public key. Please insert a private key instead. It\'s a block of text starting with "' + this.prvHeaders.begin + '"');
     } else if (await PgpKey.fingerprint(uddatedKey) !== await PgpKey.fingerprint(this.primaryKi!.public)) {
-      await Ui.modal.warning(`This key ${await PgpKey.longid(uddatedKey)} does not match your current key ${this.primaryKi!.longid}`, Ui.testCompatibilityLink);
+      await Ui.modal.warning(`This key ${await PgpKey.longid(uddatedKey)} does not match your current key ${this.primaryKi!.longid}`);
     } else if (await PgpKey.decrypt(uddatedKey, uddatedKeyPassphrase) !== true) {
-      await Ui.modal.error('The pass phrase does not match.\n\nPlease enter pass phrase of the newly updated key.', false, Ui.testCompatibilityLink);
+      await Ui.modal.error('The pass phrase does not match.\n\nPlease enter pass phrase of the newly updated key.');
     } else {
       if (await uddatedKey.getEncryptionKey()) {
         await this.storeUpdatedKeyAndPassphrase(uddatedKeyEncrypted, uddatedKeyPassphrase);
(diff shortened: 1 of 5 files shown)


extension/chrome/settings/modules/my_key_update.ts
```diff
@@ -67,13 +67,13 @@ View.run(class MyKeyUpdateView extends View {
     const { keys: [uddatedKeyEncrypted] } = await openpgp.key.readArmored(String(this.inputPrivateKey.val()));
     const uddatedKeyPassphrase = String($('.input_passphrase').val());
     if (typeof uddatedKey === 'undefined') {
-      await Ui.modal.warning(Lang.setup.keyFormattedWell(this.prvHeaders.begin, String(this.prvHeaders.end)));
+      await Ui.modal.warning(Lang.setup.keyFormattedWell(this.prvHeaders.begin, String(this.prvHeaders.end)), Ui.testCompatibilityLink);
     } else if (uddatedKey.isPublic()) {
-      await Ui.modal.warning('This was a public key. Please insert a private key instead. It\'s a block of text starting with "' + this.prvHeaders.begin + '"');
+      await Ui.modal.warning('This was a public key. Please insert a private key instead. It\'s a block of text starting with "' + this.prvHeaders.begin + '"', Ui.testCompatibilityLink);
```
tomholub: Also remove here

extension/chrome/settings/modules/my_key_update.ts
```diff
@@ -67,13 +67,13 @@ View.run(class MyKeyUpdateView extends View {
...
     } else if (await PgpKey.fingerprint(uddatedKey) !== await PgpKey.fingerprint(this.primaryKi!.public)) {
-      await Ui.modal.warning(`This key ${await PgpKey.longid(uddatedKey)} does not match your current key ${this.primaryKi!.longid}`);
+      await Ui.modal.warning(`This key ${await PgpKey.longid(uddatedKey)} does not match your current key ${this.primaryKi!.longid}`, Ui.testCompatibilityLink);
```
tomholub: Also

extension/chrome/settings/modules/my_key_update.ts
```diff
@@ -67,13 +67,13 @@ View.run(class MyKeyUpdateView extends View {
...
     } else if (await PgpKey.decrypt(uddatedKey, uddatedKeyPassphrase) !== true) {
-      await Ui.modal.error('The pass phrase does not match.\n\nPlease enter pass phrase of the newly updated key.');
+      await Ui.modal.error('The pass phrase does not match.\n\nPlease enter pass phrase of the newly updated key.', false, Ui.testCompatibilityLink);
```
tomholub: Also

extension/js/common/browser/ui.ts
```diff
@@ -210,6 +212,8 @@ export class Ui {
     },
   };
 
+  public static testCompatibilityLink = '<a href="/chrome/settings/modules/compatibility.htm" target="_blank">Test your private key compatibility</a>';
```
tomholub: Let's rename to `Test your OpenPGP key compatibility` (in case we use it for testing public keys too, which we may)

extension/js/common/browser/ui.ts
```diff
@@ -158,10 +159,11 @@ export class Ui {
         },
       });
     },
-    error: async (text: string, isHTML: boolean = false): Promise<void> => {
+    error: async (text: string, isHTML: boolean = false, footer?: string): Promise<void> => {
       text = isHTML ? Xss.htmlSanitize(text) : Xss.escape(text).replace(/\n/g, '<br>');
       await Swal.fire({
         html: `<span class="red">${text}</span>`,
+        footer: footer ? Xss.htmlSanitize(footer) : '',
```
tomholub: Thanks for the sanitize!

tomholub:
Also notice failing code style

extension/chrome/settings/modules/my_key_update.ts
limonte: Done :+1:

extension/chrome/settings/modules/my_key_update.ts
limonte: Done :+1:

extension/chrome/settings/modules/my_key_update.ts
limonte: Done :+1:

extension/js/common/browser/ui.ts
limonte: Done :+1:

tomholub:
> Could you confirm that an automated test is needed for this functionality and if yes, then which steps would you like to be included in that test?

It should be fairly rare. It would be nice to have a test, but not certain it's worth it. If you want to write one, feel free. It would verify that the link shows up, that clicking it shows the compat page, and that compat page spits out some expected strings when the key is pasted.

PR #2494, opened 2020-01-20
Close original draft window once "open draft" is clicked

### LINKED ISSUE #2041: close original draft window once I click to open draft
1) make a draft of new encrypted message
2) open the draft in drafts folder
3) click the button to open our own UI

Now the original gmail draft is behind it, ideally it would auto-close when clicked

part of #143

limonte:
Closes #2041

@tomholub do we need test for this? If so, then it should be live Gmail test, right?


tomholub:
> Closes #2041
> 
> @tomholub do we need test for this? If so, then it should be live Gmail test, right?

If we did a test, it would go in Gmail tests. But I'd say the functionality is not as crucial, so we can probably skip a test unless you think we should have one.

PR #2495, opened 2020-01-23
Add note about the expiration of password encrypted messages

### LINKED ISSUE #1573: highlight that pwd msgs expires while normal ones don't
> (uninstall feedback) 3-day forced expiration?

The expiration only affects messages sent to people who don't have encryption set up.

> Honestly, it's more the sequence of events and method through which I found out there was an expiration date.
> If the product were to expose the three-day expiration earlier (preferably before installing it), and if it gave more details on just what happens after expiration (e.g., "after expiration, recipients without FlowCrypt or other PGP decryption extensions will need to install FlowCrypt or perform their own PGP decryption"), you wouldn't have had me scrambling to remove the extension, wondering what I had just unleashed on my system. I had discovered the application through a web search, and learned mostly through the search and through the extension details, just to give some context.

> You might be surprised to know I was even looking at that subscription model and thinking, "well, if it lives up to its promise, that might just happen." But the experience unfortunately led me to believe one thing (didn't know about expiration), to learning that the subscription would allow you to change the expiration (at this point, thought the benefit was being able to have them expire (that there was no expiration by default) and I was wondering what mechanism it used and if it broke the PGP standard somehow), to seeing that there was a 3-day expiration for encrypted emails. This last part was alarming because I don't recall seeing anything that indicated it was just for the bonus feature of non-PGP recipients being able to read it.

### LINKED ISSUE #2500: storage ignores undefined
statement like this will have no effect

```
await Store.setAcct(this.acctEmail, { subscription: undefined, uuid: undefined });
```

limonte:
Closes #1573 
// tom edit: also closes #2500

The diff of `compose.html` contains some lines which I just fixed indentation of, to ignore them during CR, please use the `?w=1` param in URL: https://github.com/FlowCrypt/flowcrypt-browser/pull/2495/files?w=1


extension/chrome/elements/compose.htm
```diff
@@ -110,6 +110,11 @@ <h1 id="header_title" data-test="header-title">New Secure Message</h1>
       </tr>
       <tr id="password_or_pubkey_container" data-test="password-or-pubkey-container">
         <td>
+          <div id="expiration_note">
+            Password encrypted messages and attachments <strong>expire in 3 days</strong>.
```
tomholub: The number should be dynamically determined. You'll probably need an API request to figure this out. The logic is as follows:
1) if user does not have authInfo saved, that means they don't have any account with us, therefore must be using the default 3 days
2) if there is `authInfo` present, then it should be used to grab user account info from backend, field `default_message_expire` (see Security screen, where the user can set this property)

By default, default_message_expire may be set to `3` but the user could have changed it to various other possibilities.

One thing, please don't make the API request while composer is loading initially. But it's probably ok to do the request as this infobox is about to render, and delay rendering of the infobox while the request is being sent. I'll leave details up to you.

extension/chrome/elements/compose.htm
```diff
@@ -110,6 +110,11 @@ <h1 id="header_title" data-test="header-title">New Secure Message</h1>
...
+            <br /><br />After expiration, recipients without FlowCrypt or other PGP decryption extensions will need to install FlowCrypt
```
tomholub: note for me - to take a second look later, may want to adjust wording

extension/chrome/elements/compose.htm
limonte: Thanks! Should we also have a link to that setting in the expiation note? Something like "You can change the expiration time __in settings__." where __in settings__ is the link to security settings.

tomholub: You could just say `Recipient will have access to this message <a href=settings>for N days</a>`. Since the user will be looking at it every time he sends a similar message, but is not expected to change it often, I'd make the settings stand out a bit less.

limonte: Done :+1:

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
```diff
@@ -52,9 +72,20 @@ export class ComposerPwdOrPubkeyContainer extends ComposerComponent {
...
+    if (this.composer.S.cached('password_or_pubkey').is(':hidden')) {
+      let messageExpire: number = 3;
+      try {
+        const authInfo = await Store.authInfo(this.view.acctEmail);
+        const response = await Backend.accountGetAndUpdateLocalStore(authInfo);
+        messageExpire = response.account.default_message_expire;
+      } catch (e) {
+        // ignore, use the default 3 days message expire
+      }
+      const pluralize = (count: number, noun: string, suffix = 's') => `${count} ${noun}${count > 1 ? suffix : ''}`; // TODO: where to move this helper?
```
limonte: @tomholub Where to move this helper?

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
```diff
@@ -16,26 +18,33 @@ export class ComposerPwdOrPubkeyContainer extends ComposerComponent {
...
-      this.showHideContainerAndColorSendBtn();
+    this.composer.S.cached('input_password').blur(async () => {
+      setTimeout(() => { // timeout here is needed so <a> will be visible once clicked
+        this.composer.S.cached('expiration_note').fadeOut();
+      }, 100);
+      await this.showHideContainerAndColorSendBtn();
     });
+    this.composer.S.cached('expiration_note').find('#expiration_note_settings_link').click(this.view.setHandler((el, e) => {
+      e.preventDefault();
+      this.composer.render.renderSettingsWithDialog('security'), this.composer.errs.handlers(`render settings dialog`);
```
tomholub: I'm surprised this is valid syntax. The `this.composer.errs.handlers("render settings dialog")` should be a parameter passed into `setHandler`.

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
tomholub: I guess you could put it in `Str.plularize`, although I would not mind if it just said `for 1 days` since I suspect very few people will pay to set it lower.

(I just checked the backend, it's 4 people, I would not worry about it)

limonte: Done :+1:

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
limonte: Oops, fixed :+1:

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
```diff
@@ -33,7 +54,7 @@ export class ComposerPwdOrPubkeyContainer extends ComposerComponent {
       this.hideMsgPwdUi(); // Hide 'Add Pasword' prompt if there are no recipients or message is not encrypted
       this.composer.sendBtn.enableBtn();
     } else if (this.composer.recipients.getRecipients().find(r => r.status === RecipientStatuses.NO_PGP)) {
-      this.showMsgPwdUiAndColorBtn();
+      this.showMsgPwdUiAndColorBtn(); // tslint:disable-line:no-floating-promises
```
limonte: @tomholub is it okay to leave this promise floating? If I handle it with `await`, then I have to turn `showHideContainerAndColorSendBtn` into `async`, and that will cause a whole bunch of changes from `()` to `async ()` in the codebase.

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
```diff
@@ -53,7 +54,7 @@ export class ComposerPwdOrPubkeyContainer extends ComposerComponent {
       this.hideMsgPwdUi(); // Hide 'Add Pasword' prompt if there are no recipients or message is not encrypted
       this.composer.sendBtn.enableBtn();
     } else if (this.composer.recipients.getRecipients().find(r => r.status === RecipientStatuses.NO_PGP)) {
-      await this.showMsgPwdUiAndColorBtn();
+      this.showMsgPwdUiAndColorBtn(); // tslint:disable-line:no-floating-promises
```
tomholub: Instead of suppressing it, you can add `.catch(Catch.reportErr)` even if you don't expect any errors to be thrown.

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
tomholub: Sorry missed your question, answered in https://github.com/FlowCrypt/flowcrypt-browser/pull/2495#discussion_r370803385

extension/chrome/elements/composer/composer-pwd-or-pubkey-container.ts
limonte: Done, thanks :+1:

limonte:
> Almost there, just error handling. Thanks!

Done :+1; let me know if there's something else to do in this PR.

tomholub:
> > Almost there, just error handling. Thanks!
> 
> Done 👍 let me know if there's something else to do in this PR.

Thanks, I'll just update the wording and merge it.

tomholub:
I've updated the code to differentiate between situations where:

1) default is used because there is no account
2) there was an error trying to find out how many days it will be

It's ugly but it communicates the error to user. Users tend to send screenshots when things go wrong, so it's useful to see what's going on, even if it says a bit more than the user wants to hear (it will save us time debugging their problems:

image


image

PR #2498, opened 2020-01-25
Ensure msgBody has dummy parent after being replaced

### LINKED ISSUE #2271: Cannot read property 'replaceChild' of null"
### Step to reproduce:

1. Open any encrypted message from Inbox
2. Click the remove Inbox label button:
image
After clicking the remove label button, this error is in console: `Error in protected function: Cannot read property 'replaceChild' of null"`
3. Click the "Undo" button in the bottom toast notification:
image

### Expected behavior

Nothing happens

### Actual behavior

The encrypted message is being re-rendered:

limonte:
Closes #2271

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -473,7 +473,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   private ensureHasParentNode = (el: JQuery<HTMLElement>) => {
     if (!el.parent().length) {
       const dummyParent = $('<div>');
-      dummyParent.append(el);
+      dummyParent.append(el); // xss-direct
     }
   }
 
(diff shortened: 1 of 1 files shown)


extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -468,6 +470,13 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
   }
 
+  private ensureHasParentNode = (el: JQuery<HTMLElement>) => {
```
limonte: @tomholub because it's a common DOM helper, should it be moved somewhere else from here?

limonte:
@tomholub what do you think about this solution?

extension/js/content_scripts/webmail/gmail-element-replacer.ts
tomholub: I can only imagine ever needing it in Gmail webmail, so I think it's placed correctly.

tomholub:
Code style tests are failing due to suspected xss. In this case it's not a concert, so you can comment it with `dummyParent.append(el); // xss-direct`

limonte:
> Code style tests are failing due to suspected xss. In this case it's not a concert, so you can comment it with `dummyParent.append(el); // xss-direct`

Done :+1:

PR #2507, opened 2020-01-28
Move diagnosePubkeys() from background script to content script

### LINKED ISSUE #2158: prep work for manifest V3 - move pgp functionality from background script to content script
First should do some prep work - move what we can from background script to content script

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/1509#issuecomment-548742536_

limonte:
Part of #2158


extension/chrome/elements/pgp_block_modules/pgp-block-error-module.ts
```diff
@@ -32,7 +32,7 @@ export class PgpBlockViewErrorModule {
   }
 
   public handlePrivateKeyMismatch = async (message: Uint8Array) => { // todo - make it work for multiple stored keys
-    const msgDiagnosis = await BrowserMsg.send.bg.await.pgpMsgDiagnosePubkeys({ privateKis: await Store.keysGet(this.view.acctEmail), message });
+    const msgDiagnosis = await PgpMsg.diagnosePubkeys({ privateKis: await Store.keysGet(this.view.acctEmail), message });
```
tomholub: This should not work - PgpMsg does not have OpenPGP.js. Instead of sending PgpMsg to bg script, you should be sending PgpMsg to parent script (typically the main content script), where OpenPGP.js should be imported if not present already.

tomholub: But since the tests worked.. could you verify if this method is being hit in tests? By just adding `throw new Error('Whoops')` ? If not, then we need to add a new test before we proceed with this further. Thanks!

tomholub: I updated the first comment to say that `pgp_block.htm` frames do not have OpenPGP.js loaded.

This is for performance - there may be 50 such frames, and it takes 300ms of compute time to load OpenPGP.js in each frame. It's a fat library.

tomholub:
A side note - very happy that you did one method first, so that we can nail approach before doing the rest. 👍

extension/chrome/elements/pgp_block_modules/pgp-block-error-module.ts
limonte: > But since the tests worked.. could you verify if this method is being hit in tests? By just adding `throw new Error('Whoops')` ? If not, then we need to add a new test before we proceed with this further. Thanks!

Now, I know which test is failing and how to hit that method manually, thanks!

tomholub:
Now it failed. I'm very, very confused. I should take a look.

tomholub:
Problem found: `pgp_block.htm` contains `<script src="/lib/openpgp.js"></script>`. It is not supposed to contain openpgp.js... I don't know how it got there, but it's not meant to be there.

Please remove it and see what happens (then you can remove the error throwing too). It's supposed to call bg script (soon just whatever parent page) for to proxy OpenPGP.js functionality.

Once you remove it from there, it should all start to make more sense.

limonte:
> Problem found: `pgp_block.htm` contains `<script src="/lib/openpgp.js"></script>`. It is not supposed to contain openpgp.js... I don't know how it got there, but it's not meant to be there.
> 
> Please remove it and see what happens (then you can remove the error throwing too). It's supposed to call bg script (soon just whatever parent page) for to proxy OpenPGP.js functionality.
> 
> Once you remove it from there, it should all start to make more sense.

Roger, thanks a lot for your guidance!

tomholub:
> > Problem found: `pgp_block.htm` contains `<script src="/lib/openpgp.js"></script>`. It is not supposed to contain openpgp.js... I don't know how it got there, but it's not meant to be there.
> > Please remove it and see what happens (then you can remove the error throwing too). It's supposed to call bg script (soon just whatever parent page) for to proxy OpenPGP.js functionality.
> > Once you remove it from there, it should all start to make more sense.
> 
> Roger, thanks a lot for your guidance!

There's a possibility that I would overlook this if I did this PR, because I would have never though to do it the way you did, because it's not supposed to work.

That is to say, there is value to your struggle :+1:

extension/js/common/core/pgp-msg.ts
```diff
@@ -230,6 +230,8 @@ export class PgpMsg {
   }
 
   public static diagnosePubkeys: PgpMsgMethod.DiagnosePubkeys = async ({ privateKis, message }) => {
+    // openpgp is undefined here
```
limonte: @tomholub I feel stupid on this task :laughing: 

Where should I import OpenPGP.js so it'll be available here?

tomholub: Any place that may be the parent of a `pgp_block`:
1) in the content script
2) in encrypted inbox page

But both should already have it:
content script:
image

inbox:
image

What is the stack trace of the debugger?

limonte:
```
Error: BrowserMsg() sendRawResponse::TypeError: Cannot read property 'message' of undefined
    at Function.BrowserMsg.jsonToErr (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/js/common/browser/browser-msg.js:318:15)
    at processRawMsgResponse (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/js/common/browser/browser-msg.js:245:35)

[callerStack]
    at chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/js/common/browser/browser-msg.js:217:120
    at new Promise (<anonymous>)
    at Function.BrowserMsg.sendAwait (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/js/common/browser/browser-msg.js:215:18)
    at Object.pgpMsgDiagnosePubkeys (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/js/common/browser/browser-msg.js:69:57)
    at PgpBlockViewErrorModule.handlePrivateKeyMismatch (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/elements/pgp_block_modules/pgp-block-error-module.js:28:62)
    at async PgpBlockViewDecryptModule.decryptAndRender (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:129:29)
    at async PgpBlockViewDecryptModule.initialize (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:63:25)
    at async PgpBlockView.render (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/elements/pgp_block.js:26:17)

url: chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/elements/pgp_block.htm?frameId=frame_OLkOOTxzNo&message=&hasPassword=___cu_false___&msgId=16f29cf85aa2cd61&senderEmail=mindoutofframe%40gmail.com&isOutgoing=___cu_false___&acctEmail=flowcrypt.compatibility%40gmail.com&parentTabId=17%3A0
```

tomholub: I'll take a look, this should have worked.

tomholub: I fixed it. There was a long-present issue with loading openpgp.js in content scripts. If you look into `flowcrypt-browser/build/chrome-consumer/js/content_scripts/webmail_bundle.js`, you'll notice that webmail bundle is really just concatenated source files with import statements commented out (poor man's bundler).

Once concatenated, this happened:

```js
// ...
const requireOpenpgp = () => {
   // ...
   return openpgp; // where I really meant global window.openpgp
}
// ...
const openpgp = requireOpenpgp();
```
The `return openpgp;` would fail because it thought I'm referring to the const that's defined below, where I actually meant the global one.

Did not not happen outside of content scripts, because in other pages `const openpgp = ` is in a separate file/module.

tomholub:
Now I see that previously, in content scripts, I was also calling into bg:

image

Now you can replace these with direct calls to openpgp.js without using `BrowserMsg`, because content script now has a functioning openpgp.js of it's own.

limonte:
Thanks a lot!

One more question: now the `pgp/mime with large attachment - mismatch` is failing with `missing expected content: Your current key cannot open this message.`.

That's happening (I guess) because of `pgp_block.htm` now depends on the content (parent) script.

In the Gmail UI `pgp_block.htm` is now rendered correctly:

image

But, if I open chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/elements/pgp_block.htm?account_email=flowcrypt.compatibility%40gmail.com&frame_id=frame_yVMKFLRDiY&message=&has_password=___cu_false___&message_id=162275c819bcbf9b&sender_email=human%40flowcrypt.com&is_outgoing=___cu_false___&parent_tab_id=0 directly it returns `Error: BrowserMsg.sendAwait(pgpMsgDiagnosePubkeys) returned(null) with lastError: (no lastError)` because there's no parent and therefore no listener for `pgpMsgDiagnosePubkeys`.

tomholub:
I suspect a bunch of test cases are failing because we do many tests by just opening the frame on its own, without a parent.

That's an interesting side effect.

There are, I think, two ways to resolve it:
1) always test `pgp_block.htm` frames in `inbox.htm` frame, which serves as the host. Because you need the flexibility to open `pgp_block.htm` of any url to perform tests, you could probably write some code in `inbox.ts` or similar place to place `tabId` somewhere in the html page during load, then the tests would pick up on this and insert a `pgp_block` iframe into the page with the appropriate `parentTabId` parameter.
2) make a separate page just for this purpose, eg `extension/chrome/dev/ci_pgp_host_page.ts` that does what is described above

I don't have a better idea about how to solve it. It's a bit hacky.

tomholub:
> That's happening (I guess) because of pgp_block.htm now depends on the content (parent) script.

Exactly correct. See comment above for some thoughts on that.

limonte:
> make a separate page just for this purpose, eg extension/chrome/dev/ci_pgp_host_page.ts that does what is described above

I'll go with this one for the sake of tests execution speed.

limonte:
@tomholub I need your help here again.

In the latest commit I added two simple filed to reproduce the issue I'm facing:

image

So, even though, there's a listener here: https://github.com/FlowCrypt/flowcrypt-browser/pull/2507/commits/f059985c474cff9203f8853201d93af7c323e01c#diff-19cb0f2d245c83d6d16b5a56571ee25fR13 it's not working for some reason. Could you give me a hint what is my mistake here?

tomholub:
> @tomholub I need your help here again.
> 
> In the latest commit I added two simple filed to reproduce the issue I'm facing:
> 
> image
> 
> So, even though, there's a listener here: [f059985#diff-19cb0f2d245c83d6d16b5a56571ee25fR13](https://github.com/FlowCrypt/flowcrypt-browser/commit/f059985c474cff9203f8853201d93af7c323e01c#diff-19cb0f2d245c83d6d16b5a56571ee25fR13) it's not working for some reason. Could you give me a hint what am I missing here?

Sorry, I totally missed your question. If I didn't reply for more than a few days, I've almost certainly managed to not notice it, please ping me!

I'll take a look.

extension/chrome/dev/ci_pgp_host_page.htm
```diff
@@ -0,0 +1,26 @@
...
+<body id="settings">
+  <pre style="text-align: left; font-family: monospace;"></pre>
+
+  <script src="/lib/purify.min.js"></script>
+  <script src="/lib/jquery.min.js"></script>
+  <script src="/lib/sweetalert2.js"></script>
+  <script src="/lib/openpgp.js"></script>
+  <script src="ci_pgp_host_page.js" type="module"></script>
+
+  <iframe src="chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/elements/pgp_block.htm?account_email=flowcrypt.compatibility%40gmail.com&frame_id=frame_yVMKFLRDiY&message=&has_password=___cu_false___&message_id=162275c819bcbf9b&sender_email=human%40flowcrypt.com&is_outgoing=___cu_false___&parentTabId=29%3A0" frameborder="0"></iframe>
```
tomholub: The problem is that `parentTabId` is a dynamic value - we need to figure out what it is before adding it to the url. I know you logged it in ci_pgp_host_page, but there is a chance that it has changed as you were testing it.

Actually, now that I'm looking at it, the tested frame does not _actually_ need to be an embedded frame. That will simplify things.

To (probably) fix this:
1) at the end of `ci_pgp_host_page.ts` do something like `$('body').append('<div data-test="tab-id">${Xss.escape(tabId)}</div>')`;
2) the test runner needs to retrieve the value with something like `hostPage.waitAll('@tab-id')` and `hostPage.read('@tab-id')`;
3) whenever opening a pgp_block during tests, it currently usually includes `parentTabId=0` or something like that. These should be removed, and instead you should add the dynamic value of whatever the host tabId actually is.
4) Then you can open the pgp_block as a separate tab just like we do today. It will be easier that way. It will work as long as it has the correct parentTabId and the "parent" tab remains open.

tomholub:
I'll do some of the recommended changes to make sure there isn't something else going on.

tomholub:
There was indeed another problem. It's fixed now, although I don't fully understand the actual issue.

chrome-extension://nhlmpmiobabndkmnhocbgomijbcmiikg/chrome/dev/ci_pgp_host_page.htm

image

test/source/tests/browser-recipe.ts
```diff
@@ -70,7 +70,8 @@ export class BrowserRecipe {
   }
 
   public static async pgpBlockVerifyDecryptedContent(t: AvaContext, browser: BrowserHandle, m: TestMessage) {
-    const pgpBlockPage = await browser.newPage(t, `chrome/elements/pgp_block.htm${m.params}`);
+    const pgpHostPage = await browser.newPage(t, `chrome/dev/ci_pgp_host_page.htm${m.params}`);
+    const pgpBlockPage = await pgpHostPage.getFrame(['pgp_block.htm']);
```
tomholub: Clever. You'll need to also likely remove `parentTabId=xxx` from the tests json file, and any other places that call this method.

limonte: thank you for the detailed feedback!

I'm now trying to understand why `ci_pgp_host_page.htm?account_email=flowcrypt.compatibility%40gmail.com&frame_id=frame_yVMKFLRDiY&message=&has_password=___cu_false___&message_id=162275c819bcbf9b&sender_email=human%40flowcrypt.com&is_outgoing=___cu_false___` works, but `ci_pgp_host_page.htm?account_email=flowcrypt.compatibility%40gmail.com&frame_id=frame_yVMKFLRDiY&message=&hasPassword=___cu_false___&msgId=16a9c109bc51687d&senderEmail=flowcrypt.compatibility%40protonmail.com&isOutgoing=___cu_false___&signature=___cu_true___` throws `TypeError: Cannot read property 'key' of undefined (common/core/pgp-key.js:34:43)`

It seems that in the second case openpgp.js isn't loaded.

tomholub:
@limonte let me know if you need any help here, or if it's getting too complicated - I can cover some of this

test/source/tests/browser-recipe.ts
limonte: @tomholub ping

limonte:
> @limonte let me know if you need any help here, or if it's getting too complicated - I can cover some of this

I do need some help here, please take a look at the unresolved conversation above ^

test/source/tests/browser-recipe.ts
tomholub: Sorry, I don't understand how I'm missing these messages.

Most likely the other test is also trying to use openpgp.js directly somehow. I'll take a look.

tomholub:
> > @limonte let me know if you need any help here, or if it's getting too complicated - I can cover some of this
> 
> I do need some help here, please take a look at the unresolved conversation above ^

It was accessing some other function directly. I've refactored it so that it no longer would.

limonte:
@tomholub I believe you were missing the messages because this PR became quite big.

I changed this PR to be a part of #2158 the rest will be done in another PR(s).

extension/chrome/dev/ci_pgp_host_page.ts
```diff
@@ -0,0 +1,22 @@
...
+Catch.try(async () => {
+  const tabId = await BrowserMsg.requiredTabId();
+
+  BrowserMsg.addListener('pgpMsgDiagnosePubkeys', PgpMsg.diagnosePubkeys);
+  BrowserMsg.listen(tabId);
+
+  let src = Env.getBaseUrl();
+  src += `/chrome/elements/pgp_block.htm?${location.search}`;
+  src += `&parentTabId=${encodeURIComponent(tabId)}`;
+  $('body').append(`<iframe src="${src}" frameborder="0"></iframe>`); // xss-direct
```
tomholub: While here it doesn't really matter, since it's just for dev, the correct way to fix it here would be `<iframe src="${Xss.escape(src)}" frameborder="0"></iframe>` with `// xss-escaped`

limonte:
> change // xss-direct to // xss-direct

Is it okay to force push in cases like this one? Made a silly typo in the commit message.

tomholub:
> > change // xss-direct to // xss-direct
> 
> Is it okay to force push in cases like this one? Made a silly typo in the commit message.

The reason I don't force push is because I'm afraid to lose my work if I make a typo/mistake during the force-pushing itself. Eg I just left a commit above and if you didn't notice it, there's a chance you may have force-pushed it away. So I'd just leave a comment if I felt I needed to clarify. But if you really like to clean it up with force-pushes (a lot of people do), it's silly to stop you - so I'll leave it up to you.

PR #2552, opened 2020-02-13
Move pgpMsgType from background script

limonte:
Part of #2158

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -27,8 +27,7 @@ type JQueryEl = JQuery<HTMLElement>;
 
 export class GmailElementReplacer implements WebmailElementReplacer {
 
-  public tabId: string;
-
+  private tabId: string;
   private gmail: Gmail;
   private recipientHasPgpCache: Dict<boolean> = {};
   private sendAs: Dict<SendAsAlias>;
@@ -105,7 +104,8 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
   }
 
-  private everything = () => {
+  private everything = async () => {
+    this.tabId = await BrowserMsg.requiredTabId();
     this.replaceArmoredBlocks();
     this.replaceAtts().catch(Catch.reportErr);
     this.replaceFcTags();
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -27,6 +27,7 @@ type JQueryEl = JQuery<HTMLElement>;
 
 export class GmailElementReplacer implements WebmailElementReplacer {
 
+  private tabId: string;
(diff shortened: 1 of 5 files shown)


extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -103,7 +104,8 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
   }
 
-  private everything = () => {
+  private everything = async () => {
+    this.tabId = await BrowserMsg.requiredTabId();
```
limonte: I feel like `everything()` is the wrong place for setting `this.tabId` because it'll be assigning it every second. Is there a better place for this line?

tomholub: This would be getting the tab id every second. Since it's an async call, it may be somewhat pricey to do it every second (plus it doesn't change).

I would try to instead change the signature of `await webmailSpecific.start(acctEmail, inject, notifications, factory, notifyMurdered);` to add `tabId` there as argument, then propagate it around so that you can use it inside `everything`.

tomholub: That's right, see https://github.com/FlowCrypt/flowcrypt-browser/pull/2552#discussion_r378788507

tomholub:
If `pgpMsgType` was not used anywhere in `pgp_block`, then we would not even have to have any call handler for it.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -103,7 +104,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
   }
 
-  private everything = () => {
+  private everything = async () => {
```
tomholub: This could be removed

extension/js/content_scripts/webmail/webmail.ts
```diff
@@ -135,6 +136,7 @@ Catch.try(async () => {
       start,
     });
 
+    BrowserMsg.addListener('pgpMsgType', (r: Bm.PgpMsgType) => PgpMsg.type({ data: Buf.fromRawBytesStr(r.rawBytesStr) }));
```
tomholub: This could be removed, IF pgpMsgType is not used in pgp_block.

limonte:
> If `pgpMsgType` was not used anywhere in `pgp_block`, then we would not even have to have any call handler for it.

Not sure if I'm getting this message right, so I'll clarify. `pgpMsgType` isn't used in `pgp_block`, does that mean that we can get rid of it?

extension/js/content_scripts/webmail/webmail.ts
limonte: Done, should `pgpMsgType` be remove from `BrowserMsg` as well?

tomholub: If it's never used, is should be removed

Issue #2051, opened 2019-09-30
automated tests for keyboard accessibility / shortcuts


tomholub:
Putting "third priority" because I will have to first introduce you to our test mechanisms, and share secrets, before you can effectively write and run tests.

tomholub:
Either way, testing these things will be very valuable, to make sure behavior doesn't break in the future.

tomholub:
What to test:
 - [ ] opening & changing settings in popover using keyboard
 - [ ] attaching file using keyboard
 - [ ] ctrl+enter sends message
 - [ ] enter pass phrase form will respond to enter (after I entered pp) or esc (to close it).

PR #2559, opened 2020-02-20
bump puppeteer to 2.1.1

limonte:
Working on #2051 I felt like it'd be good to update the `puppeteer` dependency.


tomholub:
Thanks! I tried upgrading a few times but every time I did, live gmail tests started failing because Google accounts page started behaving differently (google didn't like something about the newer puppeteer version during account verification, and thought it was a bot - blocked it).

limonte:
> Thanks! I tried upgrading a few times but every time I did, live gmail tests started failing because Google accounts page started behaving differently (google didn't like something about the newer puppeteer version during account verification, and thought it was a bot - blocked it).

Quite a lot of tests are failing, I'll skip the upgrade for now.

PR #2564, opened 2020-02-20
Bump puppeteer to 1.20.0

limonte:
Another try to upgrade to the latest version from v1.

Because for #2051 I'll most probably need `waitForFileChooser` which was introduced in 1.19.0: https://pptr.dev/#?product=Puppeteer&version=v1.19.0&show=api-class-filechooser


limonte:
Good, only one test is failing, I'll take a look

tomholub:
Doesn't look like puppeteer's fault, maybe a flake, although it did fail reliably.

limonte:
The retrieving message was happening too fast for two `waitForContent`s, after replacing with one `waitForContent` all is well.

tomholub:
Looks good, thanks!

PR #2569, opened 2020-02-22
Add RTL support

### LINKED ISSUE #1114: support for typing RTL languages (hebrew, etc)
ref https://mail.google.com/mail/u/0/#inbox/163c7e9b9e76b44f, my questions:

> 1) When you compose a message in standard gmail, will it always be RTL automatically? 
> 2) When you switch to an English keyboard, will it automatically start typing LTR?
> 3) What if you want to have one paragraph in RTL and one in LTR, would Gmail handle that seamlessly for you, or is that not possible in standard Gmail?
> 
> In general, is it switching between RTL and LTR based on the keyboard map / language that you currently use?

response:

> 1) It depends on the language I type 1st. So if, for example, I start typing in English, gmail will default to LTR. If on the other hand I start typing in Hebrew, gmail defaults to RTL.
> 2) So if I type in Hebrew it will start RTL and then when I switch to English it will switch direction properly and then switch back when I switch back to Hebrew. It work similarly when typing in English and switching in the middle to Hebrew and then back.
> 3) Yes, it does switch automatically.
> 
> Yes, in general it auto switches based on the language but Chrome itself has a way to manually switch direction (right click on a text area and select writing direction). 
> 
> BTW, I had to turn on RTL support in gmail (Setting > General):

limonte:
Closes #1114 

I'm personally very excited about this feature because even though the change is small in terms of LoC, the impact could be quite big because Arabic is the 5th most popular language in the world. And people who speak Hebrew will be excited as well, Israel has the largest number of startups per capita in the world, around one startup per 1,400 people, so people out there would appreciate our effort.

You might want to announce the RTL support in release notes, blog and in the news section of `settings/index.htm`.

TODO: 
- [x] add test for 1. typing RTL message 2. saving it to drafts 3. checking that draft contains `<div dir="rtl">`

### DIFF
--- a/test/source/tests/page-recipe/compose-page-recipe.ts
+++ b/test/source/tests/page-recipe/compose-page-recipe.ts
@@ -60,8 +60,6 @@ export class ComposePageRecipe extends PageRecipe {
       await Util.sleep(1);
       await composePageOrFrame.type('@input-subject', `Automated puppeteer test: ${subject}`);
     }
-    const body = `This is an automated puppeteer test: ${subject || '(no-subject)'}`;
-    await composePageOrFrame.type('@input-body', body);
     const sendingOpts = sendingOpt as { [key: string]: boolean | undefined };
     for (const opt of Object.keys(sendingOpts)) {
       const shouldBeTicked = sendingOpts[opt];
@@ -69,6 +67,8 @@ export class ComposePageRecipe extends PageRecipe {
         await ComposePageRecipe.setPopoverToggle(composePageOrFrame, opt as PopoverOpt, shouldBeTicked);
       }
     }
+    const body = subject?.match(/RTL/) ? 'مرحبا' : `This is an automated puppeteer test: ${subject || '(no-subject)'}`;
+    await composePageOrFrame.type('@input-body', body);
     return { subject, body };
   }
 
--- a/extension/chrome/elements/compose-modules/compose-input-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-input-module.ts
@@ -106,6 +106,9 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
 
   private handleRTL = () => {
     const checkRTL = (e: KeyboardEvent) => {
+      if (!this.isRichText()) { // RTL is supported for pgp/mime (rich text) only
+        return;
+      }
       if (e.key.length > 1) {
         return; // 'Enter', 'Space', etc.
       }
--- a/extension/chrome/elements/compose-modules/compose-input-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-input-module.ts
@@ -22,6 +22,7 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
     this.initShortcuts();
     this.resizeReplyBox();
     this.scrollIntoView();
+    this.handleRTL();
     this.squire.setConfig({ addLinks: this.isRichText() });
     if (this.view.debug) {
       this.insertDebugElements();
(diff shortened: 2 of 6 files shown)


tomholub:
Sounds good, when released, I'll mention it on the blog.

limonte:
@tomholub I'm facing the issue with drafts here. Saving to drafts is replacing `<div>`s with `<br>` which will convert the RTL message back to LTR.

Do you have thoughts on what would be the best way to resolve that?

tomholub:
> Do you have thoughts on what would be the best way to resolve that?

There are two scenarios - pgp/mime (used when sending rich text) vs pgp/inline.

With pgp/mime, it encrypts a whole mime message that contains a text/plain and text/html inside of it. When this is the case, we (should) prefer to use the text/html.

With pgp/inline, it's just text.

I don't really know what we currently do for drafts. I **think** we follow the same as if we were sending it (except we skip attachments), but I'm not sure.

So fixing this depends:
 - for pgp/inline, one way to fix it would be to go through the text line by line when converting text to html, and use <p> instead of <br> (where will somewhere be a method such as Str.toHtml or similar that we use).
 - for pgp/mime (rich text), we should make sure to use the html part and don't try to convert it from text (I think we already do this, else inline images cannot work).

tomholub:
You could have a look at how `fmtDecryptedAsSanitizedHtmlBlocks` converts decrypted text to html, and try to fix it there. You'd use the same logic to figure out if line contains RTL characters.

test/source/tests/page-recipe/compose-page-recipe.ts
```diff
@@ -60,15 +60,15 @@ export class ComposePageRecipe extends PageRecipe {
...
-    const body = `This is an automated puppeteer test: ${subject || '(no-subject)'}`;
-    await composePageOrFrame.type('@input-body', body);
     const sendingOpts = sendingOpt as { [key: string]: boolean | undefined };
     for (const opt of Object.keys(sendingOpts)) {
       const shouldBeTicked = sendingOpts[opt];
       if (typeof shouldBeTicked !== 'undefined') {
         await ComposePageRecipe.setPopoverToggle(composePageOrFrame, opt as PopoverOpt, shouldBeTicked);
       }
     }
+    const body = subject?.match(/RTL/) ? 'مرحبا' : `This is an automated puppeteer test: ${subject || '(no-subject)'}`;
```
limonte: I moved typing body after setting `richtext`, so it'll take plaintext/richtext into consideration, e.g. in `checkRTL()`

limonte:
@tomholub I believe tests are failing not because of changes from this PR, please take a look, the error message is:

```
message: 'Missing gmail mock data at https://github.com/FlowCrypt/flowcrypt-bin/raw/master/gmail-mock-data/usernosubmitorgruleflowcryptcom.json',
```

tomholub:
> @tomholub I believe tests are failing not because of changes from this PR, please take a look, the error message is:
> 
> ```
> message: 'Missing gmail mock data at https://github.com/FlowCrypt/flowcrypt-bin/raw/master/gmail-mock-data/usernosubmitorgruleflowcryptcom.json',
> ```

I think getting the branch up to master should fix this

limonte:
> > @tomholub I believe tests are failing not because of changes from this PR, please take a look, the error message is:
> > ```
> > message: 'Missing gmail mock data at https://github.com/FlowCrypt/flowcrypt-bin/raw/master/gmail-mock-data/usernosubmitorgruleflowcryptcom.json',
> > ```
> 
> I think getting the branch up to master should fix this

That helped, thanks!

limonte:
Oops, wrong button :)

limonte:
@tomholub this PR in its current state adds the RTL support to richtext messages only. I didn't make changes in converting text to html.

Questions:
1. do you think we should ship the RTL support to both plain and richtext messages? Currently, we'll have to recommend using the experimental richtext for users who need RTL.
2. What is needed for removing the "experimental" status from richtext messages?
3. When the "experimental" status from richtext messages will be removed, what will be the default sending option - plain text or richtext?

tomholub:
> @tomholub this PR in its current state adds the RTL support to richtext messages only. I didn't make changes in converting text to html.
> 
> Questions:
> 
>     1. Do you think we should ship the RTL support to both plain and richtext messages? Currently, we'll have to recommend using the experimental richtext for users who need RTL.
> 
>     2. What is needed for removing the "experimental" status from richtext messages?
> 
>     3. When the "experimental" status from richtext messages will be removed, what will be the default sending option - plain text or richtext?

1. I would probably want to target RTL on both, but it does not need to be in this PR.
2. I marked it as experimental because I wanted to see if there are any compatibility problems with other software. There indeed are, and are not resolvable (Gmail API will modify outgoing messages in ways that we cannot control). So I'm trying to figure out what to do, see 3.
3. I'm thinking to set rich text as a default option when messaging other FlowCrypt users, and plain text as default option when messaging people who use other software. Because it turns out, if they use other software and message is sent with rich text, they probably won't be able to read it (which we should also find a way to communicate to the user when they manually choose that). Once all this is done, I could remove the experimental wording. It will probably take a few months to finalize and test.

test/source/tests/page-recipe/compose-page-recipe.ts
tomholub: got it

limonte:
> It looks good as it is, if you'd like to address plain text in a separate PR. Should I merge it?

Yup, go ahead. I'll make additional PRs :+1:

limonte:
Hold on, I'll commit a small change to this PR.

limonte:
From https://github.com/FlowCrypt/flowcrypt-browser/issues/1114#issuecomment-578532445

> I tried just pasting hebrew sentences from online and that would seem to keep LTR even when pasted from RTL page. But Github and Gmail do the same.

Should be working properly now, take a look.

limonte:
@tomholub FYI Tests in Semaphore are flaky recently. Now this one is failing:

```
mail.google.com - success notif after setup, click hides it, does not re-appear + can re-connect
    ℹ (attempt 1 of 3) Retrying: TimeoutError: waiting for selector "[data-test="webmail-notification"]" failed: timeout 20000ms exceeded
    ℹ (attempt 2 of 3) Retrying: TimeoutError: waiting for selector "[data-test="webmail-notification"]" failed: timeout 20000ms exceeded
    ℹ (attempt 3 of 3) Failed:   TimeoutError: waiting for selector "[data-test="webmail-notification"]" failed: timeout 20000ms exceeded
```

I'll restart them.

tomholub:
> @tomholub FYI Tests in Semaphore are flaky recently. Now this one is failing:
> 
> ```
> mail.google.com - success notif after setup, click hides it, does not re-appear + can re-connect
>     ℹ (attempt 1 of 3) Retrying: TimeoutError: waiting for selector "[data-test="webmail-notification"]" failed: timeout 20000ms exceeded
>     ℹ (attempt 2 of 3) Retrying: TimeoutError: waiting for selector "[data-test="webmail-notification"]" failed: timeout 20000ms exceeded
>     ℹ (attempt 3 of 3) Failed:   TimeoutError: waiting for selector "[data-test="webmail-notification"]" failed: timeout 20000ms exceeded
> ```
> 
> I'll trigger the rebuild.

It has been bothering me too. Please see if you can take a look sometime, if not I may have a look later.

PR #2598, opened 2020-02-27
Add RTL support for plain text messages

### LINKED ISSUE #1114: support for typing RTL languages (hebrew, etc)

limonte:
Closes #1114

### DIFF
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -11,6 +11,12 @@ export type PromiseCancellation = { cancel: boolean };
 
 export class Str {
 
+  // ranges are taken from https://stackoverflow.com/a/14824756
+  // with the '\u0300' -> '\u0370' modification, because from '\u0300' to '\u0370' there are only punctuation marks
+  // see https://www.utf8-chartable.de/unicode-utf8-table.pl
+  public static readonly ltrChars = 'A-Za-z\u00C0-\u00D6\u00D8-\u00F6\u00F8-\u02B8\u0370-\u0590\u0800-\u1FFF\u2C00-\uFB1C\uFDFE-\uFE6F\uFEFD-\uFFFF';
+  public static readonly rtlChars = '\u0591-\u07FF\uFB1D-\uFDFD\uFE70-\uFEFC';
+
   public static parseEmail = (full: string, flag: 'VALIDATE' | 'DO-NOT-VALIDATE' = 'VALIDATE') => {
     let email: string | undefined;
     let name: string | undefined;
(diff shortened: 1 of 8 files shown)


test/source/tests/tests/compose.ts
```diff
@@ -550,14 +550,24 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       expect(await body.$eval('#input_text img', el => el.getAttribute('src'))).to.eq(imgBase64);
     }));
 
-    ava.default('compose - saving and rendering a draft with RTL text', testWithBrowser('compatibility', async (t, browser) => {
+    ava.default('compose - saving and rendering a draft with RTL text (plain text)', testWithBrowser('compatibility', async (t, browser) => {
       let composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
-      const subject = `saving and rendering a draft with RTL text`;
+      const subject = `saving and rendering a draft with RTL text (plain text)`;
+      await ComposePageRecipe.fillMsg(composePage, { to: 'human@flowcrypt.com' }, subject);
```
tomholub: Should make sure to explicitly set `richtext: false`. That way when we later switch default behavior, this will continue testing plain text and not rich text.

extension/js/common/core/common.ts
```diff
@@ -92,7 +98,15 @@ export class Str {
   }
 
   public static asEscapedHtml = (text: string) => {
```
tomholub: This PR uncovered some duplicate effort - `Str.asEscapedHtml` and `Xss.escapeTextAsRenderableHtml` do more or less the same thing. Please combine them to a single method `Str.escapeTextAsRenderableHtml` (but in the Str class). 

Notice that they both replace slightly different set of characters, please combine so that it replaces any of the chars mentioned in either of the two current methods.

Careful about order of replacing, it is ordered deliberately so that characters like `&` don't get double-escaped.

Thanks!

test/source/tests/tests/compose.ts
limonte: Done :+1:

extension/js/common/core/common.ts
limonte: Done :+1:

limonte:
@tomholub is it the pull bot messing with the commit history?

image

tomholub:
> @tomholub is it the pull bot messing with the commit history?
> 
> image

Yesterday GitHub had outages, I kept getting errors when merging. Looks like it merged twice, but the second merge is empty, so it looks ok. Thanks for mentioning it!

PR #2616, opened 2020-03-03
Issue 2615 secure reply button handlers

### LINKED ISSUE #2615: The focused state on the Secure Reply button isn't working
There are '.on()` handlers followed by `.off()`:

image

in https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/content_scripts/webmail/gmail-element-replacer.ts#L159

It should be `.off('keydown')` I guess

limonte:
Fixes #2615 

Peek 2020-03-03 14-28

Also fixed the vertical alignment of the Secure Reply button in Chrome:

Before | After
-|-
image | image

### DIFF
--- a/extension/css/webmail.css
+++ b/extension/css/webmail.css
@@ -69,7 +69,7 @@ body.cryptup_gmail .inserted div.reply_message_button {
   height: auto;
 }
 
-@-moz-document url-prefix() {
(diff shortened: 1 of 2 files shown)


tomholub:
excellent, the alignment was bothering me too

PR #2630, opened 2020-03-08
Add test for offline mode

### LINKED ISSUE #2622: test offline gmail mode - decrypt
here's a 1st priority issue for you! :)

Please test decrypting message in offline mode. It's supposed to be able to decrypt messages (at least the inline ones) without going online.

It's very important for a potential customer.

limonte:
Closes #2622 

@tomholub first of all, please give your feedback on the test. I decided to add the Gmail test to simulate the real use-case, but if you think that's not the best way, let me know.

Secondly, if the first part is alright and Gmail test is fine here, I'm struggling to get the `pgp_block.htm` frame, for some reason the `getFrame()` method can't find it and throwing the `Error: Frame not found: pgp_block.htm` exception.

### DIFF
--- a/test/source/tests/tests/gmail.ts
+++ b/test/source/tests/tests/gmail.ts
@@ -116,17 +116,21 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       const composePage = await GmailPageRecipe.openSecureCompose(t, gmailPage, browser);
     }));
 
-    ava.default.only('mail.google.com - decrypt message in offline mode', testWithBrowser('compatibility', async (t, browser) => {
+    ava.default('mail.google.com - decrypt message in offline mode', testWithBrowser('compatibility', async (t, browser) => {
       const gmailPage = await BrowserRecipe.openGmailPage(t, browser);
       await gmailPage.type('[aria-label="Search mail"]', 'encrypted + signed with gpg');
       await gmailPage.press('Enter'); // submit search
       await gmailPage.page.waitFor(1000); // wait for search results
-      await gmailPage.page.setOfflineMode(true); // go OFFLINE
+      await gmailPage.page.setOfflineMode(true); // go offline mode
       await gmailPage.press('Enter'); // open the message
-      // getFramesUrls will return one iframe which is expected
+      // TODO(@limonte): use the commented line below instead of opening pgp block in a new tab
+      // once https://github.com/puppeteer/puppeteer/issues/2548 is resolved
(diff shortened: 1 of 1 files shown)


test/source/tests/tests/gmail.ts
```diff
@@ -116,6 +116,19 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
...
+      const gmailPage = await BrowserRecipe.openGmailPage(t, browser);
+      await gmailPage.type('[aria-label="Search mail"]', 'encrypted + signed with gpg');
+      await gmailPage.press('Enter'); // submit search
+      await gmailPage.page.waitFor(1000); // wait for search results
+      await gmailPage.page.setOfflineMode(true); // go OFFLINE
+      await gmailPage.press('Enter'); // open the message
+      // getFramesUrls will return one iframe which is expected
+      const urls = await gmailPage.getFramesUrls(['/chrome/elements/pgp_block.htm'], { sleep: 1 });
+      // but getFrame will fail, why?
+      const pgpBlockPage = await gmailPage.getFrame(['pgp_block.htm']);
```
tomholub: Ah, puppeteer bug https://github.com/puppeteer/puppeteer/issues/2548

The solution is to copy the url and open in separate frame. At which point.. you might as well make that a mock test where you hard code the url.. like other decrypt tests - unfortunately.

limonte: > Ah, puppeteer bug puppeteer/puppeteer#2548

Thanks, subscribed :+1:

limonte: > At which point.. you might as well make that a mock test where you hard code the url.. like other decrypt tests - unfortunately.

Yeah, that kinda ruins the point of the test. What do you think about using a visual test in this case? I'm talking about taking the screenshot of `iframe.pgp_block` and comparing it to the expected one. This is the result of `ElementHandle.screenshot`:

tomholub: That I think would be too too noisy - when we change css etc.

tomholub:
I'm ok with it as is, given our options

tomholub:
Meaning, I'd be ok to merge it

test/source/tests/tests/gmail.ts
limonte: > That I think would be too too noisy - when we change css etc.

Yeah, agree.

limonte:
> Meaning, I'd be ok to merge it

Go ahead :+1:

PR #2576, opened 2020-02-24
Downgrade puppeteer to 1.19.0

limonte:
Fixes #2575

PR #2658, opened 2020-03-14
Downgrade puppeteer to 1.19.0

### LINKED ISSUE #2575: Can't start puppeteer with devtools: true
Sometimes during writing tests it's handy to start puppeteer with `devtools: true` https://github.com/FlowCrypt/flowcrypt-browser/blob/ae45bd3684cb0a75fcf3d8f447c1400e454aaec5/test/source/browser/browser-pool.ts#L43

I was working before, but after upgrade to `v1.20.0` in #2564 it doesn't work anymore, browser just opens and closes immediately without any errors in log.

limonte:
Closes #2575

The same as #2576, for some reason GH is not allowing to reopen that PR.

I'll check if the reason for the failures in #2576 was fixed in the recently fixed flaky test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "mailparser": "2.7.1",
     "mkdirp": "^1.0.3",
     "openpgp": "4.5.4",
-    "puppeteer": "1.20.0",
+    "puppeteer": "1.19.0",
     "speakeasy": "2.0.0",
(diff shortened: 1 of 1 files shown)


limonte:
Nope, still failing with the `Error: Page crashed!` exception

PR #2660, opened 2020-03-14
Plain reply to encrypted message

### LINKED ISSUE #1735: Standard reply box gets replaced when you want to send a nonenrypted email.

limonte:
Closes #1735 

@tomholub is live Gmail test needed for this PR?

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -45,7 +45,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   max-width: 1px !important; position: absolute !important; z-index: -1000 !important`;
   private currentlyEvaluatingStandardComposeBoxRecipients = false;
   private currentlyReplacingAtts = false;
-  private keepNextNativeReplyBox = false;
+  private keepNextStandardReplyBox = false;
 
   private sel = { // gmail_variant=standard|new
     convoRoot: 'div.if',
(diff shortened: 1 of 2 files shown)


tomholub:
I don't think so, you can just test compose frame individually, set isReplyBox=true

tomholub:
@limonte sorry I got confused. Yes! Please add a Gmail test.

tomholub:
> @limonte sorry I got confused. Yes! Please add a Gmail test.

Also, the PR does not include the warning message - do you think it would be unpleasant?

limonte:
> > @limonte sorry I got confused. Yes! Please add a Gmail test.
> 
> Also, the PR does not include the warning message - do you think it would be unpleasant?

Oops, overlooked that, sorry! It'll be useful, will add it :+1:

limonte:
> Works perfectly, just left an .only behind.
> 
> Finally! This was bothering me every day fireworks fireworks fireworks

:tada: This PR can be reviewed and merged and I can add the

> very bright red box above it, which says: The last message was encrypted, but you are composing a reply without encryption.

in separated PR along with the "[switch to encrypted reply]" functionality.

Issue #2632, opened 2020-03-08
consider removing featherlight, use SweetAlert2


tomholub:
With custom styling.

I don't like the following code in featherlight:

![image](https://user-images.githubusercontent.com/6306961/76166100-d0b62800-6153-11ea-8409-b70071ca033b.png)

I don't see any good reason why would they use eval like that, makes me wonder if they made other bad decisions.

Maybe try to replace one featherlight with a SweetAlert as a test, then we can take a look.

tomholub:
Meanwhile I'll try to refactor it away.

limonte:
> With custom styling.

Does this mean that the current styling should be preserved?

limonte:
Switching to SweetAlert2 is totally possible, I was actually annoyed by inability to close featherlight with keyboard. Esc isn't working once a user started to tab through the popup, and the close button isn't focusable.

limonte:
Considering that `eval` is removed, does this issue still have the first priority importance?

tomholub:
> Considering that `eval` is now removed, does this issue still have the first priority importance?

Great question - not really, but if it sounds fun, feel free to prioritize it.

limonte:
> > With custom styling.
> 
> Does this mean that the current styling should be preserved?

you probably missed this question ^

tomholub:
What I meant is, maybe the "page" type of SweetAlert2 should look a little different than our "modal" type of SweetAlert2. 

Or maybe not?

I'll leave it up to you.

Issue #2683, opened 2020-03-22
reply box does not always get replaced (+ warning shows)


tomholub:
We somehow do this with message replies - when there's an error, a standard message replybox shows.

----------------

![image](https://user-images.githubusercontent.com/6306961/77245803-5092c680-6c19-11ea-874c-bd824300c925.png)

tomholub:
Also, I'm not able to use the plain reply, keeps switching to encrypted. So maybe we didn't handle this yet at all.

limonte:
Oops, apparently I missed something here. I'm on it :+1:

limonte:
Hm, I can't reproduce the issue for example here https://mail.google.com/mail/u/1/#search/pgp%2Fmime+encrypted-asc-inline-encrypted-header+rich-text/WhctKJVjQgpKvxTvRjBcJDqfFvRNQjWRkqCXHqfpDNNCWNcZRQCFMcsMDdSQnHtcqJVWLnb

the secure reply is rendered:

![image](https://user-images.githubusercontent.com/6059356/77246826-9eb4c380-6c33-11ea-9ecd-aa3e4b2eb7d4.png)

Could you provide more details on how to reproduce the issue? Thanks!

tomholub:
To be honest, I'm not exactly sure how to reproduce it myself - it doesn't always happen.

In the situation where it did happen, the conversation contained:
 - encrypted message (collapsed)
 - plain message (collapsed)
 - encrypted message (collapsed)
 - plain message, with some plain text and a quoted encrypted message from before, producing the error in the screenshot (opened)
 - plain reply box and a warning message, but I was not really able to use the plain reply box - once I click it, it would re-render as encrypted reply, clicking "plain reply button" would not help

Maybe we need to observe it longer, see if there is a pattern

limonte:
Thank you for the input, I'll investigate further

tomholub:
Here is another one, this time normal message.

![image](https://user-images.githubusercontent.com/6306961/77347963-8d051600-6d30-11ea-83a3-6d4d84bfe0fd.png)

If I reopen the same message, it works normally:

![image](https://user-images.githubusercontent.com/6306961/77348031-a60dc700-6d30-11ea-91f7-0ffe954b8e6e.png)

Could the new warning sign introduce a race condition causing this?

tomholub:
[...]

PR #2685, opened 2020-03-23
Replace featherlight with sweetalert2 in settings

limonte:
The first part of #2632

In this PR the most of featherlight usage cases are replaced with SweetAlert2.

Also, bumped SweetAlert2 to latest version.


limonte:
This took a bit longer than I expected with polishing all styles, now I'm switching to investigate #2683 as it looks as a nasty bug.

limonte:
> Escape doesn't close the dialog. Otherwise it looks good!

Bu default the focus will be on the close button, but I suspect you'd dislike the "ugly" outline on it by default :) That's why I was putting the focus on iframe. When iframe is focused, it'll steal all the keyboard events to itself. The solution might be to just `.blur()` the close button.

limonte:
> Escape doesn't close the dialog. Otherwise it looks good!

Fixed :+1:

Issue #2124, opened 2019-10-20
add limited s/mime compatibility when sending email - PoC


tomholub:
#1158 addresses a similar issue in even more limited ways - some enterprises have a way of converting s/mime certs to openpgp public keys, when are missing self-signatures

The above is easier, but it would be better to instead have a way to support s/mime certs natively

tomholub:
part of enterprise requests #2125

tomholub:
Could do a PoC in four steps (possibly four PRs sequentially?):
1) PoC in a separate folder, like `flowcrypt-browser/smime-poc`. Does not have to use TypeScript. Allows me to put a public certificate in one `textarea`, message in another `textarea`, click a button and shows encrypted result in a `pre`.
2) Move the PoC to become htm/ts files in `flowcrypt-browser/extension/chrome/dev`, with libraries in `flowcrypt-browser/extension/lib`, with code style closer to the rest of the repo, but still otherwise separate from all workflows.
3) Generalize and start integrating with code, possibly add to compose window but only on Firefox, marked as experimental (that's what we usually do - fewer users on Firefox, so we only allow experimental features there). Add tests.
4) Let customers try it and give us feedback, then adjust as needed (compatibilty?), then unmark as experimental and allow on Chrome.

PR #2692, opened 2020-03-26
Add S/MIME proof-of-concept to /chrome/dev directory

wiktor-k:
This works:

2020-03-26T12:34:52,276927945+01:00

but is very crude. I'll check if adding typings to node-forge is possible and continue with point 3. from #2124.


extension/chrome/dev/smime.ts
```diff
@@ -0,0 +1,44 @@
...
+Content-Disposition: attachment; filename="smime.p7m"
+Content-Description: S/MIME Encrypted Message`;
+
+Catch.try(async () => {
+  function wrap(string: string) {
+    var lines = [];
+    var index = 0;
+    var length = string.length;
+    while (index < length) {
+      lines.push(string.slice(index, index += 76));
```
tomholub: I'm a little ahead with this comment for this particular level of PoC, but you can use
```ts
(window as unknown as BrowserWindow)['emailjs-mime-codec'].foldLines(newMsg.plaintext, 76, true)
```
Take a look at `signed-msg-mail-formatter`.

extension/lib/forge.min.js
```diff
@@ -0,0 +1,2 @@
...>1&&(0===r.charCodeAt(0)&&0==(128&r.charCodeAt(1))||255===r.charCodeAt(0)&&128==(128&r.charCodeAt(1)))?r.substr(1):r}function o(e){return e<=100?27:e<=150?18:e<=200?15:e<=250?12:e<=300?9:e<=350?8:e<=400?7:e<=500?6:e<=600?5:e<=800?4:e<=1250?3:2}function c(e){return f.util.isNodejs&&"function"==typeof d[e]}function u(e){return void 0!==g.globalScope&&"object"==typeof g.globalScope.crypto&&"object"==typeof g.globalScope.crypto.subtle&&"function"==typeof g.globalScope.crypto.subtle[e]}function l(e){return void 0!==g.globalScope&&"object"==typeof g.globalScope.msCrypto&&"object"==typeof g.globalScope.msCrypto.subtle&&"function"==typeof g.globalScope.msCrypto.subtle[e]}function p(e){for(var t=f.util.hexToBytes(e.toString(16)),r=new Uint8Array(t.length),a=0;a<t.length;++a)r[a]=t.charCodeAt(a);return r}var f=r(0);if(r(3),r(12),r(6),r(26),r(27),r(2),r(1),void 0===h)var h=f.jsbn.BigInteger;var d=f.util.isNodejs?r(16):null,y=f.asn1,g=f.util;f.pki=f.pki||{},e.exports=f.pki.rsa=f.rsa=f.rsa||{};var v=f.pki,m=[6,4,2,4,2,4,6,2],C={name:"PrivateKeyInfo",tagClass:y.Class.UNIVERSAL,type:y.Type.SEQUENCE,constructed:!0,value:[{name:"PrivateKeyInfo.version",tagClass:y.Class.UNIVERSAL,type:y.Type.INTEGER
```
tomholub: I know it may sound unusual but please include a non-minified version.

Otherwise Chrome Web Store / Mozzila Addons will not be happy during reviews.

tsconfig.json
```diff
@@ -60,6 +60,10 @@
       "featherlight": [
         "lib/featherlight/featherlight.js",
         "COMMENT"
+      ],
+      "node-forge": [
```
tomholub: Sorry for this custom lib include mechanism, I'm happy to help if it gets confusing.

wiktor-k: I figured out how it works so I guess it's not bad at all :)

tomholub:
Code-style wise, I think if you install my vscode extensions, most of these will get auto-corrected on each file save:

image

image

Can you access these CI outputs?

https://flowcrypt.semaphoreci.com/jobs/c588dfb1-c698-41da-b603-d7aa73add8cd

wiktor-k:
Okay, one question though. I'm looking through `compose-send-btn-module.ts`, `general-mail-formatter.ts`, `pgp-key.ts`, `compose-draft-module.ts`, `pgp-msg.ts`, `encrypted-mail-msg-formatter.ts` and thinking: how would we approach the event of mixed S/MIME - PGP setups? That is should the e-mail be only S/MIME or PGP? What if some keys of contacts are S/MIME and some are PGP? Should we warn the user? (seems obvious but I'd rather double-check with you).

wiktor-k:
>Can you access these CI outputs?
>https://flowcrypt.semaphoreci.com/jobs/c588dfb1-c698-41da-b603-d7aa73add8cd

Nope, it gives 404 for me.

wiktor-k:
>Code-style wise, I think if you install my vscode extensions, most of these will get auto-corrected on each file save:

Sounds good thanks for the tips :+1: 

Maybe I'll add the recommendations to the repo so they'll be automatically suggested? As in here: https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions

extension/lib/forge.min.js
wiktor-k: Got it, will do so. Forge doesn't publish non-minified artifacts so that's what I took initially: https://cdn.jsdelivr.net/npm/node-forge@0.9.1/dist/

tomholub:
> > Code-style wise, I think if you install my vscode extensions, most of these will get auto-corrected on each file save:
> 
> Sounds good thanks for the tips +1
> 
> Maybe I'll add the recommendations to the repo so they'll be automatically suggested? As in here: https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions

I thought we already track them there, you can take a look if any is missing :+1:

tomholub:
> > Can you access these CI outputs?
> > https://flowcrypt.semaphoreci.com/jobs/c588dfb1-c698-41da-b603-d7aa73add8cd
> 
> Nope, it gives 404 for me.

Invited, let me know if it still gives you trouble after you log into semaphore. Eg here https://flowcrypt.semaphoreci.com/workflows/cafa5952-47fc-47aa-9987-2597a7f34021 you should see this:

image

wiktor-k:
> Invited, let me know if it still gives you trouble after you log into semaphore. Eg here https://flowcrypt.semaphoreci.com/workflows/cafa5952-47fc-47aa-9987-2597a7f34021 you should see this:

Works now, thanks!

wiktor-k:
> I thought we already track them there, you can take a look if any is missing

Right, I didn't spot the workspace settings too.

tomholub:
> Okay, one question though. I'm looking through `compose-send-btn-module.ts`, `general-mail-formatter.ts`, `pgp-key.ts`, `compose-draft-module.ts`, `pgp-msg.ts`, `encrypted-mail-msg-formatter.ts` and thinking: how would we approach the event of mixed S/MIME - PGP setups? That is should the e-mail be only S/MIME or PGP? What if some keys of contacts are S/MIME and some are PGP? Should we warn the user? (seems obvious but I'd rather double-check with you).

For the very-very early experimental integration, I'm ok to make it either-or. Since the sender only has a PGP key, no S/MIME key, the message would not even be encrypted for the sender - initially. That means, if there are mixed recipients, show `Ui.modal.error` after clicking the send button.

I think we can approach this in stages.
1) either-or, not even sender pubkey used
2) either-or, but sender pubkey is also used (somehow create an S/MIME cert from PGP prv)
3) same as 2), but also parse sent/received S/MIME messages, because we should now be able to decrypt them
4) allow user to share their own S/MIME cert with external senders, or at least export it in settings?
5) allow mixed recipients, in which case recipients who use PGP on the receiving end and don't use FlowCrypt will be screwed, because it will use S/MIME format
6) find a way to combine PGP and S/MIME in a way that most PGP as well as S/MIME clients can parse their part and ignore the other? (alternatively, note down what we learned and propose spec update one or both methods to make it work together nicely)

wiktor-k:
Okay, understood. I'll hack it together and get back with more commits ;)

By the way: the sender's S/MIME cert should be uploaded via `FlowCrypt Settings` -> `Contacts` menu?

> find a way to combine PGP and S/MIME in a way that most PGP as well as S/MIME clients can parse their part and ignore the other? (alternatively, note down what we learned and propose spec update one or both methods to make it work together nicely)

Ha, I wonder if MIME could be used to distribute one encrypted e-mail in two formats at once. This of course is not relevant now but could be interesting approach.

tomholub:
> By the way: the sender's S/MIME cert should be uploaded via `FlowCrypt Settings` -> `Contacts` menu?

There are three ways:
1) In settings -> contacts (uses `KeyImportUI`)
2) Put the email in secure compose first, it should show in gray color (assuming no pgp key found automatically). Then a link shows "Add their Public Key", see screenshot below
3) When you receive an s/mime cert in email attachment, parse and render it like we do with pgp keys

For initial demo I'd focus on method "2". Screenshot:

image

When clicked:

image

Here the user could paste the cert or load it from a file.

When storing the S/MIME cert in local contacts (keyring), you could use the `client` value which is currently `'flowcrypt' | 'pgp-other' | undefined` to become `'flowcrypt' | 'pgp-other' | 'smime' | undefined`. Also, `has_pgp` really means "uses encryption", and eventually it may be worth to do a migration to call it `has_encryption`, but not yet. Anyway for smime certs you'd also set `has_pgp: 1` (it's a number because we order by it).

wiktor-k:
I did want to wait for the tests to pass before I click it :)

tomholub:
> I did want to wait for the tests to pass before I click it :)

No worries, GitHub is pretty good about this - it won't let me merge (easily) without the tests passing, per our repo rules.

image

So I wouldn't merge something that doesn't pass, even if marked as ready.

wiktor-k:
Yeah, I must say you've got the workflow set-up here perfectly :ok_hand: :)

PR #2696, opened 2020-03-26
Add strict Content Security Policy

### LINKED ISSUE #2693: Invalid CSP used by manifest.json
Hi,

The browser extension's manifest specifies invalid CSP expression (Chrome warns about it):

```json
  "content_security_policy": "'self'; object-src 'self'; frame-ancestors https://mail.google.com 'self';"
```

The problematic part is the leading `'self'`. Was it meant to be [`default-src 'self'`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/default-src)? If so the `object-src 'self'` directive [would be redundant](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/object-src).

I'd propose a PR to change it but it may have a bigger effect than intended and maybe it should be carefully tested :)

wiktor-k:
This change enables very strict CSP that doesn't allow inline styles or scripts.

Not everything works with it though! (e.g. popup) but it seems like there are not many changes. I'll check out later if it's feasible to remove all inline styles, if so - I'll fix them, if not, I'll allow them but disallow everything else.

DO NOT MERGE this yet!

------------

close #2693

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -89,5 +89,5 @@
     "/chrome/elements/shared/footer.htm"
   ],
   "minimum_chrome_version": "61",
-  "content_security_policy": "'self'; object-src 'self'; frame-ancestors https://mail.google.com 'self';"
+  "content_security_policy": "script-src 'self'; frame-ancestors https://mail.google.com 'self';"
 }


tomholub:
The test failure was not your fault, I've re-run it.

wiktor-k:
Yep, still I can see manually that this is not yet ready for merging. Still wanted to create a draft PR to track progress. If we could achieve very strict CSP I think it would be really nice from security PoV :)

tomholub:
> Yep, still I can see manually that this is not yet ready for merging. Still wanted to create a draft PR to track progress. If we could achieve very strict CSP I think it would be really nice from security PoV :)

I agree.

I'm not sure it's feasible to remove all inline styles. We use dynamically computed height of certain elements, eg based on text content, which will set the height inline in a `style="height: 335px"`etc.

PR #2699, opened 2020-03-27
Switch from plain to encrypted reply (from toolbar)

limonte:
Connected to #2683 

Fixes the scenario described in https://github.com/FlowCrypt/flowcrypt-browser/issues/2683#issuecomment-602902847 but it will work only once:

Peek 2020-03-27 22-50

The proper (infinite) switching between plain and secure should be done separately.

This could theoretically fix the "reply box does not always get replaced" issue, but I'm not 100% sure. @tomholub I'll need your feedback in #2683 on the initial issue is it's still happening after we merge this PR.


tomholub:
:+1: you're getting pretty good at dealing with Gmail/webmail

PR #2702, opened 2020-03-28
issue 2700 deprecate account check

### LINKED ISSUE #2700: deprecate backend account/check api call
instead, require users to be logged in to access subscription

tomholub:
close #2700


tomholub:
Write a backend re-auth test as follows:

1) set up acct
2) Additional settings -> experimental -> Set a new randomly generated FlowCrypt Account UUID
3) open settings, notice modal prompt to re-auth
4) re-auth
5) no more modal in settings

PR #2708, opened 2020-03-29
Replace featherlight with SweetAlert2

### LINKED ISSUE #2632: consider removing featherlight, use SweetAlert2

limonte:
Closes #2632


extension/js/common/browser/ui.ts
```diff
@@ -205,13 +204,17 @@ export class Ui {
       });
       return typeof dismiss === 'undefined';
     },
-    page: async (htmlUrl: string): Promise<void> => {
-      const html = await (await fetch(htmlUrl)).text();
+    page: async (htmlUrl: string, replaceNewlines = false): Promise<void> => {
+      let html = await (await fetch(htmlUrl)).text();
```
tomholub: This is interesting. You pre-fetch the html from the file, and render it directly? That is, an iframe is not used?

limonte: correct

limonte:
I was expecting the `standalone - different send from, new signed message, verification in mock` test to fail. What's the deal with flaky tests, when are they triggered?

extension/js/common/browser/ui.ts
tomholub: Now I see there is a separate `iframe` method. Now I understand, you choose depending if it's static content or not. Makes sense.

tomholub:
To run these locally, you'd do `npm run-script test_local_chrome_consumer_mock_flaky`.

Anyway they do run on every CI run, eg: https://flowcrypt.semaphoreci.com/workflows/c971da53-3675-4153-af4a-9603af56b881

image

Clicking the "flaky" group there, you see:

image

The only difference in how the flaky tests are executed is that they run under a different npm script (as mentioned above) and they run strictly sequentially, one after another as opposed to other tests which use about 10 browsers in parallel (when run on CI).

limonte:
Roger, thanks. I'll check why that test gives a false positive :+1:

tomholub:
You were probably expecting this line to fail?
```
await addPrvPage.waitTillGone('.featherlight.featherlight-iframe'); // dialog closed
```
which it wouldn't, it's waiting for a featherlight selector to be gone. Since there's no featherlight, it's always "gone", so it will pass. It's not waiting for it to appear and disappear, just waiting until it's "successfully not there".

You could change the selector to some Swal alternative.

PR #2712, opened 2020-03-30
Add sending S/MIME to Secure Compose

### LINKED ISSUE #2124: add limited s/mime compatibility when sending email - PoC

wiktor-k:
// tom edit: close #2124 

 - [x] minor code updates (will make a separate review for myself)
 - [x] add a test - send to single s/mime
 - [x] add a test - send to mixed recipients (should err)
 - [x] add a test - send to two s/mime (sould err?)
 - [x] add a test - attachment (is this supposed to not work yet?)
 - [x] add a test - send with broken S/MIME cert - err
 - [ ] add a test - send with non-S/MIME cert - err
 - [x] delete earlier dev files
 - [x] manual testing
 - [ ] evaluate why shows as bytes, not attachment, on recipient end

-----------------------

This is a PR adding sending S/MIME to Secure Compose.

This is ***not** yet ready for review*.

What works:
  * adding S/MIME key if it's missing,
  * correctly encrypting content if only S/MIME keys are present

What doesn't work:
  * MIME encoding of the encrypted message... it's probably something minor I missed,
  * correctly removing own key from encrypted set if only S/MIME recipients are present,
  * correctly removing calls to PGP specific methods (that are commented out currently),
  * code style issues.


extension/chrome/elements/add_pubkey.htm
```diff
@@ -35,7 +35,7 @@ <h1>Add a pubkey to email address</h1>
       or paste below:
     </div>
     <div class="line">
-      <textarea class="pubkey" placeholder="ASCII Armored Public Key"></textarea>
+      <textarea class="pubkey" placeholder="ASCII Armored Public Key" spellcheck="false"></textarea>
```
tomholub: :+1:

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -3,15 +3,18 @@
 'use strict';
 
 import { Backend, FcUuidAuth } from '../../../../js/common/api/backend.js';
-import { BaseMailFormatter } from './base-mail-formatter.js';
-import { ComposerResetBtnTrigger } from '../compose-err-module.js';
 import { Mime, SendableMsgBody } from '../../../../js/common/core/mime.js';
 import { NewMsgData, PubkeyResult } from '../compose-types.js';
 import { Str, Value } from '../../../../js/common/core/common.js';
+
+import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
```
tomholub: It appears you have installed one of the import-ordering vscode extensions.

Me and @limonte have found it to be somewhat annoying because it causes syntax highlighting to flash while it's updating, so we dropped it.

If it's still being recommended to add it, we should probably remove that recommendation.

extension/js/common/lang.ts
```diff
@@ -12,7 +12,7 @@ export const Lang = { // tslint:disable-line:variable-name
...
-    keyFormattedWell: (begin: string, end: string) => `Private key is not correctly formated. Please insert complete key, including "${begin}" and "${end}"\n\nEnter the private key you previously used. The corresponding public key is registered with your email, and the private key is needed to confirm this change.\n\nIf you chose to download your backup as a file, you should find it inside that file. If you backed up your key on Gmail, you will find there it by searching your inbox.`,
+    keyFormattedWell: (begin: string, end: string) => `Private key is not correctly formatted. Please insert complete key, including "${begin}" and "${end}"\n\nEnter the private key you previously used. The corresponding public key is registered with your email, and the private key is needed to confirm this change.\n\nIf you chose to download your backup as a file, you should find it inside that file. If you backed up your key on Gmail, you will find there it by searching your inbox.`,
```
tomholub: :+1:

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
limonte: Yes, `sort-imports` is in recommended and I agree that we should drop it from that list: #2713

wiktor-k:
Okay, changes from previous iteration:
  * it correctly sends S/MIME encrypted e-mails if only S/MIME recipients are selected,
  * for others the old logic is used (OpenPGP),
  * cleaned up the bit about detecting keys (missing: Serial ID detection, I used random for now),
  * minimized amount of places touched by this code, still there are too many (e.g. `extension/chrome/elements/compose-modules/compose-draft-module.ts`).

The goal was to seamlessly support adding S/MIME keys and sending S/MIME e-mails while leaving the rest of features un-touched. I think it's mostly done but would like to get your opinion.

> Later you can make a file like core/smime.ts with class / methods specific to smime, whose methods could (for now, before we abstract one level up) be called from the core/*-pgp.ts files.

Yes, definitely. Abstracting S/MIME vs OpenPGP would definitely make working with the code more pleasant but my casual look indicated that PGP is interleaved in too many places to do it all at once :( Are there code coverage reports for test somewhere? Usually refactorings (even with static types) can introduce silent bugs and I'd like to avoid that.

tomholub:
> Are there code coverage reports for test somewhere? 

Nope, I don't have any such metric. If you are familiar with a way to get such metric, I'd be happy to start tracking it.

To give you an idea, most tests are integration tests, driven through the UI (puppeteer) - testing user scenarios. What that means is that most important functionality will get hit one way or another through these tests, and "hot paths" like storage get hit a lot.

As far as code line coverage.. eh.. 60%? But it's the 60% we really care about.

extension/js/common/core/pgp-msg.ts
```diff
@@ -211,6 +213,38 @@ export class PgpMsg {
...
+    // cannot decrypt e-mails that they have sent
+    const keyTypes = new Set(pubkeys.slice(1).map(PgpMsg.getKeyType));
+    if (keyTypes.size > 1) {
+      throw new Error('Mixed key types are not allowed: ' + [...keyTypes]);
+    }
+    const keyType = keyTypes.keys().next().value;
+    if (keyType === 'x509') {
+      const p7 = forge.pkcs7.createEnvelopedData();
+
+      for (const pubkey of pubkeys.slice(1)) {
```
tomholub, line 227: better filter into two groups, pgp and s/mime, in separate variables, and evaluate those

extension/js/common/core/pgp-msg.ts
```diff
@@ -253,6 +287,16 @@ export class PgpMsg {
     return diagnosis;
   }
 
+  private static getKeyType(pubkey: string): 'openpgp' | 'x509' {
```
tomholub, line 290: recommend to make this public and move it into PgpKey (for now).

extension/js/common/platform/store/contact-store.ts
```diff
@@ -103,6 +103,28 @@ export class ContactStore extends AbstractStore {
           expiresOn: null
         };
       }
+      // X.509 certificate
+      if (pubkey.startsWith('-----BEGIN CERTIFICATE-----')) {
```
tomholub, line 107: recommend to use the newly public keyType method.

extension/js/common/ui/key-import-ui.ts
```diff
@@ -146,6 +146,10 @@ export class KeyImportUi {
   }
 
   public checkPub = async (armored: string): Promise<string> => {
+    // For X.509 keys assume they're good if they start with the marker string
+    if (armored.startsWith('-----BEGIN CERTIFICATE-----')) {
```
tomholub, line 150: recommend to use the newly public keyType method.

wiktor-k:
Yep, excellent feedback, that's what I was looking for, thanks! I'll get back with the changes and then we can think of next steps.

extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -104,10 +104,12 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
         msgData.pwd = undefined; // not needed for drafts
         const sendable = await new EncryptedMsgMailFormatter(this.view, true).sendableMsg(msgData, pubkeys);
         this.view.S.cached('send_btn_note').text('Saving');
-        if (this.view.threadId) { // reply draft
-          sendable.body['text/plain'] = `[cryptup:link:draft_reply:${this.view.threadId}]\n\n${sendable.body['text/plain'] || ''}`;
-        } else if (this.view.draftId) { // new message compose draft with known draftid
-          sendable.body['text/plain'] = `[cryptup:link:draft_compose:${this.view.draftId}]\n\n${sendable.body['text/plain'] || ''}`;
+        if (!(sendable.body instanceof Uint8Array)) {
```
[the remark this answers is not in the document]
wiktor-k: Yeah actually this is a workaround, if body is a string then it's easy to append this magic markup. If it's Uint8Array I was rather thinking about converting to string, appending markup, then converting to Uint8Array?

In the long run I'd think about some other hierarchy for messages (maybe something similar to a MIME message: multiple content types + attachments) and then encrypting this entire message. Passing strings and byte arrays around starts getting out of hand (or at least I didn't internalize it yet :D ).

wiktor-k:
> Nope, I don't have any such metric. If you are familiar with a way to get such metric, I'd be happy to start tracking it.

It seems the tests cover quite a few critical functions (like sending stuff with/without attachments) so maybe refactoring this part is not as risky as I initially thought... :thinking:

tomholub:
> > Nope, I don't have any such metric. If you are familiar with a way to get such metric, I'd be happy to start tracking it.
> 
> It seems the tests cover quite a few critical functions (like sending stuff with/without attachments) so maybe refactoring this part is not as risky as I initially thought... thinking

I indeed test the compose button quite heavily functionality wise, but only some tests actually look at what was sent - eg if it's sending something unique. Eventually we'll need to dedicate the time to carefully evaluate the messages the tests compose.

I think, after my review, it should be reasonably safe.

tomholub:
@wiktor-k sorry I managed to miss this becoming ready for review! I'll take a look. Thanks!

wiktor-k:
No problem, it's still Easter here so I didn't want to push you (yet ;) ).

tomholub:
Easter here too :) I don't observe holidays much, so normally a nudge doesn't hurt.

tomholub:
Todo for me:
 - [ ] minor code updates (will make a separate review for myself)
 - [ ] add a test - send to single s/mime
 - [ ] add a test - send to mixed recipients (should err)
 - [ ] add a test - send to two s/mime (sould err?)
 - [ ] add a test - attachment (is this supposed to not work yet?)
 - [ ] manual testing

wiktor-k:
> add a test - send to mixed recipients (should err)

Yes.

> add a test - send to two s/mime (sould err?)

No, it should work. But I never tested this :)

> add a test - attachment (is this supposed to not work yet?)

Yes, it won't work but should work in general. I think S/MIME is very similary to PGP/MIME so maybe if we unified this code we would have attachment support "for free".

By the way, why isn't PGP/MIME the default? I thought [inline](https://josefsson.org/inline-openpgp-considered-harmful.html) is [considered bad](https://dkg.fifthhorseman.net/notes/inline-pgp-harmful/). Is this for some kind of backwards-compatibility reasons?

tomholub:
> By the way, why isn't PGP/MIME the default? I thought inline is considered bad. Is this for some kind of backwards-compatibility reasons?

Gmail API screws up the messages, so they don't really come out as PGP/MIME.

I'll try to talk to them to fix it, haven't had the time yet.

tomholub:
Still want to do this.. sorry for the delay!

wiktor-k:
Don't worry Tom, when this is blocked I'm doing [other stuff](https://docs.rust-embedded.org/discovery/) and will resume when you're ready :+1:

extension/chrome/elements/compose-modules/compose-draft-module.ts
tomholub, line 107: For me - clean up with `this.view.recipientsModule.isSmimeRecipientPresent()`

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -23,6 +23,8 @@ import { opgp } from '../../../../js/common/core/pgp.js';
 import { ContactStore } from '../../../../js/common/platform/store/contact-store.js';
 import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
 
+type DataArmorResult = PgpMsgMethod.OpenPGPEncryptArmorResult | PgpMsgMethod.X509EncryptResult;
```
tomholub, line 26: Should not be defined here

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -279,7 +279,7 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
     if (format === 'full') {
       const bodies = GmailParser.findBodies(gmailMsg);
       const atts = GmailParser.findAtts(gmailMsg);
-      const fromTextBody = PgpArmor.clip(Buf.fromBase64UrlStr(bodies['text/plain'] || '').toUtfStr());
+      const fromTextBody = PgpArmor.clip(Buf.fromBase64UrlStr(bodies['text/plain']?.toString() || '').toUtfStr());
```
tomholub, line 282: todo - have a look at types

extension/js/common/api/email-provider/sendable-msg.ts
```diff
@@ -83,9 +83,16 @@ export class SendableMsg {
       }
     }
     this.headers.Subject = this.subject;
-    if (this.type === 'pgpMimeSigned' && this.sign) {
+    if (this.type === 'smimePlain' && this.body['encrypted/buf']) {
```
tomholub, line 86: I would call this just `smime` or even `smimeEncrypted` instead of `smimePlain`

extension/js/common/api/email-provider/sendable-msg.ts
```diff
@@ -83,9 +83,16 @@ export class SendableMsg {
...
+      return await Mime.encodePlain(this.body['encrypted/buf'], this.headers);
+    } else if (this.type === 'pgpMimeSigned' && this.sign) {
       return await Mime.encodePgpMimeSigned(this.body, this.headers, this.atts, this.sign);
     } else {
+      // encrypted/buf is a Buf instance that is converted to one part text/plain
+      // message
+      if (this.body['encrypted/buf']) {
```
tomholub, line 93: Here somewhat confusing - is any usecase still currently using this option - encrypted/buf, if the `smime` variant is handled above? Eg as is, will this branch ever run?

@wiktor-k

extension/js/common/core/mime.ts
```diff
@@ -214,6 +219,19 @@ export class Mime {
     return rootNode.build(); // tslint:disable-line:no-unsafe-any
   }
 
+  public static encodePlain = async (body: Uint8Array, headers: RichHeaders): Promise<string> => {
```
tomholub, line 222: todo - rename to `encodeSmime`

extension/js/common/core/mime.ts
```diff
@@ -226,7 +244,7 @@ export class Mime {
     }
     const bodyNodes = new MimeBuilder('multipart/alternative'); // tslint:disable-line:no-unsafe-any
     for (const type of Object.keys(body)) {
-      bodyNodes.appendChild(Mime.newContentNode(MimeBuilder, type, body[type]!)); // tslint:disable-line:no-unsafe-any
+      bodyNodes.appendChild(Mime.newContentNode(MimeBuilder, type, String(body[type]!))); // tslint:disable-line:no-unsafe-any
```
tomholub, line 247: todo - check types

extension/js/common/core/pgp-msg.ts
```diff
@@ -211,6 +221,19 @@ export class PgpMsg {
   }
 
   public static encrypt: PgpMsgMethod.Encrypt = async ({ pubkeys, signingPrv, pwd, data, filename, armor, date }) => {
+    // slice(1) filters first key that is always own OpenPGP key
+    // in result if X.509 encryption is selected the e-mail author
+    // cannot decrypt e-mails that they have sent
+    const otherKeys = pubkeys.slice(1);
```
tomholub, line 227: I still don't like this, I would not rely on key order as indicator of which key is mine. Will refactor

extension/js/common/core/pgp-msg.ts
```diff
@@ -231,7 +254,12 @@ export class PgpMsg {
     if (signingPrv && typeof signingPrv.isPrivate !== 'undefined' && signingPrv.isPrivate()) { // tslint:disable-line:no-unbound-method - only testing if exists
       options.privateKeys = [signingPrv];
     }
-    return await opgp.encrypt(options);
+    const result: OpenPGP.EncryptBinaryResult | OpenPGP.EncryptArmorResult = await opgp.encrypt(options);
```
tomholub, line 257: should not need to clarify type - will check

extension/js/common/core/pgp-key.ts
```diff
@@ -352,6 +352,10 @@ export class PgpKey {
    * This is used to figure out how recently was key updated, and if one key is newer than other.
    */
   public static lastSig = async (key: OpenPGP.key.Key): Promise<number> => {
+    // key is undefined only for X.509 keys
+    if (!key) {
+      return Date.now();
```
tomholub, line 357: todo - give this a look

extension/js/common/ui/key-import-ui.ts
```diff
@@ -146,6 +146,9 @@ export class KeyImportUi {
   }
 
   public checkPub = async (armored: string): Promise<string> => {
+    if (PgpKey.getKeyType(armored) === 'x509') {
+      return armored;
```
tomholub, line 150: todo - add a todo

extension/chrome/elements/compose-modules/compose-draft-module.ts
tomholub: Actually this must never happen. I think. Because we are only using our own public key for drafts, which (for now) is always opgp:
```ts
const pubkeys = [{ isMine: true, email: msgData.from, pubkey: primaryKi.public }];
```

extension/js/common/core/pgp-key.ts
tomholub: Make a key abstraction, and pass that around instead, that could be either pgp or smime key behind it

I'll make another issue

tomholub:
I tried to give it a HTTPS cert instead, and it happily encrypts for it :thinking: is this expected behavior? The lib doesn't care? In that case, sounds like this is something we should be checking ourselves? (eg in next PR)

```ts
    // todo - unexpectedly works
    // ava.default.only('send non-S/MIME cert - err', testWithBrowser('compose', async (t, browser) => {
    //   const inboxPage = await browser.newPage(t, TestUrls.extensionInbox('test.ci.compose@org.flowcrypt.com'));
    //   let composeFrame = await InboxPageRecipe.openAndGetComposeFrame(inboxPage);
    //   await ComposePageRecipe.fillMsg(composeFrame, { to: 'smime@recipient.com' }, t.title);
    //   const httpsCert = '-----BEGIN CERTIFICATE-----\nMIIFZTCCBE2gAwIBAgISA/LOLnFAcrNSDjMi+PvkSbX1MA0GCSqGSIb3DQEBCwUA\nMEoxCzAJBgNVBAYTAlVTMRYwFAYDVQQKEw1MZXQncyBFbmNyeXB0MSMwIQYDVQQD\nExpMZXQncyBFbmNyeXB0IEF1dGhvcml0eSBYMzAeFw0yMDAzMTQxNTQ0NTVaFw0y\nMDA2MTIxNTQ0NTVaMBgxFjAUBgNVBAMTDWZsb3djcnlwdC5jb20wggEiMA0GCSqG\nSIb3DQEBAQUAA4IBDwAwggEKAoIBAQDBYeT+zyJK4VrAtpBoxnzNrgPMkeJ3WBw3\nlZrO7GXsPUUQL/2uL3NfMwQ4qWqsiJStShaTQ0UX1MQCBgdOY/Ajr5xgyCz4aE0+\nQeReGy+qFyoGE9okVdF+/uJhFTOkK8goA4rDRN3MrSuWsivc/5/8Htd/M01JFAcU\nEblrPkSBtJp8IAtr+QD8etmMd05N0oQFNFT/T7QNrEdItCKSS6jMpprR4phr792K\niQh9MzhZ3O+QEM+UKpsL0dM9C6PD9jNFjFz3EDch/VFPbBlcBfWGvYnjBlqKjhYA\nLPUVPgIF4CVQ60EoOHk1ewyoAyydYyFXppUz1eDvemUhLMWuBJ2tAgMBAAGjggJ1\nMIICcTAOBgNVHQ8BAf8EBAMCBaAwHQYDVR0lBBYwFAYIKwYBBQUHAwEGCCsGAQUF\nBwMCMAwGA1UdEwEB/wQCMAAwHQYDVR0OBBYEFMr4ERxBRtKNI67oIkJHN2QSBptE\nMB8GA1UdIwQYMBaAFKhKamMEfd265tE5t6ZFZe/zqOyhMG8GCCsGAQUFBwEBBGMw\nYTAuBggrBgEFBQcwAYYiaHR0cDovL29jc3AuaW50LXgzLmxldHNlbmNyeXB0Lm9y\nZzAvBggrBgEFBQcwAoYjaHR0cDovL2NlcnQuaW50LXgzLmxldHNlbmNyeXB0Lm9y\nZy8wKQYDVR0RBCIwIIIPKi5mbG93Y3J5cHQuY29tgg1mbG93Y3J5cHQuY29tMEwG\nA1UdIARFMEMwCAYGZ4EMAQIBMDcGCysGAQQBgt8TAQEBMCgwJgYIKwYBBQUHAgEW\nGmh0dHA6Ly9jcHMubGV0c2VuY3J5cHQub3JnMIIBBgYKKwYBBAHWeQIEAgSB9wSB\n9ADyAHcAb1N2rDHwMRnYmQCkURX/dxUcEdkCwQApBo2yCJo32RMAAAFw2e8sLwAA\nBAMASDBGAiEA7Omcf4+uFphcbEq19r4GoWi7E1qvsJTykvgH342x1d4CIQDSCJZK\n3zsVSw8I1GVfnIr/drVhgn4TJgacXx6+gBzfXQB3ALIeBcyLos2KIE6HZvkruYol\nIGdr2vpw57JJUy3vi5BeAAABcNnvK/kAAAQDAEgwRgIhAP7BbIkG/mNclZAVqgA0\nomAB/6xMwbu1ZUsHNBMkZG+QAiEAmZWCVdUfmFs3b+zDEaAF7eFDnz7qbDa5q6M0\n98r8In0wDQYJKoZIhvcNAQELBQADggEBAFaUhUkxGkHc3lxozCbozM7ffAOcK5De\nJGoTtsXw/XmMACBIIqn2Aan+zvQdK/cWV9+dYu5tA/PHZwVbfKAU2x+Fizs7uDgs\nslg16un1/DP7bmi4Ih3KDVyznzgTwWPq9CmPMIeCXBSGvGN4xdfyIf7mKPSmsEB3\ngkM8HyE27e2u8B4f/R4W+sbqx0h5Y/Kv6NFqgQlatEY2HdAQDYYL21xO1ZjaUozP\nyfHQSJwGHp3/1Xdq5mIkV7w9xxhOn64FXp4S0spVCxT3er1EEUurq+lXjyeX4Dog\n1gy3r417NPqQWuBJcA/InSaS/GUyGghp+kuGfIDqVYfQqU1297nThEA=\n-----END CERTIFICATE-----\n';
    //   await pastePublicKeyManually(composeFrame, inboxPage, 'smime@recipient.com', httpsCert);
    //   await composeFrame.waitAndClick('@action-send', { delay: 2 });
    //   await PageRecipe.waitForModalAndRespond(composeFrame, 'error', { contentToCheck: 'dunno', timeout: 40 });
    // }));
```

wiktor-k:
> I tried to give it a HTTPS cert instead, and it happily encrypts for it thinking is this expected behavior? The lib doesn't care? In that case, sounds like this is something we should be checking ourselves? (eg in next PR)

The library probably doesn't care but we could check it in general because S/MIME certs have "key usage flags" (simiar to E/S/C/A flags for PGP certs). I mean I didn't check if the lib exposes key flags but I hope it does... then we can reject certs that don't have correct flags. Btw interesting test case :)

tomholub:
> Btw interesting test case :)

Users!! :)

tomholub:
When opening an s/mime message on a client without s/mime (tried Gmail), it just shows the bytes.

I would have expected it to show as an attachment (I vaguely recall trying to open such message before). Can you have a look?

image

Also, have you tried to open such message in an s/mime client, after integrating it?

We're getting there.

wiktor-k:
> Also, have you tried to open such message in an s/mime client, after integrating it?

Yes, I always tested it with Thunderbird that had my certificate and it just worked:

2020-04-16T11:51:35,901000280+02:00


> When opening an s/mime message on a client without s/mime (tried Gmail), it just shows the bytes.
> I would have expected it to show as an attachment (I vaguely recall trying to open such message before). Can you have a look?

Maybe we can wrap it in an attachment. It worked in Thunderbird so I didn't bother to try out other options.

> We're getting there.

Yes :)

PR #2713, opened 2020-03-30
Remove amatiasq.sort-imports from recommended extensions

limonte:
As per https://github.com/FlowCrypt/flowcrypt-browser/pull/2712#discussion_r400281551


tomholub:
Thanks!

PR #2727, opened 2020-04-12
Fix drafts for plain replies

### LINKED ISSUE #2721: allow user to finish a plain reply draft without interfering
> I started writing a non-encrypted reply to someone who had previously sent me a *signed* email. When returning to finish writing the draft, the draft window is completely overridden by the "Encrypted Reply, Reply to All, or Forward" window. This is very annoying.

> In fact, the only way I could finish up my draft was by repeatedly reloading the tab until the "Encrypted Reply, Reply to All, or Forward"-thing glitched out and I could back to my draft. This is a glaring issue in an otherwise fantastic and polished product.

I agree, we could do better here.

limonte:
Fixes #2721


limonte:
> Thanks!
> 
> (sorry I'm taking a while to review the other PR, I'll get to it soon)

No worries at all! Sorry about sending PRs in the late evening!

tomholub:
time doesn't matter, that's why we use github asynchronously ¯\\_(ツ)\_/¯

Issue #2732, opened 2020-04-16
settle on a fingerprint format for s/mime


tomholub:
eg in contacts currently:

![image](https://user-images.githubusercontent.com/6306961/79441547-e9bdbe80-7fc6-11ea-9d3a-101eddad5788.png)

Higher priority because it affects items in storage.. better settle it sooner than later.

tomholub:
@wiktor-k I suspect this has been solved - when you get a chance, please document the outcome here, then close. Thanks!

wiktor-k:
Yes, actually I looked at issues assigned to me and I think quite a few of them have been solved. Maybe I'll review them and then get back to the attester (I'm reading sources and doing some small experiments there).

PR #2735, opened 2020-04-17
Issue 2732: Fix serial number and expiration

wiktor-k:
This is a mini PR to fix issue #2732.

Fortunately serial numbers have the same exact length as OpenPGP fingerprints v4 so it looks nice.

------------

// todo:
 - [ ] test for fingerprint / serial number
 - [ ] test expiration

### DIFF
--- a/extension/js/common/core/smime.ts
+++ b/extension/js/common/core/smime.ts
@@ -16,3 +16,8 @@ export const encrypt = (pubkeys: string[], data: Uint8Array): { data: Uint8Array
   }
   return { data: new Uint8Array(arr), type: 'smime' };
 };
+
+export const getSerialNumber = (pubkey: string) => {
+  const certificate = forge.pki.certificateFromPem(pubkey);
+  return certificate.serialNumber;
+};
--- a/extension/js/common/core/smime.ts
+++ b/extension/js/common/core/smime.ts
@@ -30,3 +30,12 @@ export const getExpirationTime = (pubkey: string) => {
   }
   return;
 };
+
+export const isEmailCertificate = (pubkey: string) => {
+  const certificate = forge.pki.certificateFromPem(pubkey);
+  const eku = certificate.getExtension('extKeyUsage');
+  if (!eku) {
+    return false;
+  }
+  return !!(eku as { emailProtection: boolean }).emailProtection;
+};
--- a/extension/js/common/core/smime.ts
+++ b/extension/js/common/core/smime.ts
@@ -21,3 +21,12 @@ export const getSerialNumber = (pubkey: string) => {
   const certificate = forge.pki.certificateFromPem(pubkey);
   return certificate.serialNumber;
 };
+
+export const getExpirationTime = (pubkey: string) => {
+  const certificate = forge.pki.certificateFromPem(pubkey);
+  const expiration = certificate.validity.notAfter;
+  if (expiration) {
+    return expiration.getTime();
+  }
+  return;
+};
--- a/extension/js/common/platform/store/contact-store.ts
+++ b/extension/js/common/platform/store/contact-store.ts
@@ -122,7 +122,7 @@ export class ContactStore extends AbstractStore {
           last_use: lastUse || null,
           pubkey_last_sig: lastSig || null,
           pubkey_last_check: lastCheck || null,
-          expiresOn: null
+          expiresOn: smime.getExpirationTime(pubkey) || null
         };
       }
       const k = await PgpKey.read(pubkey);
--- a/extension/js/common/platform/store/contact-store.ts
+++ b/extension/js/common/platform/store/contact-store.ts
@@ -8,6 +8,7 @@ import { BrowserMsg } from '../../browser/browser-msg.js';
 import { Str } from '../../core/common.js';
 import { PgpKey, Contact } from '../../core/pgp-key.js';
 import { PgpArmor } from '../../core/pgp-armor.js';
+import * as smime from '../../core/smime.js';
 
 // tslint:disable:no-null-keyword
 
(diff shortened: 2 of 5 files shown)


wiktor-k:
> send with broken S/MIME cert - err
> --
> ℹ (attempt 1 of 3) Retrying: Error: this.wait_till_gone: some of "@dialog-add-pubkey" still present after timeout:20

:thinking: hmm I guess it fails because the add pubkey dialog is not dismissed and that's because there's some error triggered when adding the key (maybe related to expiration?). Is there a way to get more details out of it? (running locally fails with missing `test-secrets.json` -> of course :man_facepalming: ).

tomholub:
> > ## send with broken S/MIME cert - err
> > information_source (attempt 1 of 3) Retrying: Error: this.wait_till_gone: some of "@dialog-add-pubkey" still present after timeout:20
> 
> thinking hmm I guess it fails because the add pubkey dialog is not dismissed and that's because there's some error triggered when adding the key (maybe related to expiration?). Is there a way to get more details out of it? (running locally fails with missing `test-secrets.json` -> of course man_facepalming ).

Unsure about why this happened - what you said sounds plausible (I can look into it tomorrow).

But yes - you can get more details. It will spam your email with details for every failure regardless of who pushes the commit. On my email I filter them and route them all straight to trash, and if I need a particular result, I'll search for it (which searches trash too for up to 30 days on Gmail).

I'll add your email there and re-run, so you should receive it.

tomholub:
I've re-run it: https://flowcrypt.semaphoreci.com/jobs/e7795d41-0d9b-429b-be4f-bf68fb9013ff

It failed again, and you should receive a debug email. Over time, you'll be getting loads of these emails (sometime with a bit of a delay), and so you'll need a way to search them. For this, take a look at the FAIL ID:

image


Here it was `r2r4i1i0gjr`. Search in Gmail:

image

The attachment is a HTML file. Download and open it:

Click on the collapsed red part:

image

image

tomholub:
As for the failure, maybe we click it too fast? Let me try to add a delay somewhere.

tomholub:
Ah! The problem is that now the faulty cert gets discovered sooner than before.

Which is a good thing.

That means we have to start looking for the error sooner.

These kinds of issues would be easier for you to debug locally. Anyway, I'll go ahead and fix it.

extension/js/common/core/smime.ts
```diff
@@ -16,3 +16,26 @@ export const encrypt = (pubkeys: string[], data: Uint8Array): { data: Uint8Array
...
 };
+
+export const getSerialNumber = (pubkey: string) => {
+  const certificate = forge.pki.certificateFromPem(pubkey);
+  return certificate.serialNumber;
+};
+
+export const getExpirationTime = (pubkey: string) => {
+  const certificate = forge.pki.certificateFromPem(pubkey);
+  const expiration = certificate.validity.notAfter;
```
tomholub, line 27: It's preferred to return a `Date` for dates, so in this case `Date | undefined` would be ideal

extension/js/common/platform/store/contact-store.ts
```diff
@@ -105,24 +106,23 @@ export class ContactStore extends AbstractStore {
       }
       // X.509 certificate
       if (PgpKey.getKeyType(pubkey) === 'x509') {
-        // FIXME: For now we return random data.
-        // Later we'll return serial ID from the certificate.
-        const longid = Math.random() + '';
+        const fingerprint = smime.getSerialNumber(pubkey);
+        const longid = fingerprint.substring(fingerprint.length / 2);
```
tomholub, line 110: Is this following any known S/MIME spec or just trying to shoehorn it into our OpenPGP code?

extension/js/common/platform/store/contact-store.ts
```diff
@@ -105,24 +106,23 @@ export class ContactStore extends AbstractStore {
...
-          fingerprint: Math.random() + '',
+          fingerprint,
           longid,
           longids: [longid],
           pending_lookup: 0,
           last_use: lastUse || null,
           pubkey_last_sig: lastSig || null,
           pubkey_last_check: lastCheck || null,
-          expiresOn: null
+          expiresOn: smime.getExpirationTime(pubkey) || null
```
tomholub, line 125: While the Date will be a bit less convenient here (need to convert to number), I still think it's generally better to work with a Date.

extension/js/common/ui/key-import-ui.ts
```diff
@@ -146,7 +147,7 @@ export class KeyImportUi {
   }
 
   public checkPub = async (armored: string): Promise<string> => {
-    if (PgpKey.getKeyType(armored) === 'x509') {
+    if (PgpKey.getKeyType(armored) === 'x509' && smime.isEmailCertificate(armored)) {
```
tomholub, line 150: I would make the error a bit more obvious, pseudo code
```
if(isx509)
  if(!isEmailCert) {
    throw UserAlert('The type of this x509 certificate is unknown.\nEither provide a valid S/MIME certificate, or an OpenPGP public key.');
  }
  return armored;
}
```

tomholub:
You could then also add unit tests, to test that it produces expected serial id (compared to some 3rd party s/mime software), "longid" and also test one key that expires, that it yields the right expiration and one that doesn't expire.

You'd add them here: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/tests/tests/unit.ts

Unit tests shouldn't really need test secrets, I think. But I'm not sure if they run without them.

wiktor-k:
Got it! I think I'll roll this expiration and serial ID thing into one class so that cert parsing will happen only once. I'm planning that after I finish the introduction of proper key type (with openpgp / x.509 variants) but changing that has ripple effects over the entire codebase so I'm slowly progressing on that front :)

tomholub:
@wiktor-k what is the status here - is it waiting for my action?

wiktor-k:
No, I'm planning to include this as soon as the refactoring thing is done (see there is a `TODO` in the refactoring to include this change too). No worries if I'll need your confirmation I'll explicitly ask.

wiktor-k:
Changes from this PR are already included in #2739 so closing this one.

PR #2739, opened 2020-04-20
Introduce Pubkey abstraction

wiktor-k:
This is very much *not done* but I'd like to show some progress on this front :)

I'm planning on having the parsing done only once (either when getting the key from external servers or IndexedDB) and then having all necessary fields (fingerprint, longid) populated.

Will see how it goes. At the end I want to remove the `unparsed` field and also the `readAsOpenPGP` method. They're introduced for now to keep the change as small as possible and to be removed ASAP.


wiktor-k:
Okay, this is now working (as in - it doesn't break keys and sends correctly encrypted PGP and SMIME emails ;) ). I'll now clean it up (there are 37 instances of `unparsed` and `readAsOpenPGP`) and fix tests. Then I'll submit this to review... phew...

tomholub:
Yeah the codebase is not exactly small these days. If you run into parts that are hard to untangle, it would be good to note these down so that I can work on streamlining them.

wiktor-k:
Okay, phew, solving these test failures takes a lot of time for me.

Some questions:

I've got one test failing (`Test: decrypt - protonmail - auto TOFU load matching pubkey first time`) because apparently when `PgpKey.parse` is called `opgp` object is not initialized. `ci_pgp_host_page.htm` has openpgp.js listed before the rest and `extension/js/common/core/pgp.ts` seems to load the object. Is this some kind of race condition? (it seems like everything is async because it's safer to delay key parsing operations...)

Also: I'm reading the test results and they're separated in pages cleanly, I see test code and it's kind of like selenium automation but it's not clear when the various pages are switched (which test line in the test output corresponds to which test line in the browser-recipe.ts). Are there any tips you could share in correlating test output with tests?

Thanks in advance!

tomholub:
> I've got one test failing (Test: decrypt - protonmail - auto TOFU load matching pubkey first time) because apparently when PgpKey.parse is called opgp object is not initialized. ci_pgp_host_page.htm has openpgp.js listed before the rest and extension/js/common/core/pgp.ts seems to load the object. Is this some kind of race condition? (it seems like everything is async because it's safer to delay key parsing operations...)

I'm shooting blindly now due to lack of time, openpgp (opgp) is not defined in `pgp_decrypt.htm` whatsoever, because there can be up to 50 of these frames per conversation, each openpgp.js takes 750 ms to load.. can be a real performance hit. So for pgp_decrypt.htm, we do operations through its parent page, through `BrowserMsg`. Maybe you were trying to access opgp directly?

> Also: I'm reading the test results and they're separated in pages cleanly, I see test code and it's kind of like selenium automation but it's not clear when the various pages are switched (which test line in the test output corresponds to which test line in the browser-recipe.ts). Are there any tips you could share in correlating test output with tests?

Are you refering to the stack trace of a failed test? Async stack traces are generally missing some of the call paths in them, because node doesn't keep track of async stacks much. Newer node versions do, maybe upgrading node for tests will give you better stack (whatever newest version there is).

I'm not 100% sure I got your question - a screenshot of what you're trying to match up would help. 

But in general, these tests are mainly meant to be debugged locally.. the CI failures can give you a pointer if you're super familiar with the tests and the tested codebase, but really, when something fails and I don't know why, I'll run it locally to see the browser fail in front of me, then I'll know what is going on.

It's not the most convenient. But yes, the tests are a lot like selenium, except puppeteer is more reliable in my experience.

wiktor-k:
> But in general, these tests are mainly meant to be debugged locally.. the CI failures can give you a pointer if you're super familiar with the tests and the tested codebase, but really, when something fails and I don't know why, I'll run it locally to see the browser fail in front of me, then I'll know what is going on.

Okay. Yeah, the CI results contain screenshots so it's possible to match what's happening but it's not intuitive. I guess this is where these `test-secrets.json` file and the NDA come in?

:thinking: because CI as good as it is I guess won't cut it for me anymore. I'll still try to make is work with the CI but in the long run having the tests run locally would probably let me work more comfortably...

tomholub:
> Okay. Yeah, the CI results contain screenshots so it's possible to match what's happening but it's not intuitive. I guess this is where these test-secrets.json file and the NDA come in?

Correct. In the long run, I'd hope to have fewer and fewer secrets in the test pipeline, and maybe eventually none, so that this would not be needed.

The test secrets contain problematic emails from our customers that we test against, and these emails may contain the user / customer PII, and that's why I need to do the NDA thing.

> thinking because CI as good as it is I guess won't cut it for me anymore. I'll still try to make is work with the CI but in the long run having the tests run locally would probably let me work more comfortably...

That's right. It's not comfortable to try to match it up from the CI without ability to run locally, and involves a lot of frustrated waiting.

tomholub:
In the long run, we could alter all of these test emails to scrub them off customer PII, and then we could just include them in the test cases directly, and then there's no more need for secrets to run 95% of tests.

tomholub:
Also, I think nearly all compose tests could be run without secrets, so we could start marking such tests so that they can ignore the secrets file. I'll make a separate issue.

wiktor-k:
Great! Thanks for thinking about this Tom I really appreciate it. I've browsed the new test results but on CI output they look fine. E.g. expired key test failures: I've reproduced them locally with a key that I made expired, they do post warnings when composing but I think that's what's supposed to happen :) (attester test fails to encrypt to that key but it's probably because the key is expired d'uh).

Maybe running them locally would give more clear answer to what's really the cause...

tomholub:
I'd love to help and run them here but I've been extremely occupied these days. Maybe I could just let it run and record a video, then send you the video to analyze.. better than nothing. Of course, best would be to have you run the tests locally.

wiktor-k:
Yeah, the video looks like a good workaround for now :) I just hope recording it won't take too much of your time. Even if not I can postpone this one and take a look at other tickets. I'd still want to merge this in ASAP because I can feel it will trigger conflicts if stale but it's not critically important.

tomholub:
Yes, this type of change will get stale real fast, so it's good to merge it. I'll record the video.

tomholub:
Now I see that it's just one failure, I imagined it's a load of them. This will be quick

tomholub:
After updating the key in settings, it's still using the expired key. That means the update form did not really update it.

wiktor-k:
D'oh! The fix for the final failure was just a one-liner :man_facepalming: Thanks for the tip Tom, after your description it was trivial to reproduce (with some GnuPG hackery to get the expired and not-expired own key)...

I'll see what other cleanup can I make and mark it as ready for review soon. After casual look I see there are quite a few places that could be simplified (esp. inside pgp-key.ts e.g. longid).

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -124,14 +124,14 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
-          const parsed = await opgp.key.readArmored(lookupResult.pubkey);
+        if (lookupResult.pubkey && lookupResult.pubkey.type === 'openpgp') {
+          const parsed = await opgp.key.readArmored(lookupResult.pubkey.unparsed);
           const key = parsed.keys[0];
           if (!key) {
             console.info('Dropping found but incompatible public key', { for: email, err: parsed.err ? ' * ' + parsed.err.join('\n * ') : undefined });
             lookupResult.pubkey = null; // tslint:disable-line:no-null-keyword
           } else if (! await PgpKey.usableForEncryption(lookupResult.pubkey) && ! await PgpKey.expired(key)) { // Not to skip expired keys
-            console.info('Dropping found+parsed key because getEncryptionKeyPacket===null', { for: email, fingerprint: await PgpKey.fingerprint(parsed.keys[0]) });
+            console.info('Dropping found+parsed key because getEncryptionKeyPacket===null', { for: email, fingerprint: parsed.keys[0].getFingerprint().toUpperCase() });
```
tomholub, line 134: Ideally getFingerprint would already be returning uppercased version, else for sure a similar bug will resurface. We always use uppercase everywhere, so it would be better to uppercase it in the common choke point.

tomholub:
> D'oh! The fix for the final failure was just a one-liner man_facepalming Thanks for the tip Tom, after your description it was trivial to reproduce (with some GnuPG hackery to get the expired and not-expired own key)...
> 
> I'll see what other cleanup can I make and mark it as ready for review soon. After casual look I see there are quite a few places that could be simplified (esp. inside pgp-key.ts e.g. longid).

Perfect, thanks!

extension/js/common/platform/store/contact-store.ts
```diff
@@ -130,7 +130,7 @@ export class ContactStore extends AbstractStore {
         throw new Error(`Could not read pubkey as valid OpenPGP key for: ${validEmail}`);
       }
       const keyDetails = await PgpKey.details(k);
-      const expiresOnMs = Number(await PgpKey.dateBeforeExpirationIfAlreadyExpired(k)) || undefined;
+      const expiresOnMs = Number(pubkey) || undefined;
```
tomholub, line 133: This looks incorrect, unless I missed something

wiktor-k: Yes, definitely. Thanks for noticing!

wiktor-k:
:thinking: I think enterprise tests fail because of timeout (keyword being `(attempt 1 of 3) Retrying: Error: Protocol error (Performance.enable): Target closed.`)...

test/source/test.ts
```diff
@@ -26,12 +26,12 @@ export const internalTestState = { expectiIntentionalErrReport: false }; // upda
...
-  TIMEOUT_EACH_RETRY: minutes(3),
-  TIMEOUT_ALL_RETRIES: minutes(13), // this has to suffer waiting for semaphore between retries, thus almost the same as below
-  TIMEOUT_OVERALL: minutes(14),
-  ATTEMPTS: testGroup === 'STANDARD-GROUP' ? oneIfNotPooled(3) : process.argv.includes('--retry=false') ? 1 : 3,
-  POOL_SIZE: oneIfNotPooled(isMock ? 12 : 2),
+  TIMEOUT_SHORT: minutes(5),
+  TIMEOUT_EACH_RETRY: minutes(15),
+  TIMEOUT_ALL_RETRIES: minutes(65), // this has to suffer waiting for semaphore between retries, thus almost the same as below
+  TIMEOUT_OVERALL: minutes(70),
+  ATTEMPTS: testGroup === 'STANDARD-GROUP' ? oneIfNotPooled(6) : process.argv.includes('--retry=false') ? 2 : 6,
```
tomholub, line 33: Don't change the retry count, that will just increase the test time.

It's probably timing out due to too many failures, which it has to keep retrying.

Increasing the timeouts but bringing back the retry count has a better chance for the tests to finish.

Also maybe cron needs a new cert, I'll look into that separately.

tomholub:
I'm trying it above, let's see.

In general, it's better to avoid force-pushing into your branch - in case I have pushed something here too - so my changes don't get accidentally overriden, which is all to easy when force-pushing. Having a long commit history with many fixes is fine, I'll be squash-merging it later anyway.

wiktor-k:
Ah, okay, thanks for the info. Different teams have different git strategies :)

The timeouts commits was just a try to see if it improves, thanks for reverting and looking at the matter! :+1:

test/source/test.ts
wiktor-k: :+1: thanks!

extension/chrome/elements/add_pubkey.ts
```diff
@@ -72,7 +72,7 @@ View.run(class AddPubkeyView extends View {
     if ($(fromSelect).val()) {
       const [contact] = await ContactStore.get(undefined, [String($(fromSelect).val())]);
       if (contact?.pubkey) {
-        $('.pubkey').val(contact.pubkey).prop('disabled', true);
+        $('.pubkey').val(PgpKey.serializeToString(contact.pubkey)).prop('disabled', true);
```
tomholub, line 75: consider calling `armor`

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -125,13 +124,9 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
-          const parsed = await opgp.key.readArmored(lookupResult.pubkey);
-          const key = parsed.keys[0];
-          if (!key) {
-            console.info('Dropping found but incompatible public key', { for: email, err: parsed.err ? ' * ' + parsed.err.join('\n * ') : undefined });
-            lookupResult.pubkey = null; // tslint:disable-line:no-null-keyword
-          } else if (! await PgpKey.usableForEncryption(lookupResult.pubkey) && ! await PgpKey.expired(key)) { // Not to skip expired keys
-            console.info('Dropping found+parsed key because getEncryptionKeyPacket===null', { for: email, fingerprint: await PgpKey.fingerprint(parsed.keys[0]) });
+          const key = lookupResult.pubkey;
+          if (!key.usableForEncryption && !PgpKey.expired(key)) { // Not to skip expired keys
+            console.info('Dropping found+parsed key because getEncryptionKeyPacket===null', { for: email, fingerprint: key.id });
```
tomholub, line 129: It may be good to propagate this info to the user as a modal or similar? (make another issue)

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -164,14 +159,14 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
         return;
       }
       if (!contact.pubkey_last_sig) {
-        const lastSig = await PgpKey.lastSig(await PgpKey.read(contact.pubkey));
+        const lastSig = +contact.pubkey.lastModified;
```
tomholub, line 162: Number() or better .getTime()

extension/chrome/elements/pgp_block.htm
```diff
@@ -31,6 +31,7 @@
   <script src="/lib/emailjs/emailjs-addressparser.js"></script>
   <script src="/lib/emailjs/emailjs-mime-builder.js"></script>
   <script src="/lib/emailjs/emailjs-mime-parser.js"></script>
+  <script src="/lib/openpgp.js"></script>
```
tomholub, line 34: This is an error

wiktor-k:
I'm going to stare at this all-green status for a few minutes... :)

tomholub:
Notice the LGTM report - this may also be causing the failed tests.

wiktor-k:
Yep, I'll be working on it today, noticed that there is a new bot. On the first sight it looked like there are good old regular test failures :)

extension/js/common/core/openpgp-key.ts
```diff
@@ -79,12 +136,12 @@ export class OpenPGPKey {
...
       // tslint:disable-next-line: no-unsafe-any
       identities: pubkey.users.map(u => u.userId).filter(u => !!u && u.userid && Str.parseEmail(u.userid).email).map(u => u!.userid).filter(Boolean) as string[],
-      lastModified: new Date(await PgpKey.lastSigOpenPGP(pubkey)),
+      lastModified,
       expiration: exp instanceof Date ? exp : undefined,
       created: pubkey.getCreationTime(),
       checkPassword: _text => Promise.resolve(false),
-      fullyDecrypted: pubkey.isFullyDecrypted(),
-      fullyEncrypted: pubkey.isFullyEncrypted(),
+      fullyDecrypted: pubkey.isPublic() ? true /* public keys are always decrypted */ : pubkey.isFullyDecrypted(),
```
tomholub, line 145: I think, if anything, public keys should show "undefined", as in not applicable

wiktor-k: Yep, sounds good. I saw that these two only make sense for private keys but if I'm not mistaken some code checks for these properties on public keys too (maybe this is a side effect). I'll diagnose it further when I get the tests in shape.

Also: there are some optimizations when a key is `encrypt`ed or `decrypt`ed as I think only these two properties change.... this is WIP still :)

wiktor-k:
:thinking: the same expiry issue I had previously... I'll continue tomorrow... it's just *one* failing test...

wiktor-k:
Oh, it seems get-stdin requires Node 10 since [a week ago](https://github.com/sindresorhus/get-stdin/commit/96ebc2ee15b3a77fc1b6e7eca86ac72d490e83ca). That makes code quality tests not happy:

```
/home/semaphore/git/flowcrypt-browser/node_modules/get-stdin/index.js:13
for await (const chunk of stdin) {
    ^^^^
SyntaxError: Unexpected reserved word
```

I'll try to fix that...

tomholub:
> Oh, it seems get-stdin requires Node 10 since [a week ago](https://github.com/sindresorhus/get-stdin/commit/96ebc2ee15b3a77fc1b6e7eca86ac72d490e83ca). That makes code quality tests not happy:
> 
> ```
> /home/semaphore/git/flowcrypt-browser/node_modules/get-stdin/index.js:13
> for await (const chunk of stdin) {
>     ^^^^
> SyntaxError: Unexpected reserved word
> ```
> 
> I'll try to fix that...
> 
> Edit: phew, okay, that did it.

I think we ran into the same issue in https://github.com/FlowCrypt/flowcrypt-browser/pull/2772 . CC @seisvelas I think you can use the fix above in your PR too:

image

@wiktor-k not sure if this should also go in the setup instructions in project readme? I had trouble building locally, too (you'll notice if you delete node_modules and try to reinstall locally)

wiktor-k:
> @wiktor-k not sure if this should also go in the setup instructions in project readme? I had trouble building locally, too (you'll notice if you delete node_modules and try to reinstall locally)

Yeah, sounds like a good idea :+1: I actually had node 14.3.0 installed but the official minimum supported version should definitely be stated somewhere.

wiktor-k:
Okay, I'll provide a quick update as I think this PR will very soon be ready to be merged:
- `readAsOpenPGP` and `unparsed` temporary measures have been removed,
- most of the code uses generic `PgpKey.parse` function,
- there are couple of places that still relay on raw `opgp.key.readArmored` function: key-import-ui, compatibility and pgp_pubkey. I'm investigating how to delegate to `PgpKey.parse` and avoid direct `opgp` calls. They use internal info about subkeys and/or read multiple keys. `PgpKey.parse` doesn't expose or support that for now,
- there are a couple of functions that currently just forward to OpenPGPKey for implementation, after this PR is merged I'll add implementation for S/MIME keys as part of #2735,
- after I'm done with the quick fixes I need to do manual tests - the CI tests are thorough and passing them all takes some time but I want to be double-sure it really works as the changes here are extensive.

wiktor-k:
Okay, I think this is in a shape that you can review. I've done manual tests and for my cases it works fine. I also did fix the import of HTTPS certs: it displays warning that the key cannot be used for encryption (just like if you imported an OpenPGP key without encryption subkey).

I'll be AFK tomorrow but I'll read comments. See you later! :wave:

tomholub:
Thanks, I'll have a look!

tomholub:
I've attempted to draw a rough line to separate pgp methods from general methods, and a folder structure that should over time reinforce that separation. That way it will be easier to move in that direction in the following PRs

extension/chrome/elements/pgp_pubkey.ts
```diff
@@ -52,30 +60,29 @@ View.run(class PgpPubkeyView extends View {
...
-      $('.line.fingerprints .fingerprint').text(Str.spaced(await PgpKey.fingerprint(this.primaryPubKey) || 'err'));
+      $('.line.fingerprints .fingerprint').text(Str.spaced(this.primaryPubKey?.id || 'err'));
     } else {
       $('.line.fingerprints').css({ display: 'none' });
     }
     if (this.primaryPubKey) {
-      const isUsableButExpired = await PgpKey.usableButExpired(this.primaryPubKey);
-      if (!isUsableButExpired && await Catch.doesReject(this.primaryPubKey.getEncryptionKey()) && await Catch.doesReject(this.primaryPubKey.getSigningKey())) {
+      const isUsableButExpired = this.primaryPubKey.usableButExpired;
+      if (!isUsableButExpired && await !this.primaryPubKey.usableForEncryption && !this.primaryPubKey.usableForSigning) {
```
tomholub, line 69: Here refactored correctly

extension/chrome/settings/modules/compatibility.ts
```diff
@@ -33,11 +34,11 @@ View.run(class CompatibilityView extends View {
     $('pre').text('').css('display', 'block');
     try {
       this.testIndex = 1;
-      const { keys, errs } = await PgpKey.readMany(Buf.fromUtfStr(keyString));
+      const { keys, errs } = await KeyUtil.readMany(Buf.fromUtfStr(keyString));
       for (const err of errs) {
         this.appendResult(`Error parsing input: ${String(err)}`);
       }
-      await this.outputKeyResults(keys);
+      await this.outputKeyResults(await Promise.all(keys.map(key => opgp.key.readArmored(KeyUtil.armor(key)).then(result => result.keys[0]))));
```
tomholub, line 41: Todo - check if we have a test for this compatibility page

extension/chrome/settings/modules/contacts.ts
```diff
@@ -141,7 +142,7 @@ View.run(class ContactsView extends View {
     const email = $('#edit_contact .input_pubkey').attr('email');
     if (!armoredPubkey || !email) {
       await Ui.modal.warning('No public key entered');
-    } else if (await PgpKey.fingerprint(armoredPubkey)) {
+    } else if (await PgpKey.fingerprint(await KeyUtil.parse(armoredPubkey))) {
```
tomholub, line 145: I wonder about code like this - there are quite a few similar usages. The intention of the original code is that some OpenPGP keys - when we try to parse them - return a null for fingerprint. But the type of `Key.id` is not optional, so I wonder where would such null go. Maybe it would throw earlier? Maybe the `.id` would actually contain a null, even if the type says it doesn't? This would be something to investigate, plus make sure we are testing this particular code branch.

extension/chrome/settings/modules/my_key_user_ids.ts
```diff
@@ -29,9 +29,8 @@ View.run(class MyKeyUserIdsView extends View {
     [this.primaryKi] = await KeyStore.get(this.acctEmail, [this.fingerprint]);
     Assert.abortAndRenderErrorIfKeyinfoEmpty(this.primaryKi);
     $('.action_show_public_key').attr('href', this.myKeyUrl);
-    const prv = await PgpKey.read(this.primaryKi.private);
-    const userIds = prv.users.map(u => u.userId).filter(Boolean).map(uid => uid!.userid); // todo - create a common function in settings.js for here and setup.js user_ids
-    Xss.sanitizeRender('.user_ids', userIds.map((uid: string) => `<div>${Xss.escape(uid)}</div>`).join(''));
+    const prv = await KeyUtil.parse(this.primaryKi.private);
+    Xss.sanitizeRender('.user_ids', prv.identities.map((uid: string) => `<div>${Xss.escape(uid)}</div>`).join(''));
```
tomholub, line 33: This is interesting - `identities` are the full uid strings, vs `emails` which is parsed / normalized?

extension/chrome/settings/setup.htm
```diff
@@ -185,7 +185,7 @@ <h1></h1>
         <div class="source_paste_container setup_page display_none">
           <div class="line left">
             <textarea class="input_private_key" placeholder="ASCII Armored Private Key"
-              data-test="input-step2bmanualenter-ascii-key"></textarea>
+              data-test="input-step2bmanualenter-ascii-key" spellcheck="false"></textarea>
```
tomholub, line 188: :+1:

extension/js/common/core/crypto/key.ts
```diff
@@ -0,0 +1,238 @@
...
+    } catch (error) {
+      Catch.reportErr(error);
+      return { normalized: '', keys: [] };
+    }
+  }
+
+  public static getKeyType = (pubkey: string): 'openpgp' | 'x509' | 'unknown' => {
+    if (pubkey.startsWith('-----BEGIN CERTIFICATE-----')) {
+      return 'x509';
+    } else if (pubkey.startsWith('-----BEGIN PGP ')) {
```
tomholub, line 212: It may be worth to test for the full pgp header (once for public, once for private) so that pgp messages end up being `unknown`.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -0,0 +1,417 @@
...
+      throw new Error(`Cannot encrypt a key that has ${encryptedPacketCount} of ${secretPackets.length} private packets still encrypted`);
+    }
+    await prv.encrypt(passphrase);
+    if (!prv.isFullyEncrypted()) {
+      throw new Error('Expected key to be fully encrypted after prv.encrypt');
+    }
+    await OpenPGPKey.wrap(prv, key);
+  }
+
+  public static decrypt = async (message: OpenPGP.message.Message, privateKeys: Key[], passwords?: string[]) => {
```
tomholub, line 88: Method `OpenPGPKey.decrypt` sounds like it would be decrypting a key, but it's actually decrypting a message

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -0,0 +1,417 @@
...
+    return await message.decrypt(privateKeys.map(key => OpenPGPKey.unwrap(key)), passwords, undefined, false);
+  }
+
+  public static encrypt: PgpMsgMethod.Encrypt = async ({ pubkeys, signingPrv, pwd, data, filename, armor, date }) => {
```
tomholub, line 92: Same here, originally I think these were in `PgpMsg` where the naming matched

extension/js/common/core/crypto/pgp/openpgp-key.ts
tomholub: Additionally, I remember this method being very large, but maybe I'm remembering another one - to look into

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -0,0 +1,417 @@
...
+  }
+
+  public static reformatKey = async (privateKey: Key, passphrase: string, userIds: { email: string | undefined; name: string }[], expireSeconds: number) => {
+    const origPrv = OpenPGPKey.unwrap(privateKey);
+    const keyPair = await opgp.reformatKey({ privateKey: origPrv, passphrase, userIds, keyExpirationTime: expireSeconds });
+    return await OpenPGPKey.wrap(keyPair.key, {} as Key);
+  }
+
+  // TODO: should be private, will change when readMany is rewritten
+  public static wrap = async (pubkey: OpenPGP.key.Key, pkey: Key, raw?: string): Promise<Key> => {
```
tomholub, line 139: Maybe we could find another name for this, but maybe not. I'm thinking `toKeyObj` which may be a worse name.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -0,0 +1,417 @@
...
+      .filter(userId => userId !== null)
+      .map((userId: OpenPGP.packet.Userid) => {
+        try {
+          return opgp.util.parseUserId(userId.userid).email || '';
+        } catch (e) {
+          // ignore bad user IDs
+        }
+        return '';
+      })
+      .filter(email => email)
```
tomholub, line 167: First I would also `trim`, just in case

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -0,0 +1,417 @@
...
+      throw new Error('Unsupported key type: ' + pubkey.type);
+    }
+    const extensions = pubkey as unknown as { raw: string };
+    if (!extensions.raw) {
+      throw new Error('Object has type == "openpgp" but no raw key.');
+    }
+    return extensions.raw;
+  }
+
+  private static unwrap = (pubkey: Key) => {
```
tomholub, line 244: Maybe wrap/unwrap is indeed a good name

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -0,0 +1,417 @@
...
+    try {
+      await key.getEncryptionKey(undefined, oneSecondBeforeExpiration);
+      return true;
+    } catch (e) {
+      return false;
+    }
+  }
+}
+
+export class PgpKey {
```
tomholub, line 279: I moved it here to have pgp key things together, we could do something about this in another PR

extension/js/common/core/crypto/pgp/pgp-msg.ts
```diff
@@ -222,53 +215,28 @@ export class PgpMsg {
   }
 
   public static encrypt: PgpMsgMethod.Encrypt = async ({ pubkeys, signingPrv, pwd, data, filename, armor, date }) => {
```
tomholub, line 217: Looks like this class should be also renamed, maybe to just `MsgUtil` instead of `PgpMsg`.. actually, almost everything in the core class should become some sort of XxxUtil, the way we are using it.

This would be for another PR.

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -0,0 +1,56 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
+import * as forge from 'node-forge';
+import { Key, KeyUtil } from '../key.js';
+
+export class SmimeKey {
+
+  public static parse = async (text: string): Promise<Key> => {
+    const certificate = forge.pki.certificateFromPem(text);
+    const email = (certificate.subject.getField('CN') as { value: string }).value;
```
tomholub, line 9: Here would be good to try to parse the email using `Str.parseEmail` to make sure it's valid, and to provide a lowercased / normalized version of it. Else consider it an invalid pubkey

extension/js/common/platform/store/contact-store.ts
```diff
@@ -103,50 +102,21 @@ export class ContactStore extends AbstractStore {
...
-      }
-      const keyDetails = await PgpKey.details(k);
-      if (!lastSig) {
-        lastSig = await PgpKey.lastSig(k);
-      }
-      const expiresOnMs = Number(await PgpKey.dateBeforeExpirationIfAlreadyExpired(k)) || undefined;
+      const pk = await KeyUtil.parse(pubkey);
+      const expiresOnMs = Number(pk.expiration) || undefined;
       return {
         email: validEmail,
```
tomholub, line 108: This whole object probably needs refactoring as it duplicates info, that would be for another PR

extension/js/common/platform/store/contact-store.ts
```diff
@@ -194,10 +164,8 @@ export class ContactStore extends AbstractStore {
         throw new Error('contact not found right after inserting it');
       }
     }
-    if (update.pubkey && update.pubkey.includes(PgpArmor.headers('privateKey').begin)) { // wrongly saving prv instead of pub
-      Catch.report('Wrongly saving prv as contact - converting to pubkey');
-      const key = await PgpKey.read(update.pubkey);
-      update.pubkey = key.toPublic().armor();
+    if (update.pubkey) {
```
tomholub, line 167: Please add back the `      Catch.report('Wrongly saving prv as contact - converting to pubkey');` - this is meant to inform us so that we can fix such bugs, especially after large refactors

extension/js/common/settings.ts
```diff
@@ -153,9 +153,9 @@ export class Settings {
   }
 
   public static renderPrvCompatFixUiAndWaitTilSubmittedByUser = async (
-    acctEmail: string, containerStr: string | JQuery<HTMLElement>, origPrv: OpenPGP.key.Key, passphrase: string, backUrl: string
-  ): Promise<OpenPGP.key.Key> => {
-    const uids = origPrv.users.map(u => u.userId).filter(u => !!u && u.userid && Str.parseEmail(u.userid).email).map(u => u!.userid).filter(Boolean) as string[];
+    acctEmail: string, containerStr: string | JQuery<HTMLElement>, origPrv: Key, passphrase: string, backUrl: string
+  ): Promise<Key> => {
+    const uids = origPrv.identities;
```
tomholub, line 158: Todo - double check that something similar to `     const uids = origPrv.users.map(u => u.userId).filter(u => !!u && u.userid && Str.parseEmail(u.userid).email).map(u => u!.userid).filter(Boolean) as string[];` is done when creating `identities`

extension/chrome/elements/backup.ts
```diff
@@ -73,7 +71,7 @@ View.run(class BackupView extends View {
   }
 
   private testPassphraseHandler = async () => {
-    if (await PgpKey.decrypt(await PgpKey.read(this.armoredPrvBackup), String($('#pass_phrase').val())) === true) {
+    if (await (await KeyUtil.parse(this.armoredPrvBackup)).checkPassword(String($('#pass_phrase').val())) === true) {
```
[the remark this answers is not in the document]
tomholub: But should be called `checkPassPhrase`

tomholub:
I've broken the content scripts, which I'll go ahead and fix

tomholub:
We need a test for searching contacts when accessing Secure Compose from Gmail. The IDBCursor does not like async to be mixed while a transaction is ongoing.

image

Earlier this was passing all tests, meaning we are not testing this.

extension/js/common/settings.ts
tomholub: Yup, looks good

extension/js/common/platform/store/contact-store.ts
tomholub: Ok for now

extension/chrome/settings/modules/compatibility.ts
tomholub: We don't, I'll test it manually + created issue https://github.com/FlowCrypt/flowcrypt-browser/issues/2789

tomholub: Shows an exception when I paste public key from https://flowcrypt.com/pub/tom@flowcrypt.com

```
(17) PK 0 > SK 0 > Verify: [TypeError: Cannot read property 'Symbol(byValue)' of undefined]  
```

extension/chrome/settings/modules/my_key_user_ids.ts
tomholub: identities have full uids but only those that have valid email in them. I'll write in a comment

tomholub:
I've tried to clean up my own PR comments, and there's a few left:

https://github.com/FlowCrypt/flowcrypt-browser/pull/2739#discussion_r439373389
https://github.com/FlowCrypt/flowcrypt-browser/pull/2739#discussion_r439490758
https://github.com/FlowCrypt/flowcrypt-browser/pull/2739#discussion_r439380214
https://github.com/FlowCrypt/flowcrypt-browser/pull/2739#discussion_r439384106
https://github.com/FlowCrypt/flowcrypt-browser/pull/2739#discussion_r439384295

I may still be pushing a commit or two, depending how the tests turn out.

tomholub:
These errors seem to be due to failing google login again.

I'll eventually set up Google login to delegate the auth to a 3rd party IdP, which will be easier to script for.

extension/chrome/elements/backup.ts
```diff
@@ -34,14 +32,14 @@ View.run(class BackupView extends View {
...
-    const fingerprint = await PgpKey.fingerprint(prvBackup);
+    const prvBackup = await KeyUtil.parse(this.armoredPrvBackup);
+    const fingerprint = prvBackup.id;
     if (!fingerprint) {
       throw new Error('Missing backup key fingerprint');
     }
     if (prvBackup) {
       $('.line.fingerprints .fingerprint').text(Str.spaced(fingerprint));
-      if (await Catch.doesReject(prvBackup.getEncryptionKey()) && await Catch.doesReject(prvBackup.getSigningKey())) {
+      if (prvBackup.usableForEncryption && prvBackup.usableForSigning) {
```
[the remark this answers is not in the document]
wiktor-k: Ugh, I already caught a couple of mistakes like that but to see yet another one... :(

Thanks for your review!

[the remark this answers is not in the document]
wiktor-k: Ugh, I already caught a couple of mistakes like that but to see yet another one... :(

Thanks for your review!

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -85,14 +83,16 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
+    const armoredPubkeys = [{ pubkey: await KeyUtil.parse(senderKi.public), email: senderEmail, isMine: true }];
     const emailsWithoutPubkeys = [];
     for (const i of contacts.keys()) {
       const contact = contacts[i];
       if (contact && contact.has_pgp && contact.pubkey) {
         armoredPubkeys.push({ pubkey: contact.pubkey, email: contact.email, isMine: false });
       } else if (contact && this.ksLookupsByEmail[contact.email] && this.ksLookupsByEmail[contact.email].pubkey) {
-        armoredPubkeys.push({ pubkey: this.ksLookupsByEmail[contact.email].pubkey!, email: contact.email, isMine: false }); // checked !null right above. Null evaluates to false.
+        const pubkey = this.ksLookupsByEmail[contact.email].pubkey!;
+        const key = typeof pubkey === 'string' ? await KeyUtil.parse(pubkey) : pubkey;
```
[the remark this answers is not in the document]
wiktor-k: Good idea, I'll take a look at it!

Or... did you already took care of it? Not sure what "resolved conversation" means in this case but I'm reading through your comments one-by-one :)

extension/chrome/settings/modules/compatibility.ts
wiktor-k: Good catch. I'll take a look at it.

extension/chrome/settings/modules/contacts.ts
wiktor-k: :thinking: do you have such a key for tests? I'd like to see how would the app behave.

If we need to support them then yeah `id` needs to be nullable. If not, I'd rather throw on parse rather to check for nulls in the entire codebase...

extension/chrome/settings/modules/my_key_user_ids.ts
wiktor-k: Yeah, it seems there are two usages of UIDs: one is for display (`identities`) and one is for lookup (`emails`). This is usually grouped in one property that's an array of objects that have full-UID and e-mail properties inside but based on the usages I saw it seems identities/emails split works well too...

extension/js/common/core/crypto/key.ts
wiktor-k: Ugh, right! :facepalm:

extension/js/common/core/crypto/pgp/openpgp-key.ts
wiktor-k: Yes! This should be renamed as there already is `encryptKey` method. And yeah, there is one in pgp-msg but that got split into two that you see now (OpenPGP/SMIME).

extension/chrome/elements/backup.ts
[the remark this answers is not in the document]
tomholub: The important thing to do here is to write a test for this scenario. Else it will keep happening.

extension/chrome/settings/modules/contacts.ts
tomholub: I'd also rather throw early. I don't have any such key now (I may have encountered this before starting to write tests..) but as long as it does throw early, that's cool with me - there is nothing useful we can do with such keys.

extension/js/common/core/crypto/pgp/openpgp-key.ts
wiktor-k: :thinking: or `toPubkey` or however will that interface be called. I meant this to be used internally but as the comment explains this is still not yet possible.

extension/chrome/elements/backup.ts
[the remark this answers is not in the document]
wiktor-k: Haha, yeah, this part obvious, I was just mentioning that this already caused me problems so I could scan for it proactively. Now I'll look for `Catch.doesReject` with more suspicion! Or better wrap in tests any logic that touches them before refactoring...

extension/js/common/core/crypto/pgp/openpgp-key.ts
wiktor-k: I still have some half-baked ideas in my head that I'm not sure if they make sense... for example during refactorings I saw that the Pubkey interface is *mostly* immutable. The only thing that could change it was encrypting and decrypting key. That required some changes in the internal API but I'm thinking whether it would be a good idea to just have separate interfaces for encrypted and decrypted keys and return them from the encrypt/decrypt methods. This is tricky because OpenPGP.js keys are internally mutable so we would either clone the objects or...

The other thing is Pubkeys travelling across realms: they lose the associated OpenPGP.js keys because (obviously) this is not JSON serializable. This doesn't seem to be a big problem now but still something that keeps me awake at night... ;)

extension/js/common/core/crypto/smime/smime-key.ts
wiktor-k: Yes, good idea. We don't see this error normally because in the UI we check for the `emailProtection` extended key usage but a self-signed cert with this bit but malformed e-mail can exist.

I'll write a unit test for that as this (parsing) seems to be small and relatively self-contained.

extension/js/common/platform/store/contact-store.ts
wiktor-k: Yes, agreed fully! I even thought about removing this method altogether but did not measure the impact yet :)

extension/js/common/platform/store/contact-store.ts
wiktor-k: Whops, sorry, removed by mistake.

tomholub:
I'm just doing my approvals here since it's easier to approve individual commits now, compared to going through the whole PR again, especially after I rearranged the folders.

wiktor-k:
By the way: I experimented with running parsing unit tests in node without browsers (temporary branch `scratch-smime-tests`). Good news is that it works, bad news is that we'd need to extract code that doesn't use any browser-isms (like `window`) to either separate classes/modules or `if`-guard them.

Take a look and tell me what do you think about it. I think it's worth pursuing as we'd have very quick tests of critical functionality but this would require some further refactorings (like extracting `Key` interface to separate file and splitting `Str` into two classes: simple one that doesn't touch browser APIs and the rest).

tomholub:
> (temporary branch `scratch-smime-tests`

The whole core folder is supposed to be able to be used on node, without a window environment.

If a `window` is used somewhere in them, inside a particular method, that should not affect the ability to use remaining stuff on node.

tomholub:
If you notice, we have a symbolic link on the core folder that also copies it into tests folder. Tests run on node, and it uses functionality found there.

wiktor-k:
> The whole core folder is supposed to be able to be used on node, without a window environment.

I think there maybe a leak in this scenario: after including full Str I get compilation errors on node. It looks like Str (`core/common.ts`) is importing `xss.ts` that uses `dompurify` that conflicts with definitions node uses. :thinking: maybe it's possible to adjust what node.js is using to avoid conflicts... I'll check it out.

tomholub:
> > The whole core folder is supposed to be able to be used on node, without a window environment.
> 
> I think there maybe a leak in this scenario: after including full Str I get compilation errors on node. It looks like Str (`core/common.ts`) is importing `xss.ts` that uses `dompurify` that conflicts with definitions node uses. thinking maybe it's possible to adjust what node.js is using to avoid conflicts... I'll check it out.

You're importing Str from the wrong path - from the one made for browser.

When woring from node, import from `test/source/core/common.ts`, not from `extension/js/common/core/common.ts`.

The files are the same. But you'll notice `../platform/xss.ts` files are different (from the perspective of the `common.ts` file.

wiktor-k:
> You're importing Str from the wrong path - from the one made for browser.
> When woring from node, import from test/source/core/common.ts, not from extension/js/common/core/common.ts.

`smime-key.ts` is importing common.ts via `../../common.ts` path:

https://github.com/FlowCrypt/flowcrypt-browser/pull/2739/files#diff-0dd065ae44fee7c015604a4f26055350R4

> The files are the same. But you'll notice ../platform/xss.ts files are different (from the perspective of the common.ts file.

Do you think it should work? Because I can't modify smime-key.ts for the purpose of tests and if I remove it (like in the `scratch-...` branch) `npm run pretest` works correctly...

tomholub:
> > You're importing Str from the wrong path - from the one made for browser.
> > When woring from node, import from test/source/core/common.ts, not from extension/js/common/core/common.ts.
> 
> `smime-key.ts` is importing common.ts via `../../common.ts` path:
> 
> https://github.com/FlowCrypt/flowcrypt-browser/pull/2739/files#diff-0dd065ae44fee7c015604a4f26055350R4
> 
> > The files are the same. But you'll notice ../platform/xss.ts files are different (from the perspective of the common.ts file.
> 
> Do you think it should work? Because I can't modify smime-key.ts for the purpose of tests and if I remove it (like in the `scratch-...` branch) `npm run pretest` works correctly...

Then import `smime-key.ts` from the right path: from `test/source/core/crypto/smime/smime-key.ts` and not from `extension/js/common/core/crypto/smime/smime-key.ts`.

Of course I could be wrong and you may be actually running into an issue I don't know about, but more likely you're simply importing from the browser core folder instead of the tests core folder.

wiktor-k:
> Then import smime-key.ts from the right path: from test/source/core/crypto/smime/smime-key.ts and not from extension/js/common/core/crypto/smime/smime-key.ts.

Ah, right! Now I think I get why there are two copies of every file... :)

Thanks for the pointer, I'll check it out.

> Of course I could be wrong and you may be actually running into an issue I don't know about, but more likely you're simply importing from the browser core folder instead of the tests core folder.

Nah, I think this is it.

extension/chrome/settings/modules/compatibility.ts
wiktor-k: It seems OpenPGP.js changed some APIs quite some time ago. Previously [`Subkey.prototype.verify`](https://openpgpjs.org/openpgpjs/doc/key.js.html#line1085) returned status codes but [currently](https://raw.githubusercontent.com/openpgpjs/openpgpjs/master/dist/openpgp.js) it throws on errors and just returns `undefined` (nothing) on success.

`keyStatus` enum appears to be completely removed from the source code.

Even if it was unchanged I'm still not sure if this code:

```ts
const verifyResNum = await subKey.verify(key.primaryKey);
const veryfyResWord = opgp.enums.read(opgp.enums.keyStatus, 1);
```

shouldn't lookup by the Num result and maybe it should look like this:

```ts
const verifyResNum = await subKey.verify(key.primaryKey);
const veryfyResWord = opgp.enums.read(opgp.enums.keyStatus, verifyResNum);
```

Still, given the API changes this is not needed. I simplified the test. Blame shows unrelated changes due to `openpgp` -> `opgp` renaming and I didn't dig deeper on how and what the second like was inserted...

extension/chrome/settings/modules/compatibility.ts
```diff
@@ -93,9 +93,8 @@ View.run(class CompatibilityView extends View {
         this.appendResult(`${skn} Created: ${await this.test(async () => this.formatDate(subKey.keyPacket.created))}`);
         this.appendResult(`${skn} Algo: ${await this.test(async () => `${subKey.getAlgorithmInfo().algorithm}`)}`);
         this.appendResult(`${skn} Verify: ${await this.test(async () => {
-          const verifyResNum = await subKey.verify(key.primaryKey);
-          const veryfyResWord = opgp.enums.read(opgp.enums.keyStatus, 1);
-          return `${verifyResNum}: ${veryfyResWord}`;
+          await subKey.verify(key.primaryKey);
```
tomholub, line 96: I figured the new OpenPGP.js api thorws on verify? In this case this was an issue even before this PR.

tomholub:
Let me know when ready to review again :)

extension/chrome/settings/modules/compatibility.ts
wiktor-k: Yes, this was an issue before this PR. I tested that on stable Flowcrypt from Chrome Web Store and the result was the same. I'm not sure which version introduced this bug. (It wasn't very prominent so I'm not surprised that this slipped by. I actually glanced at the compliance output and didn't notice there was a failure :D ).

wiktor-k:
I've got a one thing I'd like to fix before another round and thought it would take just a moment but clearly didn't... :)

wiktor-k:
Tom, could I ask you about the following test (in `test/source/tests/tests/unit.ts`):

```typescript
ava.default(`[unit][MsgBlockParser.detectBlocks] does not get tripped on non-pgp certs`, async t => {
  expect(MsgBlockParser.detectBlocks("This text breaks email and Gmail web app.\n\n-----BEGIN CERTIFICATE-----\n\nEven though it's not a vaild PGP m\n\nMuhahah")).to.deep.equal({
    "blocks": [
      MsgBlock.fromContent("plainText", "This text breaks email and Gmail web app.\n\n-----BEGIN CERTIFICATE-----\n\nEven though it's not a vaild PGP m\n\nMuhahah"),
    ],
    "normalized": "This text breaks email and Gmail web app.\n\n-----BEGIN CERTIFICATE-----\n\nEven though it's not a vaild PGP m\n\nMuhahah"
  });
  t.pass();
});
```

I added the ability to parse S/MIME certs with MsgBlockParser (actually S/MIME and OpenPGP can now be freely mixed in the text) but this text fails now because `BEGIN CERTIFIATE` is a start of S/MIME cert.

Do you think this test should be changed or is there a real bug in MsgBlockParser now that I've added S/MIME parsing there?

Thanks for your time!

wiktor-k:
Okay, I think this completes changes that I wanted to make (minus the failing test). Compatibility test now supports S/MIME keys, `readMany` also (it can still be improved further but I think it's a good baseline) and I got rid of the `PgpKey.fingerprint`.

I think a good plan would be to review it and fix as many bugs as possible and merge it to master. I still have some ideas for further refactorings but I think we already used up quite a bit of our risk budget so let's test it thoroughly and roll it to users as our next step. :)

tomholub:
> Do you think this test should be changed or is there a real bug in MsgBlockParser now that I've added S/MIME parsing there?

I'll update the test, to reflect that we now have cert parsing functionality

> Okay, I think this completes changes that I wanted to make (minus the failing test). Compatibility test now supports S/MIME keys, `readMany` also (it can still be improved further but I think it's a good baseline) and I got rid of the `PgpKey.fingerprint`.
> 
> I think a good plan would be to review it and fix as many bugs as possible and merge it to master. I still have some ideas for further refactorings but I think we already used up quite a bit of our risk budget so let's test it thoroughly and roll it to users as our next step. :)

Yes, I'll do another review. Thanks!

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -5,20 +5,31 @@ import { Catch } from '../../../platform/catch.js';
...
 const internal = Symbol('internal public key');
 
 // todo - OpenPGPKey and PgpKey should be merged into one
 
 export class OpenPGPKey {
 
+  private static readonly encryptionText = 'This is the text we are encrypting!';
+
   public static parse = async (text: string): Promise<Key> => {
+    // TODO: Should we throw if more keys are in the armor?
```
tomholub, line 19: That's a good idea, for another issue.

tomholub:
> but I think we already used up quite a bit of our risk budget

I think we are about 300% over it. I'll be releasing this gradually, first 1% of users, a few days later 3% and so on. Then if we find a major bug, we'd need to fix it and I'd have to turn around and update 100%.

This means followup development should be extremely low risk, so that I can afford to do a full rollout after we have to release the inevitable bugfix after this.

PR #2745, opened 2020-04-26
Do not show 'switch to encrypted reply' warning when replying to signed message

### LINKED ISSUE #2720: don't show plain warning when replying to signed-only message
> I don't like the red notification that says "The last message was encrypted, but you are composing a reply without encryption. Switch to encrypted reply". Not all communication has to be encrypted and I am well aware that I am not writing an encrypted message. Your interface is good enough (and intuitive enough) that such in-your-face messages are not necessary.

You are a bright and knowledgeable user.. not all of our users are.This is a result of us making encryption more accessible - a whole bunch of people who would otherwise not be able to use encryption can use it now, and they span a whole spectrum in terms of their technical abilities, to the extent of not knowing how to attach a file.But we still want even these people to stay secure.

> What makes matters worse is that the mail I am replying to is just signed, not even encrypted.

That's a great point. we'll fix it. There's no need to show the alert when replying to a signed message.

limonte:
Fixes #2720 

I recommend using `?w=1` for CR: https://github.com/FlowCrypt/flowcrypt-browser/pull/2745/files?w=1

### DIFF
--- a/extension/js/common/xss-safe-factory.ts
+++ b/extension/js/common/xss-safe-factory.ts
@@ -8,7 +8,7 @@ import { Dict, Str, Url, UrlParams } from './core/common.js';
 import { Att } from './core/att.js';
 import { Browser } from './browser/browser.js';
 import { Catch } from './platform/catch.js';
-import { MsgBlock } from './core/msg-block.js';
+import { MsgBlock, MsgBlockType } from './core/msg-block.js';
 import { MsgBlockParser } from './core/msg-block-parser.js';
 import { PgpArmor } from './core/pgp-armor.js';
 import { Ui } from './browser/ui.js';
@@ -190,7 +190,7 @@ export class XssSafeFactory {
     return Ui.e('span', { class: 'pgp_attachment', html: this.iframe(this.srcPgpAttIframe(meta, isEncrypted)) });
   }
 
-  public embeddedMsg = (type: string, armored: string, msgId?: string, isOutgoing?: boolean, sender?: string, signature?: string | boolean) => {
+  public embeddedMsg = (type: MsgBlockType, armored: string, msgId?: string, isOutgoing?: boolean, sender?: string, signature?: string | boolean) => {
     return this.iframe(this.srcPgpBlockIframe(armored, msgId, isOutgoing, sender, signature), ['pgp_block', type]) + this.hideGmailNewMsgInThreadNotification;
   }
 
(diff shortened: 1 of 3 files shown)


extension/js/common/xss-safe-factory.ts
```diff
@@ -190,8 +190,8 @@ export class XssSafeFactory {
     return Ui.e('span', { class: 'pgp_attachment', html: this.iframe(this.srcPgpAttIframe(meta, isEncrypted)) });
   }
 
-  public embeddedMsg = (armored: string, msgId?: string, isOutgoing?: boolean, sender?: string, signature?: string | boolean) => {
-    return this.iframe(this.srcPgpBlockIframe(armored, msgId, isOutgoing, sender, signature), ['pgp_block']) + this.hideGmailNewMsgInThreadNotification;
+  public embeddedMsg = (type: string, armored: string, msgId?: string, isOutgoing?: boolean, sender?: string, signature?: string | boolean) => {
```
tomholub, line 193: Please don't use string - `typeof MsgBlock.type` or similar would be better

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -324,20 +326,20 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
-              msgEl = this.updateMsgBodyEl_DANGEROUSLY(msgEl, 'append', this.factory.embeddedMsg('', msgId, false, senderEmail)); // xss-safe-factory
+              msgEl = this.updateMsgBodyEl_DANGEROUSLY(msgEl, 'append', this.factory.embeddedMsg('encryptedMsg', '', msgId, false, senderEmail)); // xss-safe-factory
             }
           } else if (treatAs === 'publicKey') { // todo - pubkey should be fetched in pgp_pubkey.js
             nRenderedAtts = await this.renderPublicKeyFromFile(a, attsContainerInner, msgEl, isOutgoing, attSel, nRenderedAtts);
           } else if (treatAs === 'privateKey') {
             nRenderedAtts = await this.renderBackupFromFile(a, attsContainerInner, msgEl, attSel, nRenderedAtts);
           } else if (treatAs === 'signature') {
-            const embeddedSignedMsgXssSafe = this.factory.embeddedMsg('', msgId, false, senderEmail, true);
+            const embeddedSignedMsgXssSafe = this.factory.embeddedMsg('signature', '', msgId, false, senderEmail, true);
```
tomholub, line 342: Here you would use `signedMsg`

PR #2757, opened 2020-04-30
Add 'decrypt - [gpg] signed fully armored message' test [waiting for Tom]

limonte:
Closes #2589

PR #2760, opened 2020-05-04
DO NOT MERGE: Issue 2589 test wiktor

wiktor-k:
This is a test for #2757. Do *not* merge.

### DIFF
--- a/extension/js/common/core/pgp-msg.ts
+++ b/extension/js/common/core/pgp-msg.ts
@@ -150,7 +150,7 @@ export class PgpMsg {
         // .. as it is now this could allow an attacker to append bogus signatures to validly signed messages, making otherwise correct messages seem incorrect
         // .. which is not really an issue - an attacker that can append signatures could have also just slightly changed the message, causing the same experience
         // .. so for now #wontfix unless a reasonable usecase surfaces
-        sig.match = true; // (sig.match === true || sig.match === null) && await verifyRes.verified;
+        sig.match = (sig.match === true || sig.match === null) && await verifyRes.verified;
         if (!sig.signer) {
           // todo - currently only the first signer will be reported. Should we be showing all signers? How common is that?
           sig.signer = await PgpKey.longid(verifyRes.keyid.bytes);
--- a/extension/js/common/core/pgp-msg.ts
+++ b/extension/js/common/core/pgp-msg.ts
@@ -145,6 +145,13 @@ export class PgpMsg {
       // While this looks like bad method API design, it's here to ensure execution order when 1) reading data, 2) verifying, 3) processing signatures
       // Else it will hang trying to read a stream: https://github.com/openpgpjs/openpgpjs/issues/916#issuecomment-510620625
       const verifyResults = Array.isArray(msgOrVerResults) ? msgOrVerResults : await msgOrVerResults.verify(pubs);
+      if ('getLiteralData' in msgOrVerResults) {
+        const data = msgOrVerResults.getLiteralData();
+        if (data) {
+          (msgOrVerResults as unknown as { data: Uint8Array }).data = await opgp.stream.readToEnd(data);
+        }
+      }
+      console.log(msgOrVerResults);
       for (const verifyRes of verifyResults) {
         // todo - a valid signature is a valid signature, and should be surfaced. Currently, if any of the signatures are not valid, it's showing all as invalid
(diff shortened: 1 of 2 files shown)


limonte:
`<offtopic>`
@wiktor-k a small tip which you can use to _really_ prevent merging, use drafts instead of pull requests: https://github.blog/2019-02-14-introducing-draft-pull-requests/

Also, you can now convert PRs to drafts:

image
`</offtopic>`

wiktor-k:
:facepalm: I pressed the wrong button, damn, sorry :bow:

limonte:
> facepalm I pressed the wrong button, damn, sorry bow

No worries at all! you can convert pull request to draft and vice versa at any time.

wiktor-k:
> No worries at all! you can convert pull request to draft and vice versa at any time.

Today I learned something, thanks! :+1:

tomholub:
> u can convert pull request to draft and vice versa at any time.

I think this wasn't possible in the past, or maybe I overlooked it. Thanks!

limonte:
> > u can convert pull request to draft and vice versa at any time.
> 
> I think this wasn't possible in the past, or maybe I overlooked it. Thanks!

Correct, that feature was added recently: https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/

wiktor-k:
Okay, this is really interesting case. I also did a small PoC in pure openpgpjs but could reproduce it on the given data.

Could someone tell me how to check what's exactly in `msgOrVerResults` and `pubs`:
`   public static verify = async (msgOrVerResults: OpenpgpMsgOrCleartext | OpenPGP.message.Verification[], pubs: OpenPGP.key.Key[], contact?: Contact): `

I think msgOrVerResults is just the `openpgp.message.readArmored` of the `message` query parameter that you have in the test and pubs is a key from https://attester.flowcrypt.com/lookup/flowcrypt.compatibility@gmail.com I did test that locally but it works :(

Now, maybe we can create a smaller example or just run my mini-demo on the CI to check if the small example also fails? (I've re-used Tom's code from the linked openpgp.js)

And yeah... I guess this need to be debugged step-by-step and currently CI builds don't allow me to check what's inside which variable.

In the mean time I'll browse the OpenPGP.js source code ... it seems this cannot be avoided ;)

wiktor-k:
Ha:

```typescript
const verifyResults = keys.signedBy.length ? await decrypted.verify(keys.forVerification) : undefined; // verify first to prevent stream hang
const content = new Buf(await opgp.stream.readToEnd(decrypted.getLiteralData()!)); // read content second to prevent stream hang
const signature = verifyResults ? await PgpMsg.verify(verifyResults, [], keys.verificationContacts[0]) : undefined; // evaluate verify results third to prevent stream hang
```

It seems now OpenPGP.js is mentioning the need to read it in the README: https://github.com/openpgpjs/openpgpjs/blob/master/README.md#streaming-sign-and-verify-uint8array-data

`openpgp.verify` seems to be written in a way that can also protect against cleartext messages being read in a streaming way (as compared to running `verify` on Message objects directly).

tomholub:
Did you say you can or cannot reproduce this locally?

I can try to run the stuff here and see if it reproduces it, and give you
the values if yes.

wiktor-k:
Cannot reproduce it locally. I took the same data that the test is using and wrote some minimal code that should hang but it doesn't:

```typescript
const keys = (await openpgp.key.readArmored(publicKeyArmored)).keys;
const msg2 = await openpgp.message.readArmored(msg);
console.log(msg2);

const verifyResults = await msg2.verify(keys, new Date(), false);
console.log(verifyResults);
console.log(23);
for (const verifyRes of verifyResults) {
    console.log(24);
    console.log(await verifyRes.verified); // stuck here
    console.log(25);
}
```

The main culprit of hanging should be that the message has not been read to the end (`openpgp.verify` supposedly protects against it). But without having a local reproduction it takes too long to investigate it. (I did try inserting reading the stream but it apparently doesn't work). In the openpgp ticket it seems like you were able to reproduce verification hang locally in Chrome.

tomholub:
There's no good reason for it to fail in ci while it works locally. Other
than dependency drift, so deleting node modules may be worth a try.

Also browser versions, but locally run tests use same puppeteer version,
and that one is tied to particular chromium build which is included.
@limonte run out locally and it worked, so.. don't know.

I'll ask you both to drop that PR for now - I'll play with it later this
month.

Thanks!

PR #2772, opened 2020-05-21
Sanitize error messages by removing all tags

### LINKED ISSUE #2770: HTML Injection issue via msgId
This is issue "FLO-02-004 Extension: HTML Injection in error message on certain pages" from the Cure53 report. From the report itself:

> A HTML injection was found on pgp_block.htm via the error message. This is becausethe returned message from the API call is rendered as HTML. However, this issue doesnot introduce XSS because the rendered content is sanitized and only allowed safe tags.Besides, CSP of WebExtension will block this attack nevertheless. At the same time, thiscould still be used for Phishing.PoC (Replace the logged-in email address in the highlighted area):chrome-extension://emghkmengffbgcnpcajidfdkffaikghk/chrome/elements/pgp_block.htm?frameId=&hasPassword=___cu_false___&msgId=%3Cs%3E&senderEmail=&isOutgoing=___cu_true___&acctEmail=EMAIL_ADDRESS&parentTabId=

Screenshot_2020-05-21 FLO-02-report pdf

Currently we are sanitizing this, which prevents it from being a much more serious issue. Nevertheless, this should be fixed by escaping the output so user input is not rendered as HTML in the error message.

seisvelas:
I don't think there is any need to allow tags here, but if I'm wrong please let me know and we'll find a different way. This should prevent HTML injection via error messages per FLO-02-004 of the Cure53 report.

// tom
close #2770

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -36,7 +36,7 @@ blocks:
 
       prologue:
         commands:
-          - sudo sh -c "echo '209.250.232.81 cron.flowcrypt.com' >> /etc/hosts" && sudo sh -c "echo '127.0.0.1 google.mock.flowcrypt.com' >> /etc/hosts"
+          - sudo sh -c "echo '209.250.232.81 cron.flowcrypt.com' >> /etc/hosts" && sudo sh -c "echo '127.0.0.1 google.mock.flowcryptlocal.com' >> /etc/hosts"
           - cache restore sources-$SEMAPHORE_PIPELINE_ID
           - cd ~/git/flowcrypt-browser
           - cache restore node-modules-$SEMAPHORE_GIT_BRANCH-$(cat package.json.sum)
@@ -45,6 +45,8 @@ blocks:
 
         - name: code quality
           commands:
+            - nvm install 10.19.0
+            - nvm alias default 10.19.0
             - npm run-script test_tslint
             - npm run-script test_eslint
             - npm run-script test_stylelint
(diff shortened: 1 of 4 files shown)


seisvelas:
I'm failing nearly every test. I can't imagine this is caused by that small change in error rendering, there must be more going on here than I understand.

seisvelas:
> I think a better way to fix this would be to find where the error gets stringified, like `String(err)` or similar somewhere, and Xss.escape that. I think we do need to sometimes render tags in errors intentionally, but not coming from exception descriptions.

Alright, I changed

seisvelas:
> I think a better way to fix this would be to find where the error gets stringified, like `String(err)` or similar somewhere, and Xss.escape that. I think we do need to sometimes render tags in errors intentionally, but not coming from exception descriptions.

Okay, I changed the line

    await this.renderErr(String(e), this.view.encryptedMsgUrlParam ? this.view.encryptedMsgUrlParam.toUtfStr() : undefined);

To

    await this.renderErr(Xss.escape(String(e)), this.view.encryptedMsgUrlParam ? this.view.encryptedMsgUrlParam.toUtfStr() : undefined);

Built it, and ran it in Chrome. I browsed to `      await this.renderErr(Xss.escape(String(e)), this.view.encryptedMsgUrlParam ? this.view.encryptedMsgUrlParam.toUtfStr() : undefined);` and it is correctly escaping the tag:

Screenshot from 2020-05-21 15-21-18

However, I am still failing lots of tests, which I am investigating right now

seisvelas:
All tests are passing :) I copied Wiktor's changes to appease Semaphore

Issue #2592, opened 2020-02-26
SweetAlert2 is adding padding to the document in Chrome


limonte:
![Peek 2020-02-26 14-40](https://user-images.githubusercontent.com/6059356/75347084-b43f0380-58a8-11ea-9676-259f05e4a49f.gif)

PR #2793, opened 2020-06-23
Bump sweetalert2 to v9.15.2

### LINKED ISSUE #2792: Confirm dialog inside settings page cause horizontal jumpiness
While working on #2764 I bumped into this visual issue:

This is happening because of https://github.com/FlowCrypt/flowcrypt-browser/issues/2592

limonte:
Fixes #2792

### DIFF
--- a/extension/js/common/browser/ui.ts
+++ b/extension/js/common/browser/ui.ts
@@ -416,6 +416,5 @@ export class Ui {
   private static swal = () => Swal.mixin({
     showClass: { popup: 'swal2-noanimation', backdrop: 'swal2-noanimation' },
     hideClass: { popup: '', backdrop: '' },
-    scrollbarPadding: false,
   })
 }
--- a/extension/css/sweetalert2.css
+++ b/extension/css/sweetalert2.css
@@ -311,7 +311,7 @@
   border-radius: 100%;
   border-color: transparent;
   background-color: transparent !important;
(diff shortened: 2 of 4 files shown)


limonte:
@tomholub the same issue with live gmail tests as in #2778 is happening here:

```
> node ./node_modules/ava/cli --verbose --concurrency=1 build/test/test/source/test.js -- CONSUMER-LIVE-GMAIL STANDARD-GROUP

  ✖ Couldn't find any files to test
```

So I guess it's something unrelated to both PRs and should be investigated separately, reported here: #2794

extension/js/common/browser/ui.ts
```diff
@@ -416,6 +416,5 @@ export class Ui {
   private static swal = () => Swal.mixin({
     showClass: { popup: 'swal2-noanimation', backdrop: 'swal2-noanimation' },
     hideClass: { popup: '', backdrop: '' },
-    scrollbarPadding: false,
```
limonte, line 419: Here I'm undoing #2593 which wasn't the clear fix for #2592

`scrollbarPadding: false` was kinda workaround which is not needed after https://github.com/sweetalert2/sweetalert2/pull/2006 was released

tomholub: Got it, thanks!

PR #2802, opened 2020-06-26
issue 2771 fix key rm, keep fingerprint html attr

### LINKED ISSUE #2771: key remove button does not work
Looks like we must be missing a test for this, too.

image

tomholub:
close #2771


tomholub:
Should also add a test for this, but I won't have time for it now :(

PR #2805, opened 2020-06-29
Fix Firefox account recovery / Move `checkPassPhrase` outside of Key object

### LINKED ISSUE #2804: Unable to recover an account in Firefox
### Steps to reproduce:

1. Reset FlowCrypt account (Settings -> Additional -> Experimental -> Reset)
2. Recover FlowCrypt account

### Expected behavior

Ability to recover account in Firefox

### Actual behavior:

image

---

The issue is happening here: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/platform/store/contact-store.ts#L131

`Contact.pubkey` is `Key` which can't be cloned by Firefox. 

> I think there are only two suspects to why [structured clone](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) fails: `checkPassPhrase` (it's a function) or symbols on the object. My casual skim over MDN suggests it's the former - and that's good because I wanted to move this function to a helper, static function anyway.

I tried removing checkPassPhrase and it doesn't seem to help, so the internal public key symbol is most probably the cause.

wiktor-k:
This aligns `checkPassPhrase` with other key functions and also allows the Key object to be serialized using structured clone algorithm in Firefox.

Fixes #2804.

I've tested this in Firefox while first reproducing it. It failed on each time when setting up an already used account but if you can test it too I'd be really glad ;)

### DIFF
--- a/extension/chrome/elements/backup.ts
+++ b/extension/chrome/elements/backup.ts
@@ -71,7 +71,7 @@ View.run(class BackupView extends View {
   }
 
   private testPassphraseHandler = async () => {
-    if (await (await KeyUtil.parse(this.armoredPrvBackup)).checkPassPhrase(String($('#pass_phrase').val())) === true) {
+    if (await KeyUtil.checkPassPhrase(this.armoredPrvBackup, String($('#pass_phrase').val())) === true) {
       await Ui.modal.info('Success - your pass phrase matches this backup!');
     } else {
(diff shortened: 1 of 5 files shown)


wiktor-k:
Thanks @limonte :+1:

PR #2816, opened 2020-06-30
Fix unexpected value when sending e-mail

wiktor-k:
To reproduce:
  - Use Firefox,
  - Open Secure Compose on gmail.com (*not* internal inbox),
  - Send an e-mail to a key that has expiration set.

The bug is caused by Date objects arriving from another realm (extension vs gmail.com page). They are full-fledged Date objects but they are not `instanceof Date` because `Date` points to the Date constructor from current frame (gmail.com).

The fix checks the existance of the function we need (`getTime`).

This was kind of hard to debug until I found a toggle in Firefox devtools to allow debugging extensions :)

### DIFF
--- a/extension/js/common/core/crypto/key.ts
+++ b/extension/js/common/core/crypto/key.ts
@@ -184,7 +184,9 @@ export class KeyUtil {
     if (!exp) {
       return false;
     }
-    if (exp instanceof Date) {
+    // exp instanceof Date does not work if the date objects
+    // are from another realm (e.g. iframe)
+    if (typeof exp.getTime === 'function') {
       return Date.now() > exp.getTime();
     }
     throw new Error(`Got unexpected value for expiration: ${exp}`);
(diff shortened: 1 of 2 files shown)


extension/js/common/core/crypto/key.ts
```diff
@@ -184,7 +184,9 @@ export class KeyUtil {
     if (!exp) {
       return false;
     }
-    if (exp instanceof Date) {
+    // exp instanceof Date does not work if the date objects, are from another realm (e.g. iframe)
+    // see https://github.com/FlowCrypt/flowcrypt-browser/pull/2816
+    if (typeof exp.getTime === 'function') {
```
tomholub, line 189: Here you will need to disable a linter check for tests to pass. Also, the check here is different from the check in openpgp-key, is that by design?

wiktor-k:
Okay, check out the simplification commit (tested it locally in firefox).

extension/js/common/core/crypto/key.ts
tomholub: When I do very verbose checks like this, it is typically because I used to get some other value in the past and I didn't know what it was (regardless of the types show - after all, weird stuff can happen in real life that the type system may not know about).

But I'll merge this, maybe the reason I added that is no longer relevant.

Issue #2780, opened 2020-06-09
consider using SAML login to Auth0 to bypass Google login in live tests


tomholub:
Google (enterprise) has an option where they forward a login request to IdP of our choice. We could choose an IdP that is much easier to script for, like Auth0, and finally have a robust login mechanism for live CI tests.

This would ideally be done on a separate domain.

PR #2818, opened 2020-07-02
Do not show gray border for unencryped unsinged messages

### LINKED ISSUE #2149: plain response should not render as secure
1) click "reply"
2) write message
3) select "Send plain"
4) send

Sent message will render in green box, even though it's sent plain.It should show without any secure frame.

limonte:
Fixes #2149


limonte:
@tomholub Live Gmail tests are failing and I think I know the reason. Recently I faced captcha on the login screen when running Gmail tests. It happens sometimes, not always. I do not know for now what is the trigger for the login form to add captcha, but `TimeoutError: waiting for selector "input[type="password"]"` definitely seems like there's a captcha on the previous step (entering login).

tomholub:
Got it, when I have some time I'll resolve it with https://github.com/FlowCrypt/flowcrypt-browser/issues/2780

PR #2819, opened 2020-07-02
Add ability to download attachment with keyboard

### LINKED ISSUE #2418: It's impossible to download the attachment using keyboard
image

1. `div#download`should be in the taborder. 
2. When focused, there should be an outline so users will be able to see the focused element
3. When focused, `Space` or `Enter` should trigger downloading
4. Once downloaded, the focus should be on `.file-download-right-click-link` so users will be able to press `Shift` + `F10` to trigger the context menu and proceed with `Save Link As...`

A small detail that can be included in this PR: add the ellipsis to match the item from the system context menu:

image

as well as in

image

limonte:
Closes #2418

Peek 2020-07-02 21-13

### DIFF
--- a/extension/css/cryptup.css
+++ b/extension/css/cryptup.css
@@ -2290,7 +2290,7 @@ body.attachment:hover { border-color: #eaeaea; }
 
 body.attachment #download {
   display: flex;
-  visibility: hidden;
+  opacity: 0;
   position: absolute;
   width: 32px;
   height: 32px;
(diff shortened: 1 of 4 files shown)


extension/css/cryptup.css
```diff
@@ -2398,7 +2408,7 @@ body.attachment .file-download-right-click-link {
   text-decoration: none;
 }
 
-.file-download-right-click-link {
+#attachment-preview-download .file-download-right-click-link {
```
limonte, line 2411: @tomholub I made a mistake in this line in https://github.com/FlowCrypt/flowcrypt-browser/pull/2778 which caused the regression when downloading attachments. The font color of the `Right-click here and choose 'Save Link As…' to save encrypted file` link was white on white background which made it invisible for users :/

If possible, please release a new patch version, this PR fixed the issue.

tomholub: Got it - will do!

PR #2845, opened 2020-07-15
fix caclulating attachments height in setInputTextHeightManuallyIfNeeded()

### LINKED ISSUE #2844: Compose windows is oveflowing in Firefox when adding an attachment
The regression from my PR: https://github.com/FlowCrypt/flowcrypt-browser/commit/bd09d27b365babc3e3ae1096ea5b28fec559041d#diff-ad5677b9ec2465f5bba80e6765439724R87

It seems that `#att_list` is dynamic element and its cached version returns `undefined` for `.height()`.

limonte:
Fixes #2844

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-input-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-input-module.ts
@@ -57,7 +57,7 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
   }
 
   public extractAttachments = () => {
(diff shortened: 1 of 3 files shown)


limonte:
@tomholub IMO this could be merged because it's a bugfix for a bug which was introduced during the feature-freeze period. And also, despite its size, it's a critical bug for Firefox users because they can't click the Send button because it's outside the viewport.

tomholub:
Got it. I was worried it may introduce other bugs (I'll have to do a 100% release today on chrome). I'll merge it then.

PR #2875, opened 2020-07-29
Add offline drafts

### LINKED ISSUE #2636: Add ability to save drafts in offline mode
It is possible to use FlowCrypt for writing encrypted messages in offline mode. Once online, the message written offline can be sent the usual way.

The only missing thing which might be very inconvenient is that the message draft can not be saved in offline mode and if the user will close tab/browser, (s)he'll lose the data.

In order to fully support offline mode, it should be possible to store drafts in `localStorage` when offline and once the user gets back online, check `localStorage` for saved drafts and store them normally (`ComposeDraftModule.draftSave()`)

limonte:
Closes #2636

Hopefully the code is good now, I tried to make local drafts as close as possible to Gmail drafts so the saving/restoring/deleting processes would be the same.

Added the ability to save drafts when offline for both compose and reply use-cases.

7U19rxmNYf

---

Also fixed the centering of reply loader:

before | after
-|-
image | image

And also this typo:

image

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -46,6 +46,9 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
     this.view.recipientsModule.onRecipientAdded(async () => await this.draftSave(true));
   }
 
+  /**
+   * Returns true when a draft was loaded
+   */
   public initialDraftLoad = async (draftId: string): Promise<boolean> => {
     if (this.view.isReplyBox) {
       Xss.sanitizeRender(this.view.S.cached('prompt'), `Loading draft.. ${Ui.spinner('green')}`);
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -174,7 +174,9 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
     }
   }
 
-  private isLocalDraftId = (draftId: string) => !!draftId.match(this.localDraftPrefix);
+  private isLocalDraftId = (draftId: string) => {
+    return !!draftId.match(this.localDraftPrefix);
+  }
 
   private localDraftCreate = async (mimeMsg: string, threadId: string, recipients: string[], subject: string) => {
     const draftId = `${this.localDraftPrefix}${threadId}`;
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -46,19 +46,21 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
     this.view.recipientsModule.onRecipientAdded(async () => await this.draftSave(true));
   }
 
-  public initialDraftLoad = async (draftId: string): Promise<void> => {
+  public initialDraftLoad = async (draftId: string): Promise<boolean> => {
     if (this.view.isReplyBox) {
       Xss.sanitizeRender(this.view.S.cached('prompt'), `Loading draft.. ${Ui.spinner('green')}`);
     }
     try {
       const draftGetRes = this.isLocalDraftId(draftId) ? await this.localDraftGet(draftId) : await this.view.emailProvider.draftGet(draftId, 'raw');
       if (!draftGetRes) {
-        return await this.abortAndRenderReplyMsgComposeTableIfIsReplyBox('!draftGetRes');
+        await this.abortAndRenderReplyMsgComposeTableIfIsReplyBox('!draftGetRes');
+        return false;
       }
       const decoded = await Mime.decode(Buf.fromBase64UrlStr(draftGetRes.message.raw!));
       const processed = Mime.processDecoded(decoded);
       await this.fillAndRenderDraftHeaders(decoded);
       await this.decryptAndRenderDraft(processed);
+      return true;
     } catch (e) {
       if (ApiErr.isNetErr(e)) {
         Xss.sanitizeRender('body', `Failed to load draft. ${Ui.retryLink()}`);
@@ -77,6 +79,7 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
         await this.abortAndRenderReplyMsgComposeTableIfIsReplyBox('exception');
       }
     }
+    return false;
   }
 
   public draftDelete = async () => {
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -13,6 +13,7 @@ import { EncryptedMsgMailFormatter } from './formatters/encrypted-mail-msg-forma
 import { Env } from '../../../js/common/browser/env.js';
 import { GmailRes } from '../../../js/common/api/email-provider/gmail/gmail-parser.js';
 import { MsgBlockParser } from '../../../js/common/core/msg-block-parser.js';
+import { NewMsgData } from './compose-types.js';
 import { PgpMsg } from '../../../js/common/core/crypto/pgp/pgp-msg.js';
 import { storageLocalGet, storageLocalSet, storageLocalRemove } from '../../../js/common/api/chrome.js';
 import { Ui } from '../../../js/common/browser/ui.js';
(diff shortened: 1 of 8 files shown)


limonte:
Tests are failing, I'll check them later today.

limonte:
@tomholub Ready for review 🚀

tomholub:
Thanks! I should be able to have a look tomorrow

extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -111,23 +121,43 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
           sendable.body['text/plain'] = `[cryptup:link:draft_compose:${this.view.draftId}]\n\n${sendable.body['text/plain'] || ''}`;
         }
         const mimeMsg = await sendable.toMime();
-        if (!this.view.draftId) {
-          const { id } = await this.view.emailProvider.draftCreate(mimeMsg, this.view.threadId);
+        if (!this.view.draftId || this.isLocalDraftId(this.view.draftId)) {
```
tomholub, line 124: If I understand this code correctly, if I'm loading a draft from storage and I'm now back online, it will save the local draft to the cloud. Is that correct? That would be a good move, and may be good to add a short comment. Maybe this part could actually be moved into a separate method, this `draftSave` is growing big. Something like `private doUploadDraftWithLocalStorageFallback`

extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -111,23 +121,43 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
...
+          let draftId: string;
+          try {
+            ({ id: draftId } = await this.view.emailProvider.draftCreate(mimeMsg, this.view.threadId));
+            if (this.isLocalDraftId(this.view.draftId)) { // delete local draft if there is one
+              await storageLocalRemove([this.view.draftId]);
+            }
+          } catch (e) {
+            if (ApiErr.isNetErr(e)) {
+              draftId = await this.localDraftCreate(mimeMsg, this.view.threadId, msgData.recipients.to || [], msgData.subject);
```
tomholub, line 133: could add `this.view.S.cached('send_btn_note').text('Draft saved locally (offline)');`

extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -150,6 +180,38 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
     }
   }
 
+  private isLocalDraftId = (draftId: string) => {
+    return !!draftId.match(this.localDraftPrefix);
+  }
+
+  private localDraftCreate = async (mimeMsg: string, threadId: string, recipients: string[], subject: string) => {
```
tomholub, line 187: There may be an error in this method. A few lines below there is an `if (threadId)` and `else`. But regardless if there is thread id or not, this `threadId` is used as part of `draftId`. It appears that all new messages have the same draft id, because there is no thread to differentiate them. Is this intentional? If yes, it would be good to add a comment with the reasoning / intended behavior

extension/chrome/elements/compose.ts
```diff
@@ -167,8 +167,14 @@ export class ComposeView extends View {
...
     }
-    if (this.isReplyBox && this.threadId && !this.ignoreDraft && storage.drafts_reply && storage.drafts_reply[this.threadId]) {
-      this.draftId = storage.drafts_reply[this.threadId]; // there may be a draft we want to load
+    if (this.isReplyBox) { // reply
+      if (this.threadId && !this.ignoreDraft && storage.drafts_reply && storage.drafts_reply[this.threadId]) {
+        this.draftId = storage.drafts_reply[this.threadId]; // there may be a draft we want to load
+      }
+    } else { // compose
+      if (!this.draftId) {
+        this.draftId = this.draftModule.localDraftPrefix;
```
tomholub, line 176: Now I see this may be intentional

tomholub: Maybe we could then be more explicit about it, and have a `this.draftModule.localNewMessageDraftId` also, to make it clear there is only one such new message draft possible

test/source/tests/tests/compose.ts
```diff
@@ -541,6 +542,19 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       expect(await body.$eval('#input_text img', el => el.getAttribute('src'))).to.eq(imgBase64);
     }));
 
+    ava.default('compose - saving and rendering a draft when offline', testWithBrowser('compatibility', async (t, browser) => {
+      let composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
+      await (composePage.target as Page).setOfflineMode(true); // go offline mode
+      await ComposePageRecipe.fillMsg(composePage, { to: 'human@flowcrypt.com' }, 'offline test', {});
+      expect(await composePage.read('@action-send')).to.eq('Re-enter recipient..'); // ensure offline mode
+      await composePage.type('@input-body', `This is a test of saving a draft when offline`);
+      await ComposePageRecipe.waitWhenDraftIsSaved(composePage);
```
tomholub, line 551: Here could test for presense of the different "offline" draft saved note, proposed above

tomholub:
And thanks for the little bonus fixes :+1:

extension/chrome/elements/compose-modules/compose-draft-module.ts
limonte: > If I understand this code correctly, if I'm loading a draft from storage and I'm now back online, it will save the local draft to the cloud. Is that correct?

Correct.

> That would be a good move, and may be good to add a short comment.

Done :+1:

> Maybe this part could actually be moved into a separate method, this draftSave is growing big. Something like `private doUploadDraftWithLocalStorageFallback`

Done 👍

extension/chrome/elements/compose-modules/compose-draft-module.ts
limonte: Done 👍

extension/chrome/elements/compose-modules/compose-draft-module.ts
limonte: Yup, that's intentional because the new message (not reply) will always have the same local draft ID, which is now `this.draftModule.localNewMessageDraftId`.

extension/chrome/elements/compose.ts
limonte: Done :+1:

test/source/tests/tests/compose.ts
limonte: Done :+1:

limonte:
The false negative from Semaphore:

image

seisvelas:
> @seisvelas after the GSKM vuln report doc is prepared - please evaluate security - to cross check that drafts are indeed being stored encrypted locally, just like when storing them in the cloud. Thanks!

Sure! I loaded this PR in Firefox using web-ext and created a locally saved draft. Everything worked per @limonte's example. To verify that the locally stored draft is encrypted, I navigated to `about:debugging -> This Firefox`. Then, I found the FlowCrypt extension and clicked 'inspect'. From there I could peek at the extension's storage, and select the local draft. Here's that that looks like:

ruto

If I copy the contents of the stored message there and base64 decode it, I get this:
```

Content-Type: multipart/mixed;
 boundary="----sinikael-?=_1-15961457536010.9590969725101605"
Openpgp: id=00C75FDC84F8EBF2
From: ruto@jabujabu.belly
To: link98@kokiri.forest
Subject: Bring fish
Date: Thu, 30 Jul 2020 21:49:13 +0000
Message-Id: <1596145753603-bddfbaa6-81ff7235-722dcac0@gmail.com>
MIME-Version: 1.0

------sinikael-?=_1-15961457536010.9590969725101605
Content-Type: text/plain
Content-Transfer-Encoding: quoted-printable

-----BEGIN PGP MESSAGE-----
Version: FlowCrypt Email Encryption 7.8.8

wV4DQmNngpBe1wESAQdARreUymd1u6FAj8kBGeCTt+GNOSKF6eWQAbesd0cu
yjgwZ4e5jarqRlxYO0RlKNJSEAxuYgrSMLtUAmyksmlHnLi18xS590MSpyzo
g7ernPbY0j4BjUJzU8qOlkv8RDUXD9lP3ECelub/0sF+6aYK6mSwwCRrJ6qM
p3odi6VuZawMycZnZYIrhCrnqICe4mCKrA=3D=3D
=3D6Nbn
-----END PGP MESSAGE-----

------sinikael-?=_1-15961457536010.9590969725101605--
```

So we see that the message **really is** encrypted in local storage 🎉

PR #2876, opened 2020-07-30
Fix storing drafts

### LINKED ISSUE #2822: Draft opens in regular plain text editor, instead of Secure Compose editor
First, I create a message but don't send it, so it gets saved as a draft:

Screenshot_2020-07-04_10-44-38

I close the Secure Compose editor and click on the draft. But instead of opening in the Secure Compose editor, the ciphertext is opened in a plaintext Gmail editor:

Screenshot_2020-07-04_10-45-30

limonte:
Fixes #2822

### DIFF
--- a/test/samples/mock-data.ts
+++ b/test/samples/mock-data.ts
@@ -88,7 +88,7 @@ const data: UserMessages = {
       "id": "draft-3",
       "threadId": "16cfb0e25821733b",
       "labelIds": ["DRAFT"],
-      "snippet": "[flowcrypt:link:draft_reply:16cfa9001baaac0a] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 6.9.9 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/zmLZUBAQ/",
+      "snippet": "[cryptup:link:draft_reply:16cfa9001baaac0a] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 6.9.9 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/zmLZUBAQ/",
       "historyId": "1044100",
       "internalDate": "1567580104000",
       "payload": {
--- a/test/samples/mock-data.ts
+++ b/test/samples/mock-data.ts
@@ -16,7 +16,7 @@ const data: UserMessages = {
       "id": "draft-0",
       "threadId": "16eec6ebc087faa7",
       "labelIds": ["DRAFT"],
-      "snippet": "[cryptup:link:draft_compose:r304765387393056602] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 7.3.6 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/zmLZUBAQ//",
+      "snippet": "[flowcrypt:link:draft_compose:r304765387393056602] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 7.3.6 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/zmLZUBAQ//",
       "historyId": "1105694",
       "internalDate": "1575924710000",
       "payload": {
@@ -62,7 +62,7 @@ const data: UserMessages = {
       "id": "draft-1",
       "threadId": "16d6cbeb73bd2a9b",
       "labelIds": ["DRAFT"],
-      "snippet": "[cryptup:link:draft_compose:r-8909860425873898730] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 7.0.2 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/",
+      "snippet": "[flowcrypt:link:draft_compose:r-8909860425873898730] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 7.0.2 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/",
       "historyId": "1058042",
       "internalDate": "1569487501000",
       "payload": {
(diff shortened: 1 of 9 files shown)


limonte:
@tomholub the fix is working, but the test is not. I'm unable to run Gmail tests locally because the sign in form is asking me to use the 2SA device:

image

1. As I understand correctly the issue, it has to be covered by the live test. Please correct me if I'm wrong here.
2. There's `cryptup:` in drafts content, should we replace it with `flowcrypt:` and support `cryptup:` for backward compatibility?
3. We'll have to recommend a solution in #2822 for users with broken drafts.

tomholub:
> There's the cryptup: prefix in drafts content, should we replace it with flowcrypt: and support cryptup: for backward compatibility?

I'd probably leave it as is, since the change is not visible to the user, it doesn't bother me.

limonte:
> 
> 
> > There's the cryptup: prefix in drafts content, should we replace it with flowcrypt: and support cryptup: for backward compatibility?
> 
> I'd probably leave it as is, since the change is not visible to the user, it doesn't bother me.

It is visible in the "Drafts" section:

image

tomholub:
> > > There's the cryptup: prefix in drafts content, should we replace it with flowcrypt: and support cryptup: for backward compatibility?
> > 
> > 
> > I'd probably leave it as is, since the change is not visible to the user, it doesn't bother me.
> 
> It is visible in the "Drafts" section:
> 
> image

I see. Yup, we could do as you proposed.

limonte:
> 
> 
> I'll merge this so that the underlying issue can get fixed and released. We could do the name change in another PR.

While the fix seems to be correct, tests are failing here, I'm currently working on them.

tomholub:
I've re-run it, just to confirm what exactly is up with the tests.

tomholub:
Ok now I see it. I think in the encrypted inbox, we don't show the body content in the message list the way Gmail does. But after clicking, you should be able to see the draft prefix.

However given that you are testing with a mock (a good thing), you could instead write a test on the mock-server side. When the mock backend receives a draft, you can test that it always has the prefix in the body.

Such check would go somewhere here where the mock receives the data: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/mock/google/google-endpoints.ts#L183

Maybe you would not even need to then write a special test with ava - since other tests are probably also storing drafts, they should likely start failing if the mock backend starts checking for this.

limonte:
> 
> 
> Ok now I see it. I think in the encrypted inbox, we don't show the body content in the message list the way Gmail does. But after clicking, you should be able to see the draft prefix.
> 
> However given that you are testing with a mock (a good thing), you could instead write a test on the mock-server side. When the mock backend receives a draft, you can test that it always has the prefix in the body.
> 
> Such check would go somewhere here where the mock receives the data: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/mock/google/google-endpoints.ts#L183
> 
> Maybe you would not even need to then write a special test with ava - since other tests are probably also storing drafts, they should likely start failing if the mock backend starts checking for this.

Very good point, thank you! Are you ok with showing this to users instead of "Not saved (error)":

image

I did that change in https://github.com/FlowCrypt/flowcrypt-browser/pull/2876/commits/c3a8a7d69f86cb9a752ea87263e655dda57c820d the styling is to be improved

tomholub:
> > Ok now I see it. I think in the encrypted inbox, we don't show the body content in the message list the way Gmail does. But after clicking, you should be able to see the draft prefix.
> > However given that you are testing with a mock (a good thing), you could instead write a test on the mock-server side. When the mock backend receives a draft, you can test that it always has the prefix in the body.
> > Such check would go somewhere here where the mock receives the data: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/mock/google/google-endpoints.ts#L183
> > Maybe you would not even need to then write a special test with ava - since other tests are probably also storing drafts, they should likely start failing if the mock backend starts checking for this.
> 
> Very good point, thank you! Are you ok with showing this to users instead of "Not saved (error)":
> 
> image
> 
> I did that change in [c3a8a7d](https://github.com/FlowCrypt/flowcrypt-browser/commit/c3a8a7d69f86cb9a752ea87263e655dda57c820d)
> The styling is to be improved

Normally I like to show errors to users in detail. But there's not much space here. You could continue showing the original short note, and do a toast with the full error, maybe that's a better compromise.

test/samples/mock-data.ts
```diff
@@ -16,7 +16,7 @@ const data: UserMessages = {
       "id": "draft-0",
       "threadId": "16eec6ebc087faa7",
       "labelIds": ["DRAFT"],
-      "snippet": "[cryptup:link:draft_compose:r304765387393056602] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 7.3.6 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/zmLZUBAQ//",
+      "snippet": "[flowcrypt:link:draft_compose:r304765387393056602] -----BEGIN PGP MESSAGE----- Version: FlowCrypt 7.3.6 Gmail Encryption Comment: Seamlessly send and receive encrypted email wcFMA0taL/zmLZUBAQ//",
```
tomholub, line 19: I would actually leave some of these with "cryptup" still - to make sure both versions work (in case you haven't).

tomholub:
I just realized, the method I suggested will probably give you a bunch of trouble because:

when creating a new draft, we create it in two steps. First step we make it without an id in the body, because we don't yet know what the id of the newly created draft will be (or some such reason). As soon as draft gets created, without this tag, we update it properly with the tag.

I'll have a look at the code, maybe I can fix it easily by adding a dummy tag on the first creation.

limonte:
> 
> 
> I just realized, the method I suggested will probably give you a bunch of trouble because:
> 
> when creating a new draft, we create it in two steps. First step we make it without an id in the body, because we don't yet know what the id of the newly created draft will be (or some such reason). As soon as draft gets created, without this tag, we update it properly with the tag.
> 
> I'll have a look at the code, maybe I can fix it easily by adding a dummy tag on the first creation.

thank you!

tomholub:
It bothers me how complex these things have become. I'm not sure it's possible to reduce the complexity, so we'll really have to double down on testing. The fix above should address it, if it passes, I'll be merging this.

limonte:
> 
> 
> It bothers me how complex these things have become. I'm not sure it's possible to reduce the complexity, so we'll really have to double down on testing. The fix above should address it, if it passes, I'll be merging this.

Thank you for helping me with this PR, I appreciate that!

tomholub:
Glad to help. You have discovered that it has something to do with the encrypted/buf property, which would probably take me a long time and a lot of frustration to find out.

PR #2883, opened 2020-08-10
Issue 2732: fix S/MIME import on contacts screen

### LINKED ISSUE #2732: settle on a fingerprint format for s/mime

wiktor-k:
S/MIME certs couldn't be imported on contacts screen due to missing libraries and internal support to display the certs.

This change adds a test to catch this scenario, fixes the underlying code and adds missing libraries.

I've also improved the naming of `needle` as it took me some time to figure out why string was not found on the page when I see it's there :)

2020-08-10T13:09:16,299609298+02:00

As for #2732 this is now using `serialNumber` that is defined as:

```
hex encoded value of an ASN.1 INTEGER.
Conforming CAs should ensure serialNumber is:
  - no more than 20 octets
  - non-negative (prefix a '00' if your value starts with a '1' bit)
```

it looks like OpenPGP fingerprint and is displayed automatically in groups so I think this fixes #2732.

### DIFF
--- a/extension/chrome/elements/pgp_pubkey.htm
+++ b/extension/chrome/elements/pgp_pubkey.htm
@@ -29,6 +29,7 @@
   <script src="/lib/jquery.min.js"></script>
   <script src="/lib/sweetalert2.js"></script>
   <script src="/lib/openpgp.js"></script>
+  <script src="/lib/forge.js"></script>
   <script src="pgp_pubkey.js" type="module"></script>
 
 </body>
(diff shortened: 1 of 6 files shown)


wiktor-k:
:thinking: I see red cross on github but when I click Details it's all green... what gives?

tomholub:
> thinking I see red cross on github but when I click Details it's all green... what gives?

That means it's good. Sometimes it gets weird like that. I'll go report it to Semaphore.

PR #2884, opened 2020-08-11
escape labelId parameter for HTML

seisvelas:
Escapes HTML from the labelId parameter:

Screenshot_2020-08-10_23-41-43

I'll leave the issue open for similar problems (of which I suspect there might be a few).

### DIFF
--- a/extension/js/common/platform/xss.ts
+++ b/extension/js/common/platform/xss.ts
@@ -37,6 +37,15 @@ export class Xss {
     return $(selector as any).replaceWith(Xss.htmlSanitize(dirtyHtml)); // xss-sanitized
   }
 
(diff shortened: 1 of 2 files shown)


extension/js/common/platform/xss.ts
```diff
@@ -37,6 +37,15 @@ export class Xss {
     return $(selector as any).replaceWith(Xss.htmlSanitize(dirtyHtml)); // xss-sanitized
   }
 
+  public static htmlEscape = (str: string) => {
```
tomholub, line 40: We already have a method like this - please use the Xss.escape method below

seisvelas: Thanks, how could I have missed that! Fixed.

PR #2888, opened 2020-08-15
issues 2776 2886 deps pulled from npm

### LINKED ISSUE #2776: mechanism to keep up to date with recent lib versions
Our current lib update procedure does not make it easy for LGTM and other bots to find vulnerabilities, because they don't know which versions of deps are we using.

Recommendation:

> Implement and document a software update procedure

This really means, we should be putting deps in `package.json` and then copying them from there during build, instead of the current "batteries included" approach.

Then vulns like #2775 and #2717 wouldn't happen.

### LINKED ISSUE #2886: update all dependencies
All libs except emailjs (many conflicts, plus we have custom edits)

Can be done manually or by addressing #2776

tomholub:
close #2776 
close #2886


tomholub:
still need to make forge.js not minified

PR #2904, opened 2020-08-20
Attempt to fix key without identities (User IDs)

wiktor-k:
This is a draft PR.

I've tested it and it works given that the check in OpenPGP.js is manually removed. It correctly fixes (reformats) the key.

I'll add the test case (disabled until OpenPGP.js v5) here and then mark it as ready for review.

### DIFF
--- a/extension/js/common/ui/key-import-ui.ts
+++ b/extension/js/common/ui/key-import-ui.ts
@@ -136,6 +136,9 @@ export class KeyImportUi {
     await this.decryptAndEncryptAsNeeded(decrypted, encrypted, passphrase);
     await this.checkEncryptionPrvIfSelected(decrypted, encrypted);
     await this.checkSigningIfSelected(decrypted);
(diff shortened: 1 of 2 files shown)


extension/js/common/ui/key-import-ui.ts
```diff
@@ -136,6 +136,9 @@ export class KeyImportUi {
     await this.decryptAndEncryptAsNeeded(decrypted, encrypted, passphrase);
     await this.checkEncryptionPrvIfSelected(decrypted, encrypted);
     await this.checkSigningIfSelected(decrypted);
+    if (await KeyUtil.isWithoutIdentities(encrypted)) {
```
tomholub, line 139: I think in this situation we can just directly test for `key.identities.length === 0`, the added method does not seem necessary (since it's not a complicated or tricky or otherwise confusing thing to test for).

Thanks!

PR #2907, opened 2020-08-20
Use @flowcrypt/fine-uploader

### LINKED ISSUE #2889: deprecate fine-uploader
We use fine-uploader to add a file upload paperclip icon, and then render the list of uploaded files and retrieve their content.

Fine-uploader further has a bunch of other functionality like uploading to various cloud solution, but we don't use any of that.

As browser support for what fine-uploader does has improved, it's likely not needed anymore. Plus it's deprecated and has a known vulnerability.

Given how little we use of the library, we could either write our own, or create a mini-library with bits copied from fine-uploader, and release that under the same license as the original library.

limonte:
Closes #2889


tomholub (approved):
Thanks!

Issue #2723, opened 2020-04-08
rel=noopener for added link safety


tomholub:
We should add a similar thing to all links sent inside emails, that would be something for flowcrypt-browser repo - it would go somewhere in the xss sanitisation class. I'll make an issue and move it there.

source: https://github.com/FlowCrypt/flowcrypt-account/pull/17

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-account/pull/17#issuecomment-610963762_

PR #2910, opened 2020-08-20
Prevent tabnabs in PGP MIME emails with Xss class

seisvelas:
Fixes https://github.com/FlowCrypt/flowcrypt-browser/issues/2723


seisvelas:
Built locally and tried with a rich text email, it worked for me. If it passes the tests, I should also probably make a test for this (should be a straightforward test, just send an email with a link, and check the email and see if the appropriate rel attribute is included).

PR #2913, opened 2020-08-24
Use filesize.js for formatting size of attachments

### LINKED ISSUE #2656: The filesize of attached files isn't human-friendly enough
I just sent the message with a file attached, the filesize was `1226843 bytes`

It was rendered as `1,199KB`, which was confusing to me at the first sight, my eyes interpreted that as `1.2KB`

Proposal 1: improve existing filesize calculation in https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/chrome/elements/pgp_block_modules/pgp-block-attachmens-module.ts#L28

Proposal 2: use existed well-tested solution like https://github.com/avoidwork/filesize.js 

image

limonte:
Fixes #2656


tomholub (approved):
Thanks!

PR #2914, opened 2020-08-24
Autofocus recover input

### LINKED ISSUE #2535: more keys to recover input focus + enter event
image

1) input should be focused
2) hitting enter should submit it

limonte:
Fixes #2535 

Only autofocus was missing, `Enter` hit was working as expected.

Also fixed SweetAlert2's font-size for Bootstrap 3 because after upgrading deps #2105 was undone. #2105 was my mistake, the lesson to myself: never edit deps commited to the repo.

image


tomholub (approved):
Thank you. The dep issue is also partly our fault, for having checked them directly into the repo instead of a more conventional approach.

PR #2916, opened 2020-08-26
Make the 'Renew' button accessible with kbd

### LINKED ISSUE #2468: Settings: the "RENEW" button isn't accessible with keyboard
Replace `<span>` with `<button>`, make sure the focused state is visible.

limonte:
Closes #2468 

G903G7jdSB

Also, I removed the `0.7` opacity on that button, because with the opacity it was looking like `<button disabled />`


tomholub (approved):
Looks good, thank you!

Issue #2753, opened 2020-04-28
transition mock data from external repo into this repo


tomholub:
Our existing system downloads mock data in bulk from https://github.com/FlowCrypt/flowcrypt-bin/tree/master/gmail-mock-data

This is not convenient when team members want to add more tests, and I'd like to get rid of it.

Instead I'd like to use a newer mechanism - storing needed messages in `exported-messages` folder.

Relevant code:
```
  /**
   * This is the proper way to add messages to mock api for testing:
   *   1) log into flowcrypt.compatibility@gmail.com
   *   2) go to Settings -> Inbox and find your message
   *   3) click "download api export"
   *   4) save the json file to exported-messages folder
   */
  private exportedMsgsPath = './test/source/mock/google/exported-messages/';

  /**
   * The original way was to export whole account at once, encrypt it and download it later.
   * There are several problems with this approach:
   *  - need to re-export whole inbox each time we want to add a message
   *  - many messages are included that are not needed for testing
   *  - lack of visibility or ability to easily edit messages
   *  - messages may contain sensitive information, so we cannot freely include them
   *      in the repo - they are encrypted and new developers cannot run tests without
   *      a test-secrets.json file, which is inconvenient
   *  This list of message ids will be skipped when importing the bulk import, instead
   *    they will be imported from the newer mechanism - exported-messages folder.
   *  We will be transferring used messages to this mechanism one by one, until there
   *    are no more messages in the bulk import, at which point we can remove this prop
   *    as well as all functionality that relates to it.
   */
  private skipBulkImportMessages = ['171d138c8750863b'];
```

A thing to note is that the old mechanism just exported the whole inbox, even messages we don't need. These are bulky. For `exported-messages` we should only store mocks we are actually using.

PR #2918, opened 2020-08-27
Display all e-mails directly on my_key screen

### LINKED ISSUE #2547: clearer userids for keys screen
If I have an account `acct@gmail.com` and alias `alias@gmail.com` and create a key for `alias@gmail.com`, on the `Settings -> Additional Settings -> Public Key` screen it will show as key for `acct@gmail.com` which may be confusing.

It's trying to say that it's a key for certain account, but the user may expect to see their primary user id there.

Maybe just show all keyids right on that screen without having to click "show user ids".

wiktor-k:
Fixes #2547.

PR submitted for an effort to close my assigned long standing issues :)

I'll get back to you with a proposal to replace the XSS sanitize render with something safe by design but I need a little bit of time to prepare proof-of-concept.

For the private keys that are not really private maybe it would also be good to move them to some separate not-secret file...

### DIFF
--- a/extension/chrome/settings/modules/my_key.htm
+++ b/extension/chrome/settings/modules/my_key.htm
@@ -17,7 +17,7 @@
 <body id="settings">
   <div id="content" class="my_key page" style="text-align: left;">
     <div class="line">
-      <b class="key_type">Public Key Info</b> for <span class="email"></span>&nbsp;&nbsp;<a href="#"
+      <b class="key_type">Public Key Info</b> for <span class="email" data-test="content-emails"></span>&nbsp;&nbsp;<a href="#"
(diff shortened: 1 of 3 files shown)


tomholub (approved):
Looks good, thank you

tomholub:
> For the private keys that are not really private maybe it would also be good to move them to some separate not-secret file...

This one could be a part of #2753

> I'll get back to you with a proposal to replace the XSS sanitize render with something safe by design but I need a little bit of time to prepare proof-of-concept.

Not sure what you mean here, but it sounds interesting. Before the proof of concept - please make an issue describing the problem, and roughly what shape would the ideal solution have, so that we can align.

PR #2923, opened 2020-09-01
Change #header-row's margin to padding so body won't have margins

### LINKED ISSUE #2922: The settings layout has unneeded scrollbar which causes horizontal jumpiness for popups
ZwqFyfLH4x

limonte:
Fixes #2922 

The horizontal jumpiness caused by body with top margin was bothering my eye, so I pushed this one-liner bypassing the queue.

### DIFF
--- a/extension/css/settings.css
+++ b/extension/css/settings.css
@@ -8,7 +8,7 @@ html {
   width: auto;
 }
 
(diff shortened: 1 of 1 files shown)


tomholub (approved):
Thank you!

PR #2927, opened 2020-09-01
Get rid of 'Save link as...' for attachments

### LINKED ISSUE #2820: Get rid of "Save Link As..." (research)
The "Save link as..." bothers me quite a bit in general, but it's what I've done to work around browser restrictions (which only started popping up in browsers about 1 year ago). I hope to one day re-visit it and see if it can work without it., like before, by just clicking it, then waiting for the download.

At one point it stopped working in Gmail but still worked in flowcrypt inbox, suggesting it's a restrictions on frames inside host pages on different origin.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/2418#issuecomment-568164135_

limonte:
Fixes #2820

It seems that there's no restrictions anymore and both browsers allow to download blob links from iframes, so in this PR I'm simply removing the 'Save link as...' workaround without adding new tricks.

The better review experience will be with `?w=1`: https://github.com/FlowCrypt/flowcrypt-browser/pull/2927/files?w=1


tomholub:
That would be great news! I just hope it doesn't come back :) code looks good. I'll test it later.

tomholub (approved):

tomholub:
It really does work nicely.

PR #2932, opened 2020-09-03
fix eslint header/header

limonte:
Fix eslint `header/header` rule


tomholub (approved):
Thank you for looking into this

PR #2935, opened 2020-09-08
issue 2934 fix enterprise CORS issue

### LINKED ISSUE #2934: CORS issue when sending email on enterprise extension
image

 - there are two differences between enterprise extension release and regular: 1) the enterprise release has a more conservative release schedule 2) enterprise has a more restricted set of permissions in the extension manifest
 - some time ago, browser behavior has changed on Chrome, to enforce CORS for requests initiated from extension frames (compose window) inside other frames (Gmail). That time we only updated the consumer version, by mistake.
 - the CORS does not prevent the request from reaching Google, and therefore the actual messages are in fact sent. However the browser blocks the response of the request, causing the rendered error.
 - our automated tests do equal testing of both standard and enterprise release (all tests run on both versions). However the compose tests are run in a separate tab, not directly in the Gmail tab.
 - the CORS issue only affects frames inside the Gmail tab, so automated tests didn't uncover this
 - so far enterprise customers preferred using the standard version over the enterprise version (against my strong recommendation to use the enterprise one) to get features faster. This meant this issue in the enterprise version went undiscovered for longer than it should.

The fix was tested today locally by changing the manifest file of the extension to include "https://www.googleapis.com/*" as one of the permissions.

Today the following steps will be taken to fully reconcile this:
 - [ ] merge the fix
 - [ ] automated tests will be updated to include an email-sending test that runs directly on Gmail (we call these "live tests", and already have similar ones due to previous experience), on both regular and enterprise versions
 - [ ] code will be audited for possible opportunities for similar errors on other APIs when accessed on the enterprise version from inside the Gmail page
 - [ ] we'll switch to run the enterprise version during the regular feature development process (currently developers tend to spin up the regular one)
 - [ ] once automated tests and the fix are merged, and the code audited, we'll release a new version 7.9.1 later today

tomholub:
close #2934


tomholub (commented):
lgtm

PR #2943, opened 2020-09-15
Update recepients in compose.htm after updating key in pgp_block.htm

### LINKED ISSUE #2666: imported public keys should be live-reloaded in compose window
1) receive a message with a public key
2) click reply which loads public key (presumably outdated/wrong one)
3) click "UPDATE CONTACT" which imports public key
4) hover over recipient green bubble in the reply form. Currently will show old fingerprint. Should be showing fingerprint of of the key that was just imported.

Should add a mock test.

limonte:
Fixes #2666

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -9,7 +9,7 @@ import { ProviderContactsQuery, Recipients } from '../../../js/common/api/email-
 import { RecipientElement, RecipientStatus, RecipientStatuses } from './compose-types.js';
 import { Str, Value } from '../../../js/common/core/common.js';
 import { ApiErr } from '../../../js/common/api/error/api-error.js';
-import { BrowserMsg } from '../../../js/common/browser/browser-msg.js';
+import { Bm, BrowserMsg } from '../../../js/common/browser/browser-msg.js';
 import { Catch } from '../../../js/common/platform/catch.js';
 import { Google } from '../../../js/common/api/google.js';
 import { GoogleAuth } from '../../../js/common/api/google-auth.js';
@@ -86,6 +86,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     this.view.S.cached('compose_table').click(this.view.setHandler(() => this.hideContacts(), this.view.errModule.handle(`hide contact box`)));
     this.view.S.cached('add_their_pubkey').click(this.view.setHandler(() => this.addTheirPubkeyClickHandler(), this.view.errModule.handle('add pubkey')));
     BrowserMsg.addListener('addToContacts', this.checkReciepientsKeys);
+    BrowserMsg.addListener('reRenderRecipient', async ({ contact }: Bm.ReRenderRecipient) => await this.reRenderRecipientFor(contact));
     BrowserMsg.listen(this.view.parentTabId);
   }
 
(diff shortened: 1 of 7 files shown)


limonte:
@tomholub I can't figure out how to write the test case for this fix. Could you tell me how would you proceed with it?

limonte:
The `do not search attester for recipients on particular domains` test case is failing for this PR, but I can't debug it, for some reason it's failing for me on the `master` branch:

image

image

Do you maybe know what might be the issue?

tomholub (commented):
Code looks good. I'll have a look at the failure

tomholub:
I found the problem. We enabled WKD on our domain, which is serving public keys - for example for human@flowcrypt.com at https://openpgpkey.flowcrypt.com/.well-known/openpgpkey/flowcrypt.com/hu/0000?l=human

This has a side effect on the test, because the test expects no public key to be found (because attester search is skipped) when one is in fact found from a new source - the recently installed WKD at flowcrypt.com.

So this is something outside of browser codebase that has a side effect.

I'll do something about it.

tomholub:
The test case would look something like this:

```
    ava.default('expired key will turn green when manually updated in different window', testWithBrowser('compose', async (t, browser) => {
      const composePage = await ComposePageRecipe.openStandalone(t, browser, acct);
      await ComposePageRecipe.fillMsg(composePage, { to: 'expired.on.attester@domain.com' }, t.title);
      await composePage.waitForContent('.email_address.has_pgp', 'expired.on.attester@domain.com');
      const recipient = await expandContainer.$('.email_preview span');
      expect(await PageRecipe.getElementPropertyJson(recipient!, 'className')).to.include('expired');
      // now open a pubkey frame and update the pubkey
      const pubkeyFrame = // open a pgp_pubkey.htm frame that renders some non-expired pubkey in it. Eg you can maybe export `const somePubkey` from `attester-endpoints.ts` and then form the frame url with `pubkey=${somePubkey}`
      await pubkeyFrame.waitForContent('@action-add-contact', 'UPDATE KEY');
      // here make sure to change the email input field in the frame to `expired.on.attester@domain.com` before clicking below
      await pubkeyFrame.waitAndClick('@action-add-contact');
      await pubkeyFrame.waitForContent('@container-pgp-pubkey', `${recipientEmail} added`);
      // now you could wait a second then close the frame
      // finally, check that the compose has turned green
      expect(await PageRecipe.getElementPropertyJson(recipient!, 'className')).to.not.include('expired');
    }));
```

Bonus points: if you could also add `another.expired@domain.com` to `attester-endpoints.ts` to return the same expired pubkey as `expired.on.attester@domain.com`. Then in the above test, do `{ to: 'expired.on.attester@domain.com,another.expired@domain.com' }`. Then you check both of them show as expired. After you update one of them in the pubkey frame, you could then check that one of them has turned green, but the other one stayed orange.

limonte:
Thank you for helping me out @tomholub ! 

Something weird is happening here: tests are failing in CI but passing locally, e.g.

image

limonte:
Let's see if increasing the `timeout` here https://github.com/FlowCrypt/flowcrypt-browser/blob/42e5fd376d08b6f1b45dc20be37f4ad3aab7fee8/test/source/tests/page-recipe/setup-page-recipe.ts#L181 will help:

tomholub:
I think the timeout is already well long enough, there will be something else going on. I'll take a look.

tomholub:
I broke it, let's see now

tomholub:
It appears to be happy now, sorry for the mess.

limonte:
@tomholub thanks a lot for your help! The PR is ready for review now.

tomholub (approved):
Excellent, I am happy to see this one merge!

PR #2944, opened 2020-09-16
Fix webmailNamePattern regex

### LINKED ISSUE #2928: The "Getting file info ..." status is applied only for even attachments (2nd, 4th, ...)
Actual: 

image

Expected: the "Getting file info ..." status is applied to all attachments so their alignment won't look broken.

limonte:
Fix #2928 

I guess this is the smallest PR in this repo :)

My knowledge of regexes wasn't enough and I asked for the help from the community: https://stackoverflow.com/q/63928569

Fortunately, the help was prompt and TIL that the `RegExp` object keeps track of the `lastIndex` where a match occurred, so on subsequent matches it will start from the last used index, instead of `0`: https://stackoverflow.com/a/1520853

### DIFF
--- a/extension/js/common/core/att.ts
+++ b/extension/js/common/core/att.ts
@@ -15,7 +15,7 @@ export type FcAttLinkData = { name: string, type: string, size: number };
 
 export class Att {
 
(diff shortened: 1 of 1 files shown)


tomholub (approved):

tomholub:
Thanks, this is a tricky one.

PR #2946, opened 2020-09-18
Allow preview for pgp/mime attachments

### LINKED ISSUE #2929: pgp/mime attachments (when using rich text) have no preview
eg 

image

appropriate after #2820

limonte:
Closes #2929

Encrypted attachments have preview + download options:

image

Signed attachments don't need preview because it's available in Gmail UI, only download:

image

All elements are accessible with keyboard.


tomholub (commented):
Excellent, looking forward! I forgot to mention tests - I imagine they can be mostly copied from the regular attachment tests, they should even be a bit simpler.

tomholub (approved):

PR #2952, opened 2020-09-20
Add allow-downloads to attachment preview iframes

### LINKED ISSUE #2950: investigate download button not working above file preview
> Separately, I noticed (on chrome, enterprise version build) that the download button on top of the preview doesn't work (for any file type). The download button in the mini attachment frame works. I haven't tested if it's related to this PR, or if it happens on master too. Please also have a look and add tests in case it's not tested.

from #2949 in case it's a general issue

limonte:
Fixes #2950 

Google Chrome 83 began to disallow iframes from downloading unless the `allow-downloads` key is added to an iframe’s `sandbox` attributes: https://www.chromestatus.com/feature/5706745674465280

### DIFF
--- a/extension/chrome/elements/attachment_preview.htm
+++ b/extension/chrome/elements/attachment_preview.htm
@@ -12,7 +12,7 @@
 </head>
 
 <body id="attachment-preview">
-  <button id="attachment-preview-download" class="display_none">
+  <button id="attachment-preview-download" class="display_none" data-test="attachment-preview-download">
     <img src="/img/svgs/download-link-white.svg" alt="download attachment" width="28">
     DOWNLOAD
   </button>
--- a/extension/js/common/browser/ui.ts
+++ b/extension/js/common/browser/ui.ts
@@ -269,7 +269,7 @@ export class Ui {
           $(Swal.getContent()).attr('data-test', 'attachment-dialog');
           $(Swal.getCloseButton()).attr('data-test', 'dialog-close');
         },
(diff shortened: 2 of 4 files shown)


tomholub:
Excellent, thank you for debugging this!

limonte:
Actually the test would work with or without `allow-downloads` because we're using old puppeteer with old (v6x) Chromium. I'll write the anyways because this feature should definitely be covered with a test, but eventually we should upgrade puppeteer and fix tests which will be failing with the latest puppeteer.

I created separated issue for that: #2953

tomholub (approved):
:+1:

limonte:
While reading the 3 yo https://github.com/puppeteer/puppeteer/issues/299 I noticed the reference to https://github.com/FlowCrypt/flowcrypt-browser/pull/2222 in which I didn't find the testing of downloading.

At the end, it's simpler that I thought it would be. Let me know if you think the test case should remove the `7 years.jpeg` file before or after itself.

tomholub:
> While reading the 3 yo [puppeteer/puppeteer#299](https://github.com/puppeteer/puppeteer/issues/299) I noticed the reference to #2222 in which I didn't find the testing of downloading.
> 
> At the end, it's simpler that I thought it would be. Let me know if you think the test case should remove the `7 years.jpeg` file before or after itself.

I liked the test a lot the way you wrote it. Yes - it would be good to delete the file, I think both before and after, in both cases ignoring error if the file is not there.

tomholub (approved):

tomholub (approved):

Issue #2834, opened 2020-07-11
Attachments for signed-only messages are rendered as encrypted


limonte:
While working on #2832 I noticed that attachments for signed-only messages are rendered as encrypted. That's wrong because attachments are not encrypted for signed-only messages.

Because of that, the attachment preview feature is not working correctly because it's trying to decrypt attachments:

![3](https://user-images.githubusercontent.com/6059356/87234685-24739f00-c3dc-11ea-9aa0-47bc64bd0933.gif)

tomholub:
Now I see this properly - this happens when the sent message is rendered immediately following sending it. That is why the signed message is rendered with our own UI (normally, it would be rendered with Gmail UI if you exit the conversation and come back - exactly because it's not encrypted.

I think that's a fair usecase to cover - allow user to immediately click on sent attachment to assure themselves that they sent the right thing. I do this a lot myself.

PR #2988, opened 2020-09-30
quick fix for signed unencrypted attachments bug

fatimaavilesfa:
This is a quick fix for https://github.com/FlowCrypt/flowcrypt-browser/issues/2834.

The first time you open an email with a signed, unencrypted attachment the extension thinks it is encrypted. I tested this with a quick console log, and the property `this.isEncrypted` is true in attachment_preview.ts, even though the attachment is *not* encrypted. However, if you refresh the page or return to the email after already having opened it, everything works fine. The email preview works normally. Firefox does not have this problem, only Chrome.

In this PR I've added a quick fix using the attachment's media type, because encrypted attachments always have a media type of `application/octet-stream`. This fixes the vast majority of cases, but the original bug would resurface were a user to send an attachment that actually does have a media type of `application/octet-stream`. 

While this fix is imperfect and the original source of the bug is still unknown to me, this fix should cover most practical cases in the meantime.

### DIFF
--- a/extension/chrome/elements/attachment_preview.ts
+++ b/extension/chrome/elements/attachment_preview.ts
@@ -27,7 +27,7 @@ View.run(class AttachmentPreviewView extends AttachmentDownloadView {
       Xss.sanitizeRender(this.attachmentPreviewContainer, `${Ui.spinner('green', 'large_spinner')}<span class="download_progress"></span>`);
       this.att = new Att({ name: this.origNameBasedOnFilename, type: this.type, msgId: this.msgId, id: this.id, url: this.url });
       await this.downloadDataIfNeeded();
-      const result = this.isEncrypted ? await this.decrypt() : this.att.getData();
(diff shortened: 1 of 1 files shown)


limonte (approved):
LGTM as a quick fix, but I'd still like to keep #2834 opened because

- the signed-only attachment shouldn't be rendered as "ENCRYPTED"
- there should be a fix for replying with signed-only message with attachment, then previewing the sent attachement

tomholub:
Thank you for the contribution. I'd rather not check this in - because it's the kind of code that we may forget we checked in, and will re-surface as a bug later.

I think `isEncrypted` needs to be properly set to false wherever it's coming from. Tests need to be added covering the newly fixed behavior.

When a fix is not accompanied by a test that confirms the fix, it's as if the fix didn't exist. Because next time, someone may refactor the fix away by mistake, and we'd never notice.

fatimaavilesfa:
Yes, I agree it's better to find and solve the real issue and add the proper test. I keep looking in to it :)

PR #2991, opened 2020-10-03
Fix retry attachment download

### LINKED ISSUE #2989: [Offline mode] The "Retry" button shows the attachmnet preview instead of re-trying to download
### Actual behavior: The attachment preview is shown when clicking the "Retry" button:

vLfAwWs5Eq

### Expected behavior:

Another attempt to download the attachment.

limonte:
Fixes #2989 

Also, improved styles a bit:

image

@tomholub do we need a test for this fix? I was trying to add one, but for adding a test there should be some kind of going offline possibility in `GoogleData`. What do you think?


tomholub:
The test would be problematic because the mock API is shared across all running tests, so cannot just take it all down. There would of course be a way around that, but I'd say maybe we won't test this particular issue.

tomholub (approved):

tomholub:
Thanks!

PR #2997, opened 2020-10-06
Add test for removing a key

### LINKED ISSUE #2803: test remove key button
Should also add a test for this, but I won't have time for it now :(

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/2802#issuecomment-650091880_

limonte:
Closes #2803

### DIFF
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -426,7 +426,7 @@ View.run(class SettingsView extends View {
       const created = new Date(prv.created);
       const date = Str.monthName(created.getMonth()) + ' ' + created.getDate() + ', ' + created.getFullYear();
       const escapedFp = Xss.escape(ki.fingerprint);
-      const escapedPrimaryOrRm = (ki.primary) ? '(primary)' : `(<a href="#" class="action_remove_key" fingerprint="${escapedFp}">remove</a>)`;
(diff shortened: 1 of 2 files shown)


tomholub (approved):
Thanks!

tomholub:
I suspected the test may be imperfect - my intuition was that maybe `await settingsPage.notPresent('@action-remove-key');` could run too soon, before the page is fully reloaded, and and therefore test it before the button would normally show up.

I ran it locally and the test failed, but for some other reason ¯\\\_(ツ)\_/¯ so I've added a few more lines to make sure it waits a bit longer.

Sometimes, writing tests is more magic/art then science. Hopefully this will be better.

tomholub (approved):

limonte:
> I suspected the test may be imperfect - my intuition was that maybe `await settingsPage.notPresent('@action-remove-key');` could run too soon, before the page is fully reloaded, and and therefore test it before the button would normally show up.
> 
> I ran it locally and the test failed, but for some other reason ¯\_(ツ)_/¯ so I've added a few more lines to make sure it waits a bit longer.
> 
> Sometimes, writing tests is more magic/art then science. Hopefully this will be better.

I believe the flakiness (and the amount of magic) will be decreased once we update puppeteer to latest. Thanks for making a test more robust!

PR #3008, opened 2020-10-11
issue 2788 renamed PgpMsg to MsgUtil

### LINKED ISSUE #2788: Rename PgpMsg class to MsgUtil class?
Looks like this class should be also renamed, maybe to just `MsgUtil` instead of `PgpMsg`.. actually, almost everything in the core class should become some sort of XxxUtil, the way we are using it.

This would be for another PR.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/2739_

tomholub:
close #2788

### DIFF
--- a/extension/chrome/elements/attachment.ts
+++ b/extension/chrome/elements/attachment.ts
@@ -3,7 +3,7 @@
 'use strict';
 
 import { Bm, BrowserMsg } from '../../js/common/browser/browser-msg.js';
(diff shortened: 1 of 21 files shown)


tomholub (commented):
lgtm, if tests pass

tomholub (commented):
lgtm

PR #3009, opened 2020-10-11
issue 3006 unit tests in browser + issue 2686 WellKnownHostMeta

### LINKED ISSUE #2689: prepwork for FlowCrypt Enterprise Server integration
Today, stack traces are sent to FlowCrypt API - done in `Cather.report` and `Catcher.handleErr`. There should be a way to send them to a configured FES instance.

### LINKED ISSUE #3006: a way to run Unit tests in the browser
Our current unit tests at https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/tests/tests/unit.ts are useful for testing code that runs cross-platform. However, most of our code is written to run in the browser, and these tests are run by NodeJS.

We should update these unit tests to run in the browser. If AVA is not an appropriate tool for these type of tests, we can use a better suited tool for the unit tests.

This will allow us to write more pointed, faster and specific tests. Currently, 95% of our tests are heavy UI integration tests.

tomholub:
issue #2689 - dynamic fes url (class only, not integrated into codebase)
close #3006 - unit tests in browser

 - [x] run all unit tests on both enterprise and consumer flavors
 - [x] unit test - no server


tomholub (commented):
lgtm - will merge it now to prevent conflicts. The new class is not used in the codebase yet, in upcoming PR I'll add more tests and integrate it with the codebase.

Issue #2813, opened 2020-06-29
Russian letters are omitted in attachments


limonte:
Steps to reproduce:

1. Create a file with `привет_мир.txt` filename
2. Attach it to encrypted message:

![image](https://user-images.githubusercontent.com/6059356/86030546-4add2600-ba3d-11ea-830a-a26db5737696.png)

3. Send

### Expected result

![image](https://user-images.githubusercontent.com/6059356/86030608-5f212300-ba3d-11ea-92e9-e3cfa02e4bff.png)

### Actual result

![image](https://user-images.githubusercontent.com/6059356/86030625-647e6d80-ba3d-11ea-8d90-ab689429d04c.png)

@tomholub not sure if this is a dup of #2809 if so, feel free to mark as duplicate and close this one.

tomholub:
It probably is a duplicate. The other one was specifically for password protected messages, but it looks like the issue happens on both types, so it will be good to test both types.

Issue #2809, opened 2020-06-29
problems with utf attachment names - pwd protected messages


tomholub:
> Hi,
> there is an issue with umlauts in attached filenames.
> e.g. "führerschein.jpg" becomes fu_hrerschein.jpg, which causes errors on customer side.
> thanks!
> FlowCrypt chrome 7.5.8

he said it happens for password protected messages

![image](https://user-images.githubusercontent.com/6306961/86024451-87e4f080-ba1c-11ea-82dd-ed5dda769507.png)

user can be reached by me at https://mail.google.com/mail/u/8/#search/password+is%3Ainbox/FMfcgxwGCtPkNdgvtkZfxXKlvbkKHVKk

PR #3013, opened 2020-10-12
Allow Unicode filenames

rrrooommmaaa:
made a separate function `Att.sanitizeName` to sanitize file name from `AttUi`,
allowed Unicode characters in file names
and added a bunch of unit tests.

// tom: issues #2813 #2809


tomholub:
All good, just need to turn off a nonsensical linter. I'll take a look

tomholub:
I think it make sense to replace it. I'll disable the rule

tomholub (approved):

PR #3029, opened 2020-10-16
Remove sweetalert2 types from repo, use types from node_modules

limonte:
As per conversation in https://github.com/FlowCrypt/flowcrypt-browser/pull/3028#issuecomment-709953053


limonte:
Good, tests are failing with my expectation from https://github.com/FlowCrypt/flowcrypt-browser/pull/3028#issuecomment-709930113

tomholub (approved):

PR #3033, opened 2020-10-16
Wkd now uses advanced request where possible and checks policy file

### LINKED ISSUE #3003: should also search on advanced WKD url + use policy file
Currently, the `Wkd` class only searches using the `direct` WKD method.

To fix: add the `advanced` method which should be tried first, only after which the basic WKD method should be attempted.
Further, before trying any WKD lookup, should pull WKD policy file first.

The spec is at https://tools.ietf.org/html/draft-koch-openpgp-webkey-service-10

relevant code
```
    // todo - could also search on `https://openpgpkey.{domain}/.well-known/openpgpkey/{domain}/hu/{hu}?l={user}`
    const url = `https://${recipientDomain}/.well-known/openpgpkey/hu/${hu}?l=${encodeURIComponent(user)}`;
```

Need to add tests for the WKD class. This can be done by adding all of these expected WKD endpoints to the mock, and testing lookup using browser unit tests.

For the mock, you will want to add one more set of endpoints here:
```
  const api = new LoggedApi<{ query: { [k: string]: string }, body?: unknown }, unknown>('google-mock', {
    ...mockGoogleEndpoints,
    ...mockBackendEndpoints,
    ...mockAttesterEndpoints,
    ...mockKeyManagerEndpoints,
    '/favicon.ico': async () => '',
  });
```
that could maybe be called `mockWkdEndpoints` where you add a couple of WKD paths. To create these test paths, use https://metacode.biz/openpgp/web-key-directory where you input a test email and it will give you the paths where the policy files and public keys should be.

rrrooommmaaa:
Closes #3003

First checks for policy file, then extracts the key and checks for email match.
Advanced method is now supported.

### DIFF
--- a/extension/js/common/api/key-server/wkd.ts
+++ b/extension/js/common/api/key-server/wkd.ts
@@ -60,14 +60,15 @@ export class Wkd extends Api {
         if (ApiErr.isNotFound(e) || ApiErr.isNetErr(e)) {
           continue;
         }
-        Catch.report(`Wkd.lookupEmail error retrieving policy file ${url}/policy: ${String(e)}`);
         return { pubkey: null, pgpClient: null };
       }
     }
     try {
       binary = await Wkd.download(`${validUrl!}/${userPart}`, undefined, 4);
     } catch (e) {
-      Catch.report(`Wkd.lookupEmail error retrieving key ${validUrl!}/${userPart}: ${String(e)}`);
+      if (!ApiErr.isNotFound(e)) {
+        Catch.report(`Wkd.lookupEmail error retrieving key ${validUrl!}/${userPart}: ${String(e)}`);
+      }
       return { pubkey: null, pgpClient: null };
     }
     const { keys: [key], errs } = await KeyUtil.readMany(binary!);
--- a/extension/js/common/api/key-server/wkd.ts
+++ b/extension/js/common/api/key-server/wkd.ts
@@ -19,12 +19,16 @@ export class Wkd extends Api {
   // https://www.sektioneins.de/en/blog/18-11-23-gnupg-wkd.html
   // https://metacode.biz/openpgp/web-key-directory
 
-  constructor(private myOwnDomain: string) {
+  public port: number | undefined;
+  private protocol: string;
+
+  constructor(private myOwnDomain: string, protocol = 'https') {
     super();
+    this.protocol = protocol;
   }
 
   public lookupEmail = async (email: string): Promise<PubkeySearchResult> => {
-    const parts = email.toLowerCase().split('@');
+    const parts = email.split('@');
     if (parts.length > 2) {
       return { pubkey: null, pgpClient: null };
     }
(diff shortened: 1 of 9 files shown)


extension/js/common/api/key-server/wkd.ts
```diff
@@ -36,24 +40,41 @@ export class Wkd extends Api {
...
+        }
+        Catch.report(`Wkd.lookupEmail error retrieving policy file ${url}/policy: ${String(e)}`);
         return { pubkey: null, pgpClient: null };
       }
-      Catch.report(`Wkd.lookupEmail err: ${String(e)}`);
+    }
+    try {
+      binary = await Wkd.download(`${validUrl!}/${userPart}`, undefined, 4);
+    } catch (e) {
+      Catch.report(`Wkd.lookupEmail error retrieving key ${validUrl!}/${userPart}: ${String(e)}`);
```
tomholub, line 70: Here should first make sure to return null/null on 404/notFound, because that's an expected behavior (not every user I try to write may have a public key even if the domain has WKD set up).

Only when it's some other type of error, then it may be worth reporting.

tomholub (changes requested):
Looks great, a small change regarding reporting

extension/js/common/api/key-server/wkd.ts
tomholub: Actually, the tests are failing possibly because they noticed that Catch.report was run. I think it may be ok to change these to `console.error` - so that they are only logged locally. After all, we (as a company) don't have to know about all of these reports.

rrrooommmaaa: I will.
And for some reason `https://metacode.biz/.well-known/openpgpkey/policy` fails to load. I'll debug it tomorrow.

rrrooommmaaa: The problem is when loading a policy file there is no `Access-Control-Allow-Origin: *` header,
whereas when getting a key there is.
Does this mean that reading policy file shouldn't be included in key extraction process, @wiktor-k ?

tomholub: I think this may be a bug in his WKD implementation. 

While spec doesn't mention CORS configuration (Wiktor tried to get it into the spec), in practice it is necessary to allow CORS. He may have forgotten to do it on the policy file.

tomholub:
It's really likely misconfiguration on his end. Policy has no CORS:
```
$ curl --header 'Host: somehost.com' -I https://metacode.biz/.well-known/openpgpkey/policy
HTTP/2 200 
server: nginx
date: Sat, 17 Oct 2020 12:22:00 GMT
content-type: application/octet-stream
content-length: 0
last-modified: Fri, 16 Feb 2018 12:16:51 GMT
etag: "5a86cbb3-0"
strict-transport-security: max-age=31536001; includeSubDomains; preload
x-xss-protection: 1; mode=block; report=https://metacode.biz/collector/log
x-content-type-options: nosniff
x-frame-options: DENY
content-security-policy-report-only: default-src https: data:; object-src 'none'; script-src https: 'unsafe-inline'; style-src https: 'unsafe-inline'; report-uri https://metacode.biz/collector/log
expect-ct: max-age=86400,report-uri="https://metacode.biz/collector/log"
vary: Accept
accept-ranges: bytes
```
But the end file has proper cors:
```
$ curl --header 'Host: somehost.com' -I https://metacode.biz/.well-known/openpgpkey/hu/gebusffkx9g581i6ch4t3ewgwd6dctmp?l=wiktor
HTTP/2 200 
server: nginx
date: Sat, 17 Oct 2020 12:23:26 GMT
content-type: application/pgp-keys
content-length: 6141
last-modified: Wed, 04 Dec 2019 11:02:39 GMT
etag: "5de7924f-17fd"
content-disposition: inline; filename="D8E8F074.gpg"
access-control-allow-origin: *
accept-ranges: bytes
```
Therefore we should still be polling the policy file if we want to follow the spec.

To be fair, if a lot of WKD servers are misconfigured also, we may drop reading the policy file. For now, let's try to follow.

tomholub:
I will switch it to test another email

extension/js/common/api/key-server/wkd.ts
```diff
@@ -36,24 +40,41 @@ export class Wkd extends Api {
...
+      try {
+        await Wkd.download(`${url}/policy`, undefined, 4);
+        console.info(`Policy found: ${url}/policy`);
+        validUrl = url;
+        break;
+      } catch (e) {
+        if (ApiErr.isNotFound(e) || ApiErr.isNetErr(e)) {
+          continue;
+        }
+        Catch.report(`Wkd.lookupEmail error retrieving policy file ${url}/policy: ${String(e)}`);
```
tomholub, line 63: I'd ignore this too - if the policy is returning a 500, it's probably because they don't have WKD set up in the first place and it's some other backend responding / getting confused.

extension/js/common/api/key-server/wkd.ts
wiktor-k: > He may have forgotten to do it on the policy file.

It's exactly like you've said. Additionally I'll need to put the check into WKD checker so thanks for testing that.

wiktor-k: The spec doesn't mention CORS but fortunately due to the checker printing warning on missing CORS a lot of people have it on. The policy issue is a interesting edge case I missed.

tomholub (approved):

tomholub:
Thanks! I'll have a look at the remaining error report, it may be hard for you to hunt down.

extension/js/common/api/key-server/wkd.ts
```diff
@@ -36,24 +40,42 @@ export class Wkd extends Api {
...
+        console.info(`Policy found: ${url}/policy`);
+        validUrl = url;
+        break;
+      } catch (e) {
+        if (ApiErr.isNotFound(e) || ApiErr.isNetErr(e)) {
+          continue;
+        }
+        return { pubkey: null, pgpClient: null };
+      }
+    }
```
tomholub, line 58: @rrrooommmaaa  I think the error was here: if I'm reading it correctly, it was possible for the for loop to finish and `validUrl` would remain undefined. Then the lines below would continue calling that URL like `undefined/hu/...` causing the error reports.

I recommend to try to avoid for loops that do accounting in this manner unless one really cannot avoid them. It can be hard to spot bugs. I've rewritten it as a method that gets called two times explicitly with two different parameters (advanced vs direct url), which I think will make for an easier to debug structure.

tomholub:
Btw I like this test

image

tomholub:
I'll do some more improvements to the browser tests in general while I'm at it (our tests are a bit complex, sorry about that)

tomholub (approved):

PR #3036, opened 2020-10-17
Make the 'End Pass Phrase Session' button position: fixed in gmail

### LINKED ISSUE #2912: test for session lock icon
https://github.com/FlowCrypt/flowcrypt-browser/issues/2873

I have fixed the authentication in our live tests, this would be a good thing to test so that we know to fix it in case it breaks again.

limonte:
Fixes #2912

### DIFF
--- a/extension/css/settings.css
+++ b/extension/css/settings.css
@@ -781,14 +781,13 @@ body#settings .settings-border .key_list {
 }
 #settings.client > #banner > img.action_open_settings { cursor: pointer; }
 
-#settings.client > #banner > div.action_finish_session {
+#settings.client > #banner > .action_finish_session {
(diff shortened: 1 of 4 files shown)


limonte:
@tomholub Since now we're adding the 'End Pass Phrase Session' button to Gmail's `body`, does it make sense to add the live test for it? I mean it would be heavy and it would test something which is quite robust now. The button functionality is already tested in the `decrypt - by entering pass phrase + remember in session` test case.

tomholub:
I think you're right, no need to test it separately in this case since it's a robust placement mechanism and we have a test for the functionality using a mock.

tomholub (approved):

PR #3039, opened 2020-10-18
Change the reply box label from 'Encrypted' to 'Secure'

### LINKED ISSUE #2979: when replying to signed message, the reply box wording is mismatched
Before activating compose box it says "Encrypted" 
image

But once clicking, it is actually signed:
image

We should change the wording of the "Encrypted" maybe to simply "Secure"

limonte:
Fixes #2979


limonte:
@tomholub couldf you help me to understand what's failing in this PR? The output in semaphoreci is really heavy for some reason and takes forever to load. I opened the raw output, but it wasn't helpful also.

tomholub:
Yeah the semaphore ui has regressed lately being super slow. Let me see.

limonte:
The `master` branch is failing as well, so I assume the issue isn't in this PR

tomholub:
Yup I messed something up on our backend, glad you discovered it. Redeploying..

tomholub (approved):

PR #3044, opened 2020-10-19
Add "philhindle.errorlens" to recommended extensions.

yantakus:


tomholub (approved):
You're right, we do use this one.

PR #3063, opened 2020-10-20
Fix auth prompt issues.

### LINKED ISSUE #2749: Cannot read property 'key' of undefined
```
catch.js:77 TypeError: Cannot read property 'key' of undefined
    at Function.PgpKey.read (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/js/common/core/pgp-key.js:37:40)
    at SettingsView.addKeyRowsHtml (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/settings/index.js:367:42)
    at SettingsView.initialize (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/settings/index.js:187:32)
    at async SettingsView.render (chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/chrome/settings/index.js:48:13)
    at async chrome-extension://dfbpklhlbpikpllkojnfhdhgbomoponf/js/common/view.js:26:17
```

Steps to reproduce:

1. go to https://myaccount.google.com/permissions 
2. remove FlowCrypt from the "Third-party apps with account access" list
3. open FlowCrypt settings, this modal should appear:

image

4. click OK
5. without granting permissions, close the permissions popup

yantakus:
Closes #2749.


extension/js/common/settings.ts
```diff
@@ -276,6 +276,7 @@ export class Settings {
           window.location.href = Url.create('/chrome/settings/setup.htm', { acctEmail: response.acctEmail, idToken: response.id_token });
         }
       } else if (response.result === 'Denied' || response.result === 'Closed') {
+        console.log({ response, settingsTabId })
         if (settingsTabId) {
           await Settings.renderSubPage(acctEmail, settingsTabId, '/chrome/settings/modules/auth_denied.htm');
```
yantakus, line 281: @tomholub Here, how do I check if the subPage is already rendered? ATM it renders another iframe inside the existing iframe and so on.

extension/chrome/settings/index.ts
```diff
@@ -380,6 +384,7 @@ View.run(class SettingsView extends View {
     } catch (e) {
       if (ApiErr.isAuthPopupNeeded(e) || ApiErr.isAuthErr(e)) {
         $('#status-row #status_google').text(`g:?:auth`).addClass('bad');
+        console.log('reconect confirm')
         if (await Ui.modal.confirm(`FlowCrypt must be re-connected to your Google account.`)) {
```
yantakus, line 388: @tomholub here, how do I check that `'/chrome/settings/modules/auth_denied.htm'` modal is rendered not to render this confirm on top?

extension/js/common/settings.ts
tomholub: I think it's supposed to replace whatever subPage is currently rendered. If it renders subpage inside a subpage, then maybe `newGoogleAcctAuthPromptThenAlertOrForward` is currently getting called from some subpage (just guessing)? In which case the caller may be errorneously giving `newGoogleAcctAuthPromptThenAlertOrForward` it's own `tabId` instead of the `tabId` of the settings page, which from the perspective of the calling subpage would be `parentTabId`.

Then this calls the parent, which is the settings page, which will render a subpage which will replace whatever subpage called it.

NOW, to get out of this whole mess another way, we could consider rendering a modal instead of a subpage. The modal can have the same information and same options. Problem solved, it's easier to render a modal in all sorts of context than it's to switch to a subpage.

extension/chrome/settings/index.ts
tomholub: I think the above modal solution would affect this as well, so you'd no longer need to check, since no redirect took place. Feels cleaner overall. Let me know what you think.

tomholub:
I removed Wiktor from review since he's busy shuffling pgp packets around :upside_down_face:

extension/js/common/settings.ts
yantakus: Is there existing functionality for rendering a modal? Can't find it.

tomholub: You can `await Ui.modal.warning`

tomholub: It actually uses a library made by @limonte here

yantakus: Yes, I saw this, but it just renders plain text inside the modal while we need it to render HTML with custom functionality bound to buttons:

image

Something like `await Ui.modal.warning('/chrome/settings/modules/auth_denied.htm')` would be great, but it obviously doesn't support this.

tomholub: Indeed we don't have that yet -  you can add a method like `Ui.modal.html` that will be like `info`, but accepts html. Make sure to use the Xss class to sanitize the html. It will be useful for other stuff too I'm sure.

tomholub: I see `error` has `isHtml` param - you can do the same for info

yantakus: Please, have a look now. The only thing we don't get the "Update permissions" button in the modal, because the scripts get sanitized.

tomholub: Looking at the page, the whole thing can get simplified. You can throw out all of the auth_denied javascript and html file. Use `Ui.modal.confirm` but update the method to also accept html using `isHtml` param. Then use the text as above with the link(make sure the link opens in a new tab), but at the end add a question: `Would you like to proceed to the permissions screen again?` Which then works nicely with the modal options: ok | cancel. 

That way you can throw away a bunch of useless code, which is always a win.

tomholub: That means you don't need any js inside the modal, you just await the user response (true or false) and based on that initiate again or not.

yantakus: Done, please check.

extension/js/common/assert.ts
```diff
@@ -37,7 +37,7 @@ export class Assert {
     if (acctEmail) {
       const [primaryKi] = await KeyStore.get(acctEmail, ['primary']);
       const { setup_done } = await AcctStore.get(acctEmail, ['setup_done']);
-      if (setup_done && primaryKi && !(await KeyUtil.parse(primaryKi.private)).fullyEncrypted) {
+      if (setup_done && primaryKi && !(await KeyUtil.parse(primaryKi.private))?.fullyEncrypted) {
```
tomholub, line 40: This looks suspect - if there is primaryKi, then KeyUtil parse should always return it. I'll have to look why is it Optional, and either bang! it (I'd rather it fails) or address it differently.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -19,7 +19,8 @@ export class OpenPGPKey {
   }
 
   public static parseMany = async (text: string): Promise<Key[]> => {
-    const result = await opgp.key.readArmored(text);
+    const result = await opgp?.key?.readArmored(text);
```
tomholub, line 22: This looks definitely suspect - if there's opgp missing, it needs to be added somewhere as a `<script>` tag

extension/js/common/settings.ts
```diff
@@ -276,8 +277,14 @@ export class Settings {
           window.location.href = Url.create('/chrome/settings/setup.htm', { acctEmail: response.acctEmail, idToken: response.id_token });
         }
       } else if (response.result === 'Denied' || response.result === 'Closed') {
-        if (settingsTabId) {
-          await Settings.renderSubPage(acctEmail, settingsTabId, '/chrome/settings/modules/auth_denied.htm');
+        const authDeniedHtml = await Api.ajax({ url: '/chrome/settings/modules/auth_denied.htm' }, Catch.stackTrace()) as string; // tslint:disable-line:no-direct-ajax
```
tomholub, line 280: Here, since it's just a bunch of text, I'd just make a private method like `getAuthDeniedHtml` that does something like
```ts
return [
  'whatever',
  'second line',
  ...
].join('<br>');
```
I think it will be "more together" then keeping it in a separate file.

Then you can delete auth_denied.html and auth_denied.ts

tomholub: If the two files are still used elsewhere, we should ideally refactor these situations too.

tomholub: You could put more stuff in the method to make it `renderAuthDeniedHtmlWithOptionToRetry` or whatever. The method we are in now is big enough already, anyway.

yantakus: image
it is used all across the app, that's why I didn't want to duplicate the text.

tomholub: Ah - didn't know. We can evaluate it in another PR. Ideally this would be just modals everywhere.

extension/js/common/assert.ts
yantakus: You can remove it and check yourself - it throws errors when a user closes google auth modal. I'm not familiar with the project so close to fix it properly, that's why I did it like this. If you have other ideas, please let me know.

extension/js/common/core/crypto/pgp/openpgp-key.ts
yantakus: The same as above.

tomholub:
I'm not sure what went wrong before - I have simply removed the `?` around openpgp and the other code, and it seems to work. Your fix appears to be preserved, so all is good.

I took the opportunity to simplify the auth_denied javascript so that it has no dynamic functionality and is easy to replace with a modal.

Also fixed the actual link that was in the modal, as it was now, clicking it did nothing.

Finally there was a code style error reported by the linter.

Now all looks ok.

tomholub (approved):

PR #3068, opened 2020-10-20
Signed only attachments can be viewed after send

### LINKED ISSUE #2834: Attachments for signed-only messages are rendered as encrypted

rrrooommmaaa:
Fixes #2834


tomholub (commented):
The fix and the test both look great. Not sure why it's failing yet, but it can probably be debugged.

tomholub (approved):
Excellent, looks great.

Issue #3051, opened 2020-10-19
potentially flaky test - attach with keyboard


tomholub:
https://flowcrypt.semaphoreci.com/jobs/647e0cc2-3efa-4e12-bd11-3f83a806b224

![image](https://user-images.githubusercontent.com/6306961/96438768-1660c600-11f6-11eb-87c0-7094f32d6a2d.png)

tomholub:
one more https://flowcrypt.semaphoreci.com/jobs/8e084e54-66fa-4bd9-8233-7a9b33bdada9

I may have a look later

tomholub:
Seems to still happen https://flowcrypt.semaphoreci.com/jobs/d2dea9d2-7824-4969-9c61-41d20ca85e18

PR #3078, opened 2020-10-23
'Remove key' button is hidden when using EKM

### LINKED ISSUE #2933: hide remove button when using EKM
Currently when OrgRules.usesKeyManager returns truthy value, we hide the "add key" button because keys are managed by the EKM. https://flowcrypt.com/docs/technical/email-key-manager/latest/email-key-manager.html

Without EKM it looks like this:

image

With EKM currently looks like this (no "Add Key"):

image

This issue is to also hide the "remove" button since user has no business doing manual actions on their keys when the keys are managed with EKM.

You will want to add a test similar to this one:

```ts
    ava.default('settings - my key page - remove key', testWithBrowser('compatibility', async (t, browser) => {
      const settingsPage = await browser.newPage(t, TestUrls.extensionSettings('flowcrypt.compatibility@gmail.com'));
      await SettingsPageRecipe.ready(settingsPage);
      await SettingsPageRecipe.toggleScreen(settingsPage, 'additional');
      await settingsPage.waitAll('@action-open-add-key-page');
      await settingsPage.waitAndClick('@action-remove-key');
      await settingsPage.page.waitForNavigation({ waitUntil: 'networkidle0' });
      await Util.sleep(1);
      await settingsPage.waitAll('@action-open-add-key-page');
      await settingsPage.notPresent('@action-remove-key');
    }));
``` 

But you want to create another mock account, similar to `get.key@key-manager-autogen.flowcrypt.com` which will return two keys in `test/source/mock/key-manager/key-manager-endpoints.ts` so that you have a chance to see the remove button on the second key. 

Further, the beginning of the test should look more like this:
```
    ava.default('get.key@key-manager-autogen.flowcrypt.com - automatic setup with key found on key manager', testWithBrowser(undefined, async (t, browser) => {
      const acct = 'get.key@key-manager-autogen.flowcrypt.com';
      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acct);
      await SetupPageRecipe.autoKeygen(settingsPage);
```

You are passing `undefined` into `testWithBrowser` (meaning you want to set up the account manually) and then you are using `SetupPageRecipe.autoKeygen` to tell it to go through the automatic keygen / key retrieval setup scenario where it grabs keys from the EKM.

Then you write finish the test to check the "remove" button is not there, and then you can fix the actual issue.

The actual fix will be somewhere here:

image

rrrooommmaaa:
Fixes #2933


test/source/tests/settings.ts
```diff
@@ -106,6 +107,18 @@ export let defineSettingsTests = (testVariant: TestVariant, testWithBrowser: Tes
       await settingsPage.notPresent('@action-remove-key');
     }));
 
+    ava.default('settings - my key page - remove button should be hidden when using key manager', testWithBrowser(undefined, async (t, browser) => {
+      const acct = 'two.keys@key-manager-autogen.flowcrypt.com';
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acct);
+      await SetupPageRecipe.autoKeygen(settingsPage);
+      await SettingsPageRecipe.toggleScreen(settingsPage, 'additional');
+      // check imported key at index 1
+      const myKeyFrame = await SettingsPageRecipe.awaitNewPageFrame(settingsPage, `@action-show-key-1`, ['my_key.htm', 'placement=settings']);
```
tomholub, line 116: Here can also add `await SettingsPageRecipe.waitAll('@action-show-key-2')` just to make sure it's rendered too.

tomholub (commented):
Looks great, just a suggestion for an improved test

tomholub:
The failed test is a flaky one - https://github.com/FlowCrypt/flowcrypt-browser/issues/3051 we'll be fixing it soon. You can ignore it

test/source/tests/settings.ts
rrrooommmaaa: there are only two. key-0 and key-1

tomholub: Got it, then it's good

tomholub (approved):

Issue #3071, opened 2020-10-21
replace removing auth_denied.html|ts with a modal


tomholub:
![image](https://user-images.githubusercontent.com/2179726/96745222-904ba780-13ce-11eb-8583-7a038f0239a9.png)
it is used all across the app, that's why I didn't want to duplicate the text.

_Originally posted by @yantakus in https://github.com/FlowCrypt/flowcrypt-browser/pull/3063#discussion_r509405148_

tomholub:
You can address these one by one, giving you a chance to fix up the tests after each change (if needed). It can be one PR (if easy enough) or it can be several.

yantakus:
@tomholub I can't reproduce any of these cases. Could you please help me with the reproduction steps (as [here](https://github.com/FlowCrypt/flowcrypt-browser/issues/2749) for example)?
1) https://github.com/FlowCrypt/flowcrypt-browser/blob/a74b43931075ea5063331197870459dd3777e1f9/extension/chrome/elements/compose-modules/compose-render-module.ts#L61
2) https://github.com/FlowCrypt/flowcrypt-browser/blob/74fc23465059661bcb65aafdfcf68b2c0dd4465e/extension/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.ts#L64
3) https://github.com/FlowCrypt/flowcrypt-browser/blob/2ffd45be40f12ea1a22f3df8c5745eccb054ff40/extension/chrome/settings/index.ts#L155
4) https://github.com/FlowCrypt/flowcrypt-browser/blob/9ea1217dcbf6edec4f11c3359f9370b8c5019d09/extension/chrome/settings/setup.ts#L145
5) https://github.com/FlowCrypt/flowcrypt-browser/blob/a74b43931075ea5063331197870459dd3777e1f9/extension/js/content_scripts/webmail/gmail-element-replacer.ts#L289

tomholub:
I'm not aware of any problem with these - I just figured we could take the opportunity to refactor them all away. Since it's just text we're showing, it could be a modal instead of an actual page - less code overall.

Actually, maybe you are saying that you cannot get the extension to this code path.

I think that's a valid point too - a long time ago, we used to allow user to proceed to set up the extension without the `read` permission. They could only send (let's say it's a paranoid user, doesn't want to give read perm). Therefore we had to keep checking for this. The first link is a good example:

```ts
if(this.view.scopes.read || this.view.scopes.modify) {
[...]

PR #3080, opened 2020-10-23
Remove all "action_go_auth_denied" usages

yantakus:
Fixes https://github.com/FlowCrypt/flowcrypt-browser/issues/3071#issuecomment-715460210


extension/chrome/settings/index.htm
```diff
@@ -66,7 +66,6 @@
             <b class="red">Following features are disabled for <span class="email-address"></span>:</b>
```
tomholub, line 66: Eventually we should remove the whole `auth_denied_warning`. I'll make another issue for that.

extension/chrome/settings/modules/auth_denied.htm
```diff
@@ -34,19 +34,12 @@ <h1>Explaining FlowCrypt webmail permissions</h1>
...
-    <div style="display: none;" class="display_if_rendered_as_subpage">
-      <div class="line">&nbsp;</div>
-      <div class="line">
-        <button class="button green action_auth_proceed">PROCEED</button>&nbsp;&nbsp;<button class="button gray close_page">CLOSE</button>
-      </div>
-    </div>
+    <div class="line">Would you like to proceed to the permissions screen again?</div>
   </div>
 
   <script src="/lib/purify.js"></script>
```
tomholub, line 40: I think all scripts can go. In fact the whole file can just become a few lines of js.

extension/chrome/settings/setup.htm
```diff
@@ -277,9 +277,6 @@ <h1></h1>
         <br />
```
tomholub, line 277: This would also eventually go, for another issue

extension/chrome/settings/setup.htm
```diff
@@ -307,9 +304,6 @@ <h1></h1>
...
```
tomholub, line 304: And this

extension/chrome/settings/modules/auth_denied.htm
tomholub: But it's ok to leave it as html, but in that case you don't need the HEAD, BODY tags etc. Just a bunch of content.

yantakus: But there's `<a href="#" data-swal-page="/chrome/texts/email_provider_auth_security.htm">`, that's why I left `sweetalert2.js`. And it obviously depends on `jquery.min.js`. The only thing is `purify.js`, but I thought it's also needed because goes first.

tomholub: Try removing all of it. It should continue working. I have added this functionality into the methods that render the modal.

image

yantakus: Yes, you are right. It works without the scripts. Removed.

tomholub (approved):

PR #3086, opened 2020-10-24
Remove primary key.

### LINKED ISSUE #409: ability to remove any private key when > 1 used
I walked through the wizard, let it create a key, now can't remove it to add my real key—the `(primary)` shows up instead of `(remove)`

---------------

// edited by tom: see details for implementing this in https://github.com/FlowCrypt/flowcrypt-browser/issues/409#issuecomment-662899507

yantakus:
Closes #409.


yantakus:
https://github.com/FlowCrypt/flowcrypt-browser/blob/da921ab4d80a836e3550ce4f88ee87828788f334/extension/js/common/core/crypto/key.ts#L75-L80

https://github.com/FlowCrypt/flowcrypt-browser/blob/a79419e99687c0021a769fdebd2a2d3646f09f13/test/source/platform/store/key-store.ts#L9-L13

Is there any particular reason for duplicating these?

tomholub:
I would remove the one in `flowcrypt-browser/test/source/platform/store/key-store.ts`. Then you can import in in tests from `test/source/core` which is a symlink to the `extension/js/common/core` (make sure not to import from the `extension` folder directly from tests, it would mess up build).

Thanks for noticing

yantakus:
@tomholub I'm not getting this:

> This following one is probably the trickiest. First, you'll have to figure out if any code is opening that frame with primary as one of the longid strings. If yes, please make that calling code instead give it the longid of the first stored key.

Could you please explain this in detail?

yantakus:
I'm signing off till Monday. Just removed the `primary` key from everywhere and pushed in case you want to see the code in order to help me with the above question.

tomholub:
> @tomholub I'm not getting this:
> 
> > This following one is probably the trickiest. First, you'll have to figure out if any code is opening that frame with primary as one of the longid strings. If yes, please make that calling code instead give it the longid of the first stored key.
> 
> Could you please explain this in detail?

You found a code that you want to refactor which is the following:

```ts
    const allPrivateKeys = await KeyStore.get(this.acctEmail);
    this.myPrivateKeys = allPrivateKeys
      .filter(ki => this.longids.includes(ki.longid) || (ki.primary && this.longids.includes('primary')));
```

You no longer want to deal with having a `primary` string because you are refactoring key storage to have no concept of primary keys. You'd rather keys be identified directly by id.

This is in `passphrase.ts` which is the main script in `passphrase.htm` which itself is rendered inside other pages as a modal. Also, `this.longids` comes from url params:

```ts
this.longids = Assert.urlParamRequire.string(uncheckedUrlParams, 'longids').split(',');
```

Some page is hosting / opening / rendering that `passphrase.htm` page that's settings that value. Let's take a look:

image

Brings us to:

```ts
  public srcPassphraseDialog = (longids: string[] = [], type: PassphraseDialogType) => {
    return this.frameSrc(this.extUrl('chrome/elements/passphrase.htm'), { type, longids });
  }
```

Brings us to:

image

Example

```ts
  public dialogPassphrase = (longids: string[], type: PassphraseDialogType) => {
    return this.divDialog_DANGEROUS(this.iframe(this.srcPassphraseDialog(longids, type), ['medium'], { scrolling: 'no' }), 'dialog-passphrase'); // xss-safe-factory
  }
```

Notice the `longids`, so far it's just passing the values around, let's go deeper:

```ts
    BrowserMsg.addListener('passphrase_dialog', async ({ longids, type }: Bm.PassphraseDialog) => {
      if (!$('#cryptup_dialog').length) {
        const factory = new XssSafeFactory(this.acctEmail!, this.tabId);
        $('body').append(factory.dialogPassphrase(longids, type)); // xss-safe-factory
      }
    });
```

still not there, now look for the `passphrase_dialog` browser message - whoever calls that. BrowserMsg is how scripts from various frames communicate together in the context of a browser extension.

```ts
    passphraseDialog: (dest: Bm.Dest, bm: Bm.PassphraseDialog) => BrowserMsg.sendCatch(dest, 'passphrase_dialog', bm),
```

See who calls that:

image

Quite a few places (I'm just going along a single path here for example, you'll need to ivestigate all of the other code paths above where I'm just using one for demonstration).

If I randomly pick one of them, what I'll see is eg this:

```ts
BrowserMsg.send.passphraseDialog(this.view.parentTabId, { type: 'message', longids: result.longids.needPassphrase });
```

That looks fine, `result.longids` are really a list of longids. I'm looking for some code to say something like `longids: ['primary']`. Let's see:

image

At least I found one, but I'm only looking causally for this demo

This code:

```ts
    this.view.S.cached('prompt').find('a.action_open_passphrase_dialog').click(this.view.setHandler(() => {
      BrowserMsg.send.passphraseDialog(this.view.parentTabId, { type: 'draft', longids: ['primary'] });
    }));
```

You'd want to instead grab keys from local storage, and get the first one, then just say `longids: [firstKey.longid]`

You don't want the `'primary'` string be co-mingled with actual ids.

tomholub:
There's one more problem we may like to address while we're at it.

image

The `KeyStore` also takes a `primary` string. Since you are removing `KeyInfo.primary`, then this code will be affected too.

```ts
  public static get = async (acctEmail: string, fingerprints?: string[]): Promise<KeyInfo[]> => {
    const stored = await AcctStore.get(acctEmail, ['keys']);
    const keys: KeyInfo[] = stored.keys || [];
    if (!fingerprints) {
      return keys;
    }
    return keys.filter(ki => {
      if (fingerprints.includes('primary') && ki.primary) {
        return true;
      }
      if (fingerprints.includes(ki.fingerprint)) {
        return true;
      }
      return false;
    });
  }
```

Instead of all of the `KeyStore.get(this.view.acctEmail, ['primary']);` calls, you could add a method `KeyStore.getFirst(this.view.acctEmail);` (which is really what it did) and refactor the `KeyStore.get` method accordingly.

extension/chrome/elements/passphrase.ts
```diff
@@ -35,7 +35,7 @@ View.run(class PassphraseView extends View {
     Ui.event.protect();
     await initPassphraseToggle(['passphrase']);
     const allPrivateKeys = await KeyStore.get(this.acctEmail);
-    this.myPrivateKeys = allPrivateKeys.filter(ki => this.longids.includes(ki.longid) || (ki.primary && this.longids.includes('primary')));
+    this.myPrivateKeys = allPrivateKeys.filter(ki => this.longids.includes(ki.longid));
```
tomholub, line 38: I just noticed, this `myPrivateKeys` has unfortunate name. It should say `keysWeNeedPassPhraseFor` or something like that. Please help me rename it.

extension/chrome/settings/index.ts
```diff
@@ -425,18 +425,16 @@ View.run(class SettingsView extends View {
       const created = new Date(prv.created);
       const date = Str.monthName(created.getMonth()) + ' ' + created.getDate() + ', ' + created.getFullYear();
       const escapedFp = Xss.escape(ki.fingerprint);
-      let escapedPrimaryOrRm = '';
-      if (ki.primary) {
-        escapedPrimaryOrRm = '(primary)';
-      } else if (canRemoveKey) {
-        escapedPrimaryOrRm = `(<a href="#" class="action_remove_key" data-test="action-remove-key" fingerprint="${escapedFp}">remove</a>)`;
+      let removeKeyBtn = '';
+      if (canRemoveKey) {
```
tomholub, line 429: Careful - `canRemoveKey` here means it's not forbidden by any policy, but that doesn't mean we should let users remove their very last key. Even if they normally are allowed to remove keys, they should not be allowed to remove the key if they only have one loaded - then the extension would become unusable.

extension/js/common/platform/store/key-store.ts
```diff
@@ -18,9 +18,6 @@ export class KeyStore extends AbstractStore {
       return keys;
     }
     return keys.filter(ki => {
-      if (fingerprints.includes('primary') && ki.primary) {
```
tomholub, line 21: You'll have to compensate for this somehow - see earlier comments

tomholub:
> I'm signing off till Monday. Just removed the primary key from everywhere and pushed in case you want to see the code in order to help me with the above question.

:+1: good approach - definitely helps to see something specific. Enjoy your weekend.

extension/js/common/platform/store/key-store.ts
```diff
@@ -25,6 +25,12 @@ export class KeyStore extends AbstractStore {
     });
   }
 
+  public static getFirst = async (acctEmail: string): Promise<KeyInfo> => {
```
tomholub, line 28: This returns `KeyInfo | undefined` - it's better to make it explicit so that someone doesn't get surprised when using it later.

tomholub: Actually, we could put `Assert.abortAndRenderErrorIfKeyinfoEmpty(keys[0]);` at the end - that will ensure that indeed the returned type is `KeyInfo`. Then anytime we call `Assert.abortAndRenderErrorIfKeyinfoEmpty` manually after calling `KeyStore.getFirst`, we could remove these lines.

extension/js/common/platform/store/key-store.ts
```diff
@@ -18,16 +18,19 @@ export class KeyStore extends AbstractStore {
       return keys;
     }
     return keys.filter(ki => {
```
tomholub, line 20: Can now be simplified further as `return keys.filter(ki => fingerprints.includes(ki.fingerprint));`

tomholub (commented):
All looks good - I have added last few adjustments to make. This change snowballed quite a bit, but it's good that this work is getting done.

tomholub:
I've added you to semaphore so you can also address the failing tests. https://flowcrypt.semaphoreci.com/workflows/3cceb1b5-98bd-43ba-b064-a27d9dcd564d

image

This is a result of the typing mismatch mentioned above.

extension/js/common/platform/store/key-store.ts
yantakus: If it's good now, maybe rename the method to `getFirstOrRenderError`?

tomholub: Good suggestion, we can rename it for clarity :+1:

yantakus:
> I've added you to semaphore so you can also address the failing tests. https://flowcrypt.semaphoreci.com/workflows/3cceb1b5-98bd-43ba-b064-a27d9dcd564d
> 
> image
> 
> This is a result of the typing mismatch mentioned above.

This error happens here: https://github.com/FlowCrypt/flowcrypt-browser/blob/a74b43931075ea5063331197870459dd3777e1f9/extension/chrome/settings/modules/my_key.ts#L48

This is the result of `KeyStore.get`, not the new method. The method's signature hasn't changed (it could return an empty array before as well and the above test could fail as well). What do I do in such a case?

tomholub:
> This is the result of KeyStore.get, not the new method. The method's signature hasn't changed (it could return an empty array before as well and the above test could fail as well). What do I do in such a case?

I see. I think you'll have to look at the context of the test, and understand why does it get to that state.

Check this out:

```ts
    this.fingerprint = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'fingerprint') || 'primary';
```
Maybe it's getting `primary` as a string from somewhere again?

tomholub:
Or rather, primary is used as a default value, so you'd want to change that to a required string, and make sure it's provided by the page that renders it.

extension/chrome/settings/modules/my_key.ts
```diff
@@ -36,17 +36,17 @@ View.run(class MyKeyView extends View {
     super();
     const uncheckedUrlParams = Url.parse(['acctEmail', 'fingerprint', 'parentTabId']);
     this.acctEmail = Assert.urlParamRequire.string(uncheckedUrlParams, 'acctEmail');
-    this.fingerprint = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'fingerprint') || 'primary';
+    this.fingerprint = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'fingerprint');
```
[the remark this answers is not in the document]
yantakus: How do I find `whichever places call it`? Made it required and now I have a failing test: https://github.com/FlowCrypt/flowcrypt-browser/blob/22a49cd0869343c50c92cdef22343517c1ec7435/test/source/tests/settings.ts#L49

I'm not sure how to fix it properly, could you please help me?

extension/chrome/settings/modules/my_key_update.ts
```diff
@@ -31,14 +30,13 @@ View.run(class MyKeyUpdateView extends View {
...
     this.acctEmail = Assert.urlParamRequire.string(uncheckedUrlParams, 'acctEmail');
-    this.fingerprint = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'fingerprint') || 'primary';
     this.showKeyUrl = Url.create('my_key.htm', uncheckedUrlParams);
   }
 
   public render = async () => {
     this.orgRules = await OrgRules.newInstance(this.acctEmail);
     this.pubLookup = new PubLookup(this.orgRules);
-    [this.primaryKi] = await KeyStore.get(this.acctEmail, [this.fingerprint]);
+    this.primaryKi = await KeyStore.getFirst(this.acctEmail);
```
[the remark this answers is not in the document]
yantakus: > Further, whichever places call this must be updated to always add the fingerprint to the URL.

How do I find these places?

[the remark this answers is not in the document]
tomholub: You can search it easily - by the name of the file:

image

extension/chrome/settings/modules/my_key.ts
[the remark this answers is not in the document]
tomholub: It was showing a `my_key.htm` page complain that it's missing fingerprint. Find the places that open that page:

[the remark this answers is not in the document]
yantakus: Here: https://github.com/FlowCrypt/flowcrypt-browser/blob/2a58e6322fe96b087d5d5caa0a408b2354fe93d4/extension/chrome/settings/index.ts#L34
`acctEmail` is optional, so I can't fetch the primary (first) key to get its fingerprint and pass it to `my_key` page. Any ideas? Can we somehow make the `acctEmail` required? It's obvious it may be missing because there's a check: https://github.com/FlowCrypt/flowcrypt-browser/blob/2a58e6322fe96b087d5d5caa0a408b2354fe93d4/extension/chrome/settings/index.ts#L241

extension/chrome/settings/index.ts
```diff
@@ -149,6 +149,16 @@ View.run(class SettingsView extends View {
         Catch.report(`Unknown target page in element: ${target.outerHTML}`);
       }
     }));
+    $('.action_open_public_key_page').click(this.setHandler(async target => {
+      const page = $(target).attr('page');
+      if (page) {
+        const ki = await KeyStore.getFirst(this.acctEmail!);
```
yantakus, line 155: @tomholub Just used `!` for testing purposes and it seems to work fine and the test passes. So is there a case when `acctEmail` is missing?

tomholub: `acctEmail` is missing when the account is not configured yet. Actually, it would be good to split this into two pages - the part of settings that are visible before an account is set up, and the part where acctEmail is always required. But that will be for another day, so it's ok to `!` bang it.

But this is not necessary:
```
 const page = $(target).attr('page');
      if (page) {
```
You can remove the `page` and other attributes from the html tag. You are now rendering it with code directly, so you don't need that. Hard-code the path into this click handler, and remove the path from the tag. That will simplify it.

tomholub (dismissed):

extension/chrome/settings/modules/my_key.ts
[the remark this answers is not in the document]
tomholub: I made an issue for this #3112 but for this current issue the `!` will work.

tomholub:
I'll go do some manual testing.

Code looks good. There seems to still be one test failure left.

tomholub:
Manual testing looks good

tomholub:
I'll fix it

tomholub (approved):

PR #3104, opened 2020-10-29
carefully wrap DOM element to fix #3076

### LINKED ISSUE #3076: FlowCrypt does not recognize signatures from recent Thunderbird versions
Thunderbird implemented their own OpenPGP now, and will stop relying on Enigmail.

There is something about their signatures that causes FlowCrypt to not realize the messages are signed.

Samples available on flowcrypt.compatibility@gmail.com account.

image

Use this to bring the sample into our mock tests:
```
  /**
   */
  private exportedMsgsPath = './test/source/mock/google/exported-messages/';
```

rrrooommmaaa:
Fixes #3076

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -472,8 +472,9 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     } else if (method === 'append') {
       if (replace) {
         const parent = msgBody.parent();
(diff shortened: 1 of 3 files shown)


tomholub (commented):
Test looks great. The fix is hard to follow but this is normal when fiddling with the Gmail DOM. Looks good - as long as it works.

rrrooommmaaa:
@tomholub The problem was that `msgBody.html()` returned `div` element without `id` attribute, while with DOM manipulations it was possible to preserve it.

tomholub (dismissed):

rrrooommmaaa:
@tomholub Let me add the second (plain) message to the test

tomholub (approved):

Issue #3091, opened 2020-10-26
ability to send s/mime attachments


tomholub:
Currently we have limited s/mime support - you can enter someone's email in Secure Compose, and if that person does not have any known OpenPGP key, you can click "import openpgp key" on the bottom. While the interface all only mentions OpenPGP, it will actually let you import an s/mime key.

If you import an s/mime key, you can then send out the email and it will be s/mime encrypted.

However, attachments don't work yet. Please add attachment support, and test compatibility by reading the s/mime messages in Thunderbird.

(FlowCrypt does not yet support opening s/mime emails, so have to use Thunderbird for testing)

You can get s/mime certs for testing here https://extrassl.actalis.it/portal/uapub/freemail?lang=en

tomholub:
This issue is for @vnikolaou

tomholub:
Questions from @vnikolaou 

> More questions rather important.
> After debugging a little bit the code ... i understand that for attachments at least based on x509 (sending an s/mime message).. we have calls to:
> 1. <att-ui.ts>.collectEncryptAtts()
> 2. <msg-utils.ts>.encryptMessage()
> 3. <smime-keys.ts>.encryptMessage()
> Currently there is a check inside att-ui.ts that checks for pubkey.type == 'x509' and if it is true break the flow and the message is not sent from composer.

> I tried to bypass this just to see what happens.. it seems that nothing is encrypted of course normally.. and i get an exception about not found packets over encrypted.message variable

> Generally... and given that i see that we use node-forge .. what do you think it is the library(another library maybe?) or the methods from this library (node-forge) that i should focus? On the other hand in order to have it so far disabled.. does it indicate any missing functionality or limitation in this library? Do you know?

Most likely it's disabled because the s/mime functionality is basically a proof of concept. We didn't spend time on attachments yet. The lib can probably do it.

One thing to keep in mind about design: there is MsgUtil class that is general and will generally branch into pgp-specific or smime-specific functionality based on need. Then there are classes like `OpenPGPKey` which are pgp-specific, and then there should also be s/mime specific classes somewhere. What you'd want to do is look into wherever mime-encryption of messages happens (in our code that uses node-forge) and see what happens when you add attachments, and/or how to fix it.

vnikolaou:
> Questions from @vnikolaou
> 
> > More questions rather important.
> > After debugging a little bit the code ... i understand that for attachments at least based on x509 (sending an s/mime message).. we have calls to:
> > 
[...]

PR #3123, opened 2020-11-03
issue 3091 s/mime attachments

vnikolaou:
// tom: this is related to issue #3091 

I have a created a simple JS file that you can run with node (v12+). The file is inside  `/extension/chrome/dev/smime_exp.js` and you can run it with `node extension/chrome/dev/smime_exp.mjs` (from the project root folder in command line or terminal).

So far and after experimenting a lot with messages and files, the node-forge doesn't seem able to produce smime messages with attachments (ie. multipart MIME messages) out of the box.

At the moment, what we are doing is to pass into the p7 content a simple text. This seems to be working though the size of the encrypted message is smaller than the size of a message being sent with thunderbird (that could indicate that we don't fully encrypt the message accordingly to the spec although the smime receiver is able to decrypt it).
In the case of multipart messages (eg with attachments), what should be passed into p7's content is a secured outer MIME entity that includes inner ones (if i understand well ). It is not clear how we this can be done. The API for node-forge doesn't  include any documented related methods that could be used for MIME messages. NOTE: If you try to pass a text that is the raw mime content .. then simply it is considered as text and not as mime content, so this doesn't work as well.

Googling also doesn't seem to provide any useful information for creating complex smime messages with node.js. There is a couple of libraries / modules but even these demonstrate simple examples as our (ie. no attachments). We may check them in action but i am not sure if we want to use such new modules anyway. Let me know for this..

Finally, the smime spec 3.1 doesn't provide necessarily clear explanation for the medium reader about how the encryption process works in action. Personally i found it partially useful in clearing some doubts about the algorithms used etc. but the encryption process is still a bit of mystery for me at least. 

Any thoughts or comments or suggestions? Any help on how we could potentially move on either with node-forge or other library ?


tomholub:
If node-forge has methods on how to encrypt raw bytes, but doesn't give you anything when it comes to then encoding the attachments into a mime message, then it sounds like we have to do the encoding ourselves.

One thing to get an idea is to encrypt an s/mime message with attachment in Thunderbird, and observe the content. That way you'll understand how it's supposed be structured. But there surely must be an RFC describing it somewhere. Maybe @wiktor-k will have some thoughts

vnikolaou:
> If node-forge has methods on how to encrypt raw bytes, but doesn't give you anything when it comes to then encoding the attachments into a mime message, then it sounds like we have to do the encoding ourselves.
> 
> One thing to get an idea is to encrypt an s/mime message with attachment in Thunderbird, and observe the content. That way you'll understand how it's supposed be structured. But there surely must be an RFC describing it somewhere. Maybe @wiktor-k will have some thoughts

@tomholub , actually an encrypted s/mime message with attachments (in thunderbird let's say) doesn't differ from a simpler one a lot. Only the size of the encrypted content differs usually (given that your attachments are large). Practically the headers are the same. I have already compared messages sent from our test code WITH same messages sent from thunderbird (both with the same CERT)and for text / text+attachments / text+rich text. Only the content size differs. The headers were the same in all submissions per case. So we are correct at least in the headers section ...

The RFC is not so useful at a first glance definitely. It is a reference key point for continuous reading... Thunderbird as i saw implements the s/mime v3.1 spec. I'll try to review the source code from existing libraries (open source) and check how they do smime for simple text (at least this is what they do in their examples.. just simple text). Maybe they use some other library which could offer bigger flexibility.

wiktor-k:
Hi guys!

From what I understood S/MIME is "just" MIME message that is encrypted. The simple case of only text is already a MIME message just a degenerate one that only has `Content-Type` header + `\r\n\r\n` + content of the message.

I did not verify that but the first thing I'd try for S/MIME + attachments is taking the MIME message with attachments and passing that through `p7.encrypt`.

Check out `smime-key.ts` and `encryptMessage` function.

I'll verify this later with a simple PoC message and report back (but wanted to give you an idea earlier).

tomholub:
That sounds simple, we already have all of the functionality for mime encoding, and we have the s/mime encryption method, so it should be really easy for @vnikolaou  to get it working at least as a PoC.

vnikolaou:
> Hi guys!
> 
> From what I understood S/MIME is "just" MIME message that is encrypted. The simple case of only text is already a MIME message just a degenerate one that only has `Content-Type` header + `\r\n\r\n` + content of the message.
> 
> I did not verify that but the first thing I'd try for S/MIME + attachments is taking the MIME message with attachments and passing that through `p7.encrypt`.
> 
> Check out `smime-key.ts` and `encryptMessage` function.
> 
> I'll verify this later with a simple PoC message and report back (but wanted to give you an idea earlier).

Hi @wiktor-k 

We already do something like:

```
    p7.content = forge.util.createBuffer('\r\n\r\n' + CONTENT);
    p7.encrypt();
```
So when you are saying passing the MIME into encrypt(), i guess to pass it into` p7.content` as above. If yes, what do you think to pass into? If it is a string that represents the MIME construction.. then will not work. The p7 will consider that as text and not as real MIME definition. I already did that and doesn't work..

Except of course if you mean something else..

tomholub:
You put text as well as attachments into `Mime.encode` and that gives you the string, give that to the s/mime encryption and you have attachments supported. There doesn't seem to be much more to it than that.

vnikolaou:
Just an early comment about what i did.. I used code from nodemailer + nodemailer-smime modules... and i managed to send an encrypted s/mime message with attachments after all. 
My implementation is quite dirty right now.. and outside the project ... I created a simple node project for this experiment locally (but i can share it as it is if needed). Anyway, i need to perform some checks in comparison with smime messages sent via thunderbird (to see if there are any differences in headers etc per case) and then integrate it somehow to current test page in our project.
Please mention that, we may probably need to extract code from these modules (not so much in my opinion; just a few classes perhaps).. so we may need to discuss if this is an option generally considering licensing or other similar issues.
I'll come back tomorrow with a clean example in this project (but again let me know if you want to see my node project somehow).

tomholub:
There shouldn't be any of that needed. We already have a method that encodes Mime - the `Mime.encode`. We also have a method that encrypts mime messages, and finally we have existing code that will then format the resulting encrypted message.

This is only a matter of grabbing the attachments and making sure they find their way into the encrypt method, where they get mime-encoded before getting encrypted. All of the rest is already handled.

vnikolaou:
> There shouldn't be any of that needed. We already have a method that encodes Mime - the `Mime.encode`. We also have a method that encrypts with s/mime, and finally we have existing code that will then format the resulting encrypted message.
> 
> This is only a matter of grabbing the attachments and making sure they find their way into the encrypt method, where they get mime-encoded before getting encrypted. All of the rest is already handled.

Well I didn't mean that we'll need to replace our code with these 2 libs.  But now i saw the Mime.Encode method. The funny thing is that has many common concepts about handling Mime subparts.. as the solution i have.... . Tomorrow I'll try to use that method !

wiktor-k:
> i guess you mean to pass it into p7.content as above

Yep, sorry for the mental shortcut. I've verified that this works. Create a regular MIME message that has attachment then do the `content` + `encrypt` dance and append regular headers.

For a quick PoC: https://metacode.biz/sandbox/smime set headers to:

```
MIME-Version: 1.0
Content-Type: multipart/mixed;
 boundary="------------50740425FD5ED0E2BF0AAB96"
Content-Language: en-US
```

and content to:

```
This is a multi-part message in MIME format.
--------------50740425FD5ED0E2BF0AAB96
Content-Type: text/plain; charset=utf-8
Content-Transfer-Encoding: 7bit

this is content

-- 
https://metacode.biz/@wiktor

--------------50740425FD5ED0E2BF0AAB96
Content-Type: text/plain; charset=UTF-8;
 name="file.txt"
Content-Transfer-Encoding: base64
 filename="file.txt"

dGhpcyBpcwpzYW1wbGUKZmlsZQoK
--------------50740425FD5ED0E2BF0AAB96--
```

This is properly displayed by Thunderbird as encrypted message with one attachment:

2020-11-03T21:38:19+01:00

By the way long term I think it would be good to migrate to PGP/MIME too so that all attachments are inside one envelope file. This better protects some metadata (file names etc) and I don't think there are clients nowadays that don't have support for PGP/MIME but @tomholub may want to chime in. For me the unification of S/MIME and PGP/MIME would be a good step but this may be non-trivial (I've looked at the code back then and it didn't seem like a quick fix).

tomholub:
pgp/mime is a story for another day https://partnerissuetracker.corp.google.com/issues/171715824

wiktor-k:
Can't access the resource even with my @flowcrypt.com e-mail Tom:

> Google ÜberProxy
> You do not have access to the requested resource
> Error Code 4:
> Access to this application has been denied. If you are an unrestricted FTE, contact the application owner for access. If you are a Restricted User (e.g. TVCs, GOC or Specialized Workforce Employees): Ask your manager to request access for you via (go/xw-access).

But I understand it's problematic :)

vnikolaou:
Thanks @wiktor-k , though i couldn't see the correct result after changing the public key of course. Anyway, no need to spend time in verifying this in my end.  The important thing is that you also managed to create the s/mime with attachments and confirmed that is possible. The fact also here is that the MIME message is passed into the enrypt method as a whole (all the sub-parts) and not separately.

vnikolaou:
Hi @tomholub , 

I want to express what i understand so far considering also the current implementation (at least from what i have seen):

We all managed to confirm that we are able to construct a valid s/mime message in our examples, by providing the correct MIME multipart representation including attachments. That's fine. However the integration part with the current code looks harder to me. 
The code separates the encryption of attachments. It encrypts the text in different methods and the attachments in other methods and then tries somehow to combine results. This works for PGP obviously. But at least from my example the MIME representation that is to be encrypted in S/MIME should be encrypted as a whole. I have the whole message, i get the proper representation in MIME format.. and all this is passed into the encrypt function based on node-forge. Correct me if I am wrong but you cant' encrypt partially the text and partially the attachments and combine results by merging them or something.

So if I am right, the abstraction that exists for PGP and S/MIME (ie the code for both modes pass through the same methods) should change in many points probably so as somewhere the encryption of S/MIME to take place taking the whole message including attachments. 
If this is true then breaking the code is very likely to happen if i do the changes. For a person that just came into the project may be not trivial and that is my consideration. The existing tests of course are a very important guard in making changes since they ensure the integrity of functionality (hopefully if they have a good code coverage) but still there are other factors that may be not desirable. For example the fact that i am not very familiar with the code could expand the time spent a lot on this task, something that maybe doesn't sound very good..

So let me know if I am wrong and where and how to deal with the integration.

wiktor-k:
>  though i couldn't see the correct result after changing the public key of course. 

Could you tell me your S/MIME public key and e-mail address? I can use the tool and send you some dummy message. If you have Thunderbird you could decrypt it there automatically.

> Anyway, no need to spend time in verifying this in my end.

This doesn't take a lot of time as I already had some tooling from the initial phase of S/MIME integration.

> The fact also here is that the MIME message is passed into the enrypt method as a whole (all the sub-parts) and not separately.

Yes, when I worked on the extension it wasn't that easy to do (encrypt all parts) but yeah, it works in general so there is hope :) :+1:

extension/chrome/dev/smime_exp.mjs
```diff
@@ -0,0 +1,77 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
```
tomholub, line 1: To delete

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -36,6 +37,7 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
     } else if (this.richtext) { // rich text: PGP/MIME - https://tools.ietf.org/html/rfc3156#section-4
       return await this.sendableRichTextMsg(newMsg, pubkeys, signingPrv);
     } else { // simple text: PGP/Inline with attachments in separate files
+      console.log('--- inside sendableMsg ---');
```
tomholub, line 40: To delete

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -56,6 +58,7 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
   }
 
   private sendablePwdMsg = async (newMsg: NewMsgData, pubs: PubkeyResult[], short: string, signingPrv?: Key) => {
+    console.log('-- Inside sendablePwdMsg --');
```
tomholub, line 61: To delete

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -65,11 +68,36 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
...
 
-  private sendableSimpleTextMsg = async (newMsg: NewMsgData, pubs: PubkeyResult[], signingPrv?: Key) => {
-    const atts = this.isDraft ? [] : await this.view.attsModule.attach.collectEncryptAtts(pubs);
-    const { data: encryptedBody, type } = await this.encryptDataArmor(Buf.fromUtfStr(newMsg.plaintext), undefined, pubs, signingPrv);
-    const mimeType = type === 'smime' ? 'smimeEncrypted' : undefined;
-    return await SendableMsg.create(this.acctEmail, { ...this.headers(newMsg), body: { "encrypted/buf": Buf.fromUint8(encryptedBody) }, type: mimeType, atts, isDraft: this.isDraft });
+  private sendableSimpleTextMsg = async (newMsg: NewMsgData, pubs: PubkeyResult[], signingPrv?: Key): Promise<SendableMsg> => {
+    console.log('-- Inside sendableSimpleTextMsg --');
+    const x509Pubs = pubs.filter(pub => pub.pubkey.type === 'x509').map(pub => pub.pubkey);
+    if (x509Pubs) {
```
tomholub, line 74: Here have to use `.length`

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -65,11 +68,36 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
...
+      // smime case
+      const atts: Att[] = this.isDraft ? [] : await this.view.attsModule.attach.collectAtts(); // collects attachments
+      const msgBody = this.richtext ? { 'text/plain': newMsg.plaintext, 'text/html': newMsg.plainhtml } : { 'text/plain': newMsg.plaintext };
+      const mimeEncodedPlainMessage = await Mime.encode(msgBody, { Subject: newMsg.subject }, atts);
+      const encryptedMessage = await SmimeKey.encryptMessage({ pubkeys: x509Pubs, data: Buf.fromUtfStr(mimeEncodedPlainMessage) });
+      console.log('-- End sendableSimpleTextMsg --');
```
tomholub, line 80: Rm comments

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -65,11 +68,36 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
...
+      /*
+            rootNode.setContent(body); // tslint:disable-line:no-unsafe-any
+            rootNode.addHeader('Content-Transfer-Encoding', 'base64'); // tslint:disable-line:no-unsafe-any
+            rootNode.addHeader('Content-Disposition', 'attachment; filename="smime.p7m"'); // tslint:disable-line:no-unsafe-any
+            rootNode.addHeader('Content-Description', 'S/MIME Encrypted Message'); // tslint:disable-line:no-unsafe-any
+      */
+      //const att = new Att({ data: Buf.fromUint8(encryptedMessage.data), name: 'smime.p7m', contentDescription: 'S/MIME Encrypted Message' });
+      return await SendableMsg.create(this.acctEmail, { ...this.headers(newMsg), body: { "encrypted/buf": Buf.fromUint8(encryptedMessage.data) }, type: 'smimeEncrypted', atts: [], isDraft: this.isDraft });
```
tomholub, line 88: Would be good to eventually have a specialized method like `createSmimeEncrypted` that would not have all the noise

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -65,11 +68,36 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
...
+      //const att = new Att({ data: Buf.fromUint8(encryptedMessage.data), name: 'smime.p7m', contentDescription: 'S/MIME Encrypted Message' });
+      //return await SendableMsg.create(this.acctEmail, { ...this.headers(newMsg), body: {}, type: 'smimeEncrypted', atts: [att], isDraft: this.isDraft });
+    } else {
+      // pgp case
+      const atts: any = this.isDraft ? [] : await this.view.attsModule.attach.collectEncryptAtts(pubs);
+      const { data: encryptedBody, type } = await this.encryptDataArmor(Buf.fromUtfStr(newMsg.plaintext), undefined, pubs, signingPrv);
+      const mimeType = type === 'smime' ? 'smimeEncrypted' : undefined;
+      const msg = await SendableMsg.create(this.acctEmail, { ...this.headers(newMsg), body: { "encrypted/buf": Buf.fromUint8(encryptedBody) }, type: mimeType, atts, isDraft: this.isDraft });
+      console.log(JSON.stringify(msg));
```
tomholub, line 97: Clean up

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -34,13 +34,14 @@ export class SmimeKey {
     return key;
   }
 
+  // NOTE: the methods accepts an already encoded mime message in data parameter
```
tomholub, line 37:
```
/**
 *
*/
```

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -34,13 +34,14 @@ export class SmimeKey {
...
   public static encryptMessage = async ({ pubkeys, data }: { pubkeys: Key[], data: Uint8Array }): Promise<{ data: Uint8Array, type: 'smime' }> => {
     const p7 = forge.pkcs7.createEnvelopedData();
     for (const pubkey of pubkeys) {
       p7.addRecipient(forge.pki.certificateFromPem(KeyUtil.armor(pubkey)));
     }
-    const headers = `Content-Type: text/plain`;
-    p7.content = forge.util.createBuffer(headers + '\r\n\r\n' + data);
+    // const headers = `Content-Type: text/plain`;
```
tomholub, line 43: remove this

tomholub (changes requested):

tomholub (approved):

PR #3124, opened 2020-11-04
Improve #input_from styles, add custom outline to it and #cc/#bcc buttons

### LINKED ISSUE #3037: Increase the clickable area height of `#input_from`
The clickable area of `#input_from` is only `18px`:

image

That's too little and easy to miss with a mouse cursor. The small height is easy to spot in Chrome when focusing it:

image

For example, in Gmail the height is `24px`:

image

---

Here's what I'd expect: 

image

limonte:
Fixes #3037

Also, added custom outlines as native outline were missing in Firefox for some reason:

0fix5BmFFA

Also, increased the height of `#cc` and `#bcc` buttons to `24px` so it won't be too small (prev height was `17px`)


tomholub (approved):
Thanks!

PR #3128, opened 2020-11-05
verifyDetached for signed only message tests and fixes

### LINKED ISSUE #3106: Pubkeys in signed messages from thunderbird show doubled
image

This should be improved. Somehow the message is being wrongly interpreted, with the public key rendered twice - that's a bug. It may have something to do with `Att.treatAs`, or it may be something else. I'll also make a separate issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/3076#issuecomment-719010367_

rrrooommmaaa:
Closes #3106


extension/js/common/core/att.ts
```diff
@@ -105,6 +105,8 @@ export class Att {
       return 'encryptedFile';
     } else if (this.name.match(/(cryptup|flowcrypt)-backup-[a-z0-9]+\.(key|asc)$/g)) {
       return 'privateKey';
+    } else if (this.type === 'application/pgp-keys') {
```
tomholub, line 108: Interesting this was missing :+1:

tomholub (approved):
Looks good code-wise

extension/js/common/core/att.ts
rrrooommmaaa: yes, it was identified as `encryptedMsg` and later the message was analyzed to detect it's a public key.
I decided that we'd better idenitify it as `publicKey` right away

tomholub: Excellent, thanks!

tomholub:
@rrrooommmaaa is this ready for review / merging or did you want to still improve it?

rrrooommmaaa:
@tomholub This looks ok in gmail for signed messages with detached signature,
but we haven't tested: both encrypted and signed message from Thunderbird.
Also, inbox-active-thread-module incorrectly displays plain text signed message (see picture)
This happens because `signedHtml` and `signedMsg` go different paths, the latter apparently without support for detached signature. I wonder if we should introduce `signedText` type as `signedMsg` may imply that the signature is inside it.
Or should we file separate tickets for any of these items?
image

tomholub (dismissed):

tomholub (commented):
By fixing up all of these tests, you will understand all of the internals - which means later you can help us all by simplifying all of this. This is very valuable :+1:

tomholub (approved):

tomholub (approved):

PR #3129, opened 2020-11-07
Add additional directives to CSP

seisvelas:
Adds the following directives:

    font-src (none)
    manifest-src (none/self)
    media-src (none)
    worker-src (self)
    form-action (none)

I left out `style-src` because we have a bunch of inline styles. Ideally, we should move that code to .css files and then add `style-src 'self'`. But for now I just wanted to include directives that don't break anything.

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -92,5 +92,5 @@
     "/chrome/elements/shared/footer.htm"
   ],
   "minimum_chrome_version": "61",
-  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self';"
+  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; worker-src 'none';"
 }
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -92,5 +92,5 @@
     "/chrome/elements/shared/footer.htm"
   ],
   "minimum_chrome_version": "61",
-  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; worker-src 'self';"
+  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; worker-src 'self'; object-src 'self';"
 }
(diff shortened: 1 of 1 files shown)


seisvelas:
Built and ran the extension in Chrome, things seem to work. I also tried violating the updated CSP (by inserting a `<video>` tag into the dom, the browser correctly blocked the attempt.

extension/manifest.json
```diff
@@ -92,5 +92,5 @@
     "/chrome/elements/shared/footer.htm"
   ],
   "minimum_chrome_version": "61",
-  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self';"
+  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; worker-src 'none';"
```
tomholub, line 95: The worker-src here says none but should say self. But that's the interesting part - I think we use a worker in the background page, so I would have expected this to fail.

seisvelas: I see what you mean, here:

https://github.com/FlowCrypt/flowcrypt-browser/blob/a74b43931075ea5063331197870459dd3777e1f9/extension/js/background_page/background_page.ts#L21

And here:

https://github.com/FlowCrypt/flowcrypt-browser/blob/fa582dbcb14895537292e02cfa5609956b407685/extension/chrome/elements/compose.ts#L139

Correct? And yet, when I compose and send, nothing goes wrong. I'll look deeper into why that might be the case.

Good catch!

tomholub:
Yup. Try decrypting - I think that one uses the background page.
But the tests should have caught this - I think.

tomholub:
What did you find - does it seem to have no effect?

Or maybe the worker gets denied but OpenPGP works fine without it? Then you would see error in background page console. This is my guess now, that would explain it.

tomholub (approved):

seisvelas:
>What did you find - does it seem to have no effect?

Correct, it makes no difference whether I put it as 'self' or 'none'. In an (ultimately vain) attempt to figure this out, I tried our CSP in Nginx and it worked exactly as it should. Calls to workers were blocked, for example:

```
>new Worker("/work.html")
VM11:1 Refused to create a worker from 'http://localhost/work.html' because it violates the following Content Security Policy directive: "worker-src 'none'".

(anonymous) @ VM11:1
VM11:1 Uncaught DOMException: Failed to construct 'Worker': Access to the script at 'http://localhost/work.html' is denied by the document's Content Security Policy.
    at <anonymous>:1:1
```

Then I created a very simple extension and experimented with worker-src and found the same behavior we're experiencing in FlowCrypt. So it looks like this directive is behaving oddly only when used with a browser extension. This is true in both Chrome and Firefox, which makes sense since they both use the WebExtensions API.

I'll look into worker-src separately. If someone else has any idea why extensions interact with worker-src this way, please let me know!

PR #3133, opened 2020-11-08
Re-enable offline setup test

### LINKED ISSUE #2661: resolve failing offline setup test
```ts
    ava.default.failing('setup - import key - submit - offline - retry', testWithBrowser(undefined, async (t, browser) => {
      // puppeteer seems to ignore `.setOfflineMode(true)` for localhost
      // don't want to start using non-localhost local host to avoid setup complications and keep tests straight forward
      // so for now disabling, hopefully future versions will fix this, in which case this failing test should start failing CI (because it will start passing)
      // an alternative solution could be to be able to dynamically switch mock settings to start/stop responding to a request (instead of setting puppeteer off/on line)
      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'flowcrypt.test.key.used.pgp@gmail.com');
      await SetupPageRecipe.manualEnter(settingsPage, 'flowcrypt.test.key.used.pgp', { submitPubkey: true, usedPgpBefore: true, simulateRetryOffline: true });
    }));

```

limonte:
Closes #2661 

Puppeteer indeed works as expected with offline mode and localhost.

### DIFF
--- a/extension/chrome/settings/setup.ts
+++ b/extension/chrome/settings/setup.ts
@@ -250,11 +250,6 @@ export class SetupView extends View {
     } else {
       addresses = [this.acctEmail];
     }
-    if (this.acctEmailAttesterPubId && this.acctEmailAttesterPubId !== pub.id) {
-      // already submitted another pubkey for this email
-      // todo - offer user to fix it up
-      return;
-    }
     await this.submitPubkeys(addresses, armoredPubkey);
   }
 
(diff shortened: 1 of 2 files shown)


limonte:
The re-enabled test should fail because the entered key isn't being submitted:

image

limonte:
@tomholub what would be the right way to prevent entering that `if` that prevents submitting the pubkey to attester? `this.acctEmailAttesterPubId` should be either `undefined` or `=== pub.id`.

tomholub:
This seems to be an oversight after refactoring - since we now use OIDC to authenticate public key submissions, we can freely submit a conflicting key which will get automatically replaced on Attester (since Attester can verify the authenticity of the request with OIDC).

This is different from before when attester would only accept the first public key, and then reject conflicting ones because it had no way to tell the second request is coming from the owner.

Try to just remove the if and the return. That should fix it :+1: thanks for debugging this.

tomholub (approved):

tomholub:
Thanks you!

PR #3136, opened 2020-11-09
Fix the recipients order

### LINKED ISSUE #2790: collapsed recipients not properly ordered
I'm writing an encrypted email:
````
to: recip@corp.co
cc: cc1@corp.co, cc2@corp.co
```

When I click away from the recipient field, it shows: [`cc1@corp.co`, `recip@corp.co`, `1 more`]

Which made me think that I actually added cc1 into `to` field, which wasn't true. Therefore, when ordering:
 - all to: recipients should come first
 - all cc: should follow
 - all bcc: should go last

Would be nice to have a test for this, too.

limonte:
Fixes #2790

### DIFF
--- a/test/source/tests/compose.ts
+++ b/test/source/tests/compose.ts
@@ -107,18 +107,17 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       await ComposePageRecipe.sendAndClose(composePage);
     }));
 
-    ava.default.only(`compose - recipients are properly ordered`, testWithBrowser('compose', async (t, browser) => {
+    ava.default(`compose - recipients are properly ordered`, testWithBrowser('compose', async (t, browser) => {
       const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
       composePage.page.setViewport({ width: 540, height: 606 });
       await ComposePageRecipe.fillMsg(composePage, { to: 'recip1@corp.co', cc: 'сс1@corp.co', bcc: 'bсс1@corp.co' }, 'recipients are properly ordered');
       await composePage.waitAndType(`@input-to`, 'recip2@corp.co');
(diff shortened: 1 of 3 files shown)


test/source/tests/compose.ts
```diff
@@ -107,6 +107,20 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      composePage.page.setViewport({ width: 540, height: 606 });
+      await ComposePageRecipe.fillMsg(composePage, { to: 'recip1@corp.co', cc: 'сс1@corp.co', bcc: 'bсс1@corp.co' }, 'recipients are properly ordered');
+      await composePage.waitAndType(`@input-to`, 'recip2@corp.co');
+      await composePage.waitAndType(`@input-cc`, 'cc2@corp.co');
+      await composePage.waitAndType(`@input-bcc`, 'bcc2@corp.co');
+      await composePage.waitAndFocus('@input-body');
+      await composePage.waitTillGone('@spinner');
+      const emailPreview = await composePage.waitAny('@recipients-preview');
+      const recipients = await PageRecipe.getElementPropertyJson(emailPreview, 'textContent');
+      expect(recipients).to.eq(['recip1@corp.co', 'recip2@corp.co', 'сс1@corp.co', 'сс2@corp.co'].join(''));
```
limonte, line 121: @tomholub is the expectation correct here? There should be no `bcc` emails according to #2790, is that so?

tomholub: Ah - I made a typo! " - all bcc: should go **lost**" should be " - all bcc: should go **LAST**"

limonte: Roger 👍

tomholub (approved):
Thanks! Excellent.

PR #3138, opened 2020-11-09
Use non-numeric enum for RecipientStatus

### LINKED ISSUE #2785: consider using non-numeric enums for RecipientStatus
I'm hoping there is a better way to encode enums in typescript than using numeric values explicitly:

```
export type RecipientStatus = 0 | 1 | 2 | 3 | 4 | 5;

export class RecipientStatuses {
  public static EVALUATING: RecipientStatus = 0;
  public static HAS_PGP: RecipientStatus = 1;
  public static NO_PGP: RecipientStatus = 2;
  public static EXPIRED: RecipientStatus = 3;
  public static WRONG: RecipientStatus = 4;
  public static FAILED: RecipientStatus = 5;
}
```

Please see if it can be easily refactored. If you hit a roadblock, then we can just let it be.

limonte:
Closes #2785

Also, fixed the spinner size inside of `.recipients-inputs`, it was too big and produced jumpiness:

Y6SgK1zvqh

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
@@ -3,7 +3,7 @@
 'use strict';
 
 
-import { RecipientStatuses, SendBtnTexts } from './compose-types.js';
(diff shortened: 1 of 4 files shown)


tomholub (approved):
Great!

PR #3143, opened 2020-11-11
Remove unused code and exports

### LINKED ISSUE #2711: run utility to find (and potentially remove) unused methods
While doing this, ignore the `core` folder because it includes methods potentially used on other platforms.

limonte:
Closes #2711

The best tool that I was able to find is https://github.com/nadeesha/ts-prune

The check is now the npm script:

```sh
npm run find_unused
```

I removed a couple of unused types and exports. The rest of the codebase seems to be fine 👍 

The only one small inconvenience is that `npm run find_unused` will always return status code `1` because there are some ignored findings:

image

It's possible to fix it with `loglevel=silent` in `.npmrc`, but I'm afraid that might cause confusions with running other scripts.


tomholub (approved):
Nice! It always bothered me that once you set a type as exported, it's forever "used" even if nothing uses it.

This particular tool focuses purely on types. It would be nice to also figure out which of our methods are unused? There are for sure some unused (public / static) methods.

PR #3153, opened 2020-11-13
Fix ordering items in #contacts dropdown

### LINKED ISSUE #3150: Invalid behavior of contacts search/autocomplete
### Expected behavior

1. When I write my own email
2. I expect only one suggestion - my own email

Here's Gmail:

image

### Actual behavior

In fact, I'm getting emails I was using before to test out things:

SzOM0l0bb9

In this case, if I hit `Enter` the totally valid email address I entered will be replaced by the first one from the suggestions list.

limonte:
Fixes #3150

### DIFF
--- a/test/source/mock/google/google-endpoints.ts
+++ b/test/source/mock/google/google-endpoints.ts
@@ -60,6 +60,22 @@ export const mockGoogleEndpoints: HandlersDefinition = {
             ]
           }
         };
+      } else if (q === 'testsearchorder') {
+        return {
+          feed: {
+            entry: [
+              { gd$email: [{ address: 'testsearchorder1@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder2@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder3@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder4@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder5@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder6@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder7@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder8@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder9@flowcrypt.com', primary: "true" }] },
+            ]
+          }
+        };
       } else {
         return empty;
       }
--- a/extension/chrome/elements/add_pubkey.ts
+++ b/extension/chrome/elements/add_pubkey.ts
@@ -87,12 +87,7 @@ View.run(class AddPubkeyView extends View {
     try {
       const keyImportUi = new KeyImportUi({ checkEncryption: true });
       const normalized = await keyImportUi.checkPub(String($('.pubkey').val()));
-      await ContactStore.save(undefined, await ContactStore.obj({
-        email: String($('select.email').val()),
-        client: 'pgp',
-        pubkey: normalized,
-        lastUse: Date.now(),
-      }));
+      await ContactStore.save(undefined, await ContactStore.obj({ email: String($('select.email').val()), client: 'pgp', pubkey: normalized }));
       this.closeDialog();
     } catch (e) {
       if (e instanceof UserAlert) {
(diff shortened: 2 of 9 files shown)


limonte:
@tomholub how would you test this?

tomholub:
You would pick a descriptive search term like `testsearchorder`. Then you would add an appropriate response in the mock:

```ts
  '/m8/feeds/contacts/default/thin': async ({ query: { q } }, req) => {
    if (!isGet(req)) {
      throw new HttpClientErr(`Method not implemented for ${req.url}: ${req.method}`);
    }
    const empty = { feed: { entry: [] } };
    const acct = oauth.checkAuthorizationHeaderWithAccessToken(req.headers.authorization);
    if (acct === 'ci.tests.gmail@flowcrypt.dev') {
      if (q === 'contact') {
        return {
          feed: {
            entry: [
              { gd$email: [{ address: 'contact.test@flowcrypt.com', primary: "true" }] }
            ]
          }
        };
      } else {
        return empty;
      }
    } else {
      return empty;
    }
  },
```
Where you ad `if(q === 'testsearchorder')` and you return more than 8 contacts in some "wrong" order.

Then you test that the browser extension renders them the way you expect when you search for the term in mock tests.

limonte:
@tomholub the sorting by `has_pgp` + `last_use` was also broken, I fixed it by making the sorting callback longer but easy to read and debug.

Question: I don't know how to add `has_pgp` and `last_use` to the mocked `'/m8/feeds/contacts/default/thin'` endpoind, could you advice me on that? I'd like to cover that in the test.

test/source/mock/google/google-endpoints.ts
```diff
@@ -60,6 +60,22 @@ export const mockGoogleEndpoints: HandlersDefinition = {
...
+            entry: [
+              { gd$email: [{ address: 'testsearchorder1@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder2@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder3@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder4@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder5@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder6@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder7@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder8@flowcrypt.com', primary: "true" }] },
+              { gd$email: [{ address: 'testsearchorder9@flowcrypt.com', primary: "true" }] },
```
tomholub, line 75: Yup. The complete test will be more difficult - for example you'll have to add public keys to some users (eg using manual key import - take a look at other compose code that does it) and these would then always be on the top.

tomholub:
> Question: I don't know how to add has_pgp and last_use to the mocked '/m8/feeds/contacts/default/thin' endpoind, could you advice me on that? I'd like to cover that in the test.

These don't come from the remote information, they come from the browser extension database.

To update last_use, you send message to that email.

To update has_pgp, you import any key for that email.

limonte:
Thank you for your help @tomholub 

I'm almost done, except one small detail.

> To update last_use, you send message to that email.

This doesn't seem to be 100% correct. `last_use` is somehow being updated without sending.

Here's the proof: https://github.com/FlowCrypt/flowcrypt-browser/pull/3153/files#diff-d1eb118e0deac42913d10eae1ea3b031ac10cdca408f8fc0242789e24c6dd7fdR538

The second time when I type `testsearchorder` it gives me the reversed list:

image

Which means that `last_use` of contacts is being updated somewhere without sending.

Is it something that should be investigated and fixed?

tomholub:
> Which means that last_use of contacts is being updated somewhere without sending.
> Is it something that should be investigated and fixed?

Definitely, sounds like a bug! No wonder your recipients were jumping around each time you searched.

limonte:
Tests are failing for the reason outside of this PR, the master branch is failing as well.

tomholub (approved):
Excellent! I'm extremely happy you discovered this, fixed it and wrote some good tests - it looks like it was completely broken.

tomholub:
Yes, it's testing to pull something from our server, which I'm currently fiddling with https://openpgpkey.flowcrypt.com/.well-known/openpgpkey/flowcrypt.com/hu/14bi5sf9whcm43n83ur8i6r4hfui5msk?l=tom

I'll fix it this morning I hope.

tomholub:
I think it should work now, re-running.

PR #3154, opened 2020-11-14
Fix ctrl+enter for chrome on windows

### LINKED ISSUE #3151: [Compose] Ctrl + Enter doesn't seem to work in Chrome
Strange issue, I remember this was working in Chrome, but not anymore. We have the test case `compose - keyboard - Ctrl+Enter sends message` which should cover this. Also, all is well in Firefox.

Can anyone confirm?

limonte:
Fixes #3151 

https://bugs.chromium.org/p/chromium/issues/detail?id=79407

> the keyCode is 10 instead of 13. a normal enter reports 13, like other browsers, only ctrl+enter is broken, and **only on windows**.


tomholub (approved):
Thanks!

Issue #3157, opened 2020-11-15
outgoing rich text signed pgp/mime message broken


tomholub:
possibly caused by #3127 - at least it worked on last version

it only affects rich text, which is only allowed on firefox

To reproduce:
1) use firefox
2) using send button chevron inco switch to "Rich text" and disable "encrypt" so you have a signed + rich text message
3) add text, you can make some part bold with ctrl+b just to make sure, then attach an image
4) view the sent message using FlowCrypt
5) view the sent message using Thunderbird (I haven't tried this)

![image](https://user-images.githubusercontent.com/6306961/99191460-c1688f00-2764-11eb-9b8f-5d9bcb415abb.png)

![image](https://user-images.githubusercontent.com/6306961/99191427-92521d80-2764-11eb-959b-325d521621df.png)

tomholub:
Here is the message it produced

[refactor - signed rich.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/5542790/refactor.-.signed.rich.txt)

rrrooommmaaa:
broken.txt
-- this is how it looks in gmail.
![image](https://user-images.githubusercontent.com/23335944/99654492-7489fe80-2a6b-11eb-8a05-9fd4e4953dc1.png)

in FlowCrypt "inbox":
![image](https://user-images.githubusercontent.com/23335944/99654566-8f5c7300-2a6b-11eb-9fea-47ab00916b8a.png)

So the error is that image is recognized as "plain file" in "inbox"?

tomholub:
The error is that on my end, I'm actually seeing this in Gmail:

![image](https://user-images.githubusercontent.com/6306961/99657063-924c6900-2a55-11eb-88b6-142fe29da4fb.png)

tomholub:
It's interesting that it rendered correctly for you, but not for me. Tried on both Chrome and Firefox.

![image](https://user-images.githubusercontent.com/6306961/99657348-f5d69680-2a55-11eb-8d3f-0ba26176b277.png)

tomholub:
I just noticed - you were looking at the message in the sent:folder.

I'm looking at the received message, which is the one I exported.

Try to import the received message, then you should be able to see the problem.

rrrooommmaaa:
@tomholub I see that the `broken.txt` file you attached has different delimeters as compared with Flowcrypt's "sent". So it is sort of "before processing by google", correct?

tomholub:
[...]

PR #3176, opened 2020-11-20
Added verification test for signed rich text message with image

rrrooommmaaa:
As part of #3157 investigation


tomholub (approved):
Thanks. For the actual Gmail API issue, I reported it. I hope Google will do something about it soon.

tomholub (approved):

PR #3182, opened 2020-11-23
Add ability to remove focused recipients with Del and Backspace keys

### LINKED ISSUE #3165: Ability to remove any contact with keyboard
### Given

There are 2 contacts in `#input_to`: `alice@domain.com`, `bob@domain.com`.

### The problem

Users can't delete `alice@domain.com` and keep `bob@domain.com` with the keyboard. The deletion only works with `Backspace` when the focus is in `#input_to` and not in any of the contacts. Users have to delete `bob@domain.com` first, then `alice@domain.com`. 

### Possible solution

The focused contact should remove itself when the user press ~~`Esc`~~ `Del` or `Backspace`.

limonte:
Closes #3165

### DIFF
--- a/test/source/tests/compose.ts
+++ b/test/source/tests/compose.ts
@@ -572,6 +572,27 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       ]);
     }));
 
+    ava.default('compose - delete recipients with keyboard', testWithBrowser('compatibility', async (t, browser) => {
+      const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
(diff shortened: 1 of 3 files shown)


test/source/tests/compose.ts
```diff
@@ -572,6 +572,27 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       ]);
     }));
 
+    ava.default('compose - delete recipients with keyboard', testWithBrowser('compatibility', async (t, browser) => {
```
tomholub, line 575: Great, thank you for the test :+1:

tomholub (approved):

PR #3197, opened 2020-11-27
Replace deprecated <font> tag.

### LINKED ISSUE #3130: Use of deprecated <font> tag
We use this tag here:

https://github.com/FlowCrypt/flowcrypt-browser/blob/9d1b2e1528d13925a061544aae99c8146cd0f8d8/test/source/browser/controllable.ts#L579-L584

It's part of tests and not a big deal imo *but* strictly speaking this tag is deprecated as of HTML 4 (and is not even included in HTML 5), and it's generally recommended to style text without the font tag.

yantakus:
Closes #3130

### DIFF
--- a/test/source/browser/controllable.ts
+++ b/test/source/browser/controllable.ts
@@ -576,12 +576,12 @@ export class ControllablePage extends ControllableBase {
     let html = '';
     for (const msg of this.consoleMsgs) {
       if (msg instanceof ConsoleEvent) {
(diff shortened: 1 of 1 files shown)


tomholub:
Thank you. Did you look for any html linter we could use for similar issues? I mentioned in on chat, but forgot to add it to issue description.

yantakus:
Never used or even heard of a linter for html.

tomholub (approved):

PR #3198, opened 2020-11-27
Show warnings on unusable keys

### LINKED ISSUE #2887: warn during import when key unusable for signing or encryption (Missing private key parameters)
Currently we allow user to import a key if it can be used for at least something (encryption or signing).

We should update `KeyImportUI` to show a warning if a key is not usable for both of these, so that the user is aware.

Further, what happened to the user below may be a common problem. Normally private keys are structured as follows:

1) primary key for signing
2) subkey for encryption

When they use `--export-secret-subkeys`, it will only export the subkey and therefore FlowCrypt was missing a key to do signatures with. We should provide further guidance to the user when we detect this problem during import, with the suggestions I'm mentioning below.

------------------------------------

A few interesting parts in OpenPGP compatibility tests of the affected key:

```
(15) PK 0 > Encrypt/Decrypt test: Encryption with key was successful  
(16) PK 0 > Encrypt/Decrypt test: Decryption with key succeeded  
(17) PK 0 > Sign/Verify test: [-] Exception: Error: Missing private key parameters  
```

It was using the subkey for encryption and decryption, but it seems like it wanted to use the primary key for signing. But the primary key is not present because you used `--export-secret-subkeys`.

This is because your subkey is missing signing key flag:
```
(22) PK 0 > SK 0 > SIG 0 > Key flags: [-] 12  
```
If I'm reading this correctly, you have 0x04 and 0x08 flags (both for encryption) but not 0x02 (signing). This is standard, typically signing is done by the primary key.

So you have a choice of either exporting the full key:
```
gpg --output private-key.asc --armor --export-secret-key your@email.com
```

Or modifying your key to include the signing flag on the subkey, too, then exporting the same way as before.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/2743#issuecomment-674379035_

rrrooommmaaa:
Fixes #2887


tomholub (commented):
Looks good so far

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -192,13 +192,37 @@ export class OpenPGPKey {
     }
     const algoInfo = opgpKey.primaryKey.getAlgorithmInfo();
     const key = keyToUpdate || {} as Key; // if no key to update, use empty object, will get props assigned below
+    const encryptionKey = await Catch.undefinedOnException(opgpKey.getEncryptionKey());
+    const encryptionKeyIgnoringExpiration = encryptionKey ? encryptionKey : await OpenPGPKey.getExpiredKey(opgpKey.getEncryptionKey.bind(opgpKey), exp, expired);
+    const signingKey = await Catch.undefinedOnException(opgpKey.getSigningKey());
+    const signingKeyIgnoringExpiration = signingKey ? signingKey : await OpenPGPKey.getExpiredKey(opgpKey.getSigningKey.bind(opgpKey), exp, expired);
+    const paramCountByAlgo = {
```
tomholub, line 199: Since this is a constant, it should be brought out of this method, to be a private property on this class

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -421,27 +445,24 @@ export class OpenPGPKey {
     return raw;
   }
 
-  private static usableButExpired = async (key: OpenPGP.key.Key, exp: Date | number | null, expired: () => boolean): Promise<boolean> => {
-    if (!key) {
-      return false;
-    }
-    if (! await Catch.doesReject(key.getEncryptionKey())) {
-      return false;
+  private static getExpiredKey = async (
```
tomholub, line 448: I'm suggesting `getKeyIgnoringExpiration` for consistency

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: I tried to, but failed to initialize it because `opgp` was undefined at the moment of constant initialization....

tomholub: Ah. Let's consider a private method - maybe `checkMissingPrivateKeyPackets`. That way you can remove the whole content of lines 199 to 216 and put it there, with a comment to explain why is `paramCountByAlgo` not a class constant. The method could return `{missingPrivateKeyForSigning, missingPrivateKeyForDecryption}`. Also, the method should have a link to this issue 2887 so that we can refer to it in the future.

rrrooommmaaa: @tomholub If we take it out to a separate method, than `usableForSigning` may return `true` while it's impossible to sign (missing packets). Well, it's up to you to decide, of course. We can also try to make lazy initialization on first access to `paramCountByAlgo`, filling it if it is null

tomholub: I leave it up to you then, you're the one working on the PR, so you'll know better.

rrrooommmaaa: UPD: yes, a private method called from `convertExternalLibraryObjToKey` -- good idea, I'll place some logic there.

tomholub (commented):
Looks much better :+1:

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -473,6 +454,25 @@ export class OpenPGPKey {
     return secondTry ? secondTry : null; // tslint:disable-line:no-null-keyword
   }
 
+  private static arePrivateParamsMissing = (packet: OpenPGP.packet.BaseKeyPacket): boolean => {
```
tomholub, line 457: Are you opening the vscode project using the project file in the repo, or simply as a folder?

If you opened it as a project, it should prompt you to install needed extensions. One of them should do auto-formating on save, which would have moved the method for you automatically.

rrrooommmaaa: Workspace file? I'll try that.

tomholub: Yup. The IDE should be a whole bunch smarter if you do that.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -196,19 +196,24 @@ export class OpenPGPKey {
     const algoInfo = opgpKey.primaryKey.getAlgorithmInfo();
     const key = keyToUpdate || {} as Key; // if no key to update, use empty object, will get props assigned below
     const encryptionKey = await Catch.undefinedOnException(opgpKey.getEncryptionKey());
-    const getEncryptionKey = opgpKey.getEncryptionKey.bind(opgpKey) as (keyid?: OpenPGP.Keyid | null, date?: Date, userId?: OpenPGP.UserId | null) => Promise<OpenPGP.key.Key | OpenPGP.key.SubKey | null>;
+    const getEncryptionKey = opgpKey.getEncryptionKey.bind(opgpKey) as
```
tomholub, line 199: Is there a way to achieve the same without using `.bind`? I'm trying to stay away from behavior that is too dynamic. Maybe it's just because I don't understand bind enough. But there is a preference for "dumb" code whenever possible, so that even I can read it :)

rrrooommmaaa: @tomholub Yes, actually in some other programming languages (C#) when you pass a reference to a method (opgpKey.getEncryptionKey) this reference actually includes `this` reference for the non-static method (`opgpKey`). For Javascript, I have to bind `opgpKey` explicitly so that `getEncryptionKey` will see it as `this`.
Of course we can bind it via lambda like `getEncryptionKey = (params) => opgpKey.getEncryptionKey(params)` if you're more comfortable with this programming-style binding.

tomholub: Yes that definitely looks better. I think it's really just a matter of being comfortable with a certain style, this way it will be the same as the rest of codebase.

tomholub: I thought I have forbidden `.bind` with a linter, but it looks like I didn't. I'll make another issue.

rrrooommmaaa:
@tomholub I made some change that seemed logical to me.
That is, set `usableButExpired` to false for an expired encryption key missing private key parameters.
(It can be used only for decryption of older messages only, right? -- so no point having it without private key).
As a consequence, some tests broke -- for example, 'decrypt - load key - expired key' test. Why does this test load public expired key? Is there any sense in that?

tomholub:
> @tomholub I made some change that seemed logical to me.
> That is, set `usableButExpired` to false for an expired encryption key missing private key parameters.
> (It can be used only for decryption of older messages only, right? -- so no point having it without private key).

Sounds ok, but only for private key. If it's public key, and it contains the necessary information to encrypt for it, but it's just expired, then it should say `true`.

You could still -encrypt- for such -public- key if you pretend to encrypt at an older date. We have such mechanism where user can choose to do it, after they confirm a warning that it's not recommended.

> As a consequence, some tests broke -- for example, 'decrypt - load key - expired key' test. Why does this test load public expired key? Is there any sense in that?

Sure, maybe you want to verify old signatures made by corresponding private key.

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: > 
> You could still -encrypt- for such -public- key if you pretend to encrypt at an older date. We have such mechanism where user can choose to do it, after they confirm a warning that it's not recommended.

Ok then, I'll restore `usableButExpired` calculation. I just derived that logic from key-import-ui.

> 
> Sure, maybe you want to verify old signatures made by corresponding private key.

`usableButExpired` is deduced based on encryption key only (signing key isn't taken into consideration). Maybe it's time to introduce 2 properties: `usableForEncryptionButExpired` and `usableForSigningButExpired`?

rrrooommmaaa:
> 
> You could still -encrypt- for such -public- key if you pretend to encrypt at an older date. We have such mechanism where user can choose to do it, after they confirm a warning that it's not recommended.

Ok then, I'll restore `usableButExpired` calculation. I just derived that logic from key-import-ui.

> 
> Sure, maybe you want to verify old signatures made by corresponding private key.

`usableButExpired` is deduced based on encryption key only (signing key isn't taken into consideration). Maybe it's time to introduce 2 properties: `usableForEncryptionButExpired` and `usableForSigningButExpired`?

tomholub:
> usableButExpired is deduced based on encryption key only (signing key isn't taken into consideration). Maybe it's time to introduce 2 properties: usableForEncryptionButExpired and usableForSigningButExpired?

good idea

tomholub (approved):
Excellent. Thank you.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -192,13 +195,26 @@ export class OpenPGPKey {
     }
     const algoInfo = opgpKey.primaryKey.getAlgorithmInfo();
     const key = keyToUpdate || {} as Key; // if no key to update, use empty object, will get props assigned below
+    const encryptionKey = await Catch.undefinedOnException(opgpKey.getEncryptionKey());
+    const getEncryptionKey = (keyid?: OpenPGP.Keyid | null, date?: Date, userId?: OpenPGP.UserId | null) =>
+      opgpKey.getEncryptionKey(keyid, date, userId);
+    const encryptionKeyIgnoringExpiration = encryptionKey ? encryptionKey : await OpenPGPKey.getKeyIgnoringExpiration(getEncryptionKey, exp, expired);
```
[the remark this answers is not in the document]
rrrooommmaaa: I just forgot about this operator 😊

PR #3203, opened 2020-11-30
Focus tab after opening it with window.open()

### LINKED ISSUE #2998: Firefox: unclear settings opening when already open in another tab
1) click browser extension icon -> settings (settings load)
2) click on another tab, leave settings in background tab
3) click browser extension icon -> settings (settings page reloads in background, but does not get switched to)

Ideally in step 3 it would change tabs so that the user finds themselves in the settings again

should also test on chrome

limonte:
Fixes #2998


tomholub (approved):
Thanks!

PR #3209, opened 2020-12-02
Evaluate atts container only once, detect new atts in it with mutation observer

### LINKED ISSUE #3180: sometimes a pgp block is not rendered at all from pgp/mime
No error showing in console either. Observed myself on Firefox. It was my message that was missing. Refreshing the page fixed it. Messages above in the conversation were showing without issues.

image

limonte:
Hopefully, fixes #3180

This PR is a blind shot since I wasn't able to reproduce the issue. 

There was kind of a TODO task in the method description which I implemented.
Here's what has been done:

1. `attsContainerInner` is now evaluated only once. After detecting a new atts container (which is still happening every 1s), it's being processed and marked with the `.evaluated` class to prevent from next evaluations
2. Children of `attsContainerInner`, are being evaluated as before
3. In addition to 2. there's now the mutation observer which will monitor for new attachments added to `attsContainerInner`'s DOM. Attachments detected with the mutation observer will be processed the same way as in 2.

The `processNewPgpAtts` method was extracted from the `replaceAtts` method without any changes.

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -122,8 +122,8 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   }
 
   private replaceArmoredBlocks = () => {
-    const emailsEontainingPgpBlock = $(this.sel.msgOuter).find(this.sel.msgInnerContainingPgp).not('.evaluated');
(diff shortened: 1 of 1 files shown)


tomholub (approved):
Thanks!

tomholub:
This may actually fix it.

Issue #2911, opened 2020-08-21
test for pgp mime tabnabbing


tomholub:
Built locally and tried with a rich text email, it worked for me. If it passes the tests, I should also probably make a test for this (should be a straightforward test, just send an email with a link, and check the email and see if the appropriate rel attribute is included).

_Originally posted by @seisvelas in https://github.com/FlowCrypt/flowcrypt-browser/pull/2910#issuecomment-677955390_

tomholub:
Save this message to `test/source/mock/google/exported-messages/message-export-174124764bcc0494.json`


```
{
  "acctEmail": "flowcrypt.compatibility@gmail.com",
  "full": {
    "id": "174124764bcc0494",
    "threadId": "174124764bcc0494",
    "labelIds": [
      "IMPORTANT",
      "CATEGORY_PERSONAL",
      "Label_15",
      "INBOX"
    ],
    "snippet": "-----BEGIN PGP MESSAGE----- Version: FlowCrypt Email Encryption 7.9.0 Comment: Seamlessly send and receive encrypted email wV4DcPFab5oU4iESAQdAIasjMG+7kyMIVjQDouEyJqlYOSYeudFy0Xuim0ON eR4wE92Nd/",
    "payload": {
      "partId": "",
      "mimeType": "multipart/mixed",
      "filename": "",
      "headers": [
        {
          "name": "X-Gm-Message-State",
          "value": "AOAM5305ZMzgaKTItfMB1mO5CujWI9zmTPuxtMvtMQUGhtt+q9mDljGG kVexLR8CcrGqXhMVHkVHeetnt1fQK89Q249YnKHr8S4eqk8="
        },
        {
          "name": "Openpgp",
          "value": "id=B80F2D4BBA40CFE8"
        },
        {
          "name": "From",
          "value": "sender@domain.com"
        },
        {
          "name": "MIME-Version",
          "value": "1.0"
        },
        {
          "name": "Date",
          "value": "Fri, 21 Aug 2020 11:28:21 -0700"
        },
        {
          "name": "Subject",
          "value": "PGP MIME link example"
        },
        {
          "name": "To",
          "value": "flowcrypt.compatibility@gmail.com"
        },
        {
          "name": "Content-Type",
          "value": "multipart/mixed; boundary=\"000000000000d33f0905ad6766c4\""
        }
      ],
      "body": {
        "size": 0
      },
      "parts": [
        {
          "partId": "0",
[...]

PR #3210, opened 2020-12-03
Add test for tabnabbing in PGP/MIME links

seisvelas:
Resolves https://github.com/FlowCrypt/flowcrypt-browser/issues/2911


test/source/tests/decrypt.ts
```diff
@@ -420,6 +420,15 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
       expect(content).to.include('-----BEGIN PGP MESSAGE-----Version: FlowCrypt 7.4.2 Gmail\nEncryptionComment: Seamlessly send and receive encrypted\nemailwcFMA0taL/zmLZUBAQ/+Kj48OQND');
     }));
 
+    ava.default('decrypt - inbox - prevent tabnabbing in PGP/MIME links', testWithBrowser('compatibility', async (t, browser) => {
+      const inboxPage = await browser.newPage(t, 'chrome/settings/inbox/inbox.htm?acctEmail=flowcrypt.compatibility%40gmail.com&threadId=174124764bcc0494');
+      await inboxPage.waitAll('iframe.pgp_block');
+      const pgpBlock = await inboxPage.getFrame(['pgp_block.htm']);
+      await pgpBlock.waitForSelTestState('ready');
+      const htmlContent = await pgpBlock.readHtml('#pgp_block');
+      expect(htmlContent).to.include('rel="noopener noreferrer"');
```
tomholub, line 429: This tests the header but not the actual goal. If tabnabbing allows for a particular vulnerability - like clicking a link in the message to take over the original frame - then we should be clicking the link in the test, to see if that will happen or not.

It would be also good, afterwards, to see that once you have such test, if you remove the tabnabbing fix, the test should actually fail.

seisvelas: I think it's worth putting both, because Chromium could add some kind of feature that makes window.opener null in this scenario, yet we'd still want to make sure the `rel="noopener noreferrer"` attribute is present to protect users on other browsers or older versions. 

I added a test that does as you suggest, and it works! If I disable the antitabnabbing mechanism, both of these new tests fail. When I readd it, they both succeed. 

*However*, my implementation of this test is somewhat of a mess, and it doesn't merit accession into the codebase as-is. Mostly due to my noob level Puppeteer skills. I was wondering if you perhaps have tips for how I might improve my new `window.opener` test?

tomholub: I think you've done it well. It does look a bit messy, but if it properly tests the intention, I'm ok to merge it.

tomholub (approved):

PR #3213, opened 2020-12-03
Added emails and fingerprints to PrvKeyInfo #1640

### LINKED ISSUE #1640: add all subkey fingerprints, and uid emails, into KeyInfo as optional arrays
The `KeyInfo` format is missing information about subkey longids / fingerprints, and also about emails listed in the key.

If we had this information in storage, we could avoid parsing keys when we need only the subkey longids, eg:
```js
      for (const ki of kiWithPp.keys) {
        // this is inefficient because we are doing unnecessary parsing of all keys here
        // better would be to compare to already stored KeyInfo, however KeyInfo currently only holds primary longid, not longids of subkeys
        // while messages are typically encrypted for subkeys, thus we have to parse the key to get the info
        // we are filtering here to avoid a significant performance issue of having to attempt decrypting with all keys simultaneously
        const { ids } = await Pgp.key.details(await Pgp.key.read(ki.private));
        for (const { longid } of ids) {
          if (keys.encryptedFor.includes(longid)) {
            keys.prvMatching.push(ki);
            break;
          }
        }
      }
```
These new fields can be added as optional type. We always add it when creating KeyInfo, plus we'll add a migration in background_page that adds these for existing keys.

For example:

```diff
export interface PrvKeyInfo {
  private: string;
  longid: string;
+ fingerprints?: string[];
+ emails?: string[];
  passphrase?: string;
  decrypted?: Key;  // only for internal use in this file
  parsed?: Key;     // only for internal use in this file
}
```

Then in a future release, in another PR a few months later, when we're sure it went through the migration and works without issues, we can make the `fingerprints` and `emails` no longer optional, completely remove the `longid` filed from `KeyInfo`.

### LINKED ISSUE #3220: Use separate types for storage
as discussed in #1640 
for example, `Key` shouldn't be stored with `usable`, `expired` etc.
`KeyInfo` shouldn't be stored with `passphrase`

rrrooommmaaa:
Closes #1640 

// edit tom: closes #3220

### DIFF
--- a/extension/js/background_page/migrations.ts
+++ b/extension/js/background_page/migrations.ts
@@ -2,6 +2,29 @@
 
 'use strict';
 
+import { KeyInfo, KeyUtil } from '../common/core/crypto/key.js';
+import { Catch } from '../common/platform/catch.js';
+import { GlobalStore } from '../common/platform/store/global-store.js';
+import { KeyStore } from '../common/platform/store/key-store.js';
+
+const addKeyInfoFingerprints = async () => {
+  for (const acctEmail of await GlobalStore.acctEmailsGet()) {
+    const keyinfos = await KeyStore.get(acctEmail);
+    const output: KeyInfo[] = [];
+    for (const keyinfo of keyinfos) {
+      const processed = await Catch.undefinedOnException(KeyUtil.keyInfoObj(await KeyUtil.parse(keyinfo.private)));
+      if (processed) {
+        output.push(processed);
+      }
+    }
+    await KeyStore.set(acctEmail, output);
+  }
+}
+
 export const migrateGlobal = async () => {
-  // noop
+  const globalStore = await GlobalStore.get(['key_info_store_fingerprints_added']);
+  if (!globalStore.key_info_store_fingerprints_added) {
+    await addKeyInfoFingerprints();
+    await GlobalStore.set({ key_info_store_fingerprints_added: true });
+  }
 };
--- a/extension/js/background_page/migrations.ts
+++ b/extension/js/background_page/migrations.ts
@@ -11,7 +11,7 @@ const addKeyInfoFingerprints = async () => {
     const keyinfos = await KeyStore.get(acctEmail);
     const output: KeyInfo[] = [];
     for (const keyinfo of keyinfos) {
-      const processed = await KeyUtil.prvKeyInfoObj(await KeyUtil.parse(keyinfo.private));
+      const processed = await KeyUtil.keyInfoObj(await KeyUtil.parse(keyinfo.private));
       if (processed) {
         output.push(processed);
       }
--- a/extension/js/background_page/migrations.ts
+++ b/extension/js/background_page/migrations.ts
@@ -3,7 +3,6 @@
 'use strict';
 
 import { KeyInfo, KeyUtil } from '../common/core/crypto/key.js';
-import { Catch } from '../common/platform/catch.js';
 import { GlobalStore } from '../common/platform/store/global-store.js';
 import { KeyStore } from '../common/platform/store/key-store.js';
 
@@ -12,7 +11,7 @@ const addKeyInfoFingerprints = async () => {
     const keyinfos = await KeyStore.get(acctEmail);
     const output: KeyInfo[] = [];
     for (const keyinfo of keyinfos) {
-      const processed = await Catch.undefinedOnException(KeyUtil.keyInfoObj(await KeyUtil.parse(keyinfo.private)));
+      const processed = await KeyUtil.keyInfoObj(await KeyUtil.parse(keyinfo.private));
       if (processed) {
         output.push(processed);
       }
--- a/extension/js/background_page/migrations.ts
+++ b/extension/js/background_page/migrations.ts
@@ -8,15 +8,12 @@ import { KeyStore } from '../common/platform/store/key-store.js';
 
 const addKeyInfoFingerprints = async () => {
   for (const acctEmail of await GlobalStore.acctEmailsGet()) {
-    const keyinfos = await KeyStore.get(acctEmail);
-    const output: KeyInfo[] = [];
-    for (const keyinfo of keyinfos) {
-      const processed = await KeyUtil.keyInfoObj(await KeyUtil.parse(keyinfo.private));
-      if (processed) {
-        output.push(processed);
-      }
+    const originalKis = await KeyStore.get(acctEmail);
+    const updated: KeyInfo[] = [];
+    for (const originalKi of originalKis) {
+      updated.push(await KeyUtil.keyInfoObj(await KeyUtil.parse(originalKi.private)));
     }
-    await KeyStore.set(acctEmail, output);
+    await KeyStore.set(acctEmail, updated);
   }
 };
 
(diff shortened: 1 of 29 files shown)


rrrooommmaaa:
@tomholub Looks like I broke `gmail` test setup, probably with `Config.secrets` refactoring. There is no longer "Secure Compose" button during live tests. Can you point me where the test is wiring to the gmail account so I can debug it properly?

rrrooommmaaa:
@tomholub The tests now went well.
I suggest to refactor `decrypted` field out of `PrvKeyInfo`.
Should we create a migration test?

tomholub:
Agree refactor decrypted away

No for migration test, but need to do thorough manual testing.

rrrooommmaaa:
I'm looking at Chrome console, but the local storage seems empty....
@tomholub Can you please assist?
image

tomholub:
Extension storage is not visible in browser console at all. There is no
mechanism to view it using UI in the browser

rrrooommmaaa:
> No for migration test, but need to do thorough manual testing.

I have run `background_page` in the debugger, the migration passed.

rrrooommmaaa:
We should probably have distinct types for storage and operation. I assume we never want to store `passphrase` in the storage?
So we should have a type without the passphrase to be stored

extension/js/common/core/crypto/key.ts
```diff
@@ -68,9 +68,9 @@ export type Contact = {
 export interface PrvKeyInfo {
   private: string;
   longid: string;
+  fingerprints?: string[];
```
tomholub, line 71: Since every use of `fingerprints` and `emails` is banging the optional with a `!`, we should not make these optional, and then remove these bangs.

extension/js/background_page/migrations.ts
```diff
@@ -2,6 +2,29 @@
...
+import { Catch } from '../common/platform/catch.js';
+import { GlobalStore } from '../common/platform/store/global-store.js';
+import { KeyStore } from '../common/platform/store/key-store.js';
+
+const addKeyInfoFingerprints = async () => {
+  for (const acctEmail of await GlobalStore.acctEmailsGet()) {
+    const keyinfos = await KeyStore.get(acctEmail);
+    const output: KeyInfo[] = [];
+    for (const keyinfo of keyinfos) {
+      const processed = await Catch.undefinedOnException(KeyUtil.keyInfoObj(await KeyUtil.parse(keyinfo.private)));
```
tomholub, line 15: This is a difficult decision to make - let it fail or not.

I tend to lean towards fail quickly and fail loud. Therefore we should not be ignoring errors here.

However the "fail loud" part is difficult, since this happens in a background script.

Either way, let's make it fail if there is an error.

extension/js/common/core/crypto/key.ts
```diff
@@ -297,4 +297,22 @@ export class KeyUtil {
     }
   }
 
+  public static keyInfoObj = async (prv: Key, passphrase?: string): Promise<KeyInfo> => {
```
tomholub, line 300: Looking at KeyInfo and PrvKeyInfo, I'm not sure if there is any use to having both anymore. I think nearly all usecases we have are for `PrvKeyInfo`. Wherever we need `KeyInfo`, we should use a `Key` instead (without storing it in storage).

I'm not 100% sure about this, but I think we could simplify this way.

extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -214,7 +214,7 @@ export class MsgUtil {
         return { success: false, error: { type: DecryptErrTypes.usePassword, message: 'Use message password' }, longids, isEncrypted };
       }
       const passwords = msgPwd ? [msgPwd] : undefined;
-      const privateKeys = keys.prvForDecryptDecrypted.map(ki => ki.decrypted!);
+      const privateKeys = keys.prvForDecryptDecrypted.map(decrypted => decrypted.decrypted!);
```
tomholub, line 217: It looks like the bang is not needed here, since decrypted.decrypted is not optional

test/source/util/index.ts
```diff
@@ -71,6 +71,17 @@ export class Config {
     return Config.secrets().keys.filter(k => k.title === title)[0];
   }
 
+  public static setupSecrets = async (): Promise<void> => {
+    await Config.fixKeyInfo(Config._secrets);
+  }
+
+  public static fixKeyInfo = async (secrets: TestSecretsInterface): Promise<void> => {
```
tomholub, line 78: Please add a comment here. "The keys in test secrets file used to have different structure, this does a migration so that we can continue using the file as is without distributing an updated secrets file to everyone"

tomholub (commented):
Code looks good with tome improvements to make. I will also do manual testing, once updated. Thanks!

extension/js/common/core/crypto/key.ts
rrrooommmaaa: `public` property is read in many places, for example, in keyserver (`replacePublicKeyHandler`)
In other places, like composer module, do you suggest to replace parse(ki.public) with parse(ki.private).asPublicKey?

tomholub (commented):
Code looks good. Will test tomorrow. Thanks!

extension/js/common/core/crypto/key.ts
rrrooommmaaa: Done

extension/js/common/core/crypto/pgp/msg-util.ts
rrrooommmaaa: Done

test/source/util/index.ts
rrrooommmaaa: Done

extension/js/common/core/crypto/key.ts
tomholub: I'll have to have a look and maybe attempt refactoring that away. I suspect there are a bunch of usages that are no longer needed, but I'll have to see them one by one to confirm.

extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -305,7 +305,10 @@ export class MsgUtil {
         continue;
       }
       const parsed = await KeyUtil.parse(ki.private);
-      if (MsgUtil.isKeyDecryptedFor(parsed, matchingKeyids) || await MsgUtil.decryptKeyFor(parsed, ki.passphrase!, matchingKeyids) === true) {
+      // todo - the `ki.passphrase || ''` used to be `ki.passphrase!` which could have actually allowed an undefined to be passed
+      // as fixed currently it appears better, but it may be best to instead check `ki.passphrase && await MsgUtil.decryptKeyFor(...)`
+      // but that is a larger change that would require separate PR and testing
+      if (MsgUtil.isKeyDecryptedFor(parsed, matchingKeyids) || await MsgUtil.decryptKeyFor(parsed, ki.passphrase || '', matchingKeyids) === true) {
```
tomholub, line 311: I have a concern about this particular original code, I'll have a second look later in a separate PR.

tomholub (commented):
Code looks ok so far (to my eyes) but need to fix tests again after my changes, and do manual testing.

tomholub (commented):
Tests seem to be passing, manual tests look good too.

tomholub:
@rrrooommmaaa can you please also review my changes that I added? These were to join PrvKeyInfo and KeyInfo, and also fix #3220 so that the result are: `KeyInfo` and `KeyInfoWithOptionalPassPhrase`.

tomholub (approved):
I'll go ahead and merge this, but please do have a look and let me know if I overlooked anything.

rrrooommmaaa (commented):
LGTM!

tomholub:
Thanks!

PR #3216, opened 2020-12-04
prepwork for working with x509 private keys

wiktor-k:
This is a draft PR to get the CI results (some tests fail for me locally). It'll be improved with time :)


extension/js/common/ui/key-import-ui.ts
```diff
@@ -105,18 +105,18 @@ export class KeyImportUi {
...
         const utf = file.getData().toUtfStr();
         if (utf.includes(PgpArmor.headers('privateKey').begin)) {
           const firstPrv = MsgBlockParser.detectBlocks(utf).blocks.filter(b => b.type === 'privateKey')[0];
           if (firstPrv) { // filter out all content except for the first encountered private key (GPGKeychain compatibility)
-            prv = (await opgp.key.readArmored(firstPrv.content.toString())).keys[0];
+            prv = (await KeyUtil.parse(firstPrv.content.toString()));
           }
         } else {
-          prv = (await opgp.key.read(file.getData())).keys[0];
+          prv = (await KeyUtil.parseBinary(file.getData()))[0];
```
tomholub, line 116: I may input a file that contains OpenPGP armored key. Please add this test case - since it didn't stop working, we're probably not testing it.

wiktor-k: The `if` from that `else` takes care of armored input but this has issues for X.509 private keys (the code takes first armored block and in X.509 case we need all of them). No worries I'll see what we can do in this case and add test cases along the way.

tomholub: I see it now, thank you.

extension/js/common/ui/key-import-ui.ts
```diff
@@ -105,18 +105,18 @@ export class KeyImportUi {
...
     attach.initAttDialog('fineuploader', 'fineuploader_button', {
       attAdded: async file => {
-        let prv: OpenPGP.key.Key | undefined;
+        let prv: Key | undefined;
         const utf = file.getData().toUtfStr();
         if (utf.includes(PgpArmor.headers('privateKey').begin)) {
           const firstPrv = MsgBlockParser.detectBlocks(utf).blocks.filter(b => b.type === 'privateKey')[0];
           if (firstPrv) { // filter out all content except for the first encountered private key (GPGKeychain compatibility)
-            prv = (await opgp.key.readArmored(firstPrv.content.toString())).keys[0];
+            prv = (await KeyUtil.parse(firstPrv.content.toString()));
```
tomholub, line 113: But now I'm thinking of this one :) Eventually we'll make `KeyUtil.parse` throw when it encounters more than one key.

As long as we do have a test for this scenario: `// filter out all content except for the first encountered private key (GPGKeychain compatibility)` then it's no problem - whenever we change it, we'll bump into the test and resolve it.

Actually, it's not the most ideal as is - if I give it two private keys, I may actually want to import two private keys. But now we're talking about importing more than one key at once, which is a different issue, I'll make a separate ticket for it.

wiktor-k: Yeah, I think at the end of the road we'll need to think about some clever way of merging all the `parse` functions but at the moment I'm working on handling private S/MIME keys. Once I've got that maybe I'll have a nice idea on how to join all these divergent ways of handling keys :)

extension/js/common/ui/key-import-ui.ts
```diff
@@ -105,18 +105,18 @@ export class KeyImportUi {
...
           }
         } else {
-          prv = (await opgp.key.read(file.getData())).keys[0];
+          prv = (await KeyUtil.parseBinary(file.getData(), String(prompt('Key password'))))[0];
```
wiktor-k, line 116: This logic here needs to be inverted. That is node-forge requires good password but when clicking on "import file" we don't have the password field shown yet.

I think we need to do only small level of sanity checks for X.509 keys here as it seems that's what the function is doing here (reads the key and immediately armors it discarding the object) and then do the full parse in `checkPrv`.

tomholub (approved):

tomholub:
I've cleaned it up and created followup issues so that this PR doesn't keep collecting dust and merge conflicts.

PR #3217, opened 2020-12-04
issue 3108 first attempt splitting KeyStore.getFirst

### LINKED ISSUE #3108: split getFirst into two methods: getFirstOptional and getFirstRequired
Almost all calls of `KeyStore.getFirst` assume that there will be a key returned. Many times, but not always, we also test it right below:

```ts
const primaryKi = await KeyStore.getFirst(this.acctEmail);
Assert.abortAndRenderErrorIfKeyinfoEmpty(primaryKi);
```

This is error prone, many times we forget to check.

But in some situations, we actually expect that that the key may not be present. You can tell such situations in case it follows with `if(primaryKi) { ... }` or something similar.

Todo in this issue:
 - [x] rename `KeyStore.getFirst` to `KeyStore.getFirstOptional` and change return signature to `Promise<KeyInfo | undefined>`
 - [x] add method `KeyStore.getFirstRequired` that contains just the call to `KeyStore.getFirstOptional` and `Assert.abortAndRenderErrorIfKeyinfoEmpty(primaryKi);`
 - [x] in the whole codebase, replace `getFirstOptional` with `getFirstRequired` wherever it seems to make sense - that is practically the whole codebase except maybe setup screen and other places, if any, that actually don't expect key to be present. 
- [x] where you switched the call to `getFirstRequired`, if that line followed with `Assert.abortAndRenderErrorIfKeyinfoEmpty(primaryKi);`, this check can now be removed

tomholub:
close #3108


tomholub:
@danishnavid008 please merge changes from master to here, and help me finish this PR. Thanks!

danishnavid008:
@tomholub This PR is ready for review

extension/js/common/assert.ts
```diff
@@ -35,7 +35,7 @@ export class Assert {
 
   public static abortAndRenderErrOnUnprotectedKey = async (acctEmail?: string, tabId?: string) => {
     if (acctEmail) {
-      const primaryKi = await KeyStore.getFirstOptional(acctEmail);
+      const primaryKi = await KeyStore.getFirstRequired(acctEmail);
```
tomholub: Here you switched to required, but you can see below that we are testing `&& primaryKi &&` for truthiness. That means it was expected that this key may be optional in this particular method. With this in mind, you should also scrutinize other parts.

tomholub (commented):
I see one error - please have a look. Have you evaluated all usages of `getFirstRequired` vs `getFirstOptional`? Are there no situations when you're unsure and would want to double check with me?

extension/js/common/assert.ts
danishnavid008: @tomholub Yes, I've evaluated all usages of `getFirstRequired` vs `getFirstOptional`. I had mistakenly used `getFirstRequired` but then I fixed it in the next commit.

danishnavid008: Can you please check the latest commit? You'll see I fixed it.

extension/chrome/settings/modules/change_passphrase.ts
```diff
@@ -74,6 +73,7 @@ View.run(class ChangePassPhraseView extends View {
       await Ui.modal.error('Pass phrase did not match, please try again.');
       $('#current_pass_phrase').val('').focus();
     }
+    Assert.abortAndRenderErrorIfKeyinfoEmpty(this.primaryKi);
```
tomholub, line 76: This one may have been added by mistake

tomholub (commented):
It looks good now - there is one remaining line that maybe was added by mistake.

Maybe earlier I was reviewing an older version - not sure. Thanks!

extension/chrome/settings/modules/change_passphrase.ts
danishnavid008: @tomholub Yeah, It was added by mistake while resolving the merge conflicts

tomholub:
All good, thank you

PR #3221, opened 2020-12-06
Removed bulkExport and UserMessages

### LINKED ISSUE #2753: transition mock data from external repo into this repo

rrrooommmaaa:
Closes #2753


tomholub (commented):
Looking forward! :)

tomholub (commented):
:+1:

tomholub (commented):
:+1:

tomholub (commented):
No small task, but you're getting there :+1:

rrrooommmaaa:
@tomholub Interestingly, looks like composer doesn't search for contacts in Drafts.
As I moved messages tagged with `DRAFT` label to DATA[acctEmail].drafts, the test `compose - can load contact based on name different from email` started to fail. Turns out `FirstName` only appears in on message labeled as DRAFT. Actually, in gmail Flowcrypt also doesn't recognize this recipient. Should this be filed as a bug in contacts search?
And how do we fix this PR? Should we create a non-draft message to search for `FirstName`?

tomholub:
> @tomholub Interestingly, looks like composer doesn't search for contacts in Drafts.
> As I moved messages tagged with `DRAFT` label to DATA[acctEmail].drafts, the test `compose - can load contact based on name different from email` started to fail. Turns out `FirstName` only appears in a single message labeled as DRAFT. Actually, in gmail Flowcrypt also doesn't recognize this recipient. Should this be filed as a bug in contacts search?
> And how do we fix this PR? Should we create a non-draft message to search for `FirstName`?

It's supposed to search in sent messages, not in drafts.

Therefore it seems behavior of the code is correct (if it can recognize sent messages, but not drafts), and therefore the message should be moved to sent folder (and maybe slightly updated), which should fix the test.

tomholub (approved):
All looks good if tests pass

rrrooommmaaa:
5 messages still have missing or empty "attachments" property in their JSON while they do have "attachmentId"s . This may affect future testing. Do you suggest to leave as it is or fill them just in case?

tomholub:
I think ok to leave them. Most commonly, when we need to test a new message-related thing, it will come with a newly exported message to test on.

Issue #3228, opened 2020-12-07
canceling, then re-trying consent screen causes error


tomholub:
Tested on Firefox, but probably true on Chrome too

Reproduce:
1) On setup screen click `Continue with Gmail`
2) close the auth window that shows
3) Click `OK` where it says `Explaining FlowCrypt webmail permissions`
4) this time actually approve the consent
5) nothing happens. Console shows:

> Unchecked runtime.lastError: No window with id: 27.

To fix this, you can change the `Explaining FlowCrypt webmail permissions` modal from `confirm` to `warning` and then the button will just close the modal. Remove the sentence `Would you like to proceed to the permissions screen again?`.

PR #3238, opened 2020-12-08
Fixed retrying consent screen error on cancelling

danishnavid008:
isseu #3228

### DIFF
--- a/extension/js/common/api/email-provider/gmail/google-auth.ts
+++ b/extension/js/common/api/email-provider/gmail/google-auth.ts
@@ -131,7 +131,6 @@ export class GoogleAuth {
     ]);
     if (authRes.result === 'Success') {
       chrome.windows.remove(oauthWin.id); // The prompt windows is removed when the user is authorized.
-
       if (!authRes.id_token) {
         return { result: 'Error', error: 'Grant was successful but missing id_token', acctEmail: authRes.acctEmail, id_token: undefined };
       }
--- a/extension/js/common/api/email-provider/gmail/google-auth.ts
+++ b/extension/js/common/api/email-provider/gmail/google-auth.ts
@@ -129,14 +129,9 @@ export class GoogleAuth {
       GoogleAuth.waitForAndProcessOauthWindowResult(oauthWin.id, acctEmail, scopes, authRequest.csrfToken, save),
       GoogleAuth.waitForOauthWindowClosed(oauthWin.id, acctEmail),
     ]);
-    try {
-      chrome.windows.remove(oauthWin.id);
-    } catch (e) {
-      if (String(e).indexOf('No window with id') === -1) {
-        Catch.reportErr(e);
-      }
-    }
     if (authRes.result === 'Success') {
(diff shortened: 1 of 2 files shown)


extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -129,14 +129,9 @@ export class GoogleAuth {
       GoogleAuth.waitForAndProcessOauthWindowResult(oauthWin.id, acctEmail, scopes, authRequest.csrfToken, save),
       GoogleAuth.waitForOauthWindowClosed(oauthWin.id, acctEmail),
     ]);
-    try {
-      chrome.windows.remove(oauthWin.id);
-    } catch (e) {
-      if (String(e).indexOf('No window with id') === -1) {
-        Catch.reportErr(e);
-      }
```
tomholub, line 137: Why do you propose removing the try/catch? Did it give you any trouble?

tomholub (changes requested):
I prefer that the bug is fixed as suggested in the issue:

> To fix this, you can change the Explaining FlowCrypt webmail permissions modal from confirm to warning and then the button will just close the modal. Remove the sentence Would you like to proceed to the permissions screen again?.

Your solution is not bad, but to make sure it doesn't break again, you'd have to write a test for it too (which is difficult until you get familiarized with our tests). Instead of doing that, it's easier to fix it as suggested.

extension/js/common/api/email-provider/gmail/google-auth.ts
danishnavid008: I figured out the error was with `chrome.windows.remove(oauthWin.id);` method. On cancel the chrome window is already removed by itself and we don't have the windows, so it throws the error. What we need to do was to delete the windows once the result is returned from the promise. That way we don't need the try/catch.

danishnavid008: If you want me to fix it by the suggested way I can do it as well.

tomholub: Ok. That does make sense. We can go with your proposal then, but I'll need to fix a failing test - there's a missing await. Take a look. https://flowcrypt.semaphoreci.com/workflows/44c0d9e4-4bc3-49e0-8ebc-9a8ce0c337d1

tomholub (approved):

tomholub:
It's missing a semicolon. These type of issues should be auto-fixed on every save when you open the workspace using the workspace file thanks to these settings:

```
    "editor.formatOnSave": true,
    "editor.formatOnPaste": true,
```

Maybe you have some personal settings that over-rule this?

danishnavid008:
> It's missing a semicolon. These type of issues should be auto-fixed on every save when you open the workspace using the workspace file thanks to these settings:
> 
> ```
>     "editor.formatOnSave": true,
>     "editor.formatOnPaste": true,
> ```
> 
> Maybe you have some personal settings that over-rule this?

@tomholub Sorry! I'll fix it and update the PR. I had opened the app using the workspace file and yeah it might be some personal settings that over-rule this.

tomholub (approved):

PR #3254, opened 2020-12-15
Issue 3156 contacts page searching

### LINKED ISSUE #3156: searching domain.com in contacts page does not find joe@domain.com

danishnavid008:
close #3156


tomholub (changes requested):
Thank you.

This will fix it, but it will also need a test. You can find a test of the contact page that we already have and edit it or copy it to create a new one, and try to search for a full domain during the test, making sure it will find the expected result.

tomholub (approved):

PR #3256, opened 2020-12-16
Extract inline images into attachments for plain rich text messages

### LINKED ISSUE #2460: format images of signed or plain rich text messages
Still todo: inline images of signed, or plain (rich text) messages

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/2398#issuecomment-570550384_

These should use MIME email embedded image references like other email clients do.

limonte:
Fixes #2460 

TODO: 
- [ ] @seisvelas https://github.com/FlowCrypt/flowcrypt-browser/pull/3256#discussion_r545256781

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
@@ -2,9 +2,12 @@
 
 'use strict';
 
+import * as DOMPurify from 'dompurify';
+
 import { ApiErr } from '../../../js/common/api/shared/api-error.js';
 import { Attachment } from '../../../js/common/core/attachment.js';
 import { BrowserMsg } from '../../../js/common/browser/browser-msg.js';
+import { Buf } from '../../../js/common/core/buf.js';
 import { Catch } from '../../../js/common/platform/catch.js';
 import { ComposeSendBtnPopoverModule } from './compose-send-btn-popover-module.js';
 import { GeneralMailFormatter } from './formatters/general-mail-formatter.js';
@@ -138,11 +141,49 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
         a.type = 'application/octet-stream'; // so that Enigmail+Thunderbird does not attempt to display without decrypting
       }
     }
+    if (choices.richtext && !choices.encrypt && !choices.sign) { // extract inline images of plain rich-text messages (#3256)
+      const { htmlWithInlineImages, imgAttachments } = this.extractInlineImagesToAttachments(msg.body['text/html'] as string);
+      msg.body['text/html'] = htmlWithInlineImages;
+      msg.attachments.push(...imgAttachments);
+    }
     if (this.view.myPubkeyModule.shouldAttach()) {
       msg.attachments.push(Attachment.keyinfoAsPubkeyAtt(senderKi));
     }
     await this.addNamesToMsg(msg);
   }
+
+  private extractInlineImagesToAttachments = (html: string) => {
+    const imgAttachments: Attachment[] = [];
+    DOMPurify.addHook('afterSanitizeAttributes', (node) => {
+      if (!node) {
+        return;
+      }
+      if ('src' in node) {
+        const img: Element = node;
+        const src = img.getAttribute('src') as string;
+        const { mimeType, data } = this.parseInlineImageSrc(src);
+        const imgAttachment = new Attachment({ name: img.getAttribute('name') || '', type: mimeType, data: Buf.fromBase64Str(data), inline: true });
+        imgAttachment.id = Attachment.attachmentId();
+        img.setAttribute('src', `cid:${imgAttachment.id}`);
+        imgAttachments.push(imgAttachment);
+      }
+    });
+    const htmlWithInlineImages = DOMPurify.sanitize(html);
+    DOMPurify.removeAllHooks();
+    return { htmlWithInlineImages, imgAttachments };
+  }
+
+  private parseInlineImageSrc = (src: string) => {
+    let mimeType;
+    let data = '';
+    const parts = src.split(/[:;,]/);
+    if (parts.length === 4 && parts[0] === 'data' && parts[1].match(/^image\/\w+/) && parts[2] === 'base64') {
+      mimeType = parts[1];
+      data = parts[3];
+    }
+    return { mimeType, data };
+  }
+
 
   private doSendMsg = async (msg: SendableMsg) => {
     // if this is a password-encrypted message, then we've already shown progress for uploading to backend
--- a/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
@@ -142,8 +142,8 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
       }
     }
     if (choices.richtext && !choices.encrypt && !choices.sign) { // extract inline images of plain rich-text messages (#3256)
-      const { htmlWithInlineImages, imgAttachments } = this.extractInlineImagesToAttachments(msg.body['text/html'] as string);
-      msg.body['text/html'] = htmlWithInlineImages;
+      const { htmlWithCidImages, imgAttachments } = this.extractInlineImagesToAttachments(msg.body['text/html'] as string);
+      msg.body['text/html'] = htmlWithCidImages;
       msg.attachments.push(...imgAttachments);
     }
     if (this.view.myPubkeyModule.shouldAttach()) {
@@ -173,9 +173,9 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
         imgAttachments.push(imgAttachment);
       }
     });
-    const htmlWithInlineImages = DOMPurify.sanitize(html);
+    const htmlWithCidImages = DOMPurify.sanitize(html);
     DOMPurify.removeAllHooks();
-    return { htmlWithInlineImages, imgAttachments };
+    return { htmlWithCidImages, imgAttachments };
   }
 
   private parseInlineImageSrc = (src: string) => {
--- a/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
@@ -141,8 +141,10 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
         a.type = 'application/octet-stream'; // so that Enigmail+Thunderbird does not attempt to display without decrypting
       }
     }
-    if (choices.richtext && !choices.encrypt && !choices.sign) { // extract inline images of plain rich-text messages (#3256)
-      const { htmlWithCidImages, imgAttachments } = this.extractInlineImagesToAttachments(msg.body['text/html'] as string);
+    if (choices.richtext && !choices.encrypt && !choices.sign && msg.body['text/html']) {
+      // extract inline images of plain rich-text messages (#3256)
+      // todo - also apply to rich text signed-only messages
+      const { htmlWithCidImages, imgAttachments } = this.extractInlineImagesToAttachments(msg.body['text/html']);
       msg.body['text/html'] = htmlWithCidImages;
       msg.attachments.push(...imgAttachments);
     }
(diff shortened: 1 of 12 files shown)


limonte:
@tomholub I'd like to ask for your advice here about this issue I'm struggling with.

I'm trying to align the behavior with Gmail's which looks like this:

image

But what I'm getting is

image

---

I'm creating the attachment by 

```js
const imgAttNode = Mime.createAttNode(imgAttachment);
```

then getting its id like this

```js
const imgAttachmentId: string = imgAttNode._headers.find((header: Dict<string>) => header.key === 'X-Attachment-Id').value;
```

and then setting the `img`'s `src` to

```js
img.setAttribute('src', `cid:${imgAttachmentId}`);
```

But after sending the attachment id seems to be changed and I can't figure out for now how to fix that.

Let me know if you have any advice here.

tomholub:
> But after sending the attachment id seems to be changed and I can't figure out for now how to fix that.

When looking at the MIME message, as you screenshotted, is that a screenshot taken AFTER it was sent throug Gmail API, or before uploading to the API?

Please have a look at the ids just before you upload the message to gmail, to know exactly how was the message constructed, making sure it is not Gmail modifying them.

If that's what you've indeed been looking at, I'll have to take a second / more thorough look.

tomholub:
Accidentally closed and re-opened

limonte:
> is that a screenshot taken AFTER it was sent throug Gmail API, or before uploading to the API?

After sending through Gmail.

> Please have a look at the ids just before you upload the message to gmail, to know exactly how was the message constructed, making sure it is not Gmail modifying them.
>
> If that's what you've indeed been looking at, I'll have to take a second / more thorough look.

Let me double-check that I'm extracting the attachment ID right before sending it to Gmail.

limonte:
Your question was good, I found my mistake. Thanks!

limonte:
@tomholub it works now, thanks! Sometimes a good question is better than the advice :)

Tests are still TODO in this PR, but please review the functionality.

Question: in `extractInlineImagesToAttachments()` I'm iterating images with `DOMPurify`. Is that the best way to perform `string -> nodes -> edit img tags -> back to string`?

limonte:
Another question: I noticed there was `Att -> Attachment` renaming recently. 

There are still some `/Att[A-Z]/` occurrences and I'm using `Mime.createAttNode()`

image

I guess we should also do `/Att[A-Z]/ -> /Attachment$1/`, if so I'd prefer it to be done in separate PR not to mix the fix and the refactoring in one PR.

tomholub:
Well noticed! Agree gree that would be a separate pr

Have to look at the other question later when on pc

extension/js/common/core/mime.ts
```diff
@@ -208,7 +210,11 @@ export class Mime {
       } else {
         contentNode = new MimeBuilder('multipart/alternative'); // tslint:disable-line:no-unsafe-any
         for (const type of Object.keys(body)) {
-          contentNode.appendChild(Mime.newContentNode(MimeBuilder, type, body[type]!.toString())); // already present, that's why part of for loop
+          let content = body[type]!.toString();
+          if (type === 'text/html') {
+            content = Mime.extractInlineImagesToAttachments(content, rootNode);
+          }
+          contentNode.appendChild(Mime.newContentNode(MimeBuilder, type, content)); // already present, that's why part of for loop
```
tomholub, line 217: `// already present, that's why part of for loop` this comment belongs to lone 213. I know it's not obvious at all - it was referring to the `!` bang, trying to justify it

extension/js/common/core/mime.ts
```diff
@@ -208,7 +210,11 @@ export class Mime {
       } else {
         contentNode = new MimeBuilder('multipart/alternative'); // tslint:disable-line:no-unsafe-any
         for (const type of Object.keys(body)) {
-          contentNode.appendChild(Mime.newContentNode(MimeBuilder, type, body[type]!.toString())); // already present, that's why part of for loop
+          let content = body[type]!.toString();
+          if (type === 'text/html') {
+            content = Mime.extractInlineImagesToAttachments(content, rootNode);
```
tomholub, line 215: I suspect this mechanism will also apply to pgp/mime messages, which is a mime message containing HTML, then encrypted as a whole and sent out as an attachment.

Since inside that encrypted envelope is a html mime message, I think it would use this mechanism as written.

Which may be a good thing. Just need to check that for example Thunderbird can render it (probably yes).

tomholub: Need to check that FlowCrypt itself can render it (pgp/mime rich text message with inline attachments like this).

extension/js/common/core/mime.ts
```diff
@@ -403,4 +409,36 @@ export class Mime {
     return node;
   }
 
+  private static extractInlineImagesToAttachments = (html: string, rootNode: any) => {
```
tomholub, line 412: Looks good in high-level overview :+1: but will need to give this more scrutiny later.

tomholub: @seisvelas should scrutinize it for attack vectors if any

extension/js/common/core/mime.ts
```diff
@@ -403,4 +409,36 @@ export class Mime {
...
+    });
+    const htmlWithInlineImages = DOMPurify.sanitize(html);
+    DOMPurify.removeAllHooks();
+    return htmlWithInlineImages;
+  }
+
+  private static parseInlineImageSrc = (src: string) => {
+    let mimeType;
+    let data = '';
+    const matches = src.match(/data:(image\/\w+);base64,(.*)/);
```
tomholub, line 438: This may be a very expensive regex if the image has a few megabytes.

To speed it up, I'd first add `/^` at the beginning, so it doesn't search the whole thing. I'd remove `(.*)` from the end, again so it doesn't need to go that far. If these two are found, then you can extract the data with `.split(',').pop()`.. but parsing is also a source of errors, I think @seisvelas should have a look. Eg if  we do it as I proposed, what if someone produces  `data:image/png;base64,realbase64,fakebase64` causing anomalies.. so you'd have to also check, if you follow this recommendation, how many parts does it have after splitting by `.`

extension/js/common/platform/xss.ts
```diff
@@ -98,14 +98,14 @@ export class Xss {
           a.href = src;
           a.className = 'image_src_link';
           a.target = '_blank';
-          a.innerText = (title || 'show image') + (src.startsWith('data:image/') ? '' : ' (remote)');
+          a.innerText = (title || 'show image') + (src.startsWith('data:image/') || src.startsWith('cid:') ? '' : ' (remote)');
```
tomholub, line 101: Need to take a second look later

tomholub (commented):
Looks pretty good :)

extension/js/common/core/mime.ts
limonte: Moved, thanks for catching it 👍

extension/js/common/core/mime.ts
limonte: > I suspect this mechanism will also apply to pgp/mime messages

Not anymore, now it'll be applied only for plain rich-text messages.

extension/js/common/core/mime.ts
limonte: Removed the expensive regex and also verified that the format is correct, e.g. there are no unexpected parts.

extension/js/common/platform/xss.ts
limonte: No need, I reverted this addition, it's not needed after https://github.com/FlowCrypt/flowcrypt-browser/pull/3256/commits/f68b85f42d4064391a4664f2af2c0949777ab6b9

limonte:
@tomholub do you maybe have any quick guess why the test is failing, there's no `` tag in `getMessageBySubject().payload.body.data`. I checked the sending, the HTML is correct right before sending.

test/source/mock/google/google-data.ts
```diff
@@ -186,7 +186,8 @@ export class GoogleData {
     }
     let body: GmailMsg$payload$body;
     if (parsedMail.text) {
-      body = { data: parsedMail.text, size: parsedMail.text.length };
+      const data = parsedMail.html || parsedMail.text;
```
tomholub, line 189: A cleaner way to write it would be to add `if(parsedMail.html)` as the first condition above

tomholub:
> @tomholub do you maybe have any quick guess why the test is failing, there's no `` tag in `getMessageBySubject().payload.body.data`. I checked the sending, the HTML is correct (with `` tag) right before sending.

Sorry haven't been able to have a look yet, can't think of anything without investigating

Except if the img got sanitized away somewhere using Xss class

limonte:
> > @tomholub do you maybe have any quick guess why the test is failing, there's no `` tag in `getMessageBySubject().payload.body.data`. I checked the sending, the HTML is correct (with `` tag) right before sending.
> 
> Sorry haven't been able to have a look yet, can't think of anything without investigating
> 
> Except if the img got sanitized away somewhere using Xss class

Thanks, I'll investigate further.

test/source/mock/google/google-data.ts
limonte: done 👍

limonte:
@tomholub I need your help with writing a test. There's `` instead of  ``. I was tracking it down to [this place](https://github.com/FlowCrypt/flowcrypt-browser/blob/83fb99f2c6ed7cb457f63d5b1f5b9d60d25047da/extension/js/common/api/email-provider/gmail/gmail.ts#L92), `mimeMsg` is as expected here, with extracted inline images.

I guess something is happening mocked google endpoints, but I can't figure out what and where. Let me know if you have time for helping me with this, if not, then I'll continue debugging it by myself.

tomholub:
Sorry, I only do light work this week so may take a bit of time before I get to this with meaningful help.

@rrrooommmaaa - if you have the time - could you look into this?

limonte:
> Sorry, I only do light work this week so may take a bit of time before I get to this with meaningful help.
> 
> @rrrooommmaaa - if you have the time - could you look into this?

Thanks, I realized that from your GitHub activity :) @rrrooommmaaa no need for help atm, I think I'll finish this by myself. I'll let you know additionally if the help will be needed here.

limonte:
> @tomholub I need your help with writing a test. There's `` instead of ``. I was tracking it down to [this place](https://github.com/FlowCrypt/flowcrypt-browser/blob/83fb99f2c6ed7cb457f63d5b1f5b9d60d25047da/extension/js/common/api/email-provider/gmail/gmail.ts#L92), `mimeMsg` is as expected here, with extracted inline images.

Found the reason: 

`simpleParser` is the easiest way to parse emails. You only need to provide a message source to get a parsed email structure in return. **As an additional bonus all embedded images in HTML (eg. the images that point to attachments using cid: URIs) are replaced with base64 encoded data URIs, so the message can be displayed without any additional processing**

limonte:
@tomholub the PR is ready for review 🚀 

@seisvelas please evaluate the security of the `extractInlineImagesToAttachments()` method

Thanks!

test/source/tests/compose.ts
```diff
@@ -1042,6 +1047,12 @@ const sendImgAndVerifyPresentInSentMsg = async (t: AvaContext, browser: BrowserH
   await ComposePageRecipe.sendAndClose(composePage);
   // get sent msg id from mock
   const sentMsg = new GoogleData('flowcrypt.compatibility@gmail.com').getMessageBySubject(subject)!;
+  if (sendingType === 'plain') {
+    expect(sentMsg.payload?.body?.data).to.match(/<img src="cid:(.+)@flowcrypt">This is an automated puppeteer test: Test Sending Plain Message With Image/);
+    return;
+    // todo - this test case is a stop-gap. We need to implement rendering of such messages below,
+    //   then let test plain messages with images in them (referenced by cid) just like other types of messages below
```
tomholub, line 1054: I'll make an issue for this

tomholub (approved):
Looks good! It seems that FlowCrypt itself cannot render such messages yet? We may want to look into that sometime, when using Settings -> Inbox.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -138,12 +141,57 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
         a.type = 'application/octet-stream'; // so that Enigmail+Thunderbird does not attempt to display without decrypting
       }
     }
+    if (choices.richtext && !choices.encrypt && !choices.sign && msg.body['text/html']) {
+      // extract inline images of plain rich-text messages (#3256)
+      // todo - also apply to rich text signed-only messages
```
tomholub, line 146: Also todo here for later

Issue #3105, opened 2020-10-29
don't require user to click to verify signature on first pubkey load


tomholub:
The first time a key is fetched, it will show this instead of re-rendering itself. Actually, it's mostly out of laziness - we could improve this and have it render the signature immediately if it correctly fetches a matching key. I'll open a separate issue for this.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/3076#issuecomment-719009689_

This only needs a code that goes through signature verification again instead of rendering the "click me" message. Probably many tests will need to be updated.

PR #3258, opened 2020-12-17
Added Primary User ID to VerifyRes

### LINKED ISSUE #2147: improve signed message rendering consistency
> Last question -- when receiving *signed* emails it shows "Person's Name - Matching Signature" in a stylized font in the upper-right.  We received a request from the COO here asking why for some people it shows their first & last name, where for other it shows just their email address instead, and whether they can edit / change how theirs is displayed there.  Again I can provide screenshots it's not immediately obvious.

-------------

Notes
1. rework to show consistent results based on primary email of the key that signed
2. if this is different from the email that the contact is saved as, highlight the difference (show in orange)
3. if this is different from the email sender, highlight the difference (show in orange)

rrrooommmaaa:
Closes #2147


rrrooommmaaa:
@tomholub Please have a look. We also have to keep in mind that uid might be absent

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
```diff
@@ -18,11 +19,12 @@ export class PgpBlockViewSignatureModule {
 
   public renderPgpSignatureCheckResult = (signature: VerifyRes | undefined) => {
     if (signature) {
-      const signerEmail = signature.contact ? signature.contact.name || this.view.senderEmail : this.view.senderEmail;
+      const signerEmail = signature.signer?.primaryUserId ? Str.parseEmail(signature.signer.primaryUserId).email || this.view.senderEmail : this.view.senderEmail;
```
tomholub, line 22: Should we remove the `this.view.senderEmail` here? And just stick to the signer, since sender can be spoofed.

tomholub: Then if it's not defined, use `(unknown signer)`

test/source/test.ts
```diff
@@ -55,6 +57,9 @@ ava.before('set config and mock api', async t => {
     const mockApi = await mock(line => mockApiLogs.push(line));
     closeMockApi = mockApi.close;
   }
+  const pubkey = pubkey2864E326A5BE488A;
+  const contact = await ContactStore.obj({ email: 'sams50sams50sept@gmail.com', name: 'Mr Sam', pubkey, client: 'pgp' });
+  await ContactStore.save(undefined, contact);
```
tomholub, line 62: Is it possible to put these lines somewhere closer to where they are being used?

tomholub (changes requested):
Looks good, with one potential change and one comment

tomholub (approved):

rrrooommmaaa:
@tomholub Looks like the "Unknown Signer" update affected "Fetched pubkey..." and "Missing pubkey..." messages.
Do we want to have "Unknown Signer" there?

rrrooommmaaa:
And SHA1 test also returns "Unknown Signer" because the signature isn't verified.

tomholub:
> @tomholub Looks like the "Unknown Signer" update affected "Fetched pubkey..." and "Missing pubkey..." messages.
Do we want to have "Unknown Signer" there?

If there is a missing pubkey, it should show clearly that it's missing pubkey as before (in the small text below). Since we only rely on the key uid.. and we don't have the key.. then indeed it should say `Unknown Signer` above the small text.

> "Fetched pubkey..." 

This is something we should fix - there's another issue for it https://github.com/FlowCrypt/flowcrypt-browser/issues/3105
If you think it's easier to fix in this PR, then you can do it, else you can do it in separate PR.

tomholub:
> And SHA1 test also returns "Unknown Signer" because the signature isn't verified.

That's probably ok. SHA1 is deprecated

rrrooommmaaa:
"Fetched pubkey..." and "Missing pubkey..." messages remain unchanged. It's the upper part changed (from sender e-mail to "Unknown Signer"). So I think we're good to merge it.

tomholub (approved):

PR #3279, opened 2020-12-29
[WIP] Remove storing drafts metadata into local storage

### LINKED ISSUE #2151: re-working drafts without local storage + drafts don't work from other device
#920

limonte:
Closes #2151 

~~Drafts seem to work fine after removing this code, tests are passing + tested manually.~~

Reply drafts aren't working.

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -76,7 +76,6 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
       } else if (this.view.isReplyBox && ApiErr.isNotFound(e)) {
         console.info('about to reload reply_message automatically: get draft 404', this.view.acctEmail);
         await Ui.time.sleep(500);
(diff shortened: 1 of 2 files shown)


limonte:
There's still code to remove.

limonte:
Even though tests are passing, this actually breaks reply drafts. I'll keep this draft opened so I can write the test case which proves that the removed code is actually needed. Then I'll revert the removed code.

tomholub:
> Even though tests are passing, this actually breaks reply drafts. I'll keep this draft opened so I can write the test case which proves that the removed code is actually needed. Then I'll revert the removed code.

That's a good approach.

But after writing the test, instead of reverting removed code, can you see if there is any way we can actually do without the local storage for drafts? Maybe by getting the necessary information from Google API instead?

limonte:
> But after writing the test, instead of reverting removed code, can you see if there is any way we can actually do without the local storage for drafts? Maybe by getting the necessary information from Google API instead?

Yes, I'll try that in another PR which will actually resolve #2151

tomholub:
Got it :+1:

PR #3283, opened 2021-01-01
Allow import unprotected key

### LINKED ISSUE #2937: add private key -> unprotected -> choose pp -> "cannot add unprotected"
Importing unprotected key in additional settings -> add key, is supposed to ask user to create a pass phrase for the key before importing. Which it does, but then it still complains the key is plain (it's supposed to encrypt it with the newly chosen pass phrase, and not complain).

https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgxwJXfsJJkWLKgFvCvPVrksLJtDs

image

Looks like this interface was designed for initial key import, where it works and is tested (to my knowledge - to verify), but it does not seem to work the same when using "add key" in settings.

rrrooommmaaa:
Closes #2937

1. Is adding a new private key considered a flaky operation?
2. No need to armor/parse the private key, right?


tomholub:
> Is adding a new private key considered a flaky operation?

Generally nothing is considered flaky until it starts displaying flaky behavior. Then we try to fix it. If we're not able to fix it, it may just get moved to the flaky group so it doesn't mess up our tests.

> No need to armor/parse the private key, right?

The key must definitely go through some parsing first, we need to understand it. Then we have to encrypt it with whatever pass phrase the user chooses, and then we have to armor it back after that.

rrrooommmaaa:
> Generally nothing is considered flaky until it starts displaying flaky behavior. Then we try to fix it. If we're not able to fix it, it may just get moved to the flaky group so it doesn't mess up our tests.

As a matter of fact, the first test run on https://github.com/FlowCrypt/flowcrypt-browser/pull/3283/commits/5c9241b0292bfb71c72a2de4da70e882c65682e6 resulted in some timeouts and overall failure, even though the new test was added to `flaky.ts`. Let me move it to `settings.ts` and see the difference.
 
> > No need to armor/parse the private key, right?
> 
> The key must definitely go through some parsing first, we need to understand it (we're already doing that). Then we have to encrypt it with whatever pass phrase the user chooses (I think we somehow missed this step earlier), and then we have to armor it back after that, for saving.

I understand this. I was asking about the small optimizations, I did: previously some calls to `KeyStore.add` were done with armoring the key as this method used to accept `string`, and I changed the signature to accept `Key | string` to avoid unnecessary back and forth conversions.

extension/js/common/platform/store/key-store.ts
```diff
@@ -43,10 +43,10 @@ export class KeyStore extends AbstractStore {
     return withPp;
   }
 
-  public static add = async (acctEmail: string, newKeyArmored: string) => {
+  public static add = async (acctEmail: string, newKey: string | Key) => {
```
tomholub, line 46: Now I see what you meant. Yup, no problem here. Best way to highlight this kind of question is to do a self-review on the PR code, where you add a comment on a particular line, then you click `review changes` on top and use `Comment` type of review. Then I'll know exactly what you are referring to.

If I read the code properly before reading your comment, I would have definitely put the two together, so it's also my own fault. But commenting on a particular line would have made it easier for me to understand the question.

extension/chrome/settings/modules/add_key.ts
```diff
@@ -75,7 +75,7 @@ View.run(class AddKeyView extends View {
     try {
       const checked = await this.keyImportUi.checkPrv(this.acctEmail, String($('.input_private_key').val()), String($('.input_passphrase').val()));
       if (checked) {
-        await KeyStore.add(this.acctEmail, checked.normalized); // resulting new_key checked above
+        await KeyStore.add(this.acctEmail, checked.encrypted); // resulting new_key checked above
```
tomholub, line 78: Ah! Thanks.

tomholub (approved):

PR #3284, opened 2021-01-02
ContactStore.search() now returns Promise<ContactPreview[]> #3282

### LINKED ISSUE #3282: Make ContactStore.search() return ContactPreview instead of Contact
This is needed to skip parsing the Key and thus save time.
As per discussion in #3247

rrrooommmaaa:
Closes #3282

### DIFF
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -109,8 +109,7 @@ View.run(class ContactsView extends View {
   }
 
   private actionExportAllKeysHandler = async () => {
-    const allArmoredPublicKeys = (await ContactStore.searchPubkeys(undefined, { has_pgp: true, limit: 500, substring: String($('.input-search-contacts').val()) })).
-      map(a => a!.trim()).join('\n');
+    const allArmoredPublicKeys = (await ContactStore.searchPubkeys(undefined, { has_pgp: true })).map(a => a!.trim()).join('\n');
     const exportFile = new Attachment({ name: 'public-keys-export.asc', type: 'application/pgp-keys', data: Buf.fromUtfStr(allArmoredPublicKeys) });
     Browser.saveToDownloads(exportFile);
   }
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -108,8 +108,8 @@ View.run(class ContactsView extends View {
     }
   }
 
-  private actionExportAllKeysHandler = () => {
-    const allArmoredPublicKeys = this.contacts.map(c => c.armoredPubkey).filter(Boolean).map(a => a!.trim()).join('\n');
+  private actionExportAllKeysHandler = async () => {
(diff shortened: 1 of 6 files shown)


extension/chrome/settings/modules/contacts.ts
```diff
@@ -109,7 +109,7 @@ View.run(class ContactsView extends View {
   }
 
   private actionExportAllKeysHandler = () => {
-    const allArmoredPublicKeys = this.contacts.map(c => c.pubkey).filter(Boolean).map((c: Key) => (KeyUtil.armor(c)).trim()).join('\n');
+    const allArmoredPublicKeys = this.contacts.map(c => c.armoredPubkey).filter(Boolean).map(a => a!.trim()).join('\n');
```
tomholub, line 112: Depending on performance, I would consider to not even pull armored key out of the database when searching contacts this way. And then I want to view details about a contact like here, I would make another call to the database with `ContactStore.get(...)`.

That way, when I'm pulling many records from database - searching - I'm only just getting metadata, which should always be fast.

If it's already fast now, then that may be an unnecessary optimization, in which case we can leave it as is.

tomholub (commented):
Code looks good. I like this.

extension/chrome/settings/modules/contacts.ts
rrrooommmaaa: I agree. I took a closer look and found out that pubkeys are only needed after the user presses "export all". But since our records don't have any ids, I introduced `ContactStore.searchPubkeys` method. What do you think?

extension/chrome/settings/modules/contacts.ts
```diff
@@ -108,8 +108,8 @@ View.run(class ContactsView extends View {
     }
   }
 
-  private actionExportAllKeysHandler = () => {
-    const allArmoredPublicKeys = this.contacts.map(c => c.armoredPubkey).filter(Boolean).map(a => a!.trim()).join('\n');
+  private actionExportAllKeysHandler = async () => {
+    const allArmoredPublicKeys = (await ContactStore.searchPubkeys(undefined, { has_pgp: true, limit: 500, substring: String($('.input-search-contacts').val()) })).map(a => a!.trim()).join('\n');
```
tomholub: Wow - this was a bug before, and it's still a bug now. I'm glad you highlighted this!

This functionality is supposed to export every single public key that's in the database, not the currently displayed ones. And there should be no limit.

extension/chrome/settings/modules/contacts.ts
tomholub: It's an improvement - at least there are two different methods now, which we can optimize easier later on.

extension/chrome/settings/modules/contacts.ts
rrrooommmaaa: Done. Should I create a test for it?

tomholub (approved):

extension/chrome/settings/modules/contacts.ts
tomholub: It's a bit harder to test, you'd have to import > 500 keys to see the difference. Thank you for asking - let's skip test this time.

PR #3286, opened 2021-01-03
Call reject on exception in search.onsuccess

### LINKED ISSUE #2759: database errors are not properly passed around
Report from user:

> Whenever I try to write an email, it keeps loading after I put the email adress. The button to send is not enabled. What is going on?

They were experiencing an exception that was happening probably in backend page:

> Error: BrowserMsg() sendRawResponse::NotFoundError: IDBDatabase.transaction: 'contacts' is not a known object store name

This should have resulted in a thrown exception on the compose page, which should result in an error modal. The exception got lost somewhere on the way, resulting in "endless loading".

rrrooommmaaa:
Closes #2759 

Raising an exception inside 
```
       search.onsuccess = Catch.try(() => {
...
})
```
block resulted in unresolved Promise, so `const raw: unknown[] = await new Promise(...` hang.

### DIFF
--- a/extension/js/common/platform/store/contact-store.ts
+++ b/extension/js/common/platform/store/contact-store.ts
@@ -158,7 +158,7 @@ export class ContactStore extends AbstractStore {
       const tx = db.transaction('contacts', 'readwrite');
       const contactsTable = tx.objectStore('contacts');
       contactsTable.put(prepared);
-      tx.oncomplete = Catch.try(resolve);
+      tx.oncomplete = () => resolve();
       tx.onabort = () => reject(ContactStore.errCategorize(tx.error));
     });
   }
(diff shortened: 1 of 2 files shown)


rrrooommmaaa:
or perhaps, we shouldn't pass `reject` to `reportErr`, but make `Catch.try` a bit more complicated.

extension/js/common/platform/store/contact-store.ts
```diff
@@ -297,7 +297,7 @@ export class ContactStore extends AbstractStore {
             cursor.continue();
           }
         }
-      });
+      }, reject);
```
tomholub, line 300: I think better just stop using Catch.try in this situation. Instead put our own try/catch block inside the callback method here. Something like

```ts
search.onsuccess = () => {
  try {
    ...
  } catch(e) {
    reject(e)
  }
}
```

tomholub: Then Catch.try can stay as is.

Good find btw.

tomholub (changes requested):
This looks good - assuming you have tested it to properly pass the error back to the caller.

Can you please do the same with all of the `search.onsuccess = ` in other methods? It seems the existing `Catch.try` will be swallowing up the errors (and reporting them) leaving the caller hanging.

tomholub (approved):

PR #3293, opened 2021-01-05
Focus #input_text for reply drafts

### LINKED ISSUE #3290: use font-weight normal for reply buttons
I find these slightly too loud:

image

and these more pleasant

image

Else they look great.

### LINKED ISSUE #3291: If the reply draft is present, #input_text should be focused
When opening a thread with the reply draft

### Actual behavior

The Recepients field is focused:

chrome_WbYhp5JpRK

### Expected behavior

The `#input_text` should be focused.

limonte:
Fixes #3291 
Fixes #3290 
Closes #3279 (proves that it was breaking reply drafts, the test from this PR is failing there)


tomholub (approved):
Thanks!

Issue #3297, opened 2021-01-05
Put `multipart/encrypted` or `multipart/signed` content inside `multipart/alternative` or `multipart/mixed`


rrrooommmaaa:
As part of research for #3082 
Though it may not help solve this particular issue, it may be useful in future

tomholub:
I'll close this issue as the draft problem was solved differently

PR #3298, opened 2021-01-05
[concept] wrap multipart/encrypted into multipart/alternative

rrrooommmaaa:
#3297


extension/js/common/core/mime.ts
```diff
@@ -197,26 +197,40 @@ export class Mime {
 
   public static encode = async (body: SendableMsgBody, headers: RichHeaders, attachments: Attachment[] = [], type?: MimeEncodeType): Promise<string> => {
```
tomholub, line 198: I'm hesitant to merge this type of change if it ends up being used for regular pgp/mime messages. If we only used it for drafts, that would be ok, but I'd prefer not to alter with outgoing messages unless we have to.

Is there some benefit to wrapping outgoing messages this way that I overlooked?

rrrooommmaaa: No, there is no immediate benefit. This was done as part of experiment, so I wanted to save this code.

tomholub: Got it, I'll leave this open

tomholub:
Doing a PR cleanup - will close this. Thanks!

PR #3304, opened 2021-01-08
Added test for updating non-first private key

### LINKED ISSUE #3109: create a test for updating non-first prv by fingerprint
The quite below is from a code review where the code was broken but tests didn't break, meaning this is a scenario we were not testing, and should be testing.

> This uncovers a problem. The page should be able to update any key. As written, this will only be able to update the first key regardless of which key we want to update.
> 
> It appears we don't have a test for this, else the test should fail - we need to add a test for updating keys other then the first one. The test will be hard for you to write as it requires familiarity with OpenPGP, so we'll make another issue for that.
> 
> Before we add that test, you should make `fingerprint` a required parameter, which should fix the actual issue.
> 
> ```ts
> this.fingerprint = Assert.urlParamRequire.string(uncheckedUrlParams, 'fingerprint');
> ```
> Then you grab key based on the fingerprint, not just the first one.
> 
> Further, whichever places call this must be updated to always add the fingerprint to the url.
> 
> _Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/3086#discussion_r515275516_

rrrooommmaaa:
Closes #3109


test/source/tests/settings.ts
```diff
@@ -243,6 +245,37 @@ export let defineSettingsTests = (testVariant: TestVariant, testWithBrowser: Tes
...
+
+      await SettingsPageRecipe.addKeyTest(t, browser, acctEmail, testKeyMultiple98acfa1eadab5b92, '1234');
+
+      const settingsPage = await browser.newPage(t, TestUrls.extensionSettings(acctEmail));
+      await SettingsPageRecipe.toggleScreen(settingsPage, 'additional');
+      // open key at index 1
+      const myKeyFrame = await SettingsPageRecipe.awaitNewPageFrame(settingsPage, `@action-show-key-1`, ['my_key.htm', 'placement=settings']);
+      await Util.sleep(1);
+      await myKeyFrame.waitAll('@content-fingerprint');
+      await myKeyFrame.waitAndClick('#action_update_prv');
```
tomholub, line 270: Please give it a `@action-update-prv` (which is `data-test="action-update-prv"`) to make it clear that this selector is used for tests. I like to keep them separate.

test/source/tests/settings.ts
```diff
@@ -243,6 +245,37 @@ export let defineSettingsTests = (testVariant: TestVariant, testWithBrowser: Tes
       await SettingsPageRecipe.addKeyTest(t, browser, 'ci.tests.gmail@flowcrypt.dev', unprotectedPrvKey, 'this is a new passphrase to protect previously unprotected key');
     }));
 
+    ava.default('settings - my key page - update non-first private key', testWithBrowser(undefined, async (t, browser) => {
+      const acctEmail = 'flowcrypt.test.key.multiple@gmail.com';
+      const settingsPage1 = await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);
+      await SetupPageRecipe.manualEnter(settingsPage1, 'unused', {
```
tomholub, line 250: Eventually we can refactor this `unused` away in this issue: https://github.com/FlowCrypt/flowcrypt-browser/issues/3246

tomholub (commented):
Code looks good with a small update

tomholub (approved):

PR #3308, opened 2021-01-09
Added reauth tests on uuid change

### LINKED ISSUE #2703: Write a backend re-auth test (eg when user revokes session)
Write a backend re-auth test as follows:

1) set up acct
2) Additional settings -> experimental -> Set a new randomly generated FlowCrypt Account UUID
3) open settings, notice modal prompt to re-auth
4) re-auth
5) no more modal in settings

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/2702#issuecomment-605492242_

---------------------------

I'm not entirely sure if our mock backend actually looks at the uuids or ignores them. If it's ignoring them (more likely), we'd have to add functionality where the backend actually saves and remembers uuids.

rrrooommmaaa:
Closes #2703

### DIFF
--- a/extension/chrome/settings/modules/experimental.htm
+++ b/extension/chrome/settings/modules/experimental.htm
@@ -36,7 +36,7 @@
     <a href="#" class="action_throw_obj bad">Throw Obj</a>
   </div>
   <div class="line"><a href="#" class="action_reset_fc_auth bad">Reset FlowCrypt Account Auth (if any)</a></div>
-  <div class="line"><a href="#" class="action_regenerate_uuid bad">Set a new randomly generated FlowCrypt Account UUID</a></div>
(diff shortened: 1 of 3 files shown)


tomholub (dismissed):
Thanks!

tomholub:
Actually I'd be ok to skip the Gmail live test on this one, mock should be enough. Unless you ran into a scenario that you think should be tested on Gmail too.

rrrooommmaaa:
> Actually I'd be ok to skip the Gmail live test on this one, mock should be enough. Unless you ran into a scenario that you think should be tested on Gmail too.

It ran fine on my computer, so I don't know of such scenario. Perhaps, there should be a sleep timeout adjust. Should I comment it out or remove it completely?

tomholub (approved):
I'll merge it, and we'll probably eventually remove it

PR #3309, opened 2021-01-09
do not attempt to upload to attester when orgrules forbid it

### LINKED ISSUE #3306: do not attempt to submit to attester if disabled with orgrules
To reproduce:
1) create a test using a mock where the OrgRules contain `NO_ATTESTER_SUBMIT` rule.
2) go through setup

**Expected**: attester submission is not attempted, no modal shows

**Actual**: shows error "Not submitting public key to Attester - disable for your org" which means it actually attempted to submit the public key, but failed on a last-resort check somewhere deeper in submission code.

image

rrrooommmaaa:
Closes #3306

### DIFF
--- a/test/source/tests/setup.ts
+++ b/test/source/tests/setup.ts
@@ -397,6 +397,14 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await attesterFrame.waitAndRespondToModal('error', 'confirm', 'Disallowed by your organisation rules');
     }));
 
+    ava.default('user@no-submit-org-rule.flowcrypt.com - do not submit to attester on key generation', testWithBrowser(undefined, async (t, browser) => {
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'user@no-submit-org-rule.flowcrypt.com');
+      await SetupPageRecipe.autoKeygen(settingsPage);
+      await Util.sleep(10);
+      await settingsPage.notPresent('.swal2-container');
+      await settingsPage.close();
+    }));
+
     ava.default('user@no-search-domains-org-rule.flowcrypt.com - do not search attester for recipients on particular domains', testWithBrowser(undefined, async (t, browser) => {
       // disallowed searching attester for pubkeys on "flowcrypt.com" domain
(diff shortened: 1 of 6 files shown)


test/source/tests/setup.ts
```diff
@@ -397,6 +397,14 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await attesterFrame.waitAndRespondToModal('error', 'confirm', 'Disallowed by your organisation rules');
     }));
 
+    ava.default('user@no-submit-org-rule.flowcrypt.com - do not submit to attester on key generation', testWithBrowser(undefined, async (t, browser) => {
```
tomholub, line 400: The test somehow didn't manage to reproduce the problematic behavior. Could you please try a scenario when the private key is already present on EKM + there is a no-submit rule?

test/source/tests/setup.ts
```diff
@@ -397,6 +397,14 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'user@no-submit-org-rule.flowcrypt.com');
+      await SetupPageRecipe.autoKeygen(settingsPage);
+      await Util.sleep(10);
+      await settingsPage.notPresent('.swal2-container');
```
tomholub, line 404: Instead of waiting a fixed amount and then testing for non-presense, you can wait for the next expected step directly (which is a screen that shows that setup is done). That should be faster and more reliable.

test/source/tests/setup.ts
tomholub: Sorry - maybe I have you wrong feedback! It seems you were actually able to reproduce the problem?

If so, then you could remove lines 403 and 404, because the line above is already testing for the successful setup (the modal pupped up before the setup was done).

rrrooommmaaa: No, it didn't reproduce in this setting. I just had to figure out how to run a key generation test

rrrooommmaaa: The problem was that `SetupKeyManagerAutogenModule` set the option `submit_main: true`

tomholub (approved):
Excellent! Thank you - this was exactly it.

PR #3310, opened 2021-01-11
Refactor drafts to work without local storage

### LINKED ISSUE #2151: re-working drafts without local storage + drafts don't work from other device

### LINKED ISSUE #2223: The saved draft isn't visible in the conversations list
Steps to reproduce:

1. Open any encrypted conversation in the Inbox
2. Click Encrypted Reply
3. Write down any text and wait for the "Saved" notification in the footer
4. Go back to Inbox, there should be the saved draft:

image

### Expected result

image

### Actual result

image

limonte:
This PR makes sure that drafts are not using local storage so users can open/change/send their drafts from different devices/browsers.

Closes #2223

Closes #2151

I had to leave `AcctStoreDict.drafts_reply` in the codebase in order to keep the existed (legacy) reply drafts working.


extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -140,7 +140,7 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
           }
         } else {
           await this.doUploadDraftWithLocalStorageFallback(mimeMsg, msgData, async () => {
-            await this.view.emailProvider.draftUpdate(this.view.draftId, mimeMsg);
+            await this.view.emailProvider.draftUpdate(this.view.draftId, mimeMsg, this.view.threadId);
```
limonte, line 143: Sometimes you smash your head the whole day just because someone at Google forgot to document [here](https://developers.google.com/gmail/api/reference/rest/v1/users.drafts/update) that `threadId` should be passed to  `.draftUpdate()`. Otherwise, it'll be changed after updating the draft (I still have no idea why).

tomholub (commented):
Nice! How did you discover this?

limonte:
> Nice! How did you discover this?

By long debugging and not understanding why our drafts aren't threaded. First, I was thinking that the reason is in missing `References` and `In-Reply-To` headers, but that was a wrong assumption. Then I checked the thread right after the draft creation and the thread was ok, containing the secured reply. But then I remembered that we're updating a draft right after creating it, and something is probably wrong there. That was the case indeed because the thread wasn't containing the secured reply right after the draft. Luckily, trying the undocumented `threadId` helped.

tomholub:
Excellent work, if this ends up working, I'll consider it magical.

limonte:
@tomholub there are some visual details to fix

- remove Gmail's border around reply drafts
- remove unexpected vertical jumping when loading reply drafts

but other than that, the PR is mostly ready for review. I'll handle those details after catching some zzz.

extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -206,18 +202,8 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
...
-    if (threadId) { // it's a reply
-      const drafts = draftStorage.drafts_reply || {};
-      drafts[threadId] = draftId;
-      await AcctStore.set(this.view.acctEmail, { drafts_reply: drafts });
-    } else { // it's a new message
-      const drafts = draftStorage.drafts_compose || {};
-      drafts[draftId] = { recipients, subject, date: new Date().getTime() };
-      await AcctStore.set(this.view.acctEmail, { drafts_compose: drafts });
-    }
     await storageLocalSet({ [draftId]: { message: { raw: Buf.fromUtfStr(mimeMsg).toBase64UrlStr(), threadId } } });
```
tomholub, line 207: I would like to review this in another issue and another PR - I'd rather have a more structured place where to save drafts.

For example a "localDrafts" array or object where all the drafts would be.

extension/chrome/elements/compose.ts
```diff
@@ -171,8 +171,8 @@ export class ComposeView extends View {
       await this.renderModule.fetchReplyMeta(Object.keys(storage.sendAs!));
     }
     if (this.isReplyBox) { // reply
-      if (this.threadId && !this.ignoreDraft && storage.drafts_reply && storage.drafts_reply[this.threadId]) {
-        this.draftId = storage.drafts_reply[this.threadId]; // there may be a draft we want to load
+      if (this.threadId && !this.draftId && !this.ignoreDraft && storage.drafts_reply && storage.drafts_reply[this.threadId]) {
+        this.draftId = storage.drafts_reply[this.threadId]; // there may be a legacy draft we want to load
```
tomholub, line 175: Interesting - this means we should make another issue to delete this line, and potentially others, maybe 6 months later.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -114,7 +114,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   private everything = () => {
     this.replaceArmoredBlocks();
     this.replaceAtts().catch(Catch.reportErr);
-    this.replaceFcTags();
+    this.replaceComposeDraftLinks();
```
tomholub, line 117: Thank you - this name is so much better.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -529,17 +518,13 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     const newReplyBoxes = $('div.nr.tMHS5d, td.amr > div.nr, div.gA td.I5').not('.reply_message_evaluated').filter(':visible').get();
     if (newReplyBoxes.length) {
       // cache for subseqent loop runs
-      const { drafts_reply } = await AcctStore.get(this.acctEmail, ['drafts_reply']);
       const convoRootEl = this.getGonvoRootEl(newReplyBoxes[0]);
       const replyParams = this.getLastMsgReplyParams(convoRootEl!);
-      const threadId = this.determineThreadId(convoRootEl!);
```
tomholub, line 535: I'm assuming the `threadId` is now retrieved from the API based on `replyParams.replyMsgId`?

Need to do throrough testing later, by sending emails between two accounts, to make sure they get correctly threaded on both ends.

tomholub (commented):
Looks good. This will need a bunch of testing (since it affected some non-draft code too). I may also do some testing before I merge this, presumably you already have.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
limonte: `threadId` isn't needed here anymore because the reply would be loaded automatically. Secure replies are threaded now after https://github.com/FlowCrypt/flowcrypt-browser/pull/3310/commits/45745be54b0585db215f4a7fff1b91a882ef368b

limonte:
> Looks good. This will need a bunch of testing (since it affected some non-draft code too). I may also do some testing before I merge this, presumably you already have.

Let me just manually check that legacy drafts aren't broken, I assume we can skip testing that (correct me if I'm wrong). After that I'll mark the PR as "ready for review".

limonte:
> > Looks good. This will need a bunch of testing (since it affected some non-draft code too). I may also do some testing before I merge this, presumably you already have.
> 
> Let me just manually check that legacy drafts from local storage aren't broken, I assume we can skip testing that (correct me if I'm wrong). After that I'll mark the PR as "ready for review".

They work and they even are updated and available from all devices/browsers once opened with the updates from this PR 😎

limonte:
Found a small bug, revering to draft.

extension/chrome/elements/compose.ts
limonte: Done: https://github.com/FlowCrypt/flowcrypt-browser/issues/3329

limonte:
Ready for review 🚀

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -99,22 +99,10 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     $('.reply_message_iframe_container:visible').last().append(this.factory.embeddedReply(params, false, true)); // xss-safe-value
   }
 
-  public scrollToElement = (selector: string) => {
```
limonte, line 102: This removal isn't really related to the PR, but I decided to include it here because

1. it was only used to scroll to the secured reply box
2. the secured reply box is autofocused and therefore browsers will scroll to is natively
3. when the secured reply box was higher than the viewport this method was scrolling to the bottom and the cursor wasn't visible

limonte:
While working on #3327 I discovered two issues here:

- [ ] the secured reply draft isn't rendered if the last reply is plain
- [ ] when sending the reply drafts the subject starts with "Fwd: " instead of "Re: " (this should be covered with a test)

tomholub:
> While working on #3327 I discovered two issues here:
> 
>     * [ ]  the secured reply draft isn't rendered if the last reply is plain
> 
>     * [ ]  when sending the reply drafts the subject starts with "Fwd: " instead of "Re: " (this should be covered with a test)

Are these issues already present on master? If yes, we can merge this and address it in followup PR. If not, then it will be indeed better to fix them before merging.

Thanks for noticing these!

extension/js/content_scripts/webmail/gmail-element-replacer.ts
tomholub: I originally added this because at least one browser did not scroll down at all. You'd click on reply button and if the reply box was below screen, you would not see anything happen. It would seem like it's broken.

I did notice that the scrolling is not perfect at times, let's try removing it as you did, and hopefully the old issues don't resurface.

limonte:
> Are these issues already present on master? If yes, we can merge this and address it in followup PR. If not, then it will be indeed better to fix them before merging.

Both issues are introduced in this PR and not present in master, so I'll address them here.

limonte:
> While working on #3327 I discovered two issues here:
> 
> * [x]  the secured reply draft isn't rendered if the last reply is plain
> * [x]  when sending the reply drafts the subject starts with "Fwd: " instead of "Re: " (this should be covered with a test)

Fixed.

tomholub (approved):
Looks good - I'll do some manual testing. Thanks!

tomholub:
I've tested this & unable to open/view any reply drafts. I'm sending a video to your email (it has some unrelated messages showing in it).

tomholub (changes requested):
Could not make it work, per comment above

limonte:
> I've tested this & unable to open/view any reply drafts. I'm sending a video to your email (it has some unrelated messages showing in it).

Thanks for the feedback! Just to clarify: did you reload the extension in `chrome://extensions/` after rebuilding it?

limonte:
> > I've tested this & unable to open/view any reply drafts. I'm sending a video to your email (it has some unrelated messages showing in it).
> 
> Thanks for the feedback! Just to clarify: did you reload the extension in `chrome://extensions/` after rebuilding it? It's needed because `gmail-element-replacer.ts` has been changed.

limonte:
> I'm pretty sure I did - it's my standard workflow. But I could have messed up. Are you able to reproduce what I showed in the video, or does it work on your end?
> 
> If it does work differently then the video for you, I should probably try again.

Thanks for clarifying, something is wrong here indeed. I am able to reproduce the behavior from your video on the `flowcrypt.compatibility@gmail.com` account, but not on my personal account. Let me see what's the difference between them.

tomholub:
flowcrypt compatibility account has thread / conversation view disabled - have a look at that

limonte:
> flowcrypt compatibility account has thread / conversation view disabled - have a look at that

That should be it, thanks!

image

I'll fix that!

tomholub:
This kind of sounds like we should have another account for live tests that is configured with this disabled.. not in this PR, but we'd be likely to break it sooner or later again.

limonte:
> This kind of sounds like we should have another account for live tests that is configured with this disabled.. not in this PR, but we'd be likely to break it sooner or later again. What is not tested, almost as if it didn't exist :)

Agree. If it's not tested, it's broken (c)

limonte:
> > Nice! How did you discover this?
> 
> By long debugging and not understanding why our drafts aren't threaded. First, I was thinking that the reason is in missing `References` and `In-Reply-To` headers, but that was a wrong assumption. 

These headers are needed indeed when the conversation view is off. From [gmail docs]:(https://developers.google.com/gmail/api/guides/threads#adding_drafts_and_messages_to_threads):

1. The requested `threadId` must be specified on the `Message` or `Draft.Message` you supply with your request.
2. The `References` and `In-Reply-To` headers must be set in compliance with the RFC 2822 standard.
2. The `Subject` headers must match.

tomholub (approved):
Code looks good, pending manual testing

tomholub:
Sorry to let this excellent PR wait, I hope I'll get to merge it early next week.

tomholub (approved):

Issue #3315, opened 2021-01-13
accumulating drafts cause live gmail tests to fail


tomholub:
Example failure: https://flowcrypt.semaphoreci.com/jobs/3a01ea85-c237-4140-b189-34108f5d5b08

The solution could be to call an API that deletes all drafts on Gmail before start of the test. However, the test code that runs on nodejs does not have the right permission, only the browser extension does (after the login step). There are three options how to solve it that come to mind:

a) somehow extract the necessary access token from the browser extension after login step, and then make a call from NodeJS to delete all drafts.
b) create a separate miniapp for tests where the nodejs app will use puppeteer to log in and authenticate. The resulting code will be scraped from the window after authentication, then nodejs requests access token manually, and then uses it to delete the drafts.
c) find out why are we leaving drafts behind in some other test, and adjust the test to always make sure to delete the draft using Gmail UI at the end of a test

![image](https://user-images.githubusercontent.com/6306961/104477957-a3c77280-55b9-11eb-999b-eaba715387a7.png)

tomholub:
This seems to be consistently affecting tests. @rrrooommmaaa please have a look later. Some test is leaving behind drafts on Gmail. It seems to be relatively recent.

rrrooommmaaa:
> c) find out why are we leaving drafts behind in some other test, and adjust the test to always make sure to delete the draft using Gmail UI at the end of a test

I assume this can happen as a result of a test crash.

tomholub:
That was my assumption too. But sometimes I feel even successful tests cause this, given the rate at which we are getting these.

If you need to log into that account to see, you can grab the credentials from test secrets.

PR #3318, opened 2021-01-14
Save rich text drafts as inline PGP messages

### LINKED ISSUE #3082: rich text messages drafts are broken
1) create a new secure message **on Firefox**
2) switch to rich text (pgp/mime) - must be on firefox, then click the "chevron" icon next to send button, and choose `rich text (pgp/mime)`
3) start composing, a draft will get saved
4) close the window
5) open the draft

It's broken. The issue is on the opening side, not recognizing the draft. It's possible the sending side will still have to be improved, too - normally we add "draft tags" that make the content script recognize it's a draft when opening.

image

rrrooommmaaa:
This PR saves rich text drafts as inline PGP messages to be able to pick it up later in gmail
I tested it in Gmail to pick it up and send (as RFC3156) -- it works good.

Should I do a (separate) test to 
1. make sure a draft is saved as inline message.
2. an inline draft like this is sent as `multipart/encrypted` (RFC3156)
?

close #3082

### DIFF
--- a/extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
+++ b/extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
@@ -83,6 +83,11 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
 
   private sendableRichTextMsg = async (newMsg: NewMsgData, pubs: PubkeyResult[], signingPrv?: Key) => {
     const plainAtts = this.isDraft ? [] : await this.view.attsModule.attach.collectAtts();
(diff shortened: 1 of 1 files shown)


tomholub (approved):
Good solution. I find that this particular issue is harder to test (best would be on live gmail to make sure content scripts can pick it up, but it's not ideal to have gmail test that actually alter data by storing drafts on google).

I'd be ok to skip testing for this particular issue.

tomholub:
If you really wanted to test it, then you'd want to store a rich text message into mock. Then the mock checks that it can be decrypted and contains the rich html in it.

tomholub:
The broken tests now are caused by #3315

tomholub (approved):
I'm good to merge this if you mark it as ready

tomholub (approved):

PR #3320, opened 2021-01-16
Clean drafts before each GMail test

### LINKED ISSUE #3315: accumulating drafts cause live gmail tests to fail

rrrooommmaaa:
This PR extracts google token and deletes all drafts on the account

close #3315

### DIFF
--- a/test/source/browser/controllable.ts
+++ b/test/source/browser/controllable.ts
@@ -624,6 +624,18 @@ export class ControllablePage extends ControllableBase {
       }
     }
   }
+
+  public getFromLocalStorage = async (keys: string[]): Promise<any> => {
+    const result = await new Promise((resolve, reject) => {
+      (this.target as Page).exposeFunction('saveRawStorageResult', (x: any) => {
+        resolve(x);
+      }).catch(reject).then(() =>
+        (this.target as Page).evaluate(keys =>
+          chrome.storage.local.get(keys, items => (window as any).saveRawStorageResult(items)), keys
+        ).catch(reject));
+    });
+    return result;
+  }
 }
 
 export class ControllableFrame extends ControllableBase {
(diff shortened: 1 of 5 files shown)


rrrooommmaaa:
> a) somehow extract the necessary access token from the browser extension after login step, and then make a call from NodeJS to delete all drafts.

@tomholub I suggest to extract google token like this and clear the drafts using it (if that's what you meant)

test/source/browser/controllable.ts
```diff
@@ -624,6 +624,18 @@ export class ControllablePage extends ControllableBase {
       }
     }
   }
+
+  public getFromLocalStorage = async (keys: string[]): Promise<any> => {
```
tomholub, line 628: Nice!! I did not know this was possible. This is an amazingly elegant solution.

rrrooommmaaa: Can we use https://www.npmjs.com/package/googleapis package to access gmail api?

tomholub: Sure, I'm pretty liberal with adding deps to the test infrastructure.

rrrooommmaaa:
This version does requests cleanup on each `TestWithBrowser` call with `!isMock`

tomholub (approved):
Looks good

Issue #3088, opened 2020-10-24
wait until view loaded before testing page with puppeteer


tomholub:
This line
```ts
    const slowMo = this.isMock ? 60 : 60; // todo - mock tests should be able to run without slowMo
```
is slowing down tests a lot. It should say
```ts
    const slowMo = this.isMock ? undefined : 60;
```
However when we do that, mock tests will start failing because they will be testing things too fast - looking for elements that are not yet there, or more likely - clicking elements where no click handlers were attached to them just yet.

We should have a general mechanism for marking a page as ready (rendered), and only then should the test try to look for elements, click them, etc.

PR #3328, opened 2021-01-18
issue 3088 speed up mock test slowMo

tomholub:
This PR attempts to wait until page is loaded before testing it - to increase reliability even as we speed up test execution.

issue #3088


tomholub (commented):
I didn't really manage to speed up the tests - even after this change (waiting for page loaded first), decreasing slowMo will increase amount of failures/retries it has to do, which doesn't really help the overall run time of the tests. I do think I may have made the tests a bit more reliable - by adding this check + leaving it at 60ms.

It did help local tests run a lot faster when it was set for 0, but I didn't want to set different settings for local vs CI. Later, we could consider doing that.

Also observed that tests marked as `flaky` were a lot more likely to fail when decreasing slowMo.

PR #3338, opened 2021-01-20
Invalidate RSA keys less than 2048 bits

### LINKED ISSUE #2715: weak RSA keys should not be usable
Public keys that have a RSA primary key smaller than 2048 bits should always have `usableForEncryption` and `usableForSigning` be false.

Similarly, when parsing a public key, any subkeys that use RSA key smaller than 2048 should not be taken into account. (not sure how to do this)

2048bit will therefore be the minimum required strength for RSA.

rrrooommmaaa:
This PR resets `usableForEncryption` and `usableForSigning` flags for RSA keys smaller than 2048 bits.

The simplest way is to delete all RSA subkeys smaller than 2048 bits  from OpenPGP.js `Key` object.
The primary key can't be deleted, otherwise the key can't be parsed and formed.
The `primaryKeySecure` flag is reset instead.
If the primary key is insecure, subkeys can't be used either.

However, we should think about how this will affect `KeyUtil.diagnose`, so that insecure subkeys simply disappear from the report.

close #2715


tomholub (commented):
That's a good solution.

In compatibility page, a lot of the work (most of it?) is done using OpenPGP.js properties and methods. As long as we do internal parsing which calls `convertExternalLibraryObjToKey` on a separate object than the one we evaluate using original library properties, we should be fine. In practice, to have the objects separate, I guess in compatibility page we would parse the armored key twice? Once only using openpgp.js and once using our methods that alter it.

rrrooommmaaa:
> In compatibility page, a lot of the work (most of it?) is done using OpenPGP.js properties and methods.

Right, but we parse the string key to our own `Key` object (with OpenPGP.js embedded as [internal]) and pass this object to `diagnose`. The idea was to clean up the invalid subkeys right on `parse` rather than do it before `sign`, `encrypt` etc.
So I propose to store both original and "secure" version of the key.
This PR stores the original into `raw` (previously string).

rrrooommmaaa:
@tomholub Changing type of `raw` from string to OpenPGP's `Key` causes complications with sending it to background (`armor()` function attached to it gets lost). I almost finished refactoring, but there is `keyParse` message? What is it for?
Perhaps, we should use two fields in the `Key` object (`rawKey` and `rawText` (or simply `raw`)?

tomholub:
> but there is keyParse message? What is it for?

It's only used here:

```ts
        const { key } = await BrowserMsg.send.bg.await.keyParse({ armored: pubkey });
        if (!key.allIds.map(id => OpenPGPKey.fingerprintToLongid(id)).includes(signerLongid)) {
          render(`Fetched sender's pubkey ${OpenPGPKey.fingerprintToLongid(key.id)} but message was signed with a different key: ${signerLongid}, will not verify.`, () => undefined);
          return;
        }
```

That's in `pgp_block.htm`. That page does not load openpgp.js itself (because there may be 50 message frames in a conversation, and loading the library 50 times takes a lot of time).

So the method takes in armored key and returns `Key` which is json-serializable, and therefore can be sent back over the browser message response mechanism.

tomholub:
> @tomholub Changing type of raw from string to OpenPGP's Key causes complications with sending it to background (armor() function attached to it gets lost).

Definitely. The browser will json-serialize the object before sending, and then deserialize when received. You can imagine literally doing `JSON.parse(JSON.stringify(object))`. When you do that on an object, only fields, arrays, etc are left. No methods will stay on it. For that reason we have to avoid sending any "clever" objects over the boundary.

> Perhaps, we should use two fields in the Key object (rawKey and rawText (or simply raw)?

We could definitely rename raw to `rawArmored` (string) for example.

Then here

```ts
    const raw = (pubkey as unknown as { [internal]: OpenPGP.key.Key })[internal];
    if (!raw) {
      throw new Error('Object has type == "openpgp" but no internal key.');
    }
    return raw;
```

We should rename `raw` to `openpgpObj` to avoid confusion.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -468,6 +484,50 @@ export class OpenPGPKey {
...
+            continue; // discard this packet as insecure
+          }
+        }
+      }
+      newPacketList.push(packet);
+    }
+    if (packets.length !== newPacketList.length) {
+      return { primaryKeySecure, secure: new opgp.key.Key(newPacketList) };
+    }
+    return { primaryKeySecure, secure: opgpKey };
```
tomholub, line 528: Let's rename the method to `removeWeakKeyPackets` and return values to `{isPrimaryKeyStrong, keyWithoutWeakPackets}`

Please add method comment/doc to link to the github issue 2715 for further reading / understanding

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -441,11 +449,19 @@ export class OpenPGPKey {
...
-    if (!raw) {
+    const opgpKey = (pubkey as unknown as { [internal]: OpenPGP.key.Key })[internal];
+    if (!opgpKey) {
       throw new Error('Object has type == "openpgp" but no internal key.');
     }
-    return raw;
+    return opgpKey;
+  }
+
+  private static extractInsecureExternalLibraryObjFromKey = (pubkey: Key) => {
```
tomholub, line 459: Instead of `Insecure` we could say `StrengthUnchecked`, and also add a link to the issue 2715

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -150,9 +153,10 @@ export class OpenPGPKey {
    *    is done on the original supplied object.
    */
   public static convertExternalLibraryObjToKey = async (opgpKey: OpenPGP.key.Key, keyToUpdate?: Key): Promise<Key> => {
+    const { primaryKeySecure, secure } = OpenPGPKey.removeInsecureKeyPackets(opgpKey);
```
tomholub, line 156: Here also let's use the renamed var names as in another comment

tomholub (commented):
A few notes to make the code easier to understand in the future

tomholub (commented):
:+1:

rrrooommmaaa:
@tomholub Should I create S/MIME certificates with RSA/1024 keys for testing?
Are we ever going to use a private key there?

tomholub:
We could merge this in, and then create a test as you propose.

We plan private key usage as well.

rrrooommmaaa:
> We plan private key usage as well.

Ok, so we should clarify what we mean by `usable for signing` -- is it about verifying?

tomholub:
> > We plan private key usage as well.
> 
> Ok, so we should clarify what we mean by `usable for signing` -- is it about verifying?

Usable for signing means - if a private key - that the private key can be used to sign a message.

If it's a public key, then it means that the private key that is related to this public key would be usable for signing (I don't think we look at this for public keys, but the OpenPGP.js library does a similar check when verifying).

rrrooommmaaa:
> 
> Usable for signing means - if a private key - that the private key can be used to sign a message.
> 
> If it's a public key, then it means that the private key that is related to this public key would be usable for signing (I don't think we look at this for public keys, but the OpenPGP.js library does a similar check when verifying).

Right, my concern was -- do we have to have this logic when determining `usableForSigning` flag for S/MIME certificate.
My current PR only provides rough check whether `PublicKey` is present and strong.

tomholub:
>  do we have to have this logic when determining

I think we don't have to do anything beyond what you did at this point. The S/MIME integration is still early in general. Later on we may improve it if there is need for it.

tomholub (approved):

PR #3350, opened 2021-01-27
Report issues with parsing inline images

### LINKED ISSUE #3301: check security of cid replacements PR
@tomholub the PR is ready for review 🚀 

@seisvelas please evaluate the security of the `extractInlineImagesToAttachments()` method

Thanks!

_Originally posted by @limonte in https://github.com/FlowCrypt/flowcrypt-browser/issues/3256#issuecomment-750590830_

limonte:
This PR reports issues with parsing inline images to track issues with unexpected `src` values of inline images (https://github.com/FlowCrypt/flowcrypt-browser/issues/3301#issuecomment-757591686)

close #3301

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
@@ -178,8 +178,8 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
         } else {
           throw new ComposerUserError(`
             Unable to parse an inline image <details>
-              <summary>with src="..."</summary>
-              ${src}
+              <summary>See error details</summary>
+              src="${src}"
             </details>
           `);
         }
--- a/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
@@ -164,15 +164,19 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
         const img: Element = node;
         const src = img.getAttribute('src') as string;
         const { mimeType, data } = this.parseInlineImageSrc(src);
-        const imgAttachment = new Attachment({
-          cid: Attachment.attachmentId(),
-          name: img.getAttribute('name') || '',
-          type: mimeType,
-          data: Buf.fromBase64Str(data),
-          inline: true
-        });
-        img.setAttribute('src', `cid:${imgAttachment.cid}`);
-        imgAttachments.push(imgAttachment);
+        if (mimeType && data) {
+          const imgAttachment = new Attachment({
+            cid: Attachment.attachmentId(),
+            name: img.getAttribute('name') || '',
+            type: mimeType,
+            data: Buf.fromBase64Str(data),
+            inline: true
+          });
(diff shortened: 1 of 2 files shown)


limonte:
The better review experience would be with `?w=1`: https://github.com/FlowCrypt/flowcrypt-browser/pull/3350/files?w=1

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -164,15 +164,19 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
+            cid: Attachment.attachmentId(),
+            name: img.getAttribute('name') || '',
+            type: mimeType,
+            data: Buf.fromBase64Str(data),
+            inline: true
+          });
+          img.setAttribute('src', `cid:${imgAttachment.cid}`);
+          imgAttachments.push(imgAttachment);
+        } else {
+          Catch.report(`Unable to parse an inline image with src="${src}"`);
```
tomholub, line 178: This reports it to us - not good - a privacy violation. You can instead do a toast to tell the user.

limonte: Right, will do that 👍

limonte: Instead of toast I decided to go with the error modal which was already implemented there. 

CleanShot 2021-01-28 at 14 53 20@2x

CleanShot 2021-01-28 at 14 54 32@2x

The current behavior is: we will NOT send a message if there were issues with parsing inline images, IMO it's better to prevent sending a message which is broken because of unexpected issues. @tomholub Let me know if you want some other behavior here, e.g. remove broken images and send a message without them.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -164,15 +165,24 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
+            data: Buf.fromBase64Str(data),
+            inline: true
+          });
+          img.setAttribute('src', `cid:${imgAttachment.cid}`);
+          imgAttachments.push(imgAttachment);
+        } else {
+          throw new ComposerUserError(`
+            Unable to parse an inline image <details>
+              <summary>See error details</summary>
+              src="${src}"
```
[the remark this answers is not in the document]
limonte: The error modal with `src` in it will be shown only to the user who is sending the message with evil tags. That's why I initially assumed that escaping isn't needed, but it's better to be safe than sorry, added `Xss.escape(src)` just in case.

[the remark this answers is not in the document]
seisvelas: Ahh sorry, I didn't realize it was only for the sender. My bad!

Still do agree that better safe than sorry, as you said. Imagine an attacker finds that a certain image that should work triggers this error for whatever reason. The attack could then embed some HTML into the image and convince the victim to attach it to an encrypted email via FlowCrypt, triggering this error, and BAM! The evil HTML is rendered on the victim.

seisvelas (approved):

tomholub (approved):
Looks good, thank you both :+1:

PR #3356, opened 2021-01-29
fix ⌘ + Enter shortcut for sending

### LINKED ISSUE #3355: ⌘ + Enter shortcut for sending doesn't work in macOS
macOS users would expect `⌘ + Enter` shortcut to work for triggering the sending. It doesn't work now because we're listening for the `keypress` event, and we should listen for the `keydown`: https://stackoverflow.com/questions/3902635/how-does-one-capture-a-macs-command-key-via-javascript#comment15734641_5500536

limonte:
This PR allows macOS users to send message with the expected `⌘ + Enter` shortcut

close #3355


limonte:
@tomholub this was a quick-fix, so I decided to submit it without prioritising. FYI, I recently switched to macOS as my primary OS, it still hurts because my fingers need to adjust to new keys and shortcuts, but as a front-end developer I have to be as close as possible to the majority of users.

tomholub (approved):

tomholub:
Thanks!

PR #3364, opened 2021-02-01
Issue 3363 ekm no fp

### LINKED ISSUE #3363: stop sending fingerprint to EKM

tomholub:
This PR no longer sends fingerprint to PUT /keys/private on EKM since it's no longer required.

close #3363


tomholub (commented):
lgtm if passes

PR #3372, opened 2021-02-02
Scroll to the top of reply box if it's higher than the viewport

### LINKED ISSUE #3367: The new reply should scroll itself into the viewport after clicking the "Secure Reply" button
The scrolling functionality was removed by me in #3310 and for saved reply drafts it's not wanted as it was scrolling to the bottom of the reply box and in case the reply box is high, the cursor wasn't visible.

But for the new compose boxes it does make sense to scroll into them:

CleanShot 2021-02-01 at 18 05 35

limonte:
This PR makes sure that the reply box is positioned properly in the viewport on initialisation

close #3367


tomholub (approved):
Thank you!

PR #3381, opened 2021-02-03
Limit the values the page url parameter can have

seisvelas:
I noticed that in the extension, the legitimate values passed to this parameter only ever start with either `/chrome` or `modules`. I 
tested it and everything worked fine.

I recognize this solution is pretty hacky and amatuerish - I'm very open to suggestions for better approaches.

Resolves https://github.com/FlowCrypt/flowcrypt-security/issues/98


seisvelas:
I forgot the branches in this repo have a different naming structure, give me one second to fix that

PR #3383, opened 2021-02-03
 Limit the values the page url parameter can have

seisvelas:
I noticed that in the extension, the legitimate values passed to this parameter only ever start with either /chrome or modules. I
tested it and everything worked fine.

I recognize this solution is pretty hacky and amatuerish - I'm very open to suggestions for better approaches.

Resolves FlowCrypt/flowcrypt-security#98

### DIFF
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -50,7 +50,7 @@ View.run(class SettingsView extends View {
     const uncheckedUrlParams = Url.parse(['acctEmail', 'page', 'pageUrlParams', 'advanced', 'addNewAcct']);
     this.acctEmail = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'acctEmail');
     this.page = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'page');
-    if (!/^\/chrome|modules/.test!(this.page as string)) {
+    if (!/^(\/chrome|modules)/.test!(this.page as string)) {
       this.page = undefined;
     }
     this.page = (this.page === 'undefined') ? undefined : this.page; // in case an "undefined" string slipped in
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -50,7 +50,8 @@ View.run(class SettingsView extends View {
     const uncheckedUrlParams = Url.parse(['acctEmail', 'page', 'pageUrlParams', 'advanced', 'addNewAcct']);
     this.acctEmail = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'acctEmail');
     this.page = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'page');
-    if (!/^(\/chrome|modules)/.test!(this.page as string)) {
+    if (this.page && !/^(\/chrome|modules)/.test!(this.page as string)) {
+      Ui.modal.error('An unexpected value was found for the page parameter');
       this.page = undefined;
     }
     this.page = (this.page === 'undefined') ? undefined : this.page; // in case an "undefined" string slipped in
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -51,7 +51,8 @@ View.run(class SettingsView extends View {
     this.acctEmail = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'acctEmail');
     this.page = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'page');
     if (this.page && !/^(\/chrome|modules)/.test!(this.page as string)) {
-      void Ui.modal.error('<div data-test="page-param-modal">An unexpected value was found for the page parameter</div>');
+      Ui.modal.error('An unexpected value was found for the page parameter')
+        .catch(err => console.log(err));
       this.page = undefined;
     }
     this.page = (this.page === 'undefined') ? undefined : this.page; // in case an "undefined" string slipped in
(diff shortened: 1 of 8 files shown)


seisvelas:
Ah, my approach is flawed, it's not really checking at the beginning, it's happy if the word modules or /chrome appears anywhere in the string. I'll fix that real quick.

extension/chrome/settings/index.ts
```diff
@@ -50,6 +50,9 @@ View.run(class SettingsView extends View {
     const uncheckedUrlParams = Url.parse(['acctEmail', 'page', 'pageUrlParams', 'advanced', 'addNewAcct']);
     this.acctEmail = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'acctEmail');
     this.page = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'page');
+    if (!/^(\/chrome|modules)/.test!(this.page as string)) {
+      this.page = undefined;
```
tomholub, line 54: Please show a modal in this case, to say that an unexpected value was found in the "page" parameter. We could even write a test for this.

seisvelas: Modal added! It looks like this:

Screenshot from 2021-02-03 01-50-56

Adding a test now, in the meantime let me know if you'd like any changes to the modal

seisvelas:
I'm a bit stuck with this test. It opens the page with the modal and then closes it right away, then says it never appeared!

Here's the actual error: 

```
  settings - error modal when page parameter invalid01:21
    ℹ (attempt 1 of 3) Retrying: Error: waiting failed: Elements did not appear: [data-test="page-param-modal"]01:21
    ℹ (attempt 2 of 3) Retrying: Error: waiting failed: Elements did not appear: [data-test="page-param-modal"]01:21
    ℹ (attempt 3 of 3) Failed:   Error: waiting failed: Elements did not appear: [data-test="page-param-modal"]01:21
          at ControllablePage.ControllableBase.waitAny (/home/semaphore/git/flowcrypt-browser/test/source/browser/controllable.ts:83:11)01:21
          at ControllablePage.ControllableBase.waitForContent (/home/semaphore/git/flowcrypt-browser/test/source/browser/controllable.ts:289:5)01:21
          at /home/semaphore/git/flowcrypt-browser/test/source/tests/settings.ts:251:701:21
          at /home/semaphore/git/flowcrypt-browser/test/source/test.ts:66:701:21
          at /home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:143:6101:21
          at BrowserPool.withNewBrowserTimeoutAndRetry (/home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:143:11)01:21
          at /home/semaphore/git/flowcrypt-browser/test/source/test.ts:62:501:21
          at <async> waitAny (/test/source/browser/controllable.ts:63:19)01:21
          at <async> waitForContent (/test/source/browser/controllable.ts:288:26)01:21
          at <async> <anonymous> (/test/source/tests/settings.ts:248:101)01:21
          at <async> <anonymous> (../test/source/test.ts:62:53)01:21
          at <async> <anonymous> (/test/source/browser/browser-pool.ts:143:49)01:21
  [ALL RETRIES FAILED for settings - error modal when page parameter invalid]01:21
  › BrowserPool.processTestError (test/source/browser/browser-pool.ts:171:9)01:21
  › BrowserPool.withNewBrowserTimeoutAndRetry (test/source/browser/browser-pool.ts:157:14)01:21
  › test/source/test.ts:62:5
```

@tomholub I'll get back at this when I wake up, but if you have suggestions in the meantime of what I may be doing wrong, it's much appreciated.

extension/chrome/settings/index.ts
```diff
@@ -50,6 +50,10 @@ View.run(class SettingsView extends View {
     const uncheckedUrlParams = Url.parse(['acctEmail', 'page', 'pageUrlParams', 'advanced', 'addNewAcct']);
     this.acctEmail = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'acctEmail');
     this.page = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'page');
+    if (this.page && !/^(\/chrome|modules)/.test!(this.page as string)) {
+      void Ui.modal.error('<div data-test="page-param-modal">An unexpected value was found for the page parameter</div>');
```
tomholub, line 54: I'm not sure how the void works or what it does, but the no-floating-promises error wanted you to handle possible promise rejection with a `.catch(Catch.reportErr)` - the `Ui.modal.error(...)` returns a promise that may reject silently if not handled.

test/source/tests/settings.ts
```diff
@@ -245,6 +245,12 @@ export let defineSettingsTests = (testVariant: TestVariant, testWithBrowser: Tes
       await SettingsPageRecipe.addKeyTest(t, browser, 'ci.tests.gmail@flowcrypt.dev', unprotectedPrvKey, 'this is a new passphrase to protect previously unprotected key');
     }));
 
+    ava.default('settings - error modal when page parameter invalid', testWithBrowser('ci.tests.gmail', async (t, browser) => {
+      const invalidParamModalPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/index.htm?page=invalid`));
```
tomholub, line 249: Maybe it's missing acctEmail param in the url?

seisvelas: Added. That made it not close right away! I don't know why - when I test manually it works just fine even if I leave acctEmail blank. Oh well, that did the trick! Additionally, HTML tags don't render in modals (which is good) so I had to find the test a different way than using the data-test tag. Now tests are passing!

extension/chrome/settings/index.ts
seisvelas: The void operator turns anything null, so we would explicitly ignore the error if the promise was rejected. You're correct that's a bad idea, so instead now I'm logging the error to the console.

tomholub (approved):
Thanks!

PR #3384, opened 2021-02-03
Adjust modal styles for short iframes

### LINKED ISSUE #3369: fix modal rendering in iframes with low height
To triage this, you can try the following:
 - set up an extension
 - go to settings -> Additiomal Settings -> Experimental -> Set a new randomly generated FlowCrypt Account UUID
 - now in settings, you should see "Please log in with FlowCrypt to continue." - ignore that
 - go to any encrypted conversation and find encrypted reply buttons (as per screenshot above)
 - Reply using one of the buttons
 - A new set of buttons re-render, and you should see the "Could not get information from backend" message - that is expected
 - after attempting to scroll down and clicking ok on that modal message, normally, it should allow you to re-authenticate. In my case, it showed me the "platform" error instead. Are you able to reproduce that error too?

In a separate issue, we should also fix the CSS on that message so that the modal is readable / clickable.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/3368#issuecomment-771012100_

@limonte please see if we can make this modal visually fixed / usable. The host frame has a low height so we just barely see the first line of the modal. We did some fix in attachment window. Maybe could do a css media query to do something similar when the frame height is small?

This is purely a request for a css fix, the functionality will be fixed in the originally referenced issue.

limonte:
This PR adds styles for modals in short iframes so they'll be painless to use (no scrolling) for users.

close #3369

image

image

image

### DIFF
--- a/extension/css/cryptup.css
+++ b/extension/css/cryptup.css
@@ -2842,6 +2842,38 @@ body#settings div.webmail_notifications div.webmail_notification a {
 }
 body#settings div.webmail_notifications div.webmail_notification a:hover { text-decoration: none; }
 
+/* short frames */
+@media (max-height: 200px) {
+  .swal2-container {
+    padding: 1px !important;
+  }
+
+  .swal2-popup {
+    width: auto !important;
+    flex-direction: row !important;
+    justify-content: space-around !important;
+    font-size: 0.75em !important;
+    padding: 2px 20px !important;
+  }
+
+  .swal2-header,
+  .swal2-content,
+  .swal2-actions {
+    padding: 0 !important;
+  }
+
+  .swal2-content {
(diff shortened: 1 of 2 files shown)


limonte:
Tests are failing, let me check if that's due to this PR or not.

tomholub (approved):
This is excellent! Thanks

limonte:
Apparently, there's the flaky test https://flowcrypt.semaphoreci.com/workflows/54a8f5b5-872b-4ddf-a6cb-3729341dab33?pipeline_id=c3649734-779a-4122-aabb-1c4b9c42eca4

I'll keep an eye on it and report if it happens again.

PR #3386, opened 2021-02-04
Escape subjects in extension inbox

seisvelas:
Currently XSS is mitigated like so:

https://github.com/FlowCrypt/flowcrypt-browser/blob/4e45cf6dfd80972e34ada1a7572c12e3a7276725/extension/chrome/settings/inbox/inbox.ts#L114-L118

But we don't account for HTML tags, which we don't want to render inside of subject lines.

Resolves https://github.com/FlowCrypt/flowcrypt-security/issues/100


seisvelas:
Tested this manually with Chromium and it does fix the issue. Let me know if you'd also like me to add a test that verifies that HTML tags in subjects are successfully escaped.

tomholub (approved):
Thanks!

PR #3391, opened 2021-02-04
Workaround the unavailability of chrome.windows in iframes on Firefox

### LINKED ISSUE #3368: could not get information from backend / platform not supported
This is on MacOS / Firefox but more likely it's not platform specific. Sounds like something messed up on my account.

image

Here's what likely caused that - which itself is not a problem - normally I can click ok and re-do auth.

image

If I scroll down in there and click ok:

image

That seems like a bug on our end. It would be good to know where that error is coming from.

limonte:
This PR workarounds the Firefox limitation when `chrome.windows` isn't available in iframes. The workaround opens the settings page where user can log in with the standard workflow. 

close #3368

https://user-images.githubusercontent.com/6059356/106969963-9767ad00-6754-11eb-8f19-1198e9b16cc5.mp4

### DIFF
--- a/extension/js/common/browser/browser.ts
+++ b/extension/js/common/browser/browser.ts
@@ -65,16 +65,16 @@ export class Browser {
     const basePath = chrome.runtime.getURL(`chrome/settings/${path}`);
     const pageUrlParams = rawPageUrlParams ? JSON.stringify(rawPageUrlParams) : undefined;
     if (acctEmail || path === 'fatal.htm') {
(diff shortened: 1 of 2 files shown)


tomholub (approved):
Thanks!

tomholub:
Please have a look at the conflict. Thanks!

limonte:
> Please have a look at the conflict. Thanks!

Resolved.

tomholub (approved):

PR #3398, opened 2021-02-05
Adjust OAuth popup size

### LINKED ISSUE #3373: auth window not tall enough
It wants me to scroll down to see the button lately:

image

limonte:
This PR increases the OAuth popup height so users won't have to scroll to the action button.

close #3373

image

Also, I decreased the popup width because that seems to be the cause of that padding from your screenshot (was not happening in my environment):

image


limonte:
@tomholub please manual test this PR in your environment, I'm interested if those white padding would be gone with the decreased popup width. Thanks!

tomholub:
Here: 

image

tomholub (approved):

tomholub:
My environment may be less common compared to typical Win/MacOS so I'm not worried that I still see a scrollbar. As long as the button is visible without scrolling.

PR #3400, opened 2021-02-05
Adjust OAuth popup size for Windows

### LINKED ISSUE #3373: auth window not tall enough

limonte:
This PR increases the OAuth popup height so users won't have to scroll to the action button.

close #3373

@tomholub I assumed that the previous version would work in Windows as well, but after actual checking it wasn't good. Now the popup is good in both OS. Sorry about the hassle.

image


limonte:
Tests are failing not due to this PR, but the failure isn't flaky. @tomholub PTAL

tomholub (approved):

Issue #3332, opened 2021-01-19
ability to store more than one public key per recipient


tomholub:
Right now the extension will store only one public key per recipient email address.

For a smooth user experience with key rotations we need to add the ability to store more than one public key per recipient in the extension. That, combined with the WKD improvements in #3017 and #3018 will allow for a smooth experience during key rotation.

I haven't investigated exactly how the email-to-pubkey relationship is handled on the storage level, it could be a unique index on IDBDatabase. Maybe the index could be simiply dropped without migration. And then the following will need to be updated:
 - [ ] Store method used to retrieve one public key for one recipient should start to return an array of public keys
 - [ ] all code that uses that method downstream should start expecting that too
 - [ ] when the called from compose window, additional code needs to be added to decide which of the retrieved keys to use (some may be expired, some may be revoked - I'd recommend to use whichever valid key that has the most recent signature on it)
 - [ ] when called from pgp decrypt window to verify a signed message, key should be chosen by key id (therefore presumably no change there)

tomholub:
@rrrooommmaaa please have a look to see how involved this change may be - particularly the database / unique index part.

rrrooommmaaa:
@tomholub First refactoring thought: 
In `Contact` structure we have fields `longid` and `longids` and corresponding indexes: `index_longid` and `index_longids`.
`longid` is used to search for primary key and `longids` is used to search among primary and subkeys.
As we're going to have multiple pubkeys in contact, I suggest to drop `longid` field and `index_longid` index and use `longids` field to search.

tomholub:
Instead of having multiple public keys in one contact, can we have one
contact per public key?

rrrooommmaaa:
Currently, `email` is the key field for the Contact's object store. So we'll need to refactor this store to have a different key. I'm not an expert, but it may slow down the search by e-mail. The positive effect would be that there will be less refactoring in other portions of code.

rrrooommmaaa:
> Maybe we could still make it the key (index), but not an unique one (if IDBDatabase supports that)

The key itself should be unique as the IndexedDB is a "key-value" type of the database, if I'm reading the right specs:
https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Basic_Concepts_Behind_IndexedDB#concepts
```
Each record in an object store must have a key that is unique within the same store, so you cannot have multiple records with the same key in a given object store.
```

[...]

Issue #3018, opened 2020-10-12
process / update revoked and expired keys from WKD response


tomholub:
related to  #3017 . Here we build on this issue further, to look at revoked keys that were also received through the WKD. For example, if there is a revocation for the key we currently have in contacts database locally, and then another valid key, we can safely process the revocation (throw away the old local key?) and import the fresh, working key.

needed for upcoming key life-cycle functionality

I imagine this will be done after #3332

rrrooommmaaa:
> For example, if there is a revocation for the key we currently have in contacts database locally, and then another valid key, we can safely process the revocation (throw away the old local key?) and import the fresh, working key.

If we throw away the revoked key and then the user imports the old (unrevoked) version of it? Insecure.

rrrooommmaaa:
Should we expect a situation when, for example, there are 2 subkeys of the key, and we receive 2 copies of the key, each of such copy has only a single subkey revoked? Then we need to somehow merge these packets to have both subkeys revoked in the storage.

tomholub:
> Should we expect a situation when, for example, there are 2 subkeys of the key, and we receive 2 copies of the key, each of such copy has only a single subkey revoked? Then we need to somehow merge these packets to have both subkeys revoked in the storage.

Great question - the answer is no. Our Email Key Manager does not create this situation - it revokes a whole key, not just some subkey. For users who host other type of WKD software, this situation is too esoteric to worry about right now. We could file an issue to think about it later, but for now think in terms of whole keys.

tomholub:
> > For example, if there is a revocation for the key we currently have in contacts database locally, and then another valid key, we can safely process the revocation (throw away the old local key?) and import the fresh, working key.
> 
> If we throw away the revoked key and then the user imports the old (unrevoked) version of it? Insecure.

Good point - let's store the revocation locally instead of the original pubkey. With the revocation stored locally, we can take it into account when either updating pubkeys or when choosing pubkey to use, which is more secure. Thanks!

rrrooommmaaa:
Do you mean to store revocation signature apart from the whole key? Or to store revoked keys separately?
We can just add a flag in the `Key` if there is revocation signature present

tomholub:
> We can just add a flag in the Key if there is revocation signature present

This is what I meant - add a revoked flag to ContactStore.

[...]

PR #3401, opened 2021-02-06
Wkd selects usable key from all received

### LINKED ISSUE #3017: WKD evaluate all received keys
Currently, our client implementation only looks at the first received key.

It should at least look at all of them (WKD response may contain any amount of keys in it), and pick the first key that is usable for encryption. Spec https://tools.ietf.org/html/draft-koch-openpgp-webkey-service-10

Recommend to do after #3003

Later on we will follow up with #3018

rrrooommmaaa:
This PR selects a usable key among all received from Wkd

close #3017


rrrooommmaaa:
@tomholub Looks like `human@flowcrypt.com` fails due to a problem unrelated to this PR.
https://openpgpkey.flowcrypt.com/.well-known/openpgpkey/flowcrypt.com/policy is not reachable.
As for revoked keys -- our goal might be not only to skip them from Wkd's response, but to update ContactStore so these keys won't be usable. How do you think about it?

tomholub:
>  As for revoked keys -- our goal might be not only to skip them from Wkd's response, but to update ContactStore so these keys won't be usable. How do you think about it?

Definitely - but maybe not in this PR. There is another issue #3018 to look into that as a followup.

That is also related to issue https://github.com/FlowCrypt/flowcrypt-browser/issues/3332

I think the overall sequence could be:
 - this PR
 - issue #3332 - multiple pubkeys per recipient - in this issue we would also add logic to select the right key to use if one is revoket, one is expired, one is valid etc
 - issue #3018 - also update revocations

Right now, since there is only one public key per recipient email, we'd have to create some logic that would be quite different (I think) from the final situation. Right now, let's say the recipient rotates the key. Old one is revoked, new one is available. We would have to throw out the old (revoked) one and replace it with the new one. As opposed to, after we support several pubkeys per recipient, we can independently update the revoked ones and also all of the valid ones that WKD returns - save them all. Then only once we try to email that recipient (or verify their emails) will we choose the right public key from all of the ones that are available.

tomholub:
> @tomholub Looks like human@flowcrypt.com fails due to a problem unrelated to this PR. https://openpgpkey.flowcrypt.com/.well-known/openpgpkey/flowcrypt.com/policy is not reachable.

Sorry, fixed

rrrooommmaaa:
> I think the overall sequence could be:
> 
> * this PR
> * issue #3332 - multiple pubkeys per recipient - in this issue we would also add logic to select the right key to use if one is revoket, one is expired, one is valid etc
> * issue #3018 - also update revocations

Ok, so this PR is ready to merge, as I understood.

test/source/mock/wkd/wkd-endpoints.ts
```diff
@@ -116,6 +206,9 @@ export const mockWkdEndpoints: HandlersDefinition = {
   '/.well-known/openpgpkey/localhost/hu/pob4adi8roqdsmtmxikx68pi6ij35oca?l=incorrect': async () => {
     return alice; // advanced for incorrect@localhost
   },
+  '/.well-known/openpgpkey/localhost/hu/66iu18j7mk6hod4wqzf6qd37u6wejx4y?l=some.revoked': async () => {
+    return validAmongRevoked;
```
tomholub, line 210: I just noticed - we are returning armored keys here on the WKD endpoints but that does not reflect how WKD returns keys (which should be unarmored bytes, concatenated). I'll make another issue for this.

tomholub (approved):
Looks good!

PR #3404, opened 2021-02-08
return unarmored keys from WKD mock

### LINKED ISSUE #3402: mock WKD should be returning unarmored keys
I just noticed - we are returning armored keys here on the WKD endpoints but that does not reflect how WKD returns keys (which should be unarmored bytes, concatenated). I'll make another issue for this.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/3401#discussion_r571605651_

OpenPGP.js has a method to return key bytes instead of the armored version. See example https://openpgpkey.flowcrypt.com/.well-known/openpgpkey/flowcrypt.com/hu/14bi5sf9whcm43n83ur8i6r4hfui5msk?l=tom

To return several keys, concatenate the bytes together.

rrrooommmaaa:
This PR makes Wkd mock return unarmored keys

close #3402

### DIFF
--- a/extension/js/common/core/crypto/key.ts
+++ b/extension/js/common/core/crypto/key.ts
@@ -130,6 +130,14 @@ export class KeyUtil {
     return (await KeyUtil.parseMany(text))[0];
   }
 
+  public static dearmor = async (text: string): Promise<{ type: number, data: Uint8Array }> => {
+    const decoded = await opgp.armor.decode(text);
+    let buffer = new Uint8Array();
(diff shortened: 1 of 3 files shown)


rrrooommmaaa:
@tomholub I'm a bit concerned about not having properly tested `KeyUtil.dearmor` as it's unclear how it works on incorrect or large inputs. It's good enough for test infrastructure, but we sort of exposed it to further usage within the app.
And about naming: it dearmors not only keys, should we move it to a different class?

tomholub:
Agree. Feel free to test it further and move it.

PR #3408, opened 2021-02-10
Keep the cursor of the reply box in the viewport

### LINKED ISSUE #3403: The cursor isn't staying in the viewport when inserting new lines in reply box
### Actual behavior

https://user-images.githubusercontent.com/6059356/107156286-27d90400-6986-11eb-9956-d2037dd57a7e.mp4

Cursor is below the viewport, it'll be visible again once user types any character, but `Enter/Return` for some reason doesn't keep the cursor in the viewport. 

### Expected behavior

The cursor is always in the viewport.

limonte:
This PR keeps the cursor of the reply box in the viewport so users always know where they are.

close #3403

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-input-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-input-module.ts
@@ -5,6 +5,7 @@
 import { NewMsgData, RecipientElement } from './compose-types.js';
 import { SquireEditor, WillPasteEvent } from '../../../types/squire.js';
 
+import { BrowserMsg } from '../../../js/common/browser/browser-msg.js';
 import { Catch } from '../../../js/common/platform/catch.js';
 import { Recipients } from '../../../js/common/api/email-provider/email-provider-api.js';
 import { Str } from '../../../js/common/core/common.js';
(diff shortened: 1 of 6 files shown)


limonte:
@tomholub I tried to follow the Gmail's style and fit the whole reply box with the header and the footer to the viewport with the text scrollable, but it turned out not to be the best idea because of the complications which are caused by communicating of iframe and global (parent) window.

In order to resize iframe I need to calculate it's internal height and tell that number to the parent. This caused 2 issues. The first one is internal, the complexity of code has increased and IMO it became too complex for quite a simple issue.

The second one is external and visible for end-users, notice how `111` is blinking, that's because `input_text` increases its height with adding new lines and then with a small delay the container `iframe` increases as well:

https://user-images.githubusercontent.com/6059356/107451007-1727ba00-6b4f-11eb-9e04-0bf2f2ddd5af.mp4

I will probably rewrite the whole thing tomorrow if I don't figure out how to solve the mentioned issues.

tomholub:
Thanks you for posting these here - and for thinking of resulting complexity - I think about that a lot these days.

limonte:
Closing in favor of #3459

PR #3424, opened 2021-02-15
Include DOMPurify JS map for better error output

seisvelas:
Currently, when purify.js throws an error we get this:

>DevTools failed to load SourceMap: Could not load content for chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/lib/purify.js.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME

This PR adds the map so we get better errors from this.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -47,6 +47,7 @@ fi
 mkdir -p $OUTDIR/lib/
 mkdir -p $OUTDIR/css/
 cp node_modules/dompurify/dist/purify.js $OUTDIR/lib/purify.js
+cp node_modules/dompurify/dist/purify.js.map $OUTDIR/lib/purify.js.map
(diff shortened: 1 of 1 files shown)


limonte (approved):
Usually, `.map.js` are used for minified files, but in this case `purify.js` isn't minified but compiled down to support old browsers, that's why there's the reference to `purify.map.js` in it to report issues from the original source, not from the compiled version.

limonte:
Tests are failing not due to this PR, I re-ran them, there seem to be flaky ones as well as the one which is failing constantly: `secure reply btn, reply draft`. I'm the author of that test case, I'll check it out.

tomholub (approved):

PR #3433, opened 2021-02-16
Added HIDE_ARMOR_META to OrgRules #2035

### LINKED ISSUE #2035: OrgRule to hide armor comment and version
There would be an OrgRule flag called `HIDE_ARMOR_META`. Then you'd use `if(orgRules.shouldHideArmorMeta())` - when true, set `opgp.config.show_comment = false` and `opgp.config.show_version = false` - in the following  in the following situations:
 - [ ] rendering of `compose.htm` page. This should cause any sent message to not have any comment/version in armor
 - [ ] during setup in `setup.htm` - this should cause imported keys to get imported without armor

rrrooommmaaa:
This PR adds HIDE_ARMOR_META to OrgRules

close #2035

### DIFF
--- a/extension/chrome/settings/setup.ts
+++ b/extension/chrome/settings/setup.ts
@@ -104,7 +104,7 @@ export class SetupView extends View {
     this.scopes = await AcctStore.getScopes(this.acctEmail);
     this.storage.email_provider = this.storage.email_provider || 'gmail';
     this.orgRules = await OrgRules.newInstance(this.acctEmail);
-    if (this.orgRules.shouldHideArmorMeta()) {
+    if (this.orgRules.shouldHideArmorMeta() && typeof opgp !== 'undefined') {
       opgp.config.show_comment = false;
       opgp.config.show_version = false;
     }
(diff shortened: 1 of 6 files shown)


tomholub (commented):
:+1:

rrrooommmaaa:
@tomholub Is it ok to leave the code like this? Can we put `opgp` configuration somewhere in OrgRules.newInstance?

extension/chrome/settings/setup.ts
```diff
@@ -103,6 +104,10 @@ export class SetupView extends View {
     this.scopes = await AcctStore.getScopes(this.acctEmail);
     this.storage.email_provider = this.storage.email_provider || 'gmail';
     this.orgRules = await OrgRules.newInstance(this.acctEmail);
+    if (this.orgRules.shouldHideArmorMeta()) {
+      opgp.config.show_comment = false;
+      opgp.config.show_version = false;
+    }
```
tomholub, line 110: You could say here `if (this.orgRules.shouldHideArmorMeta() && typeof opgp !== 'undefined') {` which would address your concern (situation when opgp not imported/included).

Then you can leave the rest as is.

tomholub: Because if opgp doesn't exist, then we are at no risk of successfully sending messages - with or without comments :)

tomholub (approved):
Great!

tomholub:
This needs documenting, I'll make an issue for Bilal :+1:

Issue #3432, opened 2021-02-16
password protected message: hidden button


tomholub:
I managed to confuse the compose window to not re-calculate height, or do it badly:

![image](https://user-images.githubusercontent.com/6306961/108045963-faeab800-703b-11eb-826f-9d9119e94eeb.png)

The button on the bottom is indeed cropped and barely clickable.

Typing in the compose box doesn't help. In fact, when I click on intro, it gets worse:

![image](https://user-images.githubusercontent.com/6306961/108046107-2a99c000-703c-11eb-9759-aeea04e36f04.png)

It fixed itself when I closed the window and re-opened it (adding intro now made it re-calculate height).

@limonte I know you may be shooting in the dark but please see if anything comes to your mind. Thank you!

limonte:
Will try to figure it out. Firefox, right?

tomholub:
Correct - Firefox :+1:

limonte:
I managed to reproduce the same issue in the reply box, but I'm not sure if the underlying reasons are the same for your case and for this case:

https://user-images.githubusercontent.com/6059356/108120107-70d93880-70a9-11eb-98a4-4774de308c11.mp4

tomholub:
Looks close enough - either way a worthwhile fix :+1:

tomholub:
In my case, if I started typing in the box, it would still not fix itself. I imagine in your case, typing in the box would fix it. So the cause is potentially different.

limonte:
> Looks close enough - either way a worthwhile fix 👍

Yup, I'll proceed with fixing the reply box issue.

> In my case, if I started typing in the box, it would still not fix itself. I imagine in your case, typing in the box would fix it. So the cause is potentially different.

Correct.

limonte:
@tomholub FYI: I'm aware of this issue and that it's labeled as first priority, but I wasn't able to observe it so far.

tomholub:
I think I haven't seen this in a while. Closing for now.

PR #3434, opened 2021-02-16
Trigger resizeComposeBox() after showing/hiding password UI

limonte:
This PR resizes reply box when showing/hiding password UI so the send btn is visible when the email address in recipients is marked grey

Addresses https://github.com/FlowCrypt/flowcrypt-browser/issues/3432#issuecomment-780109901


tomholub (approved):
Thanks!

PR #3437, opened 2021-02-17
Prevent HTML injection via Label ID

seisvelas:
This resolves https://github.com/FlowCrypt/flowcrypt-security/issues/106 by escaping the LabelID. Here's how this looks now when I try to inject HTML:

Screen Shot 2021-02-17 at 16 31 48

Let me know if I'm not accounting for anything!


tomholub (approved):
Thanks!

PR #3445, opened 2021-02-19
Introducing `emails` and `pubkeys` stores instead of `contacts`

### LINKED ISSUE #3332: ability to store more than one public key per recipient

rrrooommmaaa:
This PR uses `search` index on `searchable` field instead of `index_has_pgp`, so we can remove `has_pgp` field.
This particular step can help us measure performance as compared to `index_has_pgp` usage

close #3332


rrrooommmaaa:
@tomholub What do you think about it? As each array element in `searchable` is prefixed with either `t:` for `has_pgp=1` or `f:` for `has_pgp=0`, we can use range search `f:` up to `t:` exclusive (or simply up to `t:`) instead of former `has_pgp=0` search, and `t:` inclusive upwards instead of `has_pgp=1`, so we can eliminate the need for `has_pgp` field as you suggested (the prefix will be calculated based on `fingerprints` length). It may be slower due to the fact that `searchable` has many more values, do we need benchmarks?
Do you know of any tests in place that actually check correctness of search based solely on `has_pgp` field
or should I add one?

tomholub (commented):
A clever hack. This would have to be documented better, don't be afraid to write some prose because it's otherwise not obvious at all.

For tests, try to give it knowingly wrong behavior, and see if anything breaks. I imagine something should break.

rrrooommmaaa:
@tomholub I have populated the store with 12500 dummy records, half of them `has_pgp=0`, the other half -- `has_pgp=1`.
Each `searchable` array field has about 50 entries, so `search` index is 50 times larger than `index_has_pgp` (counting the number of keys).
So range search on (... up to 't:') correctly yielded 6250 records of `has_pgp=0` in about 26 seconds, same with ('t:' onwards).
The `index_has_pgp` search took a fraction of second.
However, we can optimize `searchable` in a way that we no longer have ALL letters entries, e.g. it's enough to have
"t:tom" instead of "t:t", "t:to" and "t:tom", so we can do a range search ("t:t" up to "t:u" exlusive) when user types "t", ("t:to" up to "t:tp" exclusive) when user types "to" etc., so "t:tom" will be found in all cases.
Thus the size of `search` index will decrease (and speed up) tenfold.
So the cost of not having `has_pgp` field be about 1 extra second per 1000 entries. Note that this is true for `has_pgp` only searches, used in a couple in places, like 'Export All Keys'.
Compose uses `has_pgp` AND substring search, so `search` index is used there anyway -- no performance loss happens (On the opposite -- performance gain will happen if we reduce the size of `search` index as I suggested).
Thoughts?

tomholub:
You can try this. The code should be commented thoroughly so that it's clear what is it trying to do, and why.

tomholub:
For the client field - it was used to decide whether to attach public key or not. Let's have the following logic:
 - if all recipients are on the same domain as me, don't need to attach public key automatically
 - if at least one of the recipients are on another domain, then:
      - if my own key is available on WKD, I don't need to attach automatically
      - if my own key is not available on WKD, then:
           - if I've already sent my public key to this person before, don't need to attach automatically
           - else attach automatically

All of this logic only makes the public key attach button either activated or not automatically. User can still change it by clicking on the icon before actually sending the message.

`if I've already sent my public key to this person before` is already handled by method `this.view.recipientsModule.doesRecipientHaveMyPubkey` so you'll just change contents of `reevaluateShouldAttachOrNot` as per the logic above.

Thanks!

extension/chrome/elements/compose-modules/compose-my-pubkey-module.ts
```diff
@@ -46,7 +46,7 @@ export class ComposeMyPubkeyModule extends ViewModule<ComposeView> {
     (async () => {
       const contacts = await ContactStore.get(undefined, this.view.recipientsModule.getRecipients().map(r => r.email));
       for (const contact of contacts) {
-        if (contact?.has_pgp && contact.client !== 'cryptup') {
+        if (contact?.has_pgp) {
```
tomholub, line 49: Here a bit of logic change as per comment above.

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -110,12 +110,11 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
             lookupResult.pubkey = null; // tslint:disable-line:no-null-keyword
           }
         }
-        const client = lookupResult.pgpClient === 'flowcrypt' ? 'cryptup' : 'pgp'; // todo - clean up as "flowcrypt|pgp-other'. Already in storage, fixing involves migration
         const ksContact = await ContactStore.obj({
           email,
           name,
           pubkey: lookupResult.pubkey,
-          client: lookupResult.pubkey ? client : undefined,
+          // client: // todo: decide whether we need this field
```
tomholub, line 117: Don't need

extension/chrome/settings/modules/contacts.ts
```diff
@@ -118,11 +118,7 @@ View.run(class ContactsView extends View {
     const [contact] = await ContactStore.get(undefined, [$(viewPubkeyButton).closest('tr').attr('email')!]); // defined above
     $('.hide_when_rendering_subpage').css('display', 'none');
     Xss.sanitizeRender('h1', `${this.backBtn}${this.space}${contact!.email}`); // should exist - from list of contacts
-    if (contact!.client === 'cryptup') {
-      Xss.sanitizeAppend('h1', '&nbsp;&nbsp;&nbsp;&nbsp;<img src="/img/logo/flowcrypt-logo-19-19.png" />');
-    } else {
-      Xss.sanitizeAppend('h1', '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;');
-    }
+    Xss.sanitizeAppend('h1', '&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;'); // todo: do we need this line?
```
tomholub, line 121: I think it's for some formatting. @limonte would probably want to murder me for this one :D but for now please add it into the `Xss.sanitizeRender('h1'` on line 120.

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -29,8 +29,8 @@ export class SmimeKey {
       expiration: SmimeKey.dateToNumber(certificate.validity.notAfter),
       fullyDecrypted: false,
       fullyEncrypted: false,
-      isPublic: true,
-      isPrivate: true,
+      isPublic: !!certificate.publicKey,
+      isPrivate: !!certificate.privateKey,
```
tomholub, line 33: The way we use these fields, currently I think, is either/or.

So either isPublic is true, or isPrivate, not both.

That means if there is both public and private key included, I think we'd set isPublic to false and isPrivate to true. (for consistency with pgp)

extension/js/common/platform/store/contact-store.ts
```diff
@@ -61,27 +70,23 @@ export class ContactStore extends AbstractStore {
...
+          const pubkeys = openDbReq.result.createObjectStore('pubkeys', { keyPath: 'fingerprint' });
+          emails.createIndex('search', 'searchable', { multiEntry: true });
+          emails.createIndex('index_pending_lookup', 'pendingLookup');
+          emails.createIndex('index_fingerprints', 'fingerprints', { multiEntry: true }); // fingerprints of all connected pubkeys
+          pubkeys.createIndex('index_longids', 'longids', { multiEntry: true }); // longids of all public key packets in armored pubkey
         }
-        if (event.oldVersion < 3) {
-          contacts = openDbReq.transaction!.objectStore('contacts');
-          contacts.createIndex('index_longids', 'longids', { multiEntry: true }); // longids of all public key packets in armored pubkey
+        if (event.oldVersion < 5) { // delete when migrating from 4 to 5?
```
tomholub, line 83: Yep, sounds good. Leave in for now, for safety.

tomholub (commented):
Looks directionally good :+1:  I'll scrutinize the details later.

extension/js/common/platform/store/contact-store.ts
rrrooommmaaa: In fact, I figured that we've never done db-based migration, so it would be more convenient to call some migration procedure right here. Do you have any objections against it? We can think of a nice way to inject a function here somehow.

extension/js/background_page/background_page.ts
```diff
@@ -41,6 +41,7 @@ opgp.initWorker({ path: '/lib/openpgp.worker.js' });
 
   try {
     db = await ContactStore.dbOpen(); // takes 4-10 ms first time
+    await moveContactsToEmailsAndPubkeys(db);
```
rrrooommmaaa, line 44: @tomholub this is the piece of code I was referring to when talking about contacts migration.
It would be more convenient to perform migration inside `dbOpen()` call. Thus we'll be able to delete `contacts` store right after successful conversion to `emails` and `pubkeys`.
We can, for example, pass a list of functions to `dbOpen` to nicely inject transformations so they don't clutter `ContactStore` code?

rrrooommmaaa: or, perhaps, `ContactStore.update` will cause errors when called from inside upgrade transaction

extension/js/common/platform/store/contact-store.ts
tomholub: Sure, that's fine with me - the other migrations were for local storage which works differently. Fine to put it here.

extension/js/background_page/background_page.ts
tomholub: We can do all of this in `ContactStore` itself, no need to bring the migration out to here.

rrrooommmaaa: > We can do all of this in ContactStore itself, no need to bring the migration out to here.

Looks like there is a problem with this approach. IDTransaction docs say that transaction is automatically completed when it goes out of scope, and this disallows usage of async function in `onupgradeneeded` etc.. They are fired alright, but the transaction is closed by the time the actual processing occurs (before waiting for the promise returned by `onupgradeneeded` handler to be resolved). I searched the github for projects where an async function is used as event handler, found only a small project https://github.com/SourceCodeBot/crudodb (I suspect they haven't tested it properly).
The general approach for transactional operations is (though I haven't found these exact requirement in IndexedDB specs) -- a handler must immediately open a new request, otherwise the transaction will be auto-completed.
We can't convert contacts to pubkeys/emails in a completely synchronized way because of async `KeyUtil.parse` function.
(This is the point for another discussion -- KeyUtil.parse is async only because it uses a Stream API approach for dearmoring of potentially large messages. As the keys are small, we may think about a sync implementation of KeyUtil.parse)
Still, maybe it's user-unfriendly to "hang" (will it hang?) the browser window for the process of database upgrade, converting multiple contacts, so it makes sense to keep the conversion out of the upgrade handler. The drawback is that the empty `contacts` table will be lying around until the next upgrade.

tomholub: >  Still, maybe it's user-unfriendly to "hang" (will it hang?) the browser window for the process of database upgrade, converting multiple contacts, so it makes sense to keep the conversion out of the upgrade handler. The drawback is that the empty contacts table will be lying around until the next upgrade.

It will possibly hang the whole browser when we do something like this in background page. Also depends on contacts count. For 50 contacts, not a problem. For 1000 contacts - possible problem.

I think our most reasonable chance is to do a migration of database structure itself, fire off async migration of the actual database, and hope for the best.

If the actual migration does not finish successfully, then on next migration it will be retried (like a browser restart). I think that's ok as long as there is this retry in place.

Best would be to then not migrate all contacts in one gigantic transaction, but one by one. If one of the transactions throws an error because whatever (parsing, concurency, etc), it would not fail the whole migration. 100 keys already migrated, key 101 failed, 50 more to go - the failed one gets catched and skipped, and it will chug along for the remaining 50 keys. Then next time it only has to retry one.

rrrooommmaaa:
> * if my own key is available on WKD, I don't need to attach automatically

Do we need to test this condition by actually sending a lookup request or somehow obtain this information from the settings?

tomholub:
I would actually make a request to the WKD to check for this during sending, for simplicity

rrrooommmaaa:
> I would actually make a request to the WKD to check for this during sending, for simplicity

This means, that `icon_pubkey' can be `active`, but still, the final decision to not send the pubkey will be taken in `shouldAttach` to eliminate the need to cache Wkd result -- ok, if you consider this 'simpler', so be it.

rrrooommmaaa:
UPD: if we should use `getSender()` from `compose-sender-module`, than yes, we should do the check at the moment of sending, when we have the exact senderEmail. So makes no sense to `reevaluate...` the `icon_pubkey` to be `active` on an event and then override it to be inactive, as it may be misleading to the user.
I think it's better to have this button as inactive and then override to active unless `toggledManually`

tomholub:
> This means, that icon_pubkey can be active, but still, the final decision to not send the pubkey will be taken in shouldAttach to eliminate the need to cache Wkd result -- ok, if you consider this 'simpler', so be it.

No, this would be complicated and confusing to user as you say.

The evaluation will happen at the same times that it happens today (probably during loading, and then when a recipient is added or removed, but not if the user ever toggled it manually in that session).

See this comment https://github.com/FlowCrypt/flowcrypt-browser/pull/3445#issuecomment-791414059 :

> All of this logic only makes the public key attach button either activated or not automatically. User can still change it by clicking on the icon before actually sending the message.

Meaning, if user didn't toggle it manually, the above is used to color the button. At the time of actual sending, we just follow the color of the button - there is no second guessing.

As for this condition:

> if my own key is available on WKD, I don't need to attach automatically

You can cache that in a class variable so that you only need to pull it once in one secure compose window.

I'm happy to explain further, sorry I confused you with the "during sending", I really meant "while using the secure compose window".

extension/js/background_page/background_page.ts
rrrooommmaaa: I have implemented migration in batches of 50, but it's going to throw on an error and crash the extension.
If we skip, this portion of contacts won't be available for search -- is this what you're after?

tomholub:
I've fixed up the tests on master - sorry for the mess earlier. I also fixed the WKD test.

extension/js/common/platform/store/contact-store.ts
```diff
@@ -14,14 +13,36 @@ import { OpenPGPKey } from '../../core/crypto/pgp/openpgp-key.js';
...
+  pendingLookup: number;
+  lastUse: number | null;
+};
+
+type Pubkey = {
+  fingerprint: string;
+  armoredKey: string;
+  longids: string[];
+  lastCheck: number | null,
+  lastSig: number;
```
rrrooommmaaa, line 36: @tomholub  Turns out we don't need to store this field, as we don't index it, nor search by it and never extract its value (the key is always parsed and `lastModified` of the parsed key is returned instead)

tomholub: sounds good, can drop

rrrooommmaaa:
@tomholub Looks like `pending_lookup` (aka `pendingLookup`) is always 0. I haven't found any usage of that flag.

tomholub:
I used it in the past - to mark recipients that were just loaded but presence of their pubkey was not checked yet. But maybe over the years we refactored it away? If it's unused, that's better - simpler.

rrrooommmaaa:
Should we keep it or remove? (unclear from your answer)

tomholub:
If it's not used, definitely remove it - it will make the code simpler, which is good.

tomholub:
Sorry for the lack of clarity earlier

extension/js/background_page/background_page.ts
tomholub: When there is an error, it's better to skip a batch then stop half way through the migration.

tomholub: Will the skipped batch get retried (without any special retry logic) next time I restart the browser?

extension/js/background_page/migrations.ts
```diff
@@ -26,3 +40,60 @@ export const migrateGlobal = async () => {
...
   }
 };
+
+export const moveContactsToEmailsAndPubkeys = async (db: IDBDatabase): Promise<void> => {
+  if (!db.objectStoreNames.contains('contacts')) {
+    return;
+  }
+  console.info('migrating contacts of ContactStore to emails and pubkeys...');
+  const batchSize = 50;
+  while (await moveContactsBatchToEmailsAndPubkeys(db, batchSize)) {
```
tomholub, line 50: I think it's ok as it is. If it crashes, and half of contacts don't get converted, it will try again next time I restart browser.

The only change would be to ensure that the plugin can start up and continue functioning if this method throws.

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -29,8 +29,8 @@ export class SmimeKey {
       expiration: SmimeKey.dateToNumber(certificate.validity.notAfter),
       fullyDecrypted: false,
       fullyEncrypted: false,
-      isPublic: true,
-      isPrivate: true,
+      isPublic: certificate.publicKey && !certificate.privateKey,
+      isPrivate: !!certificate.privateKey,
```
tomholub, line 33: :+1:

tomholub (commented):
This PR is massive. It's looking good.

extension/js/common/platform/store/contact-store.ts
```diff
@@ -14,14 +13,37 @@ import { OpenPGPKey } from '../../core/crypto/pgp/openpgp-key.js';
...
+  lastCheck: number | null,
+  expiresOn: number | null;
+};
+
+type PubkeyAttributes = {
+  fingerprint: string | null;
+  expiresOn: number | null;
+};
+
+export type ContactV4 = {
```
rrrooommmaaa, line 42: @tomholub this is a type I created to conveniently extract the contact info (from `emails` table) and all associated pubkeys. So far it is used internally only, but later we can use it from outside the ContactStore.

extension/js/background_page/background_page.ts
rrrooommmaaa: The conversion stop at the failed batch with the current implementation. 
as `contacts` store objects are deleted within the same transaction as populating new stores, the contacts from the failed batch will remain in the database, thus they will be retried on subsequent browser restart.

extension/js/background_page/migrations.ts
rrrooommmaaa: I wrapped the call in `try/catch` and the extension is usable after conversion failure now.

rrrooommmaaa: I tested it. The records beyond and including the failed batch are retained in `contacts`. They can't be found by the app but will be retried on the next background start.

rrrooommmaaa:
I tested most of the use cases. Do you have any other suggestions on what to test?

tomholub:
Test-wise, I think it's well covered. I'll take another look at the code.

extension/js/common/core/crypto/key.ts
```diff
@@ -50,16 +50,10 @@ export type PubkeyResult = { pubkey: Key, email: string, isMine: boolean };
 export type Contact = {
   email: string;
   name: string | null;
-  pubkey: Key | null;
+  pubkey: Key | undefined;
   has_pgp: 0 | 1;
```
tomholub, line 54: Since we are doing a migration anyway, could we rename these to be consistently camelCase? To use `hasPgp`, `lastUse` etc.

test/source/tests/browser-unit-tests/unit-ContactStore.js
```diff
@@ -0,0 +1,349 @@
...
+ *  - BROWSER_UNIT_TEST_NAME(`some test name`).enterprise;
+ *  - BROWSER_UNIT_TEST_NAME(`some test name`).consumer;
+ * 
+ * This is not a JavaScript file. It's a text file that gets parsed, split into chunks, and
+ *    parts of it executed as javascript. The structure is very rigid. The only flexible place is inside
+ *    the async functions. For the rest, do not change the structure or our parser will get confused.
+ *    Do not put any code whatsoever outside of the async functions.
+ */
+
+BROWSER_UNIT_TEST_NAME(`ContactStore is able to search by partial email address`);
```
tomholub, line 25: Let's add a check that for email `abcd.vwxyz@hello.com` we are able to find it by `vwx` and also `hel`.

test/source/tests/browser-unit-tests/unit-ContactStore.js
```diff
@@ -0,0 +1,349 @@
...
+  if (abddefBySubkeyLongid.pubkey.id !== '9E020D9B752FD3FFF17ED9B65FCC1541CF282951') {
+    throw Error(`Expected to get the key fingerprint 9E020D9B752FD3FFF17ED9B65FCC1541CF282951 but got ${abddefBySubkeyLongid.pubkey.id}`)
+  }
+  return 'pass';
+})();
+
+BROWSER_UNIT_TEST_NAME(`ContactStore gets a valid pubkey by e-mail, or exact pubkey by longid`);
+(async () => {
+  // Note 1: email differs from pubkey id
+  // Note 2: not necessary to call ContactStore.save, it's possible to always use ContactStore.update
```
tomholub, line 331: I wonder if we should stop distinguishing between `save` and `update` and always call it `set`? Do we really currently use them differently, and is there actually a difference?

tomholub (commented):
All looks good - a few comments.

(still need to review contact-store.ts closer).

Thanks!

extension/js/common/platform/store/contact-store.ts
```diff
@@ -272,38 +354,31 @@ export class ContactStore extends AbstractStore {
...
-        search = contacts.openCursor(); // no substring, already covered in `typeof query.has_pgp === 'undefined' && query.substring` above
+        search = emails.openCursor(); // no substring, already covered in `typeof query.has_pgp === 'undefined' && query.substring` above
       } else { // specific query.has_pgp value
-        if (query.substring) {
-          search = contacts.index('search').openCursor(IDBKeyRange.only(ContactStore.dbIndex(query.has_pgp, query.substring)));
-        } else {
-          search = contacts.index('index_has_pgp').openCursor(IDBKeyRange.only(Number(query.has_pgp)));
-        }
+        const indexRange = ContactStore.dbIndexRange(query.has_pgp, query.substring ?? '');
+        const range = IDBKeyRange.bound(indexRange.lowerBound, indexRange.upperBound, false, true);
```
tomholub, line 364: Please add comments explaining exactly how this range works, how we (ab)use it and why. I think you explained it well in an earlier comment on the issue or PR, can just copy to become code comment.

extension/js/common/platform/store/contact-store.ts
```diff
@@ -312,110 +387,108 @@ export class ContactStore extends AbstractStore {
...
-        const normalized = ContactStore.normalizeString(substring);
-        if (!index.includes(normalized)) {
-          index.push(ContactStore.dbIndex(hasPgp, normalized));
-        }
-      }
+  private static dbIndexRange = (hasPgp: boolean, substring: string): { lowerBound: string, upperBound: string } => {
+    const lowerBound = ContactStore.dbIndex(hasPgp, substring);
+    let copyLength = lowerBound.length - 1;
+    let lastChar = lowerBound.charCodeAt(copyLength);
+    while (lastChar >= 65535) {
```
tomholub, line 398: Please add a code comment about what this line does and why

extension/js/common/platform/store/contact-store.ts
```diff
@@ -175,72 +167,38 @@ export class ContactStore extends AbstractStore {
...
-        // or if we were searching by email, return found contact or nothing
-        return ContactStore.recreateDates([contact]);
-      } else {
-        // not found any key by primary longid, and searching by longid -> search by any subkey longid
-        // it may not find pubkeys imported before August 2019, re-importing such pubkeys will make them findable
-        return ContactStore.recreateDates([await ContactStore.dbContactInternalGetOne(db, emailOrLongid[0], true)]);
+      const contact = await ContactStore.dbContactInternalGetOne(db, emailOrLongid[0]);
+      if (!contact) {
+        return [contact];
       }
```
tomholub, line 193: This doesn't seem to do anything different compared to the line below

test/source/tests/browser-unit-tests/unit-ContactStore.js
rrrooommmaaa: This PR makes no difference between `save` and `update` operations. I left them untouched so we won't need to update a lot of client code in this massive PR.

extension/js/common/core/crypto/key.ts
rrrooommmaaa: This is a purely DTO class. We can of course refactor it in this PR, making it grow even bigger.

tomholub (commented):
Yup

test/source/tests/browser-unit-tests/unit-ContactStore.js
tomholub: Ok. This would be something to clean up in a followup PR.

extension/js/common/platform/store/contact-store.ts
rrrooommmaaa: Done

extension/js/common/platform/store/contact-store.ts
rrrooommmaaa: Done

test/source/tests/browser-unit-tests/unit-ContactStore.js
rrrooommmaaa: Done

tomholub (approved):
Thanks!

PR #3459, opened 2021-02-25
Keep the cursor of the reply box always in viewport

### LINKED ISSUE #3403: The cursor isn't staying in the viewport when inserting new lines in reply box

limonte:
This PR keeps the cursor of the reply box in the viewport so users always know where they are.

close #3403

@tomholub this small issue kept me busy, and I'm still unable to come up with the perfect solution, but this one looks good enough for me. There's one visual imperfection which I'm unable to fix - the visible vertical jumpiness of the Send button when adjusting `scrollTop` of `convoRootScrollable`:

https://user-images.githubusercontent.com/6059356/109157005-3131e100-777a-11eb-8212-4c26d1e3020b.mp4

I'm not even sure if there is a solution to keep the button stable, but if there is one it would probably be complex.

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-input-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-input-module.ts
@@ -3,7 +3,7 @@
 'use strict';
 
 import { NewMsgData, RecipientElement } from './compose-types.js';
-import { SquireEditor, WillPasteEvent } from '../../../types/squire.js';
+import { CursorEvent, SquireEditor, WillPasteEvent } from '../../../types/squire.js';
 
 import { Catch } from '../../../js/common/platform/catch.js';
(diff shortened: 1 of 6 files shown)


tomholub (approved):
Looks good!

For upcoming priority please have a look at https://github.com/FlowCrypt/enterprise-admin-panel/issues/408 . Thank you :+1:

PR #3474, opened 2021-03-06
Disallow clipboard copy of private key

seisvelas:
Remove the option to copy private key to clipboard due to formatting inconveniences for users when eg pasting into word

Resolves https://github.com/FlowCrypt/flowcrypt-security/issues/115

### DIFF
--- a/extension/chrome/settings/modules/my_key.htm
+++ b/extension/chrome/settings/modules/my_key.htm
@@ -49,8 +49,6 @@
     </div>
     <div class="line"><span class="red_label">Never send this to anyone!</span></div>
     <div class="line">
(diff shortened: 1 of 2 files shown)


seisvelas:
Built and loaded it up and it renders fine

Screen Shot 2021-03-05 at 18 48 01

tomholub (commented):
Thanks - should also remove whichever handler was associated with `action_copy_prv` click

tomholub (approved):

seisvelas:
@tomholub I am pretty sure the failing tests are related to https://github.com/FlowCrypt/flowcrypt-browser/issues/3481, although we can wait til that's fixed before merging this, just in case (if you'd prefer)

PR #3480, opened 2021-03-11
Escape HTML in password protected message intros

seisvelas:
Currently, here is what happens if I include an `<i>` tag in a password protected message's unencrypted intro section:

Screen Shot 2021-03-11 at 1 53 05

As you can see, the tag is applied and the rest of the email is italicized. Here is what happens after this PR:

Screen Shot 2021-03-11 at 1 52 51


seisvelas:
Failing lots of tests, none of which seem connected to this change in any obvious way.

seisvelas:
Okay, I reran the tests for the last commit to master (which passed when the commit was merged) and they are now failing: https://flowcrypt.semaphoreci.com/workflows/9ba160ab-2172-4a99-85ca-234edf401adc?pipeline_id=320a823f-bbcf-41db-8052-d2f11c0bbe7f

So I don't believe this is related to this PR specifically.

tomholub (approved):
I'll still merge this, thank you!

PR #3487, opened 2021-03-16
auto include pubkey based on rules discussed in #3445

rrrooommmaaa:
This PR is part of #3332 refactoring
It implements "auto include pubkey" feature based on 
this comment https://github.com/FlowCrypt/flowcrypt-browser/pull/3445#issuecomment-791414059


extension/js/common/api/key-server/wkd.ts
```diff
@@ -22,23 +23,28 @@ export class Wkd extends Api {
 
   constructor(private myOwnDomain: string) {
     super();
+    if (myOwnDomain === 'wkd.mock.flowcryptlocal.com') {
+      this.port = 8001;
```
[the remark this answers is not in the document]
tomholub: I think I was including it as a part of the mock email address. So the tests log in as `whoever@thedomain.com:8001`. It's just mocks, so they don't care.

extension/chrome/elements/compose-modules/compose-my-pubkey-module.ts
```diff
@@ -8,11 +8,13 @@ import { Lang } from '../../../js/common/lang.js';
...
 import { ViewModule } from '../../../js/common/view-module.js';
 import { ComposeView } from '../compose.js';
-import { ContactStore } from '../../../js/common/platform/store/contact-store.js';
+import { Str } from '../../../js/common/core/common.js';
+import { OpenPGPKey } from '../../../js/common/core/crypto/pgp/openpgp-key.js';
 
 export class ComposeMyPubkeyModule extends ViewModule<ComposeView> {
 
   private toggledManually = false;
+  private wkdLongids: { [acctEmail: string]: string[] } = {};
```
tomholub, line 17: It is preferable to compare fingerprints over longids whenever possible - fingerprints are slightly more secure, and there is no advantage using longids.

The only place that makes sense to use longids is when looking up public keys of a signed message, because the signature only contains a longid. Similarly encrypted message only mentions longid of the key it's encrypted for. In all other situations, let's use fingerprints.

extension/chrome/elements/compose-modules/compose-my-pubkey-module.ts
```diff
@@ -44,18 +46,40 @@ export class ComposeMyPubkeyModule extends ViewModule<ComposeView> {
...
+        if (!Array.isArray(cached)) {
+          // slow operation -- test WKD for our own key and cache the result
+          const { keys } = await this.view.pubLookup.wkd.rawLookupEmail(senderEmail);
+          const longids = keys.map(key => OpenPGPKey.fingerprintToLongid(key.id));
+          this.wkdLongids[senderEmail] = longids;
+          // check fingerprint
+          if (longids.some(id => id === senderKi.longid)) {
+            this.setAttachPreference(false);
+            return;
+          }
```
tomholub, line 71: Similarly here. If we use fingerprints, the code will even simplify.

extension/chrome/elements/compose-modules/compose-my-pubkey-module.ts
```diff
@@ -44,18 +46,40 @@ export class ComposeMyPubkeyModule extends ViewModule<ComposeView> {
...
+        }
+        for (const recipient of foreignRecipients) {
           // new message, and my key is not uploaded where the recipient would look for it
-          if (! await this.view.recipientsModule.doesRecipientHaveMyPubkey(contact.email)) {
+          if (! await this.view.recipientsModule.doesRecipientHaveMyPubkey(recipient)) {
             // either don't know if they need pubkey (can_read_emails false), or they do need pubkey
```
tomholub, line 76: This is outdated comment (already before this PR), can just say 

_// they do need pubkey_

extension/chrome/elements/compose-modules/compose-my-pubkey-module.ts
```diff
@@ -44,18 +46,40 @@ export class ComposeMyPubkeyModule extends ViewModule<ComposeView> {
...
+          // check fingerprint
+          if (longids.some(id => id === senderKi.longid)) {
+            this.setAttachPreference(false);
+            return;
+          }
+        }
+        for (const recipient of foreignRecipients) {
           // new message, and my key is not uploaded where the recipient would look for it
-          if (! await this.view.recipientsModule.doesRecipientHaveMyPubkey(contact.email)) {
+          if (! await this.view.recipientsModule.doesRecipientHaveMyPubkey(recipient)) {
```
tomholub, line 75: `doesRecipientHaveMyPubkey` seems like it would be better to place it in this file and not in recipients module, but hard to say for sure without opening the IDE, and not overly important

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -300,6 +298,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     this.onRecipientAddedCallbacks.push(callback);
   }
 
+  // todo: shouldn't we check longid?
```
tomholub, line 301: We really should - thank you for noticing. I'll file this as a separate issue.

extension/js/common/api/key-server/wkd.ts
```diff
@@ -67,7 +77,7 @@ export class Wkd extends Api {
       return { pubkey: null, pgpClient: null };
     }
     // if recipient uses same domain, we assume they use flowcrypt
-    const pgpClient = this.myOwnDomain === recipientDomain ? 'flowcrypt' : 'pgp-other';
+    const pgpClient = this.myOwnDomain === Str.getDomainFromEmailAddress(email) ? 'flowcrypt' : 'pgp-other';
```
tomholub, line 80: I suppose we'll be dropping pgpClient from everywhere. If you want to chunk up the upcoming PR, this could also be done in a separate PR.

extension/js/common/core/common.ts
```diff
@@ -33,6 +33,11 @@ export class Str {
     return { email, name, full };
   }
 
+  public static getDomainFromEmailAddress = (emailAddr: string) => {
+    // todo: parseEmail()?
```
tomholub, line 37: I think parseEmail would be sensible here, and throw if it returns "undefined"

tomholub (commented):
Excellent! Thank you for splitting it out. Some comments below.

tomholub (dismissed):
Looks good, should I merge it?

extension/js/common/api/key-server/wkd.ts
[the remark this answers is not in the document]
rrrooommmaaa: Yes, this works perfectly well...

extension/chrome/elements/compose-modules/compose-my-pubkey-module.ts
rrrooommmaaa: Done, but I had to parse the `KeyInfo.private` field as we don't currently have the primary fingerprint there, only `longid` and a set of all `fingerprints`. Or do we have to do a "must include all fingerprints" check? Also, do we need to attach our key only if "Sign" option is ticked?

extension/js/common/api/key-server/wkd.ts
rrrooommmaaa: Let's drop it in the PR #3445 that actually "deletes" the `client` field from the database. This branch still attempts to store it.

rrrooommmaaa:
> Looks good, should I merge it?

This test `compose - auto include pubkey is inactive when our key is available on Wkd` fails on the enterprise mock setup.
Do we have to configure hosts additionally?

tomholub:
> > Looks good, should I merge it?
> 
> This test `compose - auto include pubkey is inactive when our key is available on Wkd` fails on the enterprise mock setup.
> Do we have to configure hosts additionally?

From memory, enterprise extension has a more restricted manifest file which means we have to list the mock domains in manifest file. Or in OS hosts files, or both. Is it possible to use an existing mock domain that some other tests use? Just a different user on the same domain? If yes, please try that to limit amount of external configuration needed for the tests.

rrrooommmaaa:
Enterprise is still failing to access Wkd at `google.mock.flowcryptlocal.com:8001`

tomholub:
I fixed that tests on master now.

PR #3540, opened 2021-04-04
Dumping contents of 'debug' IndexedDB in browser tests #3482

### LINKED ISSUE #3482: how to extract unit test cases for MsgBlockParser.detectBlocks from integration tests?
In this repo we run many integration tests, that tend to also use the detectBlocks method.

Now we're trying to port this functionality to Kotlin at https://github.com/FlowCrypt/flowcrypt-android/issues/1052#issue-821114723 but we only have three unit tests for this method, most of the testing is done through integration tests.

@rrrooommmaaa how could we extract the data inputs/outputs for that method for a whole test run, deduplicate them and export them? This would be very valuable when porting similar functionality. We will be doing more porting in a similar way upcoming weeks/months, so it would be great to have a way to do this, even if we don't necessarily check any tooling in (depending on what you think).

rrrooommmaaa:
This PR dumps contents of `debug` IndexedDB entries.
These entries are supposed to be populated by `Debug.addMessage` calls in various test branches.
close #3482

@tomholub I tried adding this line
```
Debug.addMessage({ origText }).then(undefined, undefined);
```
inside `MsgBlockParser.detectBlocks` (as it is synchronous) and the input was successfully retrieved from the database upon test completion.
But is there a slight risk that we're trying to dump before completion of previous database transaction in this case?

### DIFF
--- a/extension/js/common/platform/debug.ts
+++ b/extension/js/common/platform/debug.ts
@@ -1,0 +1,51 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
+
+'use strict';
+
+import { AbstractStore } from './store/abstract-store.js';
+
+export class Debug {
+  public static readDatabase = async (): Promise<any[] | undefined> => {
+    const db = await Debug.openDatabase();
+    const records: any[] = [];
+    const tx = db.transaction(['messages'], 'readwrite');
+    await new Promise((resolve, reject) => {
+      tx.oncomplete = () => resolve(undefined);
+      tx.onerror = () => reject(AbstractStore.errCategorize(tx.error));
+      const messages = tx.objectStore('messages');
+      const search = messages.getAll(undefined);
+      search.onsuccess = () => {
+        records.push(...search.result);
+        messages.clear();
+      }
+    });
+    return records;
+  }
+
+  public static addMessage = async (message: any): Promise<void> => {
+    const db = await Debug.openDatabase();
+    const tx = db.transaction(['messages'], 'readwrite');
+    await new Promise((resolve, reject) => {
+      tx.oncomplete = () => resolve(undefined);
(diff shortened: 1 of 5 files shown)


tomholub (dismissed):
Very nice!

Your question about race conditions can be resolved by storing both input and output of a method as a single message. I think also easier to parse.

```ts
fun someMethod(input1, input2) {
  const output = someMethodORIGINAL(input1, input2);
  Debug.addMessage({ input: {input1, input2}, output }).catch(Catcher.reportErr);
  return output;
}
```

extension/js/common/platform/debug.ts
```diff
@@ -0,0 +1,51 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
+
+'use strict';
+
```
tomholub, line 4: Please add a comment here about what this can be used for and how - eg explain how this can be used to extract input/output of methods during tests. Then also add a note in repo readme to point to this file/comment as well. Thanks!

tomholub (dismissed):

rrrooommmaaa:
Looks like the extra open tab affects gmail setup test, We can of course open the tab afterwards, but I'm wandering if there is a bug in `doAwaitTriggeredPage` method or somewhere in the test infrastructure.

tomholub:
Probably easiest to close the tab with `finally`. But if you want to dig deeper, feel free.

rrrooommmaaa:
ok, let's just open the page to retrieve the database after the test.
It's not clear to me why the new page opens instead of previously open tab, I think spending hours to figure that out is not necessary right now,

extension/js/common/platform/debug.ts
rrrooommmaaa: Done

tomholub (approved):

PR #3551, opened 2021-04-11
Added comments for ContactStore save and update methods

### LINKED ISSUE #3536: unify ContactStore save and update
This PR makes no difference between `save` and `update` operations. I left them untouched so we won't need to update a lot of client code in this massive PR.

_Originally posted by @rrrooommmaaa in https://github.com/FlowCrypt/flowcrypt-browser/pull/3445#discussion_r605383023_

rrrooommmaaa:
This PR adds description to `save` and `update` methods of `ContactStore`.

close #3536

### To be filled by reviewers

### DIFF
--- a/extension/js/common/platform/store/contact-store.ts
+++ b/extension/js/common/platform/store/contact-store.ts
@@ -141,7 +141,17 @@ export class ContactStore extends AbstractStore {
   }
 
   /**
-   * Used to save a contact that does not yet exist
+   * Used to save a contact or an array of contacts.
+   * An underlying update operation is used for each of the provided contact.
+   * Properties `name` and `lastUse` are overwritten if the contact already exists,
+   * whereas `pubkey` and `pubkeyLastCheck` are added/updated as described in the `update` remarks.
+   *
+   * @param {IDBDatabase} db                     (optional) database to use
+   * @param {Contact | Contact[]} email            a single contact or an array of contacts
+   * @returns {Promise<void>}
+   *
+   * @async
+   * @static
    */
   public static save = async (db: IDBDatabase | undefined, contact: Contact | Contact[]): Promise<void> => {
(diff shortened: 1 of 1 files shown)


extension/js/common/platform/store/contact-store.ts
```diff
@@ -156,7 +166,24 @@ export class ContactStore extends AbstractStore {
...
-   * used to update existing contact
+   * Used to update certain fields of existing contacts or create new contacts using the provided data.
+   * If an array of emails is provided, the update operation will be performed independently on each of them.
+   * Missing properties from update object will not be overwritten in the database,
+   * e.g. `{name} as ContactUpdate` will retain the `lastUse` value of the existing record,
+   * whereas `{lastUse:null} as ContactUpdate` will reset `lastUse` to null.
+   * The `pubkey` property will be used only to add or update a pubkey record by the pubkey fingerprint.
+   * Null value of `pubkey` won't affect any pubkey records.
+   * The `pubkeyLastCheck` property can be set to a non-null value only when `pubkey` specified
+   * and will be applied only to that specific pubkey record. Missing or null `pubkeyLastCheck` will
```
rrrooommmaaa, line 177: There is a bit inconsistent behaviour. We can reset `name` and `lastUse` properties of the contact by providing null values in the `update` parameter, whereas `pubkeyLastCheck` will be ignored if `null` is provided. It's impossible to reset it. Should we do property existence check on it as with other properties or leave as it is?

tomholub: Name and lastUse should not be ever reset (no point setting name from something to null, and no point setting lastUse from something to null either). Thanks for noticing!

tomholub (approved):
I merged this one in - behavior change can be done in a followup PR.

PR #3561, opened 2021-04-14
Wkd.lookupEmail returns several keys

### LINKED ISSUE #3018: process / update revoked and expired keys from WKD response

rrrooommmaaa:
This PR processes multiple keys from Wkd and some other services

close #3018

- Tests added or updated TODO:

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -102,20 +102,35 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
-          this.ksLookupsByEmail[email] = ksContact.pubkey;
+        // save multiple pubkeys as separate operations
+        // todo: add a convenient method to storage?
+        const updates: ContactUpdate[] = [];
+        if (!pubkeys.length) {
+          if (name) {
+            // update just name
+            updates.push({ name } as ContactUpdate);
+          } else {
+            return ContactStore.obj({ email });
```
rrrooommmaaa, line 123: Do we need to save this object (containing only the `email` key) to the store?
`name` is null, `pubkey` is null, lastCheck is now set per pubkey, so it can't be set either.

tomholub: There is benefit to having email addresses of contacts in the local store even if no public key is associated with them. When looking up contacts next time, these contacts will show up in gray and I may want to still send email to them (password encrypted instead of public key encrypted).

If at the point this line executes the contact may email not be saved into storage yet, then it's good to save it.

If at this point the contact is always already stored in storage (because whichever method calls this always already did it), then you could skip this.

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -102,20 +102,35 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
-          this.ksLookupsByEmail[email] = ksContact.pubkey;
+        // save multiple pubkeys as separate operations
+        // todo: add a convenient method to storage?
+        const updates: ContactUpdate[] = [];
+        if (!pubkeys.length) {
+          if (name) {
+            // update just name
+            updates.push({ name } as ContactUpdate);
+          } else {
+            return await ContactStore.obj({ email });
```
tomholub, line 123: Is this return here still to be addressed? It looks like returning here would cause none of the pubkeys to be saved below (this is probably why it's a draft - just checking).

extension/chrome/settings/setup/setup-render.ts
```diff
@@ -98,8 +98,8 @@ export class SetupRenderModule {
     } catch (e) {
       return await Settings.promptToRetry(e, Lang.setup.failedToCheckIfAcctUsesEncryption, () => this.renderSetupDialog());
     }
-    if (keyserverRes.pubkey) {
-      const pub = await KeyUtil.parse(keyserverRes.pubkey);
+    if (keyserverRes.pubkeys.length) {
+      const pub = await KeyUtil.parse(keyserverRes.pubkeys[0]); // todo: ?
       this.view.acctEmailAttesterPubId = pub.id;
```
tomholub, line 103: Yup - please make `acctEmailAttesterPubId` an array. Then somewhere here will be `acctEmailAttesterPubId !== prv.id` or something similar, to see if the public key is consistent. Please change that to check if `acctEmailAttesterPubId.includes(prv.id)`.

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: If true, than it would be great to find a way to test this, too. Currently tests are passing even without this object being stored.

tomholub (commented):
Looking good! Adding some comments.

extension/chrome/elements/compose-modules/compose-storage-module.ts
rrrooommmaaa: this code is inside
```
if (!pubkeys.length) {
```
so no pubkeys are available here.
I'll save the dummy object {email} as per your request.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -232,6 +232,7 @@ export class OpenPGPKey {
         curve: (algoInfo as any).curve as string | undefined,
         algorithmId: opgp.enums.publicKey[algoInfo.algorithm]
       },
+      revoked: keyWithoutWeakPackets.revocationSignatures.length > 0
```
rrrooommmaaa, line 235: So this is the flag I was talking about. It's easily calculated when creating the `Key` structure.
So we don't really need to keep it in the database, do we?
The main question is: how should we prevent an `update` operation which replaces a revoked key with its undrevoked (outdated) version?

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: Got it. I think my original code was a bit of a spaghetti incident :) it's hard to follow which method has what responsibility and why, at least from reviewing the code. Your changes are good, I think it's my own code that confused me. A refactor of these methods in general may be nice, but that would be for another day.

Some of this complexity was for a reason (like responsiveness of UI as things are getting loaded from various, often remote places) and not all of these reasons were captured in tests, particularly UI responsiveness. Therefore refactoring this may not be easy, either.

Maybe a comment would help here - around that return:

> No public key found. Returning early, nothing to update in local store below.

Thanks!

extension/js/common/core/crypto/pgp/openpgp-key.ts
tomholub: > So this is the flag I was talking about. It's easily calculated when creating the Key structure.
> So we don't really need to keep it in the database, do we?

Ah. My intuition was to update the schema to add this flag to storage. Then filter based on that when pulling keys out of storage.

What you say is that you pull all keys for that user, then parse them and choose the appropriate key to use based on the parsed information. Your approach is simpler, and therefore better.

>  The main question is: how should we prevent an update operation which replaces a revoked key with its unrevoked (outdated) version?

I tend to do this in two ways:

1) a sensible logic preventing unwanted behavior in high-level code, in particular:
 - when pulling from pubkey lookup (compose)
 - when refreshing public keys from remote sources (compose)
 - when verifying messages (does that also update keys? I know it TOFU fetches a key but not sure if it would attempt to update existing key?)
 - when importing pubkey sent by another user (`pgp_pubkey.htm`) 
 - when importing pubkeys in settings (`contacts.htm`)

All of the above places would have an if to handle this situation and not update key that's already revoked. Often this will result in a ui modal or toast message to the user if it was user-initiated action. In situations when this was an automatic action (like compose window lookups after user enters recipient), the UI interaction could be skipped and the update would be skipped silently (still in high level code).

2) throw an error in low level function, in case we missed any edge cases. In this case, `save` and `update` function would check if existing stored key is revoked. If it is, it will `throw new Error("Wrongly attempted to replace a local revoked pubkey with non-revoked version")`

If the error ever throws, it will get reported to the backend, and I'll get to know about it. Then we'll have the stack trace to know which case we missed to handle in high level code.

-------------

I don't know if I'm over doing it. We could just silently skip updating in low level code and don't tell the user. But that could lead to some behavior that may appear buggy to the user.

rrrooommmaaa: > Ah. My intuition was to update the schema to add this flag to storage. Then filter based on that when pulling keys out of storage.

Actually, to make a unified solution for both OpenPGP and S/MIME revocation (we do want to support CRL later on, don't we?), the easiest would be to add a flag, or keep a table of revoked keys' fingerprints.

tomholub: I'll let you choose an approach that you deem appropriate. I'd be ok to prevent updating revoked keys in a followup PR - the situation now is already better then before (it used to not be even able to import revoked keys, I think).

extension/chrome/elements/compose-modules/compose-storage-module.ts
rrrooommmaaa: Done

tomholub (approved):

PR #3609, opened 2021-05-05
Prevent msgId path traversal workaround

seisvelas:
Prevents directory traversal by using a more robust defense. This change uses a regex to prevent the characters `.`, `/`, and 
 `\` from appearing in msgId. There was already a test for directory traversal, so I've added an additional one that tests for the workaround.

close https://github.com/FlowCrypt/flowcrypt-security/issues/134

- Tests added or updated

### To be filled by reviewers


seisvelas:
Tests fail, but they do not seem related to this , so to confirm I reran the tests of the latest to commit to master (which at the time ran successfully), and they now fail the same way. So tests are broken. I'll create a separate issue for that.

seisvelas:
close https://github.com/FlowCrypt/flowcrypt-security/issues/134

PR #3612, opened 2021-05-05
Use official stylelint vscode plugin

limonte:
We are currently using the unofficial https://github.com/hex-ci/vscode-stylelint-plus plugin because the "autofix on save" feature.

Now, the "autofix on save" feature is available in the official extension, switching to it because it's maintained by the stylelint team.


tomholub:
Thank you! The tests are currently broken for other reason.

tomholub (approved):
Thanks!

PR #3631, opened 2021-05-11
Revocation handling #3579

### LINKED ISSUE #3579: Make sure the key remains revoked when trying to update with older version of it
I'll let you choose an approach that you deem appropriate. I'd be ok to prevent updating revoked keys in a followup PR - the situation now is already better then before (it used to not be even able to import revoked keys, I think).

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/3561#r616961032_

rrrooommmaaa:
This PR handles revocations.

Introduces a separate store to keep fingerprints of revoked keys (as well as some arbitrary additional info).
As OpenPGP and X.509 can potentially have same cryptographic material,
revocation of `abcdef` also "revokes" `abcdef-X509`, as well as revocation `abcdef-X509` also "revokes" `abcdef`.
Given that, it's not to safe to load OpenPGP key if there is revocation for its X.509 sibling, even setting `revoked` flag manually,
because it will be lost during armor/parse. So the storage simply doesn't return such keys as if they are missing in the storage.

close #3579

- Tests added or updated

### To be filled by reviewers


tomholub (commented):
Looks good :)

rrrooommmaaa:
Migration may be a bit slow (it loads and parses all the OpenPGP keys in a single batch), but it will only happen if the user migrated to new 'emails' and 'pubkeys' storagewith older version of the extension

tomholub (dismissed):
Thanks!

tomholub:
We have added a requirement for authentication on the backend where there wasn't any. Some mock tests were forwarding requests to real backend which no longer works. I'm updating those tests.

tomholub (approved):

tomholub (dismissed):

tomholub (approved):

PR #3643, opened 2021-05-17
Fix rendering attachments with double whitespaces in their filenames

### LINKED ISSUE #3642: Gmail's attachment element isn't being removed
There are 4 attachments in this message, 3 default elements are removed and the 4th one isn't:

image

limonte:
This PR fixes rendering attachments with double whitespaces in their filenames.

close #3642

`innerText` is the representation and should never be used as data:

image

---

Also fixed some small styling issues:

1. Expired subscription badge:

Before | After
-|- 
CleanShot 2021-05-17 at 13 16 38@2x | CleanShot 2021-05-17 at 13 16 57@2x

2. Attachment filenames with special chars like `Č`:

Before | After
-|- 
CleanShot 2021-05-17 at 15 08 29@2x | CleanShot 2021-05-17 at 15 08 53@2x

3. Fixed vertical alignments of attachments (missed this one in #3639)

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -504,7 +504,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
 
   private filterAttachments = (potentialMatches: JQueryEl | HTMLElement, regExp: RegExp) => {
     return $(potentialMatches).filter('span.aZo:visible, span.a5r:visible').find('span.aV3').filter(function () {
-      const name = this.innerText.trim();
+      const name = this.innerHTML.trim();
       return regExp.test(name);
     }).closest('span.aZo, span.a5r');
   }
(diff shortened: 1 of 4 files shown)


extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -504,7 +504,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
 
   private filterAttachments = (potentialMatches: JQueryEl | HTMLElement, regExp: RegExp) => {
     return $(potentialMatches).filter('span.aZo:visible, span.a5r:visible').find('span.aV3').filter(function () {
-      const name = this.innerText.trim();
+      const name = this.innerHTML.trim();
```
tomholub, line 507: I feel like this could break other attachments with special characters in them, like `a<b` which will get translated to `a&lt;b` ? Or maybe `a&b` etc?

limonte: Right. Apparently, jQuery's `.text()` returns what we need here:

image

tomholub (approved):
Thanks!

PR #3645, opened 2021-05-18
Add Github Action to monitor for email addresses in comments

seisvelas:
We want to be aware of accidentally posting users' email addresses in comments. This GitHub action monitors for that.

This has already been tested on https://github.com/FlowCrypt/node-subprocess, where it works successfully.

### To be filled by reviewers

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,17 @@
+on:
+  issue_comment:
+    types: [created, edited]
+  issues:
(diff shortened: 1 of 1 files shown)


tomholub (approved):

limonte:
Late comment: the filename `main.yml` is misleading, something like `check-for-emails.yml` would be better.

tomholub:
agree

PR #3653, opened 2021-05-19
Add guidance when forgot pass phrase in already set up extension

### LINKED ISSUE #3323: provide guidance when forgot pass phrase in already set up extension
We are getting frequent emails from users about lost pass phrases. There are two places where a person may run into this:

**1) when they already have an extension set up, and they have it set to only remember their pass phrase in session. The browser may forget their pass phrase. Then they for example want to view an encrypted message, or send out a signed message, and their pass phrase is needed - which they don't know. -> addressed in this issue**

_2) when they are setting up the browser extension and they don't remember the pass phrase during setup -> addressed in #681_

-------------------

We should add a `Lost your pass phrase?` link in the pass phrase entry modal that we show user when decrypting a message or when sending a signed message.

To see the modal, you can click "forget stored pass phrases" in:

image

Then try to open an encrypted message. You will see this:

image

Here, at the bottom let's add a `Lost pass phrase?` link that will show the following modal (which is a bit different than a similar modal on a related setup issue #681):

> **Do you have at least one other working device where you can still read your encrypted email?**
>
> **If yes**: open the working device and go to `FlowCrypt Settings` -> `Security` -> `Change Pass Phrase`. It will let you change it without knowing the previous one. When done, [reset FlowCrypt on this device](#) and use the new pass phrase during the recovery step when you set up FlowCrypt on this device again.
>
> **If no**: unfortunately, you will not be able to read previously encrypted emails regardless of what you do. You can [reset FlowCrypt on this device](#) and then click `Lost your pass phrase?` during recovery step.

limonte:
This PR provides guidance when forgot pass phrase in already set up extension.

close #3323

CleanShot 2021-05-19 at 16 16 33

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/passphrase.ts
+++ b/extension/chrome/elements/passphrase.ts
@@ -75,11 +75,11 @@ View.run(class PassphraseView extends View {
     $('.action_ok').click(this.setHandler(() => this.submitHandler()));
     $('#lost-pass-phrase').click(this.setHandler((el, ev) => {
       ev.preventDefault();
+      // tslint:disable-next-line:no-floating-promises
       Ui.modal.info(`
         <div style="text-align: initial">
           <strong>Do you have at least one other working device where
           you can still read your encrypted email?</strong>
-
           <p><strong>If yes:</strong> open the working device and go to
           <code>FlowCrypt Settings</code> > <code>Security</code> >
           <code>Change Pass Phrase</code>.<br>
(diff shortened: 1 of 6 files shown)


extension/chrome/elements/passphrase.ts
```diff
@@ -75,19 +75,18 @@ View.run(class PassphraseView extends View {
     $('.action_ok').click(this.setHandler(() => this.submitHandler()));
     $('#lost-pass-phrase').click(this.setHandler((el, ev) => {
       ev.preventDefault();
+      // tslint:disable-next-line:no-floating-promises
```
tomholub, line 78: While it's a minor thing here, you can fix it better by appending `.catch(Catch.reportErr)`

tomholub (approved):

PR #3656, opened 2021-05-19
don't require user to click to verify signature on first pubkey load

### LINKED ISSUE #3105: don't require user to click to verify signature on first pubkey load

rrrooommmaaa:
This PR verifies signature as soon as a pubkey is auto-loaded from lookup etc.

close #3105

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
+++ b/extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
@@ -63,7 +63,8 @@ export class PgpBlockViewSignatureModule {
   /**
    * don't have appropriate pubkey by longid in contacts
    */
-  private renderPgpSignatureCheckMissingPubkeyOptions = async (signerLongid: string, senderEmail: string, retryVerification?: () => Promise<VerifyRes | undefined>): Promise<VerifyRes | undefined> => {
+  private renderPgpSignatureCheckMissingPubkeyOptions = async (signerLongid: string, senderEmail: string,
+    retryVerification?: () => Promise<VerifyRes | undefined>): Promise<VerifyRes | undefined> => {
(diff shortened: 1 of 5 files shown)


tomholub (commented):
:+1:

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
```diff
@@ -41,41 +51,49 @@ export class PgpBlockViewSignatureModule {
...
-      $('#pgp_signature > .cursive').remove();
-      $('#pgp_signature > .result').text('Message Not Signed');
+      if (signature) {
+        PgpBlockViewSignatureModule.setSigner(signature);
+      }
+      this.view.renderModule.doNotSetStateAsReadyYet = false;
+      Ui.setTestState('ready');
     }
-    $('#pgp_signature').css('block');
+    // $('#pgp_signature').css('block');
```
rrrooommmaaa, line 60: I don't really understand what this line was for. Can we remove it?

tomholub: Looks to be a buggy line. It should say `.css('display', 'block')` but if it worked anyway, I'd remove it as you say.

tomholub (approved):

PR #3658, opened 2021-05-20
Display revoked status of contact

### LINKED ISSUE #3636: Correctly display revoked key in compose-recipients-module
That seems ok to me. Then when user attempts to email to the revoked key, we should ensure the recipient field shows orange color (like when the key is expired) and when user hovers over it, it should show the available they was revoked.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/3579#issuecomment-841273770_

rrrooommmaaa:
This PR displays revoked status of `Contact`

close #3636

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
@@ -59,7 +59,10 @@ export class ComposePwdOrPubkeyContainerModule extends ViewModule<ComposeView> {
       this.hideMsgPwdUi(); // Hide 'Add Pasword' prompt if there are no recipients or message is not encrypted
       this.view.sendBtnModule.enableBtn();
     } else if (this.view.recipientsModule.getRecipients().find(r => [RecipientStatus.NO_PGP, RecipientStatus.REVOKED].includes(r.status))) {
-      await this.showMsgPwdUiAndColorBtn().catch(Catch.reportErr);
+      await this.showMsgPwdUiAndColorBtn(
+        this.view.recipientsModule.getRecipients().some(r => r.status === RecipientStatus.NO_PGP),
+        this.view.recipientsModule.getRecipients().some(r => r.status === RecipientStatus.REVOKED),
+      ).catch(Catch.reportErr);
     } else if (this.view.recipientsModule.getRecipients().find(r => [RecipientStatus.FAILED, RecipientStatus.WRONG].includes(r.status))) {
       this.view.S.now('send_btn_text').text(SendBtnTexts.BTN_WRONG_ENTRY);
       this.view.S.cached('send_btn').attr('title', 'Notice the recipients marked in red: please remove them and try to enter them egain.');
@@ -82,7 +85,7 @@ export class ComposePwdOrPubkeyContainerModule extends ViewModule<ComposeView> {
     return !this.view.S.cached('password_or_pubkey').is(':hidden');
   }
 
-  private showMsgPwdUiAndColorBtn = async () => {
+  private showMsgPwdUiAndColorBtn = async (anyNopgp: boolean, anyRevoked: boolean) => {
     if (!this.isVisible()) {
       const authInfo = await AcctStore.authInfo(this.view.acctEmail);
       const expirationTextEl = this.view.S.cached('expiration_note').find('#expiration_note_message_expire');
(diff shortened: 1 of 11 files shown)


tomholub (approved):
Thanks!

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -833,6 +833,12 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       Xss.sanitizePrepend(el, '<img src="/img/svgs/expired-timer.svg" class="expired-time">');
       $(el).attr('title', 'Does use encryption but their public key is expired. You should ask them to send ' +
         'you an updated public key.' + this.recipientKeyIdText(contact));
+    } else if (contact.revoked) {
+      recipient.status = RecipientStatus.REVOKED;
+      $(el).addClass("revoked");
+      Xss.sanitizePrepend(el, '<img src="/img/svgs/expired-timer.svg" class="expired-time">');
```
rrrooommmaaa, line 839: Should we have some other image here for revoked key?

extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
```diff
@@ -58,7 +58,7 @@ export class ComposePwdOrPubkeyContainerModule extends ViewModule<ComposeView> {
     if (!this.view.recipientsModule.getRecipients().length || !this.view.sendBtnModule.popover.choices.encrypt) {
       this.hideMsgPwdUi(); // Hide 'Add Pasword' prompt if there are no recipients or message is not encrypted
       this.view.sendBtnModule.enableBtn();
-    } else if (this.view.recipientsModule.getRecipients().find(r => r.status === RecipientStatus.NO_PGP)) {
+    } else if (this.view.recipientsModule.getRecipients().find(r => [RecipientStatus.NO_PGP, RecipientStatus.REVOKED].includes(r.status))) {
```
rrrooommmaaa, line 61: As revoked key makes encryption impossible (as opposed to expired key), we should affect sending button somehow, but the wording doesn't match with this change.
image
here it says "in grey":
image
Also, shouldn't the button be disabled if none of the recipients is PGP-capable?

tomholub: If none are pgp-capable, we still give the option to send password protected (the last sentence on the screenshot)

> As revoked key makes encryption impossible (as opposed to expired key), we should affect sending button somehow, but the wording doesn't match with this change.

Please change the wording as follows:

```
[emails (in grey) don't use encryption] [emails (in orange) have expired or revoked key]
Add their Public Key [or ask them to get FlowCrypt.]
Or send email password protected: _________
````

Where parts in square brackets are showing conditionally:
 - `emails (in grey) don't use encryption` -> show only if any NO_PGP present
 - `emails (in orange) have expired or revoked key` -> show only if any revoked keys present (I'm ignoring expired keys in the logic but still mentioning here because there may be mixed recipients)
 - `or ask them to get FlowCrypt.` -> show only if any NO_PGP present

Thank you for noticing this!

extension/chrome/elements/compose-modules/compose-recipients-module.ts
tomholub: @limonte could you add a similar icon into svgs that has maybe a stop sign? Or exclamation mark?

tomholub: I've made a separate issue - we can fix the icon separately afterwards. Thanks for noticing.

tomholub: The icon has been added in #3662

extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
rrrooommmaaa: Done.
Also, I noticed that class="no_pgp" has slightly different grey color (#989898) as opposed to "grey" class (#b7b7b7).
Should we make they same?

tomholub: Please choose the darker of the two. Thanks!

tomholub (dismissed):

tomholub (approved):

PR #3661, opened 2021-05-21
Add guidance when forgot pass phrase during setup

### LINKED ISSUE #681: provide guidance when missing pass phrase during setup
We are getting frequent emails from users about lost pass phrases. There are two places where a person may run into this:

_1) when they already have an extension set up, and they have it set to only remember their pass phrase in session. The browser may forget their pass phrase. Then they for example want to view an encrypted message, or send out a signed message, and their pass phrase is needed - which they don't know. This situation is addressed in issue #3323_

**2) when they are setting up the browser extension and they don't remember the pass phrase during setup -> addressed in this issue**

--------------------------------------------------------------

During setup, when my backup was found in inbox, I'm expected to enter a pass phrase.

image

 Current behavior is that if I enter a wrong pass phrase, the following link will show up:

image

That is problematic in two ways - it's not always the right advice, and the user may not even try to put any pass phrase in.

-------------------------------------------------------------

Please do the following:
 - [ ] remove the mechanism which shows this link
 - [ ] instead add a link that will always be visible that says `Lost your pass phrase?`
 - [ ] show modal text as below

> **Do you have at least one working device where you can still read your encrypted email?**
>
> **If yes**: open the working device and go to `FlowCrypt Settings` -> `Security` -> `Change Pass Phrase`. It will let you change it without knowing the previous one. When done, [reload this page](#) and use the new pass phrase.
>
> **If no**: unfortunately, you will not be able to read previously encrypted emails regardless of what you do. You can [skip recovery and create a new key instead](#). Your previous encrypted emails will remain unreadable.

The `reload this page` link should just call `window.reload()` which should end up re-loading keys from the inbox. After the page loads again, they should be facing the pass phrase input again. The `skip recovery and create a new key instead` link will do the same thing as the original link - show a modal to warn/confirm, which can be clicked to skip this step.

limonte:
This PR provides guidance when user forgot pass phrase during setup.

close #681

https://user-images.githubusercontent.com/6059356/119109385-cb2bc180-ba29-11eb-8d7f-c0e1cd33fbfb.mp4

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/passphrase.ts
+++ b/extension/chrome/elements/passphrase.ts
@@ -89,8 +89,7 @@ View.run(class PassphraseView extends View {
           <p><strong>If no:</strong> unfortunately, you will not be able to read
           previously encrypted emails regardless of what you do.
           You can <a href class="reset-flowcrypt">reset FlowCrypt on this device</a>
-          and then click <code>Skip recovery and set up encryption again</code>
-          during recovery step.
+          and then click <code>Lost your pass phrase?</code> during recovery step.
         </div>
       `, true).catch(Catch.reportErr);
       $('.reset-flowcrypt').click(this.setHandler(async (el, ev) => {
--- a/extension/chrome/settings/setup.htm
+++ b/extension/chrome/settings/setup.htm
@@ -238,9 +238,8 @@
       <div class="wide line">
         <button class="button green longer action_recover_account" data-test="action-recover-account">RECOVER ACCOUNT</button>
       </div>
-      <div class="line line_skip_recovery">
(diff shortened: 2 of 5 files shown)


tomholub:
Thank you for the video - excellent! :-)

limonte:
Video for the case when user has more than one key:

https://user-images.githubusercontent.com/6059356/119110998-522d6980-ba2b-11eb-95f4-65ef308f174f.mp4

limonte:
ping @seisvelas

seisvelas (approved):

PR #3671, opened 2021-05-23
Disable the 'Send Options' button when disabling the 'Send' button

### LINKED ISSUE #3650: The "Send Options" button should be disabled when the "Send" button is disabled
Because otherwise it creates unexpected behaviors:

https://user-images.githubusercontent.com/6059356/118666890-31cb9800-b7fc-11eb-9f94-f02cc2541697.mp4

While fixing this also make sure that `<button disabled>` doesn't have `cursor: pointer`

limonte:
This PR disables the 'Send Options' button when disabling the 'Send' button in order to prevent unexpected behaviors

close #3650

---

Also changed `autocomplete="false"` to `autocomplete="off"`, because `autocomplete` [is not a boolean attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautocomplete) and was not working in Chrome:

CleanShot 2021-05-23 at 12 24 13@2x

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-send-btn-module.ts
@@ -66,12 +66,12 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
 
   public disableBtn = () => {
     this.view.S.cached('send_btn').removeClass('green').addClass('gray').prop('disabled', true);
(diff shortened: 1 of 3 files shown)


tomholub:
That autocomplete was bothering me like forever! :-)

tomholub (approved):

Issue #3672, opened 2021-05-23
use FES to upload password protected messages when available


tomholub:
part of #2125 

- [x] remove unused API calls `messageExpiration` and `messageLink` https://github.com/FlowCrypt/flowcrypt-browser/pull/3673
- [x] remove unused API path `message/expiration` on backend https://github.com/FlowCrypt/flowcrypt-backend/issues/370
- [x] stop saving `admin_code` https://github.com/FlowCrypt/flowcrypt-browser/pull/3673
- [x] update code to stop using `short` instead use `url` response from backend https://github.com/FlowCrypt/flowcrypt-browser/pull/3673
- [x] design `message/upload` and `message/reply-token` on FES https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/issues/1269
- [ ] update following two methods to use FES when available. FES API is at https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/issues/1269

```
  public messageUpload = async (fcAuth: FcUuidAuth | undefined, encryptedDataBinary: Uint8Array, progressCb: ProgressCb): Promise<BackendRes.FcMsgUpload> => {
    return await FlowCryptComApi.messageUpload(fcAuth, encryptedDataBinary, progressCb);
  }

  public messageToken = async (fcAuth: FcUuidAuth): Promise<BackendRes.FcMsgToken> => {
    return await FlowCryptComApi.messageToken(fcAuth);
  }
```

PR #3673, opened 2021-05-23
issue 3672 cleanup for web portal

tomholub:
This PR cleans up deprecated functionality.

- [x] remove unused API calls `messageExpiration` and `messageLink` (deprecated / unused)
- [x] stop saving `admin_code` (deprecated / unused)
- [x] ~update code to stop using `short`, instead use `url` response field from backend~ compute the `url` field using returned `short` in the API client method itself

The actual backend already returns the `url` parameter that we newly use here.

issue #3672

- Tests added or updated (mock updated to reflect changes, although from user perspective there are no changes in behavior)

### To be filled by reviewers

### DIFF
--- a/extension/js/common/api/account-servers/flowcrypt-com-api.ts
+++ b/extension/js/common/api/account-servers/flowcrypt-com-api.ts
@@ -66,7 +66,13 @@ export class FlowCryptComApi extends Api {
 
   public static messageUpload = async (fcAuth: FcUuidAuth | undefined, encryptedDataBinary: Uint8Array, progressCb: ProgressCb): Promise<BackendRes.FcMsgUpload> => {
     const content = new Attachment({ name: 'cryptup_encrypted_message.asc', type: 'text/plain', data: encryptedDataBinary });
-    return await FlowCryptComApi.request<BackendRes.FcMsgUpload>('message/upload', { content, ...(fcAuth || {}) }, 'FORM', undefined, { upload: progressCb });
+    const rawResponse = await FlowCryptComApi.request<{ short: string }>('message/upload', { content, ...(fcAuth || {}) }, 'FORM', undefined, { upload: progressCb });
+    if (!rawResponse.short) {
+      throw new Error('Unexpectedly missing message upload short id');
+    }
+    // careful - this API request returns `url` as well, but that is URL of the S3 object, not of web portal page
+    // therefore we are constructing URL ourselves to point to web portal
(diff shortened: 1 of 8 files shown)


tomholub:
I may have overlooked what exactly `url` contains on backend - need to verify.

tomholub:
`url` there didn't contain url of the target web page. It contained URL of the object on S3, which is wrong. Need to update + do manual testing.

extension/js/common/api/account-servers/flowcrypt-com-api.ts
```diff
@@ -66,29 +66,20 @@ export class FlowCryptComApi extends Api {
...
   public static messageUpload = async (fcAuth: FcUuidAuth | undefined, encryptedDataBinary: Uint8Array, progressCb: ProgressCb): Promise<BackendRes.FcMsgUpload> => {
     const content = new Attachment({ name: 'cryptup_encrypted_message.asc', type: 'text/plain', data: encryptedDataBinary });
-    return await FlowCryptComApi.request<BackendRes.FcMsgUpload>('message/upload', { content, ...(fcAuth || {}) }, 'FORM', undefined, { upload: progressCb });
+    const rawResponse = await FlowCryptComApi.request<{ short: string }>('message/upload', { content, ...(fcAuth || {}) }, 'FORM', undefined, { upload: progressCb });
+    if (!rawResponse.short) {
+      throw new Error('Unexpectedly missing message upload short id');
+    }
+    // careful - this API request returns `url` as well, but that is URL of the S3 object, not of web portal page
+    // therefore we are constructing URL ourselves to point to web portal
+    return { url: `https://flowcrypt.com/${rawResponse.short}` };
```
rrrooommmaaa, line 75: I'm a bit concerned about forming an URL like this.
Shouldn't we percent-encode RFC 3986 section 2.2 Reserved Characters?
Can any of these be contained in `short`?

rrrooommmaaa (commented):
This PR is functionally relevant to the older code. Should we pay any attention to percent-encoding of URL parameter?

extension/js/common/api/account-servers/flowcrypt-com-api.ts
tomholub: Thank you for noticing. The `short` is returned by our own backend, and contains strictly 10 alphanumeric characters. So for example https://flowcrypt.com/abcde01234

rrrooommmaaa (approved):
LGTM!

Issue #3678, opened 2021-05-24
The "success notif after setup, click hides it, does not re-appear + offers to reauth" Gmail test is failing


limonte:
In `master` branch: https://flowcrypt.semaphoreci.com/jobs/c341c1de-1327-482f-9b41-7a9b24e25472

And dependabot's PRs are red because of this.

tomholub:
That would be for me to take a look.

tomholub:
@limonte would you be able to debug it? I may not be able to do it until next week

limonte:
> @limonte would you be able to debug it? I may not be able to do it until next week

Will do tomorrow 👍

PR #3680, opened 2021-05-24
Get rid of #cryptup_dialog

### LINKED ISSUE #3677: #cryptup_dialog should close itself by Esc
Currently `#cryptup_dialog` doesn't close itself when user press `Esc`. Instead it allows `Esc` to propagate, which create the unexpected behavior, e.g. closing the "Secure Compose" window while `#cryptup_dialog` stays opened.

### Steps to reproduce

1. Start wiring the password-protected message
2. Click the "Add their public key" link
3. Press `Esc`

### Actual behavior

`#cryptup_dialog` stays opened while the "Secure Compose" window is getting closed.

### Expected behavior

`#cryptup_dialog` is getting closed while the "Secure Compose" window stays opened (stop event propagation).

limonte:
This PR removes a few instances of `#cryptup_dialog` usage and replaces them with already existed functionality of showing `<iframe>` inside SweetAlert2 modal.

close #3677

---

Also fixed the `.action_finish_session` button. It was unusable when a user has a long list of emails (the button was covered by email rows, their `z-index` is `2`):

CleanShot 2021-05-24 at 22 39 43

---

Also fixed the vertical alignment in `.button`, e.g. in `decrypt.htm`:

Before | After
-|-
CleanShot 2021-05-26 at 13 26 59@2x | CleanShot 2021-05-26 at 13 28 38@2x

### To be filled by reviewers

### DIFF
--- a/extension/js/common/xss-safe-factory.ts
+++ b/extension/js/common/xss-safe-factory.ts
@@ -16,7 +16,6 @@ import { Ui } from './browser/ui.js';
 import { WebMailName } from './browser/env.js';
 import { Xss } from './platform/xss.js';
 import { SendAsAlias } from './platform/store/acct-store.js';
-import Swal from 'sweetalert2';
 
 type Placement = 'settings' | 'settings_compose' | 'default' | 'dialog' | 'gmail' | 'embedded' | 'compose';
 export type WebmailVariantString = undefined | 'html' | 'standard' | 'new';
@@ -178,7 +177,7 @@ export class XssSafeFactory {
 
   public showPassphraseDialog = async (longids: string[], type: PassphraseDialogType, initiatorFrameId?: string) => {
     const result = await Ui.modal.iframe(this.srcPassphraseDialog(longids, type, initiatorFrameId), 500, 'dialog-passphrase');
-    if (result.dismiss === Swal.DismissReason.backdrop || result.dismiss === Swal.DismissReason.esc) {
+    if (result.dismiss) { // dialog is dismissed by user interaction, not by closeDialog()
       BrowserMsg.send.passphraseEntry('broadcast', { entered: false });
     }
   }
--- a/extension/js/common/xss-safe-factory.ts
+++ b/extension/js/common/xss-safe-factory.ts
@@ -114,20 +114,27 @@ export class XssSafeFactory {
     return this.frameSrc(this.extUrl('chrome/elements/compose.htm'), { frameId: this.newId(), draftId });
   }
 
-  public srcPassphraseDialog = (longids: string[] = [], type: PassphraseDialogType, attachmentId?: string) => {
-    return this.frameSrc(this.extUrl('chrome/elements/passphrase.htm'), { type, longids, attachmentId });
+  public srcPassphraseDialog = (longids: string[] = [], type: PassphraseDialogType, initiatorFrameId?: string) => {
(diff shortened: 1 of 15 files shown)


limonte:
@tomholub ready for review. Hopefully I didn't break anything. I manually checked everything that could be related to replacing `#cryptup_dialog` with SweetAlert2.

limonte:
Rebased onto the `master` branch so the tests will pass after fixing #3678

BTW @tomholub I changed the style of the "End Pass Phrase Session" button to be less transparent and full-width. While less transparent is certainly better:

CleanShot 2021-05-26 at 11 58 18@2x

the full-width is opinionated, let me know if you like it or not. If not, I'll revert the width.

image

tomholub:
Less transparent is good and z-index is good, but I wonder if full-width wastes too much space?

limonte:
> Less transparent is good and z-index is good, but I wonder if full-width wastes too much space?

I actually agree here. It's easy to forget that a lot of users have 1366×768 screen resolution and it is important not to waste any vertical space. Reverted full-width.

extension/js/common/xss-safe-factory.ts
```diff
@@ -167,12 +176,15 @@ export class XssSafeFactory {
     return `<link class="${this.destroyableCls}" rel="stylesheet" href="${this.extUrl(`css/${file}.css`)}" />`;
   }
 
-  public dialogPassphrase = (longids: string[], type: PassphraseDialogType) => {
-    return this.divDialog_DANGEROUS(this.iframe(this.srcPassphraseDialog(longids, type), ['medium'], { scrolling: 'no' }), 'dialog-passphrase'); // xss-safe-factory
+  public showPassphraseDialog = async (longids: string[], type: PassphraseDialogType, initiatorFrameId?: string) => {
+    const result = await Ui.modal.iframe(this.srcPassphraseDialog(longids, type, initiatorFrameId), 500, 'dialog-passphrase');
+    if (result.dismiss === Swal.DismissReason.backdrop || result.dismiss === Swal.DismissReason.esc) {
```
rrrooommmaaa, line 181: If I understand correctly, this piece of code relies on some knowledge of how `Ui.modal.iframe` is implemented, namely
that `DismissReason.cancel` or `DismissReason.timer` are not possible?
Why don't we use `if (result.dismiss !== DismissReason.close) ` or simply return boolean from `Ui.modal.iframe()`?

rrrooommmaaa (changes requested):
Looks very neat to me, except for one thing that can be improved.

extension/js/common/xss-safe-factory.ts
limonte: > If I understand correctly, this piece of code relies on some knowledge of how `Ui.modal.iframe` is implemented, namely
> that `DismissReason.cancel` or `DismissReason.timer` are not possible?

Correct, we don't use timer or the SweetAlert2's cancel buttons in the passphrase dialog.

> Why don't we use `if (result.dismiss !== DismissReason.close) ` or simply return boolean from `Ui.modal.iframe()`?

Right, it can be as simple as checking `result.dismiss`. If a dialog was closed by `BrowserMsg.send.closeDialog()`, `result.dismiss` will be `undefined`. (docs: https://sweetalert2.github.io/#handling-dismissals)

rrrooommmaaa (approved):
LGTM!

PR #3693, opened 2021-05-27
Do not unnecessary trigger `evaluateRecipients()` for empty set of recipients

### LINKED ISSUE #3688: Saving draft is going into infinite loop after clicking on "Add their Public Key" link
While working on #3680 I noticed the infinite changing "Saved"/"Saving..." labels after opening the dialog for adding a pubkey:

https://user-images.githubusercontent.com/6059356/119659274-ad4dca80-be36-11eb-8541-c513c14f5d8c.mp4

limonte:
When the "Add a pubkey to email address" modal is opened, there's the `addedPubkeyDbLookupInterval` interval which calls `evaluateRecipients()`. Previously `evaluateRecipients()` was unnecessary called every second with an empty set of recipients which caused saving a draft every second.

This PR makes sure that `evaluateRecipients()` is executed once, when the contact is actually updated.

close #3688

- Not worth testing

### To be filled by reviewers


tomholub (approved):
Thanks!

PR #3719, opened 2021-06-03
Issue 3716 expiration note position

### LINKED ISSUE #3716: The "Recipient will have access to this message" block is mispositioned in secure reply
image

---

Also, while fixing this, fix also this: split this line into two lines or insert comma here:

image

limonte:
This PR fixes the position of `#expiration_note` block, as well as adds comma between `.warning_nopgp` and `. warning_revoked`.

close #3716

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-pwd-or-pubkey-container-module.ts
@@ -37,8 +37,7 @@ export class ComposePwdOrPubkeyContainerModule extends ViewModule<ComposeView> {
 
   public inputPwdFocusHandler = () => {
     const passwordContainerHeight = this.view.S.cached('password_or_pubkey').outerHeight() || 0;
-    const footerHeight = this.view.S.cached('footer').outerHeight() || 0;
-    this.view.S.cached('expiration_note').css({ bottom: (passwordContainerHeight + footerHeight) + 'px' });
+    this.view.S.cached('expiration_note').css({ bottom: passwordContainerHeight });
     this.view.S.cached('expiration_note').fadeIn();
     this.showHideContainerAndColorSendBtn(); // tslint:disable-line:no-floating-promises
   }
--- a/extension/css/cryptup.css
+++ b/extension/css/cryptup.css
@@ -2140,6 +2140,7 @@ table#compose tr#password_or_pubkey_container td {
   animation: fadeInUp 1s;
   border-top: 1px solid #cfcfcf;
   line-height: 1.5;
+  position: relative;
 }
 table#compose tr#password_or_pubkey_container td input#input_password::-webkit-input-placeholder { color: #d14836; }
 


extension/chrome/elements/compose.htm
```diff
@@ -126,7 +126,7 @@ <h1 id="header_title" data-test="header-title">New Secure Message</h1>
             (messages sent to FlowCrypt users have no limit, and don't need password exchange)
           </div>
           <div>
-            <div class="warning_nopgp" data-test="warning-nopgp">emails <span class="grey">in grey</span> don't use encryption </div>
+            <div class="warning_nopgp" data-test="warning-nopgp">emails <span class="grey">in grey</span> don't use encryption, </div>
```
rrrooommmaaa, line 129: How would it look if we have `warning_nopgp` visible and `warning_revoked` non-visible? The comma will be the terminating character in this situation. What is the simplest way to show the comma only if both `warning_nopgp` and `warning_revoked` are set? Can we do this with CSS or do we need to edit this code block?
```
    this.view.S.cached('warning_nopgp').css('display', anyNopgp ? 'inline-block' : 'none');
    this.view.S.cached('warning_revoked').css('display', anyRevoked ? 'inline-block' : 'none');
```

rrrooommmaaa (changes requested):
I suggest to gracefully handle the comma.

extension/chrome/elements/compose.htm
limonte: > How would it look if we have warning_nopgp visible and warning_revoked non-visible?

In which situation it's the case? As I understand correctly both are shown always, even if there are no orange recipients:

image

limonte: BTW @rrrooommmaaa I recommend installing the `htmlhint` extension which will help to prevent mistakes in HTML.

Right now, the markup has issues:

image

rrrooommmaaa: > In which situation it's the case? As I understand correctly both are always shown, even if there are no orange recipients

@limonte This is probably a bug, as the `warning_nopgp` is supposed to be shown if at least on recipient has NO_PGP status, whereas `warning_revoked` should be shown if at least one recipient has `REVOKED` status, as is implemented by this piece of code:
```
await this.showMsgPwdUiAndColorBtn(
        this.view.recipientsModule.getRecipients().some(r => r.status === RecipientStatus.NO_PGP),
        this.view.recipientsModule.getRecipients().some(r => r.status === RecipientStatus.REVOKED),
      )
```
And the comma is supposed to be shown only if both checks are true.

> There are two blocks with .warning_nopgp class, that should be fixed I believe

Yes, there should be two separate blocks turned on by `NO_PGP` as per this discussion
https://github.com/FlowCrypt/flowcrypt-browser/pull/3658#discussion_r636911000_

limonte: > This is probably a bug, as the warning_nopgp is supposed to be shown if at least on recipient has NO_PGP status, whereas warning_revoked should be shown if at least one recipient has REVOKED status

Reported the issue: #3726 

I removed my commit with the comma from this PR, let's handle that in #3726 as well.

rrrooommmaaa (approved):
LGTM!

PR #3763, opened 2021-06-20
Added feature to choose passphrase on import from EKM

### LINKED ISSUE #3596: OrgRule PASS_PHRASE_CHOSEN_BY_USER
In enterprise settings with EKM integration, FlowCrypt client retrieves private key decrypted from EKM (not pass phrase protected). Currently, the client requires that when `PRV_AUTOIMPORT_OR_AUTOGEN` OrgRule is used (meaning - pull keys from EKM during setup), that `PASS_PHRASE_QUIET_AUTOGEN` is also used together with it. This combination together created and stored pass phrase silently, this procedure can be seen in `setupWithEkmThenRenderSetupDone`.

We need a new OrgRule `PASS_PHRASE_CHOSEN_BY_USER` which will prompt the enterprise user to enter their own pass phrase during the setup when pulling keys from EKM.

Applicability and restrictions of `PASS_PHRASE_CHOSEN_BY_USER`:
 - only meaningful when combined with: `PRV_AUTOIMPORT_OR_AUTOGEN`. 
 - conflicts with `PASS_PHRASE_QUIET_AUTOGEN` (cannot be set simultaneously)

**Todo on docs repo:**
 - [ ] update https://flowcrypt.com/docs/business/org-rules.html

**Todo on browser repo:**
 - [x] (tom) meaningfuly refactor / classes / methods in a separate PR  #3670
 - [ ] update OrgRules class to understand this new rule, including restrictions. Add method `OrgRules.userMustChoosePassPhraseDuringPrvAutoimport()`. When the method is called:
       - check if `PASS_PHRASE_CHOSEN_BY_USER` present
       - if present, throw if  `PRV_AUTOIMPORT_OR_AUTOGEN` is **not** set at the same time
       - if present, throw if `PASS_PHRASE_QUIET_AUTOGEN` is set
       - return boolean
 - [ ] use this OrgRule during import when pulling from EKM in `setupWithEkmThenRenderSetupDone` to decide which flow to follow
 - [ ] reuse parts of `step_2a_manual_create` - just the `Choose pass phrase to protect your encrypted emails. choosing secure pass phrases` and two pass phrase fields and a button now called `continue` instead of `create and save`, to make a new block called `step_2_ekm_choose_pass_phrase`
 - [ ] connect the UI elements to pass phrase strength evaluation with `initPassphraseToggle`
 - [ ] add UI tests for this combination of OrgRules + this setup flow

rrrooommmaaa:
This PR adds possibility to choose a passphrase when getting a key from EKM

close #3596 

- Tests added or updated

### To be filled by reviewers


extension/chrome/settings/setup/setup-key-manager-autogen.ts
```diff
@@ -19,20 +18,21 @@ export class SetupWithEmailKeyManagerModule {
...
     }
-    const passphrase = PgpPwd.random(); // mustAutogenPassPhraseQuietly
-    const setupOptions: SetupOptions = { passphrase_save: true, submit_main: this.view.orgRules.canSubmitPubToAttester(), submit_all: false, passphrase };
+    const passphrase = $('#step_2_ekm_choose_pass_phrase .input_password').val();
+    await this.setupWithEkmThenRenderSetupDone(typeof passphrase === 'string' ? passphrase : '');
+  }
+
+  public setupWithEkmThenRenderSetupDone = async (passphrase: string) => {
+    const setupOptions: SetupOptions = {
+      passphrase_save: this.view.orgRules.mustAutogenPassPhraseQuietly() || Boolean($('#step_2_ekm_choose_pass_phrase .input_passphrase_save').prop('checked')),
```
rrrooommmaaa, line 31: Do we need a test to untick "Save Pass phrase" and check that the passphrase doesn't get in to the local storage?

extension/chrome/settings/setup/setup-render.ts
```diff
@@ -46,7 +48,17 @@ export class SetupRenderModule {
...
     } else if (this.view.orgRules.mustAutoImportOrAutogenPrvWithKeyManager()) {
-      await this.view.setupWithEmailKeyManager.setupWithEkmThenRenderSetupDone();
+      if (this.view.orgRules.mustAutogenPassPhraseQuietly() && this.view.orgRules.forbidStoringPassPhrase()) {
+        const notSupportedErr = 'Combination of org rules not valid: PASS_PHRASE_QUIET_AUTOGEN cannot be used together with FORBID_STORING_PASS_PHRASE.';
+        await Ui.modal.error(notSupportedErr);
+        window.location.href = Url.create('index.htm', { acctEmail: this.view.acctEmail });
+        return;
+      }
+      if (this.view.orgRules.userMustChoosePassPhraseDuringPrvAutoimport()) {
+        this.displayBlock('step_2_ekm_choose_pass_phrase');
```
rrrooommmaaa, line 58: There is small possibility that subsequent key retrieval will fail with a network error exception 
`FlowCrypt Email Key Manager at ... is down, please inform your network admin.` AFTER the passphrase is enter. 
I personally hate sites that make user enter some input too early so that the input will get lost under some circumstances, like intermittent network failures. However, moving passphrase entering to a later stage will considerably increase the code complexity. So are we going to go this way to improve UX, @tomholub ?

extension/chrome/settings/setup/setup-key-manager-autogen.ts
tomholub: When we are doing setup with EKM, I would not give the user any choice regarding pass phrase storage - instead we will infer from OrgRules. If pass phrase in memory is required then we use that, else we put it in storage without asking the user.

(big orgs usually want a consistent setup for everybody)

To anwer your question - that means there is no checkbox to uncheck.

extension/chrome/settings/setup/setup-render.ts
tomholub: I have a method that offers the user to retry the flaky step until the step succeeds, without losing variables like pass phrase.

You can see it in action here:

```ts
    const storePrvOnKm = async () => this.view.keyManager!.storePrivateKey(this.view.idToken!, KeyUtil.armor(decryptablePrv), pubArmor);
    await Settings.retryUntilSuccessful(storePrvOnKm, 'Failed to store newly generated key on FlowCrypt Email Key Manager');
```

This could allow you to ask for pass phrase early but still have good UX when things go wrong.

extension/chrome/settings/setup.htm
```diff
@@ -229,12 +223,34 @@ <h1></h1>
...
+      <div class="line left">
+        <input class="input_password2" type="text" placeholder="Repeat pass phrase" data-test="input-step2ekm-passphrase-2" id="step_2_ekm_input_password2" />
+      </div>
+
+      <div class="line left">
+        <label>
+          <input type="checkbox" class="input_passphrase_save" value="" name="check" data-test="input-step2ekm-save-passphrase" />
+          Remember Pass Phrase after closing browser
+        </label>
+      </div>
```
tomholub, line 242: Let's comment this out per another comment - in case we do want to bring it back in the future.

extension/js/common/org-rules.ts
```diff
@@ -156,6 +156,10 @@ export class OrgRules {
     return (this.domainRules.flags || []).includes('PASS_PHRASE_QUIET_AUTOGEN');
```
tomholub, line 156: Let's update this to

```ts
return this.usesKeyManager() && (this.domainRules.flags || []).includes('PASS_PHRASE_QUIET_AUTOGEN')`
```

extension/js/common/org-rules.ts
```diff
@@ -156,6 +156,10 @@ export class OrgRules {
...
   }
 
+  public userMustChoosePassPhraseDuringPrvAutoimport = (): boolean => {
+    return !(this.domainRules.flags || []).includes('PASS_PHRASE_QUIET_AUTOGEN');
```
tomholub, line 160: Let's update this to:

```ts
return this.usesKeyManager() && !this.mustAutogenPassPhraseQuietly()
```

tomholub (commented):
Looks good!

extension/chrome/settings/setup/setup-key-manager-autogen.ts
tomholub: Actually, since you already did this, it won't hurt to leave it as is.

At least it's more explicit.

I don't think it needs a distinct test where user unchecks it, looks ok to me.

extension/chrome/settings/setup/setup-render.ts
rrrooommmaaa: This is already in use in `autoGenerateKeyAndStoreBothLocallyAndToEkm()` stage.
The current implementation for the "get" stage failure `const { privateKeys } = await this.view.keyManager!.getPrivateKeys(this.view.idToken!);`
is to tell the user to "please inform your network admin." I see no point in retries here.
We'd better move the passphrase input screen after the successful completion of "get the key from ekm" stage.

tomholub (approved):
Thanks!

extension/chrome/settings/setup/setup-render.ts
tomholub: There typically is a point in retrying, but it depends on the error. Network issues, server down and 5xx codes are retriable - and I think these would be the most common.

Issue #3774, opened 2021-06-25
CSP prevents downloading of private key via FlowCrypt settings on Firefox


seisvelas:
ref: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGkXwGhDVtxsgDBqKQTLrpTdPTk

In Firefox, you can try downloading a private key via FlowCrypt Settings -> Additional Settings -> Public Key, or via the backup button on the same page - neither works, both are blocked by the CSP.

This is a bug. I should be able to fix this pretty quickly so people can conveniently download their private keys again.

Sorry!

(it works fine on Chrome btw)

Issue #3411, opened 2021-02-10
smoke tests on Firefox using Puppeteer?


tomholub:
I guess, the first thing to look into is doing a few tests on Firefox. Not the whole test suite, just checking it's not completely broken. Puppeteer I think supports this - but don't know if it supports extensions too. Will make a separate issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/2649#issuecomment-776486758_

I'm thinking just one test that logs in using a mock, and one test that logs in on live gmail and opens a message. Two tests total.

PR #3784, opened 2021-06-30
Add blob as valid frame-src to prevent Firefox key download bug

seisvelas:
Because of how Firefox interprets the frame-src CSP directive, users on Firefox are unable to download their private keys (and probably do other things). This weakens the directive to make this functionality work for FF users.

close https://github.com/FlowCrypt/flowcrypt-browser/issues/3774

- Difficult to test because we don't have Firefox specific tests (yet! https://github.com/FlowCrypt/flowcrypt-browser/issues/3411)

### To be filled by reviewers

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -91,5 +91,5 @@
     "/chrome/elements/shared/footer.htm"
   ],
   "minimum_chrome_version": "61",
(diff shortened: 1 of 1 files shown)


seisvelas:
I'm failing *all* of the live Gmail tests for some reason. But so is master, so that must be unrelated to this PR.

Do we already know why master is failing live Gmail tests, and if not, do we have an issue for it? Otherwise we should make one and look into that!

In any case, I'll consider this ready for review. In addition to passing the remaining tests (the ones that also pass on master), I also tested it manually to make sure it addresses the issue, and it does.

tomholub (approved):

tomholub:
Yes, the Gmail live tests will have to be fixed - for now I'm ignoring them.

Issue #3629, opened 2021-05-10
Gmail shortcuts request


seisvelas:
>I use a lot of the gmail keyboard shortcuts (like 'r' for reply). It would be nice if you offered some sort of similar thing (either overriding gmail's shortcuts or adding a new one).

ref: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgxwLtsvPJlpBVzkVgkbQdJsBhqbf

Is this desirable for us? Just bringing it up in case this is something we would ever want to pursue, since we received the request.

limonte:
Hey @encody and thank you for this request!

Most of shortcuts are already there, but we're still missing the docs. There's a task for me in another repo to create that page: https://github.com/FlowCrypt/flowcrypt-web/issues/262

Could you please provide list all the shortcuts that you're missing and then @tomholub will decide what to do next.

limonte:
Thank you @encody for the list of expected shortcuts.

@tomholub I would suggest to prioritize this issue because it's at the same time feature-request and bug-fix.

Here's the bug which is happening in Firefox only:



https://user-images.githubusercontent.com/6059356/120043053-eee79c80-c013-11eb-97ee-bdd7d8d6dd77.mp4




### Steps to reproduce:

1. Open any encrypted message
2. Press `r` two times

### Expected behavior

Encrypted reply with "r" in it.

### Actual behavior

Invisible unencrypted reply with "r" in it.

---

Proposed action steps:

1. In scope of this issue add support for `r` and `a` shourtcuts.
2. Create another issue if needed for "Secure Compose" action. Let's duscuss the shortcut, I don't think we should override Gmail's `c` shortcut.
3. Create another issue if needed for `?` shortcut.

tomholub:
thank you for looking into this - sounds good

limonte:
Blocked by #3789

PR #3789, opened 2021-07-03
[WIP] BrowserMsg behavior

limonte:
This PR is not aimed to be merged, its only purpose is to figure out if the `BrowserMsg` behavior is buggy or not.

I stumbled upon this issue where `BrowserMsg` works differently in Chrome and Firefox while working on https://github.com/FlowCrypt/flowcrypt-browser/issues/3629#issuecomment-850676829

### What I'm doing:

1. Sending test action from `webmail.ts`
2. Catching it in `setup-webmail-content-script.ts` along with other listeners
3. Catching it in `compose-render-module.ts`

### Expected behavior (let me know if I'm wrong with my expectations)

The test action is caught both in `setup-webmail-content-script.ts` and `compose-render-module.ts`

### Actual behavior

Behavior is different in Chrome and Firefox, and none of them matches my expectations.

### Actual behavior in Chrome

Test action is caught only in `compose-render-module.ts`:

```
➕ added test_action listener in setup-webmail-content-script.ts, tabId: 17:0
➕ added test_action listener in compose-render-module.ts, tabId: 17:0
↗️ test action sent to 17:0
✔️ caught in compose-render-module.ts
```

### Actual behavior in Firefox

Test action is caught only in `setup-webmail-content-script.ts`:

```
➕ added test_action listener in setup-webmail-content-script.ts, tabId: 1:0
➕ added test_action listener in compose-render-module.ts, tabId: 1:0
↗️ test action sent to 1:0
✔️ caught in setup-webmail-content-script.ts
```

@tomholub questions I need your help with:

- are my expectations correct? 
- is the difference in browsers known? If it is known, where can I read more about it?
- if the difference is unexpected, then we should probably fix it (I depend on this for fixing Gmails shortcuts https://github.com/FlowCrypt/flowcrypt-browser/issues/3629#issuecomment-850676829)


tomholub:
Sorry I still didn't get to this - when the question is difficult and I don't have enough brain juice left, I have a tendency to delay :-( 

It could be a bug in Firefox (or Chrome). But it could also be that our implementation is not ideal. I'll have to have a look once I can properly focus.

tomholub:
Thank you for the very clear issue demonstration

limonte:
@tomholub let me know if want me to try to fix BrowserMsg implementation. If so, then I'd need your opinion on the expected behavior. My expectations are in the PR description, I'm still not sure if they are correct or not.

tomholub:
Sorry for my continued silence here. @rrrooommmaaa could you take a look later? I didn't have a chance to properly evaluate it yet.

rrrooommmaaa:
First of all, we have separate instances of static handlers mapping `BrowserMsg.HANDLERS_REGISTERED_FRAME`.
I added a feature to chain two handlers and in running on both Firefox and Chrome there are two `adding handler to test_action` console messages.
```
➕ added test_action listener in setup-webmail-content-script.ts, tabId: 5:0 webmail_bundle.js:9844:17
adding handler to test_action webmail_bundle.js:6792:17
```
```
➕ added test_action listener in compose-render-module.ts, tabId: 5:0 compose-render-module.js:60:21
adding handler to test_action browser-msg.js:153:17
```
as stated in https://developer.chrome.com/docs/extensions/mv3/messaging/
in `chrome.runtime.sendMessage` "Note that extensions cannot send messages to content scripts using this method. To send messages to content scripts, use tabs.sendMessage." 
Browsers are indeed behaving differently, mostly because we're trying to use `chrome.runtime.sendMessage` in a way it's not designed to.
In Chrome, I even get 2 messages of 
(2) ✔️ caught in compose-render-module.ts
I can't explain this yet, but if needed we can rewrite this piece of code to use `tabs.sendMessage`, as advised

limonte:
Thank you for investigating this issue @rrrooommmaaa!

> In Chrome, I even get 2 messages of
> (2) ✔️ caught in compose-render-module.ts

Same here.

---

Before trying to fix the issue by replacing `chrome.runtime.sendMessage` with `tabs.sendMessage`, I would like to confirm what is the expected behavior because I'm still not sure about it. I'd expect that the test action can be caught both in `setup-webmail-content-script.ts` and `compose-render-module.ts`.

rrrooommmaaa:
As far as my understanding goes, `webmail_bundle.js` is part of what is called a "content script" (that is a "library" loaded to gmail page), and given that "extensions cannot send messages to content scripts using this method" from the docs, I guess indeterminate behaviour is not a bug. (Perhaps, Firefox has its own interpretation of what to do with trying to send a message to a content script -- the message is visible only within the content script).

rrrooommmaaa:
This was probably fixed with #4264

PR #3791, opened 2021-07-04
Passphrase expiration in session #3565

### LINKED ISSUE #3565: pass phrase in session to last maximum exactly 4 hours
If existing sessionStorage can be configured to have expiry, set it to auto-wipe 4 hours from storing.

If above is not possible, need our own mechanism that keeps pass phrase in memory, in one class instance in background page where each pass phrase expires in exactly 4 hours from entry - and this should be used instead of the browser session mechanism.

rrrooommmaaa:
This PR saves passphrase to expiration cache
(to expire after 4 hours)

close #3565

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/background_page/background_page.ts
+++ b/extension/js/background_page/background_page.ts
@@ -15,6 +15,7 @@ import { GlobalStoreDict, GlobalStore } from '../common/platform/store/global-st
 import { ContactStore } from '../common/platform/store/contact-store.js';
 import { SessionStore } from '../common/platform/store/session-store.js';
 import { AcctStore } from '../common/platform/store/acct-store.js';
+import { ExpirationCache } from '../common/core/expiration-cache.js';
 
 console.info('background_process.js starting');
 
@@ -24,6 +25,7 @@ opgp.initWorker({ path: '/lib/openpgp.worker.js' });
 
   let db: IDBDatabase;
   let storage: GlobalStoreDict;
+  const passphraseTracker = new ExpirationCache(4 * 60 * 60 * 1000); // 4 hours
 
   try {
     await migrateGlobal();
@@ -51,6 +53,8 @@ opgp.initWorker({ path: '/lib/openpgp.worker.js' });
 
   // storage related handlers
   BrowserMsg.bgAddListener('db', (r: Bm.Db) => BgHandlers.dbOperationHandler(db, r));
+  BrowserMsg.bgAddListener('session_passphrase_set', (r: Bm.StoreSessionSet) => passphraseTracker.set(r.key, r.value));
+  BrowserMsg.bgAddListener('session_passphrase_get', async (r: Bm.StoreSessionGet) => passphraseTracker.get(r.key));
   BrowserMsg.bgAddListener('session_set', (r: Bm.StoreSessionSet) => SessionStore.set(r.acctEmail, r.key, r.value));
   BrowserMsg.bgAddListener('session_get', (r: Bm.StoreSessionGet) => SessionStore.get(r.acctEmail, r.key));
   BrowserMsg.bgAddListener('storeGlobalGet', (r: Bm.StoreGlobalGet) => GlobalStore.get(r.keys));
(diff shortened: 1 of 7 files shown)


extension/js/background_page/background_page.ts
```diff
@@ -51,6 +53,8 @@ opgp.initWorker({ path: '/lib/openpgp.worker.js' });
 
   // storage related handlers
   BrowserMsg.bgAddListener('db', (r: Bm.Db) => BgHandlers.dbOperationHandler(db, r));
+  BrowserMsg.bgAddListener('session_passphrase_set', (r: Bm.StoreSessionSet) => passphraseTracker.set(r.key, r.value));
```
rrrooommmaaa, line 56: Should we use `acctEmail` as part of the key here?
I'm not fully acquainted with the concept of switching between accounts.
Should we forget (or make invisible) passphrases stored in this cache for another account?

extension/js/common/core/expiration-cache.ts
```diff
@@ -0,0 +1,32 @@
...
+  }
+
+  public get = (key: string): string | undefined => {
+    const found = this.cache[key];
+    if (found) {
+      if (found.expiration > Date.now()) {
+        return found.value;
+      } else {
+        // expired, so delete it and return as if not found
+        delete this.cache[key];
```
rrrooommmaaa, line 27: The simplest implementation. Entries are deleted on `get` request. We can also have a look at packages out there, like https://www.npmjs.com/package/expiring-cache

tomholub: Should additionally set up a setInterval that will go through them minute, for example. That way it's not actually kept in memory over the time limit.

extension/js/background_page/background_page.ts
tomholub: Since keys are associated with a particular account and invisible to other accounts, the same should be done for pass phrases - visible to only to a particular account.

extension/js/common/core/expiration-cache.ts
rrrooommmaaa: Done

extension/js/background_page/background_page.ts
rrrooommmaaa: Done

rrrooommmaaa:
Should I add some tests (it's not that easy to emulate 4 hours expiration)?
What should we do about this warning: `lgtm [js/clear-text-storage-of-sensitive-data]`?

tomholub:
Your code update has fixed a previous warning that was present on master. On master, we stored pass phrase in browser session, violating the LGTM security rule, and here you store it in memory, no longer violating it - no action needed here.

extension/js/common/core/common.ts
```diff
@@ -312,3 +312,7 @@ export class Url {
   }
 
 }
+
+export const emailKeyIndex = (scope: string, key: string): string => {
```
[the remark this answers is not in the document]
tomholub: Actually, even for the in memory implementation could use `singleScopeRawIndex` then this would not be needed to be separately here. If you'd like this method to be separate, I'd prefer that it's placed in AbstractStore if possible?

tomholub (commented):
Looks great. If we were to add tests, we could add unit tests of the expiring cache where we supply a short expiration and just test that class stand-alone.

extension/js/background_page/background_page.ts
```diff
@@ -26,6 +27,7 @@ opgp.initWorker({ path: '/lib/openpgp.worker.js' });
   let db: IDBDatabase;
   let storage: GlobalStoreDict;
   const passphraseTracker = new ExpirationCache(4 * 60 * 60 * 1000); // 4 hours
```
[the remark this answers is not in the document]
rrrooommmaaa: Done

extension/js/common/core/common.ts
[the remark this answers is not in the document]
rrrooommmaaa: AbstractStore is platform-dependent. It won't be possible to write a simple unit-node test, only browser tests

tomholub (approved):
Thanks!

Issue #3785, opened 2021-06-30
live Gmail tests failing


tomholub:
I'm failing *all* of the live Gmail tests for some reason. But so is master, so that must be unrelated to this PR.

Do we already know why master is failing live Gmail tests, and if not, do we have an issue for it? Otherwise we should make one and look into that!

In any case, I'll consider this ready for review. In addition to passing the remaining tests (the ones that also pass on master), I also tested it manually to make sure it addresses the issue, and it does.

_Originally posted by @seisvelas in https://github.com/FlowCrypt/flowcrypt-browser/issues/3784#issuecomment-871228524_

tomholub:
closing - live Gmail tests seem to be fine, it's other tests that are failing now

PR #3792, opened 2021-07-04
debug live Gmail tests failing

limonte:
This PR is not meant to be merged, it's the part of debugging #3785 

I don't remember where exactly, but some time ago I asked about taking screenshots from CI. I found the simple way to do that:

1. `console.log(await page.screenshot({ encoding: "base64" }))`
2. copy the output from CI log and paste it to https://codebeautify.org/base64-to-image-converter

---

As to debugging #3785 I think that `acctPassword` is not set for some reason, here's the screenshot of the state where email and password are supposed to be entered:

image

### DIFF
--- a/test/source/tests/gmail.ts
+++ b/test/source/tests/gmail.ts
@@ -68,7 +68,7 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       return gmailPage;
     };
 
-    ava.default('mail.google.com - setup prompt notif + hides when close clicked + reappears + setup link opens settings', testWithBrowser(undefined, async (t, browser) => {
+    ava.default('mail.google.com - setup prompt notif ++ hides when close clicked + reappears + setup link opens settings', testWithBrowser(undefined, async (t, browser) => {
       const settingsPage = await BrowserRecipe.openSettingsLoginButCloseOauthWindowBeforeGrantingPermission(t, browser, 'ci.tests.gmail@flowcrypt.dev');
       await settingsPage.close();
       let gmailPage = await BrowserRecipe.openGmailPage(t, browser);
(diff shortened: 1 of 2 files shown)


tomholub (commented):
Maybe the Auth0 login html has changed, no longer matching our code? Have you tried running one Gmail Live test locally?

limonte:
> Maybe the Auth0 login html has changed, no longer matching our code? Have you tried running one Gmail Live test locally?

Yes, they are working locally.

limonte:
> Maybe the Auth0 login html has changed, no longer matching our code

Our code isn't trying to fill the password field, this line https://github.com/FlowCrypt/flowcrypt-browser/pull/3792/files#diff-7f1d04918dc9bf29c051611e92209ee8476db1cdcbf6f6879a928ba01fa7e1ccR84 isn't being executed

tomholub:
Interesting!

tomholub:
I re-uploaded the secret file just in case, will now add an empty commit to re-run the tests

tomholub:
Thank you for your help debugging this. I must have pushed a bad secrets file to CI.

I've just made this correction
```diff
    "google": [
      {
-        "email": "ci.tests.gmail@flowcrypt.test",
+        "email": "ci.tests.gmail@flowcrypt.dev",
```

To match your screenshot. This should do.

tomholub:
>     console.log(await page.screenshot({ encoding: "base64" }))
>     copy the output from CI log and paste it to https://codebeautify.org/base64-to-image-converter
> 

Yes, that's an easy way to do it I didn't think about.

tomholub:
There are several problems, the secrets file was one.

Another one was that the previous selector `button` was actually clicking the show/hide button of the password, not the submit button.

tomholub (approved):
Approve if it passes tests

tomholub:
Now there is a third problem: Google captcha. I suppose I may not be able to do anything with that reliably.

tomholub:
Trying https://accounts.google.com/DisplayUnlockCaptcha but that may only work temporarily.

PR #3797, opened 2021-07-08
Disallow attester search for all domains when * as specified #3563

### LINKED ISSUE #3563: allow a wildcard * in disallow_attester_search_for_domains
Currently, we match domains exactly, eg `disallow_attester_search_for_domains: ["example.com"]`. I'd like to update this so that value `*` simply means disallow searching on all.

We don't have to support more complicated patterns like `*.domain.com` or `test.*.domain.com` - that's out of scope.

rrrooommmaaa:
This PR disallows attester search alltogether if one of the entries is '*'

close #3563

- Tests added or updated

### To be filled by reviewers

- [ ] is documented clearly and usefully, or doesn't need documentation -> issue for docs filed separately https://github.com/FlowCrypt/flowcrypt-docs/issues/161

### DIFF
--- a/extension/js/common/org-rules.ts
+++ b/extension/js/common/org-rules.ts
@@ -172,8 +172,15 @@ export class OrgRules {
    * This is because they already have other means to obtain public keys for these domains, such as from their own internal keyserver
    */
   public canLookupThisRecipientOnAttester = (emailAddr: string): boolean => {
-    const forbidden = ['*', Str.getDomainFromEmailAddress(emailAddr) || 'NONE'];
-    return !(this.domainRules.disallow_attester_search_for_domains || []).some(domain => forbidden.includes(domain));
+    const disallowedDomains = this.domainRules.disallow_attester_search_for_domains || [];
+    if (disallowedDomains.includes('*')) {
+      return false;
+    }
+    const userDomain = Str.getDomainFromEmailAddress(emailAddr);
+    if (!userDomain) {
+      throw new Error(`Not a valid email ${emailAddr}`);
+    }
+    return !disallowedDomains.includes(userDomain);
   }
 
   /**
--- a/extension/js/common/org-rules.ts
+++ b/extension/js/common/org-rules.ts
@@ -172,7 +172,8 @@ export class OrgRules {
    * This is because they already have other means to obtain public keys for these domains, such as from their own internal keyserver
    */
   public canLookupThisRecipientOnAttester = (emailAddr: string): boolean => {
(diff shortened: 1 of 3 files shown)


extension/js/common/org-rules.ts
```diff
@@ -172,7 +172,8 @@ export class OrgRules {
    * This is because they already have other means to obtain public keys for these domains, such as from their own internal keyserver
    */
   public canLookupThisRecipientOnAttester = (emailAddr: string): boolean => {
-    return !(this.domainRules.disallow_attester_search_for_domains || []).includes(Str.getDomainFromEmailAddress(emailAddr) || 'NONE');
+    const forbidden = ['*', Str.getDomainFromEmailAddress(emailAddr) || 'NONE'];
```
tomholub, line 175: I know it's a nitpick but could we dumb down the code a bit? :-) The original code was already not ideal but ok for the simple functionality, but now it's getting a bit harder for me to follow. For example:

```ts
  public canLookupThisRecipientOnAttester = (emailAddr: string): boolean => {
    const disallowedDomains = this.domainRules.disallow_attester_search_for_domains || [];
    if(disallowedDomains).includes('*')) {
      return false;
    }
    const userDomain = Str.getDomainFromEmailAddress(emailAddr);
    if(!userDomain) {
      ??? I wonder if we should throw here?
    }
    return !(disallowedDomains).includes(userDomain);
  }
```

It's more verbose, but easier to follow.

tomholub (commented):
Thank you! See an opportunity for added clarity below.

extension/js/common/org-rules.ts
rrrooommmaaa: Done

tomholub (approved):
Thank you!

rrrooommmaaa:
looks like it broke one of the tests

tomholub:
Yes, and I stupidly force-merged it thinking it was just the Gmail tests

rrrooommmaaa:
So, we actually stumbled on a situation when `lookupFingerprint()` is called.
Do we need to disallow searching by fingerprint too?
Also, do we need to look into the downloaded pubkey to find an email in the subject to check whether its domain is allowed for attester search?

tomholub:
>  Do we need to disallow searching by fingerprint too?

If `*` is specified then yes, don't call Attester even by fingerprint.

>  Also, do we need to look into the downloaded pubkey to find an email in the subject to check whether its domain is allowed for attester search?

We could file an issue for this to worry about later. It's a fair question.

PR #3799, opened 2021-07-09
Encrypt with multiple keys per recipient

### LINKED ISSUE #3769: encrypt for all of recipient's keys when sending a message
Currently, if I have the wrong key for someone, they can send me their key, and the extension will offer to update the contact to the new key. 

But really what will happen is we'll now have *both* keys for that contact. So when I write them another email, this time hoping to use their correct key - nope, it will still use the same key as before.

So here are my questions:

1. How does the extension currently decide which key to use?
2. Can't we just use the most recently added key by default?
3. If not, perhaps when a recipient has multiple keys on file, we could offer a selection for the sender to choose, instead of just assuming?

rrrooommmaaa:
This PR uses as many encryption keys as it can find per each recipient

close #3769

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -55,14 +55,16 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
   }
 
   public collectAllAvailablePublicKeys = async (senderEmail: string, senderKi: KeyInfo, recipients: string[]): Promise<CollectPubkeysResult> => {
-    const contacts = await ContactStore.get(undefined, recipients);
+    const contacts = await ContactStore.getEncryptionKeys(undefined, recipients);
```
rrrooommmaaa, line 58: I made sure `getEncryptionKeys()` returns an entry per each input recipient so that missing emails are guaranteed to make their way to `emailsWithoutPubkeys`. I suspect they may get lost with the previous version of the code if the contact store didn't have an entry for them

tomholub: Got it

extension/js/common/platform/store/contact-store.ts
```diff
@@ -235,6 +235,27 @@ export class ContactStore extends AbstractStore {
...
+      const email = emails[0];
+      const contact = await ContactStore.getOneWithAllPubkeys(db, email);
+      const encryptionKeys = (contact?.sortedPubkeys || []).filter(k => !k.revoked && (k.pubkey.usableForEncryption || k.pubkey.usableForEncryptionButExpired)).
+        map(k => k.pubkey);
+      // if non-expired present, return non-expired only
+      if (encryptionKeys.some(k => k.usableForEncryption)) {
+        return [{ email, keys: encryptionKeys.filter(k => k.usableForEncryption) }];
+      } else {
+        return [{ email, keys: encryptionKeys }];
+      }
```
tomholub, line 252: This looks good, to preserve existing notification that user key has expired. I think this may be the last step in facilitating smooth public key rotations / expirations as well - the old key will naturally expire, and both will be used in the in-between period.

tomholub (commented):
:+1:

extension/js/common/platform/store/contact-store.ts
tomholub: Actually, this could cause a little bit of trouble: we have a mechanism to keep public keys fresh, but it works higher up in the stack and probably needs all of the keys to work properly. 

I can't come up with exact scenario when it would cause user-noticeable issues, but there possibly are some.

Ideally, the storage method would be returning all public keys for that recipient, and this logic would happen somewhere in composer view code / one of its modules. But the point is that the composer should have a chance to update a newer version of an expired key, even if some other valid key is available.

As it is now, it appears maybe the updated version of the expired key would not be noticed. (actually, depending on implementation, maybe on WKD it would be noticed since we pull by email and get back an array - can process the whole array which is good. But when pulling from Attester by fingerprint, that maybe doesn't work so well, because we don't ha ve a chance to pull updated version of the expired key)

tomholub (approved):

Issue #3798, opened 2021-07-09
Dismissing 'Use of externally-controlled format string' security warning


seisvelas:
This one:

<img width="1419" alt="Screen Shot 2021-07-08 at 6 32 57 p m" src="https://user-images.githubusercontent.com/44826516/125006265-ed4dc300-e01a-11eb-81ab-15c0cf8572d9.png">

This is passing an error message and stack trace to a console log. I don't think this is dangerous the way we're using it here, so I'm dismissing the warning. But I'm creating this issue in case anyone disagrees, or I'm misunderstanding, or anything else. 

Otherwise, if you agree with the dismissal of this warning, please go ahead and close this issue :)

tomholub:
To be fair, it's not strictly necessary that we color it this way ourselves. Maybe we can use `console.error` instead of `console.log` and no longer color it?

PR #3802, opened 2021-07-09
Remove customized coloring from console.log and use console.error instead

seisvelas:
Instead of using `console.log` with custom coloring to make the text red, we can use `console.error`. Simpler *and* better!

close https://github.com/FlowCrypt/flowcrypt-browser/issues/3798

- Not worth testing

### To be filled by reviewers


seisvelas:
This will probably fail tests because of the Gmail live tests - once tests finish we have to be super mindful to check that it's actually passing everything else!

seisvelas:
Failing other tests, but I think those also might be broken? Perhaps it's the tests broken by https://github.com/FlowCrypt/flowcrypt-browser/pull/3797 (since that appears to have broken some tests)

I should mark this ready for review once tests are figured out.

tomholub:
@seisvelas master is now ok with tests

tomholub (approved):
Thanks!

PR #3803, opened 2021-07-10
Special handling of disallow_attester_search for lookupFingerprint #3800

### LINKED ISSUE #3800: fix test / disallow searching attester with * by fingerprint
>  Do we need to disallow searching by fingerprint too?

If `*` is specified then yes, don't call Attester even by fingerprint.

>  Also, do we need to look into the downloaded pubkey to find an email in the subject to check whether its domain is allowed for attester search?

We could file an issue for this to worry about later. It's a fair question.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/3797#issuecomment-877316926_

rrrooommmaaa:
This PR fixes attester's lookupFingerprint() error

close #3800

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/api/key-server/attester.ts
+++ b/extension/js/common/api/key-server/attester.ts
@@ -24,21 +24,7 @@ export class Attester extends Api {
       console.info(`Skipping attester lookup of ${email} because attester search on this domain is disabled.`);
       return { pubkey: null }; // tslint:disable-line:no-null-keyword
     }
-    try {
-      const r = await this.pubCall(`pub/${email}`);
-      // when requested from the content script, `getResponseHeader` will be missing because it's not a real XMLHttpRequest we are getting back
-      // because it had to go through background scripts, and objects are serialized when this happens
-      // the proper fix would be to send back headers from bg along with response text, and parse it here
-      if (!r.getResponseHeader) {
-        return { pubkey: r.responseText };
-      }
-      return { pubkey: r.responseText };
-    } catch (e) {
-      if (ApiErr.isNotFound(e)) {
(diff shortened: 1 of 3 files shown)


extension/js/common/api/key-server/attester.ts
```diff
@@ -53,7 +39,15 @@ export class Attester extends Api {
    * the actual api accepts either email, fingerprint or longid
    */
   public lookupFingerprint = async (fingerprintOrLongid: string) => {
-    return await this.lookupEmail(fingerprintOrLongid);
+    if (fingerprintOrLongid.includes('@')) {
+      return await this.lookupEmail(fingerprintOrLongid);
+    }
```
tomholub, line 44: Actually I think this should be a `throw new Error`, since we already have distinct method for email and fingerprint/longid. If that ever blows up, then we are calling this wrongly, which would be good to know :-)

extension/js/common/api/key-server/attester.ts
```diff
@@ -115,4 +109,22 @@ export class Attester extends Api {
     return await Api.apiCall(ATTESTER_API_HOST, resource, data, typeof data === 'string' ? 'TEXT' : undefined, undefined, hdrs, 'xhr', method);
   }
 
+  private doLookup = async (emailOrFingerprint: string): Promise<PubkeySearchResult> => {
```
tomholub, line 112: Thank you, this is clearer

tomholub (commented):
Excellent - one nitpick for a clearer distiction

tomholub (approved):

Issue #3781, opened 2021-06-29
Replace "Forgot pass phrase?" on pass phrase dialog when key_manager_url OrgRule is set


tomholub:
Instead of the "Forgot pass phrase?" link, there should be text that says "Ask your IT staff for help if you lost your pass phrase."

![image](https://user-images.githubusercontent.com/6306961/123862105-76784200-d917-11eb-9e98-d66cdec0d13c.png)

martgil:
I think I have managed to fix this issue by adding a check using `this.orgRules.usesKeyManager()` in `passphrase.ts`
![fix forget password on when keymanager enabled](https://user-images.githubusercontent.com/46025304/125911882-2ac0b068-ed58-4981-aa27-259a12e19daf.png)

tomholub:
That looks good but is that still a link? It looks like the user could click it and go somewhere we don't want them to go.

martgil:
When the link was clicked, the pop-up help for "Non-FES" user is shown. 
![image](https://user-images.githubusercontent.com/46025304/125913164-b0f5097b-13d2-4565-8623-24faf5b0a3e2.png)

Maybe we can update or add the description for "FES" user. Or maybe we could just disable the link?

tomholub:
Let's disable the link when the user uses EKM as you say

tomholub:
There's a test that uses EKM and also FORBID_STORING_PASS_PHRASE org rule. That test will be using the pass phrase dialog, I think. You can add a line to that test to verify the content has changed and is not clickable. Thanks!

martgil:
Correct me if I'm wrong. it the `settings.ts` contains the test. Maybe I should just read the Document Object Model to check if the link is not clickable when the pass phrase dialog was invoked. I am a not yet comfortable with unit testing but I am trying to keep up as fast as I can. Thanks for understand.

martgil:
I have tried to understand how ava testing was implemented in the flowcrypt-browser but I can't able to absorb what's happening inside the code. Can I help somebody once I get it, I'm going to add the test for the changes that I have made hopefully. Pardon me for any inconvenience.

tomholub:
@martgil see if "Running tests locally" in https://github.com/FlowCrypt/flowcrypt-browser/wiki will help you. After that, please ask a more specific question and we'll be able to help or demonstrate. Thanks!

martgil:
[...]

PR #3821, opened 2021-07-16
Re-phrase lost password when user uses key manager url

martgil:
issue #3781

Re-phrasing the "lost password" context when the user uses key manager which came from org rules.

### To be filled by reviewers


extension/chrome/elements/passphrase.ts
```diff
@@ -42,6 +42,9 @@ View.run(class PassphraseView extends View {
     if (!this.orgRules.forbidStoringPassPhrase()) {
       $('.forget').prop('disabled', false);
     }
+    if (this.orgRules.usesKeyManager()) {
+      $('#lost-pass-phrase').text("Ask your IT staff for help if you lost your pass phrase.")
```
rrrooommmaaa, line 46: I think it's not quite neat to have 2 alternative messages in different places ("Lost pass phrase?" in `.htm` file, "Ask your IT..." in `.ts`).
Perhaps, it'd be better to keep them in one place (`.htm`?) and switch divs like some examples below
```
.css('display', 'none');
```

tomholub: Agree, it would be neater to have both alternatives next to one another (in htm), and only switch visibility (in ts). :+1:

martgil: I have moved the two messages in a div as suggested and uses ts for controlling which message should be shown to the user. I should've learned these coding standards from now on. thanks!

extension/chrome/elements/passphrase.ts
```diff
@@ -42,6 +42,12 @@ View.run(class PassphraseView extends View {
     if (!this.orgRules.forbidStoringPassPhrase()) {
       $('.forget').prop('disabled', false);
     }
+    if (this.orgRules.usesKeyManager()) {
+      $('#lost-pass-phrase').removeAttr('id').attr('href','#');
```
tomholub, line 46: I think this way it will still appear clickable? You'd probably have to remove the `href` too. Or have two completely separate elements, one `a` (original) and one `div class=line` (updated) for example.

Finally, `pass-phrase-alt` is not so descriptive, better would be `lost-pass-phrase-with-ekm`

martgil:
Here is the underlying error that prevents me when trying to create a new test for issue #3781. 
> Hi @tomholub. Thanks for the reference, It helps me to understand it better now. I have tried what exactly we need to do but I am having an error on compose.ts for some reason:
> image
> 
> As we can see even though I haven't used any of the `.passphrase` property it still insists `TypeError: Cannot read property 'passphrase' of undefined at Function.SetupPageRecipe.recover (/home/mart/project/flowcrypt-browser/test/source/tests/page-recipe/setup-page-recipe.ts:208:69)`.
> 
> The npm command I used for testing is the `npm run test_local_chrome_consumer_mock;` So that the `if (testVariant !== 'CONSUMER-LIVE-GMAIL')` will pass since I'm running the local test for consumer and not a consumer-live variant of testing. I believed that the `flowcrypt-browser/test/source/tests/page-recipe/setup-page-recipe.ts` has nothing to do with the error. The error for `.passphrase property is undefined` keeps appearing even on the untouched version of the code from `compose.ts` so It's really strange.

test/source/tests/compose.ts
```diff
@@ -31,6 +31,27 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      await settingsPage.notPresent('.swal2-container');
+      const inboxPage = await browser.newPage(t, TestUrls.extensionInbox(acctEmail));
+      await InboxPageRecipe.finishSessionOnInboxPage(inboxPage);
+      const composeFrame = await InboxPageRecipe.openAndGetComposeFrame(inboxPage);
+      await ComposePageRecipe.fillMsg(composeFrame, { to: 'human@flowcrypt.com' }, 'should not send as pass phrase is not known', { encrypt: false });
+      await composeFrame.waitAndClick('@action-send');
+      await inboxPage.waitAll('@dialog-passphrase');
+      const passphraseDialog = await inboxPage.getFrame(['passphrase.htm']);
+      await Util.sleep(5);
+      expect(await passphraseDialog.readHtml('@lost-pass-phrase-with-ekm')).to.equal('Ask your IT staff for help if you lost your pass phrase.');
```
tomholub, line 50: No need to copy the whole test and make a new one: you should can instead edit the existing EKM test I referred to earlier.

martgil:
@tomholub I have moved the test to the existing one you're referring to and confirmed the test works properly. But there is still a build error in the CI which I don't have an idea why.

tomholub:
Live Gmail tests sometimes have problems with login, not your fault. I'm re-running them.

tomholub:
And bringing updates from master to here, in case we fixed it since.

test/source/tests/setup.ts
```diff
@@ -444,6 +444,8 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await composeFrame.waitAndClick('@action-send');
       await inboxPage.waitAll('@dialog-passphrase');
       const passphraseDialog = await inboxPage.getFrame(['passphrase.htm']);
+      await Util.sleep(5);
+      expect(await passphraseDialog.readHtml('@lost-pass-phrase-with-ekm')).to.equal('Ask your IT staff for help if you lost your pass phrase.');
```
tomholub, line 448: Please use `waitForContent` method, which will keep retrying until it can find the text, instead of wait + try once. (it will execute faster, as it doesn't always have to wait for 5 seconds)

Otherwise this looks good. You should also alter one more test - a test that doesn't use EKM, to test the opposite scenario. I'll highlight the discussion below.

tomholub:
As I mentioned in https://github.com/FlowCrypt/flowcrypt-browser/issues/3781#issuecomment-883223754 , you should also test the opposite scenario without EKM - again but adding one more test condition to the existing test.

-----------

>     5. find and edit an appropriate test that does not have OrgRules set. You've already found one: `compose - signed with entered pass phrase + will remember pass phrase in session`
> 
> 
> ```ts
>     ava.default('compose - signed with entered pass phrase + will remember pass phrase in session', testWithBrowser('ci.tests.gmail', async (t, browser) => {
>       const k = Config.key('ci.tests.gmail');
>       const settingsPage = await browser.newPage(t, TestUrls.extensionSettings('ci.tests.gmail@flowcrypt.test'));
>       await SettingsPageRecipe.forgetAllPassPhrasesInStorage(settingsPage, k.passphrase);
>       const inboxPage = await browser.newPage(t, TestUrls.extensionInbox('ci.tests.gmail@flowcrypt.test'));
>       const composeFrame = await InboxPageRecipe.openAndGetComposeFrame(inboxPage);
>       await ComposePageRecipe.fillMsg(composeFrame, { to: 'human@flowcrypt.com' }, 'sign with entered pass phrase', { encrypt: false });
>       await composeFrame.waitAndClick('@action-send');
>       await inboxPage.waitAll('@dialog-passphrase');
>       const passphraseDialog = await inboxPage.getFrame(['passphrase.htm']);
>       await passphraseDialog.waitAndType('@input-pass-phrase', k.passphrase);
>       await passphraseDialog.waitAndClick('@action-confirm-pass-phrase-entry');
>       await inboxPage.waitTillGone('@dialog-passphrase');
>       await inboxPage.waitTillGone('@container-new-message'); // confirming pass phrase will auto-send the message
>       // signed - done, now try to see if it remembered pp in session
>       const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
>       await ComposePageRecipe.fillMsg(composePage, { to: 'human@flowcrypt.com' }, 'signed message pp in session', { encrypt: false });
>       await ComposePageRecipe.sendAndClose(composePage);
>       await settingsPage.close();
>       await inboxPage.close();
>     }));
> ```
> 
> Notice that it has a generic domain: `ci.tests.gmail@flowcrypt.test`. This should mean that it doesn't have any OrgRules set up (unlike for example `user@forbid-storing-passphrase-org-rule.flowcrypt.test` which does).
> 
> Here you can similarly edit the test, but this time to test for the opposite behavior.

martgil:
Thanks, @tomholub. I will pay more attention to small details next time. I think we have the test we want for this. I have learned so much here. Thanks Again!

tomholub (approved):
Looks good. Thanks!

PR #3825, opened 2021-07-19
Throw error on email when fingerprint is expected #3805

### LINKED ISSUE #3805: adjust lookup by fingerprint / email to be stricter
Actually I think this should be a `throw new Error`, since we already have distinct method for email and fingerprint/longid. If that ever blows up, then we are calling this wrongly, which would be good to know :-)

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/3803#discussion_r667380134_

rrrooommmaaa:
This PR throws exception when email is supplied to `lookupFingerprint()`

close #3805

- Not worth testing

### To be filled by reviewers


rrrooommmaaa:
Should we throw in `sks.lookupFingerprint` too?

tomholub (dismissed):
Looks good. From prudence perspective, sure - adding to SKS as well is a bit more prudent.

tomholub (approved):
Thanks!

Issue #3735, opened 2021-06-09
cannot find contacts in settings when searching by larger string


tomholub:
searching flowcrypt in my contacts list finds plenty (at flowcrypt.com domain)

![image](https://user-images.githubusercontent.com/6306961/121305209-db2e1700-c8ec-11eb-9372-b1ffeda6b850.png)

but flowcrypt.com search yields nothing:

![image](https://user-images.githubusercontent.com/6306961/121305291-f3059b00-c8ec-11eb-87c6-f63ed271676c.png)

tomholub:
similarly cannot find `user@example.com` by searching `user@example`

tomholub:
I can't even find the recipient when I search whole email like `human@flowcrypt.com` - will not find. But just `human` will find.

tomholub:
The current logic probably splits email like `first.last.other@example.com` into chunks like `[first, last, other, example]` and then chunks them up further like `[f, fi, fir, firs, first, l, la, las, last, o, ...]` which will all work but the moment I search `first.last` or whole email `first.last.other@example.com` then it won't work.

Can reproduce by going to Settings -> additional -> Contacts and searching there.

Problem will be here, likely:

```ts
  private static dbIndexRange = (hasPgp: boolean, substring: string): { lowerBound: string, upperBound: string } => {
    // to find all the keys starting with 'abc', we need to use a range search with exlcusive upper boundary
    // ['t:abc', 't:abd'), that is, we "replace" the last char ('c') with the char having subsequent code ('d')
    // The edge case is when the search string terminates with a certain char X having the max allowed code (65535)
    // or with a sequence of these, e.g. 'abcXXXXX'. In this case, we have to remove the tail of X characters
    // and increase the preceding non-X char, hence, the range would be ['t:abcXXXXX', 't:abd')
    // If the search sequence consists entirely of such symbols, the search range will have
    // the upper boundary of 'f;' or 't;', so this algorithm always works
    const lowerBound = ContactStore.dbIndex(hasPgp, substring);
    let copyLength = lowerBound.length - 1;
    let lastChar = lowerBound.charCodeAt(copyLength);
    while (lastChar >= 65535) {
      lastChar = lowerBound.charCodeAt(--copyLength);
    }
    const upperBound = lowerBound.substring(0, copyLength) + String.fromCharCode(lastChar + 1);
    return { lowerBound, upperBound };
  }
```

[...]

PR #3835, opened 2021-07-21
fix faulty function when searching public keys

martgil:
This PR will fix the issue #3735 when searching public keys with specific search term.

### To be filled by reviewers

### DIFF
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -63,7 +63,7 @@ View.run(class ContactsView extends View {
   // --- PRIVATE
 
   private loadAndRenderContactList = async () => {
-    const contacts = await ContactStore.search(undefined, { hasPgp: true, limit: 500, substring: String($('.input-search-contacts').val()) });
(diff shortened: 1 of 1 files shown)


extension/chrome/settings/modules/contacts.ts
```diff
@@ -63,7 +63,7 @@ View.run(class ContactsView extends View {
   // --- PRIVATE
 
   private loadAndRenderContactList = async () => {
-    const contacts = await ContactStore.search(undefined, { hasPgp: true, limit: 500, substring: String($('.input-search-contacts').val()) });
+    const contacts = await ContactStore.search(undefined, { hasPgp: true, limit: 500, substring: '' });
```
tomholub, line 66: This is wrong - you don't want to pull all of the public keys from storage, and then filter them later (there may be thousands of them). To fix this, you would have to dig into `ContactStore.search` itself, and fix it there (which may be complicated).

tomholub (changes requested):

extension/chrome/settings/modules/contacts.ts
martgil: Oh, I see. Thanks for pointing this out. I'll be closing this PR now.

PR #3842, opened 2021-07-22
[Proposal] Always show cc bcc buttons

### LINKED ISSUE #3115: cc and bcc do nothing / disappear when user clicks on them
ref: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgxwKjKmVhPNlKvFMCDktbLHrJplb

The user recorded the behavior and uploaded it into a video that clearly demonstrates the problem, including JS console output (there is none)

https://youtu.be/LJSRWnGGl00

limonte:
This PR is a proposed solution for #3115

To clarify, I was not able to reproduce the reported issue in both browsers. Also, my assumption about what might be the cause of the user's issue was wrong (I was assuming that buttons clicks are not handled because buttons are getting hidden right before the click handler).

But, I had to come up with the solution, because the user clearly demonstrated with a video that the issue exists. 

Proposed solution: **always show cc and bcc buttons**. I personally think it is better to always show these buttons even though Gmail is not doing that (but Outlook for example always shows cc/bbc buttons), so users will always see them and be able to click them with one click. 

Another good part about UX is that keyboard users previously had cc/bcc buttons popping up from the invisible state when moving focus with `Tab` which is unexpected. With this change, all focusable elements are visible and expected.

close #3115

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -235,7 +235,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       container.find('r_loader').remove();
     }
     Xss.sanitizeRender(container, '<span class="rest"><span id="rest_number"></span> more</span>');
-    const maxWidth = container.parent().width()!;
+    const maxWidth = container.parent().width()! - this.view.S.cached('container_cc_bcc_buttons').width()!;
     const rest = container.find('.rest');
     let processed = 0;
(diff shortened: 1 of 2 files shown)


rrrooommmaaa (commented):
I am not an expert in CSS, so can't comment on that part.
Otherwise, all looks good.

tomholub (approved):
I trust @limonte for UX matters, so if the code looks good to you (other then css specifics), it's good to merge.

PR #3844, opened 2021-07-23
Display the number of encrypted attachments

### LINKED ISSUE #3654: The number of attachments shouldn't be removed
Gmail:

CleanShot 2021-05-19 at 16 20 53@2x

FlowCrypt:

CleanShot 2021-05-19 at 16 21 34@2x

Something in out codebase removes that useful "4 attachments" block.

limonte:
This PR displays the number of encrypted attachments:

CleanShot 2021-07-23 at 22 59 25@2x

close #3654 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -64,6 +64,8 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     standardComposeWin: '.aaZ:visible',
     settingsBtnContainer: 'div.aeH > div > .fY',
     standardComposeRecipient: 'div.az9 span[email][data-hovercard-id]',
+    numberOfAttachments: '.aVW',
+    numberOfAttachmentsDigit: '.aVW span'
   };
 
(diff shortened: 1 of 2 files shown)


tomholub (commented):
Have you tested this on a message where half of the attachments are plain, an half are encrypted? :-) I suppose it may cause some confusion. You can grab some encrypted files sent by flowcrypt extension (on a browser without the extension) and send them in a plain email as-is already (encrypted), mixed with some plain files. See if the result makes sense, it's hard to say from code alone.

limonte:
> Have you tested this on a message where half of the attachments are plain, an half are encrypted? :-) I suppose it may cause some confusion. You can grab some encrypted files sent by flowcrypt extension (on a browser without the extension) and send them in a plain email as-is already (encrypted), mixed with some plain files. See if the result makes sense, it's hard to say from code alone.

I didn't test this before you asked, but now I did. It works fine, the calculation of `nRenderedAttachments` is correct and taking this case into consideration.

CleanShot 2021-07-26 at 00 50 25@2x

tomholub (approved):
Thanks!

PR #3845, opened 2021-07-24
Import revocation information automatically #3637

### LINKED ISSUE #3637: importing revoked keys automatically
Agree. We should also show a clear UI to say "This key is revoked, meaning it should not be used. Importing this key will ensure that other versions of this key won't be used either." Let's do a separate PR for that.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/3579#issuecomment-841695457_

rrrooommmaaa:
This PR doesn't allow to import a revoked public key to 'emails' and 'pubkeys' still,
but automatically imports revocation info to 'revocations' store,
with an intention to protect against further possible import of an older version of the revoked key.

close #3637

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/pgp_pubkey.ts
+++ b/extension/chrome/elements/pgp_pubkey.ts
@@ -40,18 +40,17 @@ View.run(class PgpPubkeyView extends View {
   public render = async () => {
     Ui.event.protect();
     try {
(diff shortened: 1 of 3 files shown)


tomholub (approved):
excellent

PR #3849, opened 2021-07-26
Do not bother users with the "loading..." state of contacts suggestions

### LINKED ISSUE #3711: We should probably not bother users with "loading..." state of contacts suggestions popup
While I do agree that it's helpful in most cases to indicate the background activity to users, this particular case seems to be unnecessary, flickering "loading..." popup is kinda annoying:

https://user-images.githubusercontent.com/6059356/120302739-da86f680-c2d6-11eb-9696-f594126f237a.mp4

IMO Here, we can hide the `loading` state from users and render contacts suggestions 

- when we have them
- no results found, but Google isn't connected (suggestion to connect it)
- error occured

limonte:
This PR prevents showing the contacts suggestion popup which was containing only "loading..." message. It was blinking and unnecessary distracting users when they are typing email addresses into the `#input_to` field:

CleanShot 2021-07-26 at 22 36 46

close #3711

Also, some minor style adjustments for sweetalert2 after the latest update from #3847

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -42,12 +42,12 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
 
   private dragged: Element | undefined = undefined;
 
-  private canSearchContacts: boolean;
+  private googleContactsSearchEnabled: boolean;
   private canReadEmails: boolean;
 
   constructor(view: ComposeView) {
     super(view);
-    this.canSearchContacts = this.view.scopes.readContacts;
+    this.googleContactsSearchEnabled = this.view.scopes.readContacts;
     this.canReadEmails = this.view.scopes.read || this.view.scopes.modify;
   }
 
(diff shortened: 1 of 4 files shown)


extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -627,8 +633,13 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
         }
         ulHtml += '</li>';
       }
+      this.preventDisplaying = false;
       if (this.contactSearchInProgress) {
-        ulHtml += '<li class="loading" data-test="container-contacts-loading">loading...</li>';
+        if (this.canSearchContacts) {
+          this.preventDisplaying = true;
+        } else {
+          ulHtml += '<li class="loading" data-test="container-contacts-loading">loading...</li>';
```
tomholub, line 641: It seems counter intuitive that the loader would stay there exactly when user cannot search contacts. Is this to have some space to put the new permission prompt into?

limonte: I believe the naming of the variable is confusing here. Maybe it should be `googleContactsSearchEnabled`?

The loading state is needed for searching contacts from the storage (locally):

CleanShot 2021-07-26 at 22 51 18

But now, that I'm looking at it and the "loading..." state is so quick that it looks like a glitch, should we remove the loading state completely for both local search and google contacts search?

tomholub: That's what I was thinking - remove both, since local should be fast

limonte: Done 👍

tomholub (approved):

PR #3860, opened 2021-07-28
load, encrypt and decrypt private key for S/MIME

### LINKED ISSUE #3529: add support for encrypting pkcs12 s/mime key
Would like to implement this instead of showing error:

```ts
  public static encrypt = async (key: Key, passphrase: string) => {
    if (key.type === 'openpgp') {
      return await OpenPGPKey.encryptKey(key, passphrase);
    } else {
      throw new Error(`KeyUtil.encrypt does not support key type ${key.type}`);
    }
  }
```

The forge library we use likely has a method for this.

rrrooommmaaa:
This PR does these changes:
1. As `forge.pkcs12.pkcs12FromAsn1(asn, password)` throws an exception on invalid password 'PKCS#12 MAC could not be verified. Invalid password?', it's not possible to extract the certificate too, so I changed the internal storage format to those of a typical PEM file containing PKCS#8 (possibly encrypted) private key e.g. `-----BEGIN ENCRYPTED PRIVATE KEY-----` as well as certificate(s).
Thus, as we encounter an encrypted private key block, we load it into memory unchanged (in string form), to be later converted to a usable PrivateKey object with `forge.pki.decryptRsaPrivateKey()`
The current implementation takes the first certificate in PEM file and assumes that this is the certificate to match the private key.
Is this correct? Can certificates in PEM file be shuffled, that is can a CA certificate come first, followed by the user certificate?
To make the app more robust, should we:
- parse all the certificates to determine the lower-level one?
- when decrypting the private key, make sure it matches the certificate we previously selected?

2. Refactoring -- Certificate.privateKey is no longer used (as it's never filled in `forge` library) so perhaps having this field in `types` is an error. The benefit is that our own field may be of string (encrypted) or object (ready-to-use) type, allowing the same manipulations as OpenPGP keys.

close #3529

- Tests added or updated

### To be filled by reviewers


extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -70,6 +84,45 @@ export class SmimeKey {
...
+    }
+    if (key.fullyDecrypted) {
+      if (optionalBehaviorFlag === 'OK-IF-ALREADY-DECRYPTED') {
+        return true;
+      } else {
+        throw new Error("Decryption failed - private key was already decrypted");
+      }
+    }
+    const encryptedPrivateKey = SmimeKey.getArmoredPrivateKey(key);
+    const privateKey = await forge.pki.decryptRsaPrivateKey(encryptedPrivateKey, passphrase); // throws on password mismatch
```
rrrooommmaaa, line 99: Or doesn't throw. Need to docuble-check

rrrooommmaaa:
@tomholub I added the description on the PR. Can you comment on the general approach as well as the proposed privateKey-certificate matching logic?

tomholub:
>  Is this correct? Can certificates in PEM file be shuffled, that is can a CA certificate come first, followed by the user certificate?

I don't know. For pkcs12 just about any order is possible, and for pkcs8, I'd also say that it could be shuffled, although there may be a convention to follow certain order (hard to say).

> To make the app more robust, should we:
-    parse all the certificates to determine the lower-level one?
-    when decrypting the private key, make sure it matches the certificate we previously selected?

Sounds to me like we should probably do both of these, as you say, for robustness. (unless they are mutually exclusive? I'm not entirely sure what "determine the lower-level one" means)

tomholub:
> parse all the certificates to determine the lower-level one?

Ok now I get what you mean - you mean the lowest level certificate in the trust chain - the "user certificate" if you will as opposed to the CA certificates.

In the future, it may be good to retain the whole certificate chain, but that is out of scope for now - too much hassle at this stage, I think. Alternatively, instead of keeping whole chain, maybe we could validate the whole chain before we import the key, then forget it (also out of scope now).

Therefore, your approach to throw CAs away and preserve the leaf certificate sounds good to me.

rrrooommmaaa:
Yes, the lower-level certificate is the one in the chain that signs no other certificates.
These measures are not mutually exclusive. In theory, the certificate in the file may not match the private key, we can only find it out after decryption.

tomholub (commented):
Progressing nicely 👍

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -6,21 +6,23 @@ import { UnreportableError } from '../../../platform/catch.js';
...
 
   public static parse = (text: string): Key => {
     if (text.includes(PgpArmor.headers('certificate').begin)) {
       const blocks = MsgBlockParser.detectBlocks(text).blocks;
       const certificates = blocks.filter(b => b.type === 'certificate');
-      if (certificates.length < 1) { // todo: should we select the correct certificate if a chain is provided?
+      const leafCertificates = SmimeKey.getLeafCertificates(certificates);
+      if (leafCertificates.length !== 1) {
         throw new Error('Could not parse S/MIME key without a certificate');
       }
```
tomholub: Maybe first test for > 1 leafs to throw `Parsing S/MIME with more than one user certificate is not supported`

tomholub: Just for clarity, else it may look like there wasn't any cert when there were two

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -118,31 +197,67 @@ export class SmimeKey {
...
+    const privateKeyN = (privateKey as forge.pki.rsa.PrivateKey).n;
+    const privateKeyE = (privateKey as forge.pki.rsa.PrivateKey).e;
+    if (publicKeyN && publicKeyE && privateKeyN && privateKeyE) {
+      // compare RSA Private and Public Key material
+      if (publicKeyN.compareTo(privateKeyN) === 0 && publicKeyE.compareTo(privateKeyE) === 0) {
+        return;
+      }
+      throw new Error("Certificate doesn't match the private key");
     }
+    throw new Error("This key type is not supported");
```
rrrooommmaaa, line 256: Should we add support for ed25519 in this PR?

test/source/tests/unit-node.ts
```diff
@@ -549,6 +550,23 @@ vpQiyk4ceuTNkUZ/qmgiMpQLxXZnDDo=
...
+        instanceOf: Error, message: `Certificate doesn't match the private key`
+      });
+      t.pass();
+    });
+
+    ava.default('[unit][KeyUtil.decrypt] S/MIME key decryption of mismatching private key', async t => {
+      const encryptedKey = await KeyUtil.parse(`${testConstants.smimeEncryptedKey}
+${testConstants.smimeCert}`);
+      await t.throwsAsync(() => KeyUtil.decrypt(encryptedKey, 'AHbxhwquX5pc'), {
+        instanceOf: Error, message: `Certificate doesn't match the private key`
```
rrrooommmaaa, line 565: Or should we create `UnreportableError`?

extension/js/common/core/crypto/smime/smime-key.ts
tomholub: Can be filed as a separate issue for later

test/source/tests/unit-node.ts
tomholub: Sure, that sounds like a good idea. As long as it gets rendered to the user clearly, eg as a modal (which it probably does automatically, but would be good to double check).

tomholub (dismissed):
Looks good!

tomholub:
I'll attempt to fix the gmail tests here.

tomholub (approved):
Did various cleanup and disabled most Gmail tests due to https://github.com/FlowCrypt/flowcrypt-browser/issues/3929

PR #3861, opened 2021-07-28
allow to test manualEnter with keyPath #3528

### LINKED ISSUE #3528: test importing binary private key from file
The file import functionality during setup does not seem to be tested.

I have added some file loading functionality in `SetupPageRecipe.manualEnter` with `key.filePath` but that assumes puppeteer can directly input file into the `<input type=file>` but what actually happens during import is that the OS file picker gets automatically triggered - I don't think puppeteer has any way to handle this.

The solution may be to at least render the file input in advance, even if hidden. Then puppeteer can set it with the input file without clicking on the "import from file" label which causes the file picker.

rrrooommmaaa:
This PR fixes `SetupPageRecipe.manualEnter` to handle files

close #3528 

- Tests added or updated

### To be filled by reviewers


test/source/test.ts
```diff
@@ -30,10 +30,10 @@ const DEBUG_MOCK_LOG = false; // se to true to print mock server logs
...
 
 const consts = { // higher concurrency can cause 429 google errs when composing
-  TIMEOUT_SHORT: minutes(1),
-  TIMEOUT_EACH_RETRY: minutes(3),
-  TIMEOUT_ALL_RETRIES: minutes(13), // this has to suffer waiting for semaphore between retries, thus almost the same as below
-  TIMEOUT_OVERALL: minutes(14),
+  TIMEOUT_SHORT: minutes(1000),
+  TIMEOUT_EACH_RETRY: minutes(3000),
+  TIMEOUT_ALL_RETRIES: minutes(13000), // this has to suffer waiting for semaphore between retries, thus almost the same as below
+  TIMEOUT_OVERALL: minutes(14000),
```
tomholub, line 36: I suppose this was for local debugging

tomholub (commented):
:+1: except what seems an accidental change

tomholub (approved):

test/source/test.ts
rrrooommmaaa: Exactly

tomholub (approved):

PR #3887, opened 2021-08-04
issue 3886 build type that uses local FES

tomholub:
This PR adds a build type `chrome-enterprise-local-fes` so that we can test new FES functionality locally, without having to deploy new FES to a real server.

close #3886

----------------------------------

**Tests** _(delete all except exactly one)_:
- Not worth testing

--------------------------------

### To be filled by reviewers

I have reviewed that this PR... _(tick whichever items you personally focused on during this review)_:
- [ ] addresses the issue it closes (if any)
- [ ] code is readable and understandable
- [ ] is accompanied with tests, or tests are not needed
- [ ] is free of vulnerabilities
- [ ] is documented clearly and usefully, or doesn't need documentation

PR #3893, opened 2021-08-06
issue 3892 FES no pwd hashing

### LINKED ISSUE #3892: remove iterated hashing of password when using FES for web portal
but still keep it for flowcrypt.com/api for backwards compatibility

This is because OpenPGP.js has a native mechanism we should use instead, which doesn't rely on us keeping both recipient and sender code in sync - the amount of iterations is encoded in the message. https://datatracker.ietf.org/doc/html/rfc4880#section-3.7.1.3

More details in the PR, in code comments.

tomholub:
This PR removes manual iterated hashing when using the FES web portal, but preserves it when using flowcrypt.com api.

This PR should remain as a draft until there is a corresponding draft on the FES repo.

close #3892

 - unsure about tests. I think we may do this without automated tests (only do manual testing, once FES has been adjusted). After that, I think we'll do automated testing on the FES repo instead, with the browser extension built using https://github.com/FlowCrypt/flowcrypt-browser/pull/3887 )

### To be filled by reviewers


rrrooommmaaa (commented):
Looks good, except for one minor typo

rrrooommmaaa (approved):
LGTM!

Issue #3897, opened 2021-08-10
refactoring fetching of pubkeys on compose screen


tomholub:
in light of #3332 I think there is some more refactoring due. I'll make a draft PR to highlight it.

tomholub:
I have filed separate more specific issues linked above, will close this issue.

Also renamed a few methods for clarity in #3898

PR #3898, opened 2021-08-10
issue 3897 rename methods and add comments

tomholub:
This PR:
 - renames methods for clarity
 - adds comments for clarity
 - adds 3 notes in the code highlighting possible issues that we should be fixing in another PR

```
// note by Tom 2021-08-10: This line below looks weird in the context of recently allowed
//    more than one public key per recipient. We are blindly getting the first public key,
//    and making decisions based on that. We should be using the whole array instead.

// note by Tom 2021-08-10: I wonder if this functionality could be deprecated, it appears that
//    `lookupPubkeyFromKeyserversAndUpsertDb` below could cover this usecase, too.

* Note by Tom 2021-08-10: It looks to me like the return signature should instad
*    be `Promise<Contact[]>` since we deal with a list? And the `fail` situations
*    should probably be throwing some sort of `PubkeyLookupFailedError` or similar.
```

issue #3897 - does not actually close the issue, but aims to highlight the issues

@rrrooommmaaa please have a look in case you have some thoughts about this. I'd like to merge this PR as is (possibly with some updated comments?), and then consider doing the mentioned improvements in separate PRs one by one.

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -81,19 +81,47 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
     return await PassphraseStore.get(this.view.acctEmail, senderKi.fingerprints[0]);
   }
 
-  public lookupPubkeyFromDbOrKeyserverAndUpdateDbIfneeded = async (email: string, name: string | undefined): Promise<Contact | "fail"> => {
+  public lookupPubkeyFromKeyserversThenOptionallyFetchExpiredByFingerprintAndUpsertDb = async (
+    email: string, name: string | undefined
+  ): Promise<Contact | "fail"> => {
+    // note by Tom 2021-08-10: This line below looks weird in the context of recently allowed
+    //    more than one public key per recipient. We are blindly getting the first public key,
+    //    and making decisions based on that. We should be using the whole array instead.
```
rrrooommmaaa, line 89: This is not "blindly" the first public key. The compatibility `get` method is designed to provide the "best" key, that is non-expired (if possible) and valid as long as possible. Of course, we can replace `get` calls with `getOneWithAllPubkeys` if required.

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -81,19 +81,47 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
+      //    newer version would have been updated in db.
+      // This implementation is imperfect in that, if sender didn't pull a particular pubkey
+      //    for a long time and the local pubkey has since expired, and there actually is a
+      //    newer version available on external key server, this may unnecessarily show "bad pubkey",
+      //    until next time user tries to enter recipient in the field again, which will at that point
+      //    get the updated key from db. This could be fixed by:
+      //      - either life fixing the UI after this call finishes, or
+      //      - making this call below synchronous and using the result directly
+      //
+      // note by Tom 2021-08-10: I wonder if this functionality could be deprecated, it appears that
```
rrrooommmaaa, line 102: Looks like checkKeyserverForNewerVersionOfKnownPubkeyIfNeeded is using fingerprints, so we should refactor it to accept fingerprints, collected with `getOneWithAllPubkeys`.
`lookupPubkeyFromKeyserversAndUpsertDb` is email-based

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -81,19 +81,47 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
+  /**
+   * We are searching recipient public key by email every time we enter the recipient.
+   * This is regardless if we already have the public key stored locally or not.
+   * We process the response and if there are new public keys, we save them. If there are
+   *    newer versions of public keys we already have (compared by fingerprint), then we
+   *    update the public keys we already have.
+   * Finally, we return the updated public key back.
+   *
+   * Note by Tom 2021-08-10: It looks to me like the return signature should instad
+   *    be `Promise<Contact[]>` since we deal with a list? And the `fail` situations
```
rrrooommmaaa, line 119: `Promise<Contact>` is actually used for subsequent rendering in `renderPubkeyResult`. If we want to move the "diagnostics" logic to compose-recipients module `evaluateRecipients` method, that's ok, but we don't have to present the results as a list of `Contact`, as this is a bit misleading (`email` field will be copied all over the entries), so the perfect solution would be a special type with `email` field and a list of keys.
Currently, the diagnostics (figuring out the properties of the most appropriate key) is done by the line below.
```
const [preferred] = await ContactStore.get(undefined, [email]);
```
We can, of cause, get rid of the compatibility method `ContactStore.get` and move the key-picking logic to some other place, but we need to find all places where it used.
I see it is used in `GmailElementReplacer.evaluateStandardComposeRecipients()` too

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: That's the thing - I think we could and probably should completely drop the fingerprint-based lookup here. 

Let's say I have a `recipient@example.com` with key `ABCD`. Here, we look up by fingerprint to get the same `ABCD` key, and update it. But why do that, when `lookupPubkeyFromKeyserversAndUpsertDb` can look up by `recipient@example.com` and update the key just the same? After recent changes (when we started pulling pubkeys by email on every use), this all of this code that updates by fingerprint seems unneeded. Because what will happen if we just delete this chunk of code? The method `lookupPubkeyFromKeyserversAndUpsertDb` below will get executed, and we'll achieve the same result, just differently.

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: Got it, the key is being selected carefully to be useful. 

But it's still just one key, when we should be dealing with arrays. This has side effects. For example, when I hover over a recipient, it says `Does use encryption` and then shows only one key, but it should really be showing all public key that we will encrypt for.

I'm thinking we could use `getOneWithAllPubkeys`, which would indeed return all pubkeys including expired and revoked ones. And then have the key selection logic elsewhere.

It seems this public key selection logic is only used in compose view anyway, so we could move it there?

This means we could completely deprecate the `get` method that gets by email. I don't think there is any valid situation where we want to only get one public key anymore, except for legacy code that should be updated to work with an array meaningfully.

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: > so the perfect solution would be a special type with email field and a list of keys

That sounds good to me

> We can, of cause, get rid of the compatibility method ContactStore.get and move the key-picking logic to some other place, but we need to find all places where it used.
I see it is used in GmailElementReplacer.evaluateStandardComposeRecipients() too

Got it. Here's the code:

```ts
      // pick first usableForEncryption
      let selected = contactWithAllPubkeys.sortedPubkeys.find(entry => !entry.revoked && entry.pubkey.usableForEncryption);
      if (!selected) {
        selected = contactWithAllPubkeys.sortedPubkeys.find(entry => !entry.revoked && entry.pubkey.usableForEncryptionButExpired);
      }
      if (!selected) {
        selected = contactWithAllPubkeys.sortedPubkeys[0];
      }
```

`GmailElementReplacer.evaluateStandardComposeRecipients()` only needs to know if there is any usable key or not. So it could be as simple as:

```diff
-const [contact] = await ContactStore.get(undefined, [email]);
+const pubkeys = await ContactStore.getOneWithAllPubkeys(undefined, [email]);
-if (contact && contact.pubkey) {
+if (pubkeys.filter(p => p.usableForEncryption).length) {
```

Therefore, the actual nuanced logic that relates to choosing public keys for encryption can be actually in the compose modules somewhere. Then we can always use `.getOneWithAllPubkeys`, and never `.get`, which should produce fewer quirks in code and in behavior, I think.

tomholub:
Thanks!

extension/chrome/elements/compose-modules/compose-storage-module.ts
rrrooommmaaa: As far as I understand, our storage allows to link an e-mail to arbitrary pubkeys (with missing UserID or having an list of UserIDs etc.), so not all the keys can be found by email on the EKM, unless, of course, an OrgRule forbids to create email->pubkey link manually.

extension/chrome/elements/compose-modules/compose-storage-module.ts
rrrooommmaaa: Sure, we can create an issue to work in this direction.

extension/chrome/elements/compose-modules/compose-storage-module.ts
rrrooommmaaa: `Contact.hasPgp` just shows that a key exists, it may be revoked or expired, but generally I don't mind against moving key-selection logic out of ContactStore. It was implemented to mimick the original ContactStore with the intent of not making too much changes at once. Now that the code is looking stable, we can move on with further refactoring.

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: :+1: I'm filing https://github.com/FlowCrypt/flowcrypt-browser/issues/3905 , marking this as resolved

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: filed https://github.com/FlowCrypt/flowcrypt-browser/issues/3905

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: This particular functionality was there for customer who doesn't use EKM, and don't have any particular OrgRules affecting this. They have their public keys on our Attester.

Their keys expire, and then get renewed, with the same keys but a newer signature on it. This code would fetch them by fingerprint and get the updated version, but I'm pretty sure if we deleted this `if` and the asynchronous call in it, nothing would change, because the code right below it would update it by email anyway. Does that make sense?

limonte:
@tomholub I need one more commit for the flaky test fix. Can I do it in this PR or should I create another one?

tomholub:
No prob, you can add it here.

extension/chrome/elements/compose-modules/compose-storage-module.ts
tomholub: Maybe this particular usecase is riddled with too much customer-specific insider info :-) Maybe I'll just try to remove it, and see what happens.

limonte:
@tomholub should be good now, I ran tests twice, all is green ✅  

This flakiness is my personal enemy now 😄 Hopefully, it's gone. I've added the `waitTillFocusIsIn` helper which might be useful in the future.

tomholub:
Great, thank you. I'll go ahead and merge this now - just renamed methods and some comments and this flakiness fix.

PR #3903, opened 2021-08-11
Flaky gmail test for encrypted drafts: make sure there's no drafts before running this test

### LINKED ISSUE #3481: Flaky test - secure reply btn, reply draft
Okay, I reran the tests for the last commit to master (which passed when the commit was merged) and they are now failing: https://flowcrypt.semaphoreci.com/workflows/9ba160ab-2172-4a99-85ca-234edf401adc?pipeline_id=320a823f-bbcf-41db-8052-d2f11c0bbe7f

So I don't believe this is related to this PR specifically.

_Originally posted by @seisvelas in https://github.com/FlowCrypt/flowcrypt-browser/issues/3480#issuecomment-796663205_

limonte:
This PR is another attempt to fix the flaky Gmail test.

close #3481

### To be filled by reviewers

### DIFF
--- a/test/source/tests/gmail.ts
+++ b/test/source/tests/gmail.ts
@@ -55,7 +55,6 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       const urls = await gmailPage.getFramesUrls(['/chrome/elements/compose.htm']);
       expect(urls.length).to.equal(1);
       const replyBox = await browser.newPage(t, urls[0]);
-      console.log(await replyBox.page.screenshot({ encoding: "base64" }));
       if (expectedContent) {
         await replyBox.waitForContent('@input-body', expectedContent);
       } else {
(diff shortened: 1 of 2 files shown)


limonte:
@tomholub I just realized that the issue might be in parallelizing tests. If two instances of Gmail live tests are running simultaneously, then there are chances that something will collapse. E.g. this particular test case running in two parallel instances will fail.

Do you think we should limit the parallelization in CircleCI? Maybe there's even an option to prevent parallelization only for particular jobs (live Gmail tests in our case). Let me know your thoughts on this.

limonte:
> However, is there parallelism across PRs? That's possible, it's hard to say.

Yes, there is parallelism across PRs.

> If it's just a single tests that's giving us disproportionate amount of trouble, we may want to disable that tests, I guess.

The flaky test is quite useful, it covers creating, saving, and sending the encrypted drafts which is the important part of FlowCrypt. Let me play a bit more with it, I'll make it robust.

tomholub:
understood - thanks!

tomholub (approved):
Thanks!

PR #3938, opened 2021-08-29
initial import of S/MIME key

### LINKED ISSUE #3530: import decrypted pkcs12 private key
after #3528 and #3529 we should enable this test (and see if there is something left to implement, this is untested functionality)

```
    // ava.default.only(
    //   'setup - s/mime private key',
    //   testWithBrowser(undefined, async (t, browser) => {
    //     const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'flowcrypt.test.key.imported@gmail.com');
    //     const key = {
    //       title: 's/mime pkcs12 encrypted key',
    //       filePath: 'test/samples/smime/human-unprotected-PKCS12.p12',
    //       armored: null,
    //       passphrase: 'test pp to encrypt unprotected key',
    //       longid: null
    //     };
    //     await SetupPageRecipe.manualEnter(settingsPage, key.title, { submitPubkey: false, usedPgpBefore: false, key });
    //   })
    // );
```

rrrooommmaaa:
This PR enables importing an unencrypted S/MIME PKCS#12 file

Also implements a proper `SmimeKey.asPublicKey()` method

close #3530

- Tests added or updated

### To be filled by reviewers


test/source/tests/setup.ts
```diff
@@ -725,20 +725,20 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
-    //       title: 's/mime pkcs12 encrypted key',
-    //       filePath: 'test/samples/smime/human-unprotected-PKCS12.p12',
-    //       armored: null,
-    //       passphrase: 'test pp to encrypt unprotected key',
-    //       longid: null
-    //     };
-    //     await SetupPageRecipe.manualEnter(settingsPage, key.title, { submitPubkey: false, usedPgpBefore: false, key });
-    //   })
-    // );
+    ava.default(
```
rrrooommmaaa, line 728: @tomholub The test now passes, but we should really throw an exception when trying to import S/MIME key as the only private key? And disable this test again for now?
I'll also create an add key test like this.

tomholub: I guess we should, treating S/MIME private keys as second class citizens.

However, I'd also be ok with a warning (`Ui.modal.warn`) that the user can ignore. `Using S/MIME as the only key on account is experimental and not recommended. Continue anyway? [ok] [cancel]`

rrrooommmaaa: Looks like Ui.modal.warning isn't designed to have [ok, cancel] options. I implemented that with UI.modal.confirm, ok?

tomholub: sorry for confusion - yes - that's what I meant.

tomholub (approved):
Looks good - thanks!

PR #3979, opened 2021-09-14
S/MIME signing (WIP)

### LINKED ISSUE #2970: allow signing of outgoing s/mime emails
dependent on https://github.com/FlowCrypt/flowcrypt-browser/issues/3530

rrrooommmaaa:
This PR supports sending Signed S/MIME messages

Tasks:
- [x] encrypt/decrypt PKCS#7 message
- [x] support for armored PKCS#7
- [x] support for PKCS#7-encrypted drafts
- [x] send simple-text S/MIME message without attachments

close #2970

- Tests added or updated

### To be filled by reviewers


rrrooommmaaa:
@tomholub a simple-text S/MIME was successfully generated, also S/MIME drafts (encrypted) are supported in this PR.
Which features of the task list should I implement in this PR and which to postpone? (The PR becomes quite big).

tomholub:
Ok to merge limited functionality, for smaller PR.

The ultimate goal is to support signed+encrypted S/MIME, which is what one of our customer needs and has been asking for. That can also be in separate PR.

> add authenticated attributes to PKCS#7 message

does this mean the actual signing? Or something else?

>  send simple text with attachments

can be a separate issue for the future

>  send richtext?

can be a separate issue for the future

>  signed and encrypted?

This is the most important. Could also be tackled separately.

rrrooommmaaa:
> > add authenticated attributes to PKCS#7 message
> 
> does this mean the actual signing? Or something else?

Well, for according to docs, it is possible for SignerInfo to have some optional signed attributes, listed here: https://datatracker.ietf.org/doc/html/rfc5652#section-11

From what I understood, we can omit all of these signed attributes (and Thundertbird successfully verified the signature without them).
However, if we require at least one (e.g. "Signing Time"), we MUST include "Content Type" and "Message Digest".
As said in https://datatracker.ietf.org/doc/html/rfc5652#section-5.6 ... _If the SignedData signerInfo includes
   signedAttributes_
   
So, as far as my understanding goes, the question is whether we need `Signing Time`

rrrooommmaaa:
It's quite easy to add. Not sure if necessary. The `forge` library shows this example
```
      authenticatedAttributes: [{
        type: forge.pki.oids.contentType,
        value: forge.pki.oids.data
      }, {
        type: forge.pki.oids.messageDigest
        // value will be auto-populated at signing time
      }, {
        type: forge.pki.oids.signingTime
        // value will be auto-populated at signing time
        //value: new Date('2050-01-01T00:00:00Z')
      }]
```
`forge.pki.oids.data` is a constant `1.2.840.113549.1.7.1`, other values are placeholders to be filled by the `forge` library itself.
Took time to figure out.
Guessed you (or the client) had a postion on whether this is required.

tomholub:
Please file the above two comments in a separate issue. We'll deliver to customer without it. If they need it, they will know (their recipient system will reject the message). Then we'd add it at that point. Otherwise, we'd add it whenever we have spare time in the future (2023? :-) )

extension/js/common/core/crypto/key.ts
```diff
@@ -291,21 +291,23 @@ export class KeyUtil {
...
-    if (otherPgpPubs.length) {
-      return myPubs.concat(...otherPgpPubs).map(p => p.pubkey);
-    }
-    if (otherSmimePubs.length) { // todo - currently skipping my own pgp keys when encrypting message for S/MIME
-      return otherSmimePubs.map(pub => pub.pubkey);
+    let pgpPubs = pubs.filter(pub => pub.pubkey.type === 'openpgp');
+    let smimePubs = pubs.filter(pub => pub.pubkey.type === 'x509');
+    if (pgpPubs.length && smimePubs.length) {
+      // get rid of some of my keys to resolve the conflict
+      // todo: how would it work with drafts?
```
tomholub, line 298: Drafts could be enabled only when we have at least one OpenPGP private key? Then make them OpenPGP?

In general, it's ok to explicitly disable functionality when user only imports S/MIME private key and nothing else. It's not a scenario we are designing for, even if it's allowed.

tomholub (commented):
This is a serious effort - very happy to see this.

extension/js/common/core/crypto/key.ts
rrrooommmaaa: I implemented drafts for S/MIME because there were errors popping up when testing S/MIME user (without OpenPGP keys) editing a message (although "experimental"). I will disable such drafts completely, if you like.

tomholub: If it already works, no need to disable it. If you needed to spend a bunch of other work, then can disable as a way to save time

rrrooommmaaa:
Oh perhaps need to add a test to send an S/MIME signed message when passphrase is not entered.

extension/js/common/core/crypto/pgp/pgp-armor.ts
```diff
@@ -91,7 +91,10 @@ export class PgpArmor {
     }
     const utfChunk = new Buf(encrypted.slice(0, 100)).toUtfStr('ignore'); // ignore errors - this may not be utf string, just testing
     if (utfChunk.includes(PgpArmor.headers('pkcs7').begin)) {
-      const p7 = SmimeKey.cryptoMsgPrepareForDecrypt(encrypted); // todo: remove header check from SmimeKey's function?
+      const p7 = SmimeKey.readArmoredPkcs7Message(encrypted);
+      if (p7.type !== '1.2.840.113549.1.7.3') {
+        throw new Error('Not implemented');
+      }
```
tomholub, line 97: Could yo please add comment regarding what this means? `1.2.840.113549.1.7.3` and also what is the regular type, and why is `1.2.840.113549.1.7.3` different from whatever the regular type is.

tomholub (approved):
I'm starting to feel like we are writing some sort of an S/MIME library here. It may be worth to one day extract this functionality, including their own tests, and make an actual library from it that uses forge as dependency and get it security-reviewed. Then we'd import this s/mime library into our project. Either that, or contribute the improvements back to forge. Because these S/MIME operations are getting quite dense: ideally the browser extension code would work at a higher level of abstraction.

For now, it's more convenient to keep it in this codebase, for flexibility as you develop and add major S/MIME functionality. But once that mostly settles, it would be better to extract it into a library.

tomholub:
I'll merge this, then you can address the misisng test and a comment afterwards.

PR #3986, opened 2021-09-17
Change the storage path for local/offline drafts

### LINKED ISSUE #3337: review path of storage of local offline drafts
This is after #3310 lands

I would like to review this in another issue and another PR - I'd rather have a more structured place where to save drafts.

For example a "localDrafts" array or object where all the drafts would be.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/3310#discussion_r559498021_

limonte:
This PR puts all local drafts into 1 place in storage, under the `local_drafts` key. Also, it adds the migration, so users will not lose their local drafts which were saved in the old format.

close #3337

### To be filled by reviewers

### DIFF
--- a/extension/js/background_page/migrations.ts
+++ b/extension/js/background_page/migrations.ts
@@ -48,21 +48,22 @@ export const migrateGlobal = async () => {
     console.info('done migrating');
   }
   // migrate local drafts (#3337)
-  const storageLocal = await storageLocalGetAll();
   if (typeof globalStore.local_drafts === 'undefined') {
+    console.info('migrating local drafts in old format...');
     globalStore.local_drafts = {};
-  }
-  const oldDrafts = [];
-  for (const key of Object.keys(storageLocal)) {
-    if (key.startsWith('local-draft-')) {
-      console.info(`migrating local draft ${key}`);
-      globalStore.local_drafts[key] = storageLocal[key] as GmailRes.GmailDraftGet;
-      oldDrafts.push(key);
+    const storageLocal = await storageLocalGetAll();
+    const oldDrafts = [];
+    for (const key of Object.keys(storageLocal)) {
+      if (key.startsWith('local-draft-')) {
+        console.info(`migrating local draft ${key}`);
+        globalStore.local_drafts[key] = storageLocal[key] as GmailRes.GmailDraftGet;
+        oldDrafts.push(key);
+      }
     }
-  }
-  if (oldDrafts.length) {
-    await GlobalStore.set({ local_drafts: globalStore.local_drafts });
-    await storageLocalRemove(oldDrafts);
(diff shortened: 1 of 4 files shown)


extension/js/background_page/migrations.ts
```diff
@@ -38,13 +40,27 @@ const addKeyInfoFingerprints = async () => {
...
+  const storageLocal = await storageLocalGetAll();
+  if (typeof globalStore.local_drafts === 'undefined') {
+    globalStore.local_drafts = {};
+  }
+  for (const key of Object.keys(storageLocal)) {
+    if (key.startsWith('local-draft-')) {
+      console.log(key);
+      console.info(`migrating local draft ${key}`);
+      globalStore.local_drafts[key] = storageLocal[key] as GmailRes.GmailDraftGet;
+      await storageLocalRemove([key]);
```
rrrooommmaaa, line 60: Perhaps, it'd be better to remove old drafts after successful saving to GlobalStore?

rrrooommmaaa (changes requested):

extension/js/background_page/migrations.ts
limonte: Agree, done 👍

extension/js/background_page/migrations.ts
```diff
@@ -38,13 +40,30 @@ const addKeyInfoFingerprints = async () => {
...
+  const globalStore = await GlobalStore.get(['key_info_store_fingerprints_added', 'local_drafts']);
   if (!globalStore.key_info_store_fingerprints_added) {
     console.info('migrating KeyStorage to add fingerprints and emails of each key...');
     await addKeyInfoFingerprints();
     await GlobalStore.set({ key_info_store_fingerprints_added: true });
     console.info('done migrating');
   }
+  // migrate local drafts (#3337)
+  const storageLocal = await storageLocalGetAll();
+  if (typeof globalStore.local_drafts === 'undefined') {
```
rrrooommmaaa, line 52: Perhaps, if `globalStore.local_drafts` is not undefined, we should not call `storageLocalGetAll()`
(as with other migrations, we can consider this migration done and never return to it).
Or, perhaps, we could just load a flag like `local_drafts_moved` to not cause excessive load here?
Will there rarely be more than one draft?

rrrooommmaaa (changes requested):

extension/js/background_page/migrations.ts
limonte: > Perhaps, if globalStore.local_drafts is not undefined, we should not call storageLocalGetAll()
(as with other migrations, we can consider this migration done and never return to it).

I went with this suggestion, thank you 👍

rrrooommmaaa (approved):

PR #3997, opened 2021-09-21
Save pass phrase for all matching keys

### LINKED ISSUE #3963: need to enter my pass phrase too frequently?
I'm not 100% sure but I seem to be observing that the pass phrases kept in memory expire faster than 4 hours.

I'll be observing it further, to be sure.

limonte:
This PR unlocks all keys that match the pass phrase and keeps the pass phrase for all keys that matched. 

close #3963

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/passphrase.ts
+++ b/extension/chrome/elements/passphrase.ts
@@ -147,7 +147,8 @@ View.run(class PassphraseView extends View {
     const pass = String($('#passphrase').val());
     const storageType: StorageType = ($('.forget-pass-phrase-checkbox').prop('checked') || this.orgRules.forbidStoringPassPhrase()) ? 'session' : 'local';
     let atLeastOneMatched = false;
-    for (const keyinfo of this.keysWeNeedPassPhraseFor!) { // if passphrase matches more keys, it will save the pass phrase for all keys
+    const allPrivateKeys = await KeyStore.get(this.acctEmail);
+    for (const keyinfo of allPrivateKeys) { // if passphrase matches more keys, it will save the pass phrase for all keys
       const prv = await KeyUtil.parse(keyinfo.private);
       try {
         if (await KeyUtil.decrypt(prv, pass) === true) {


limonte:
@tomholub let me know if I'm wrong with choosing "Not worth testing" and this should be covered by tests, thanks!

extension/chrome/elements/passphrase.ts
```diff
@@ -147,7 +147,8 @@ View.run(class PassphraseView extends View {
     const pass = String($('#passphrase').val());
     const storageType: StorageType = ($('.forget-pass-phrase-checkbox').prop('checked') || this.orgRules.forbidStoringPassPhrase()) ? 'session' : 'local';
     let atLeastOneMatched = false;
-    for (const keyinfo of this.keysWeNeedPassPhraseFor!) { // if passphrase matches more keys, it will save the pass phrase for all keys
+    const allPrivateKeys = await KeyStore.get(this.acctEmail);
+    for (const keyinfo of allPrivateKeys) { // if passphrase matches more keys, it will save the pass phrase for all keys
       const prv = await KeyUtil.parse(keyinfo.private);
       try {
         if (await KeyUtil.decrypt(prv, pass) === true) {
```
tomholub, line 154: Looks good but may create side effects: `atLeastOneMatched` now means if it matched any keys at all, as opposed to any of the ones that are needed. I'd rename it to `matchedAtLeastOneKeyWeNeed`, and adjust the logic at `atLeastOneMatched = true;` appropriately to only set it to true if it has the same primary fingerprint as one of the keys in `this.keysWeNeedPassPhraseFor`.

Otherwise, you could have a situation when you have keys [A,B], you need key [A] but match [B]. The dialog will happily go away, yet the message still cannot be decrypted.

tomholub (commented):
Better write a test for this - as the product matures, more types of things should be tested.

A comment below. Thanks!

extension/chrome/elements/passphrase.ts
limonte: Done, thank you 🙏 

Also added the test case, ready for review 🚀

extension/chrome/elements/passphrase.ts
```diff
@@ -147,12 +147,15 @@ View.run(class PassphraseView extends View {
...
           await PassphraseStore.set(storageType, this.acctEmail, keyinfo.fingerprints[0], pass);
-          atLeastOneMatched = true;
+          if (this.longids.includes(keyinfo.longid)) {
```
tomholub, line 156: I'm a bit unsure myself here. I suppose, it would be more robust to check if any of the `this.longids` match any of the longids on that key (yup - there can be more than one per key - typically one for master key packet and one for subkey packet).

That looks something like this:

```ts
const keyLongids = prv.ids.map(OpenPGPKey.fingerprintToLongid)
if (this.longids.includesAnyOf(keyLongids)) {
  atLeastOneMatched = true;
}
```

But that gets even more complicated - because this will only work for OpenPGP keys but not for S/MIME keys. Plus it should be using some generic method like `KeyUtil.fingerprintToLongid` which we don't have.

@rrrooommmaaa I think this is your territory now. Could you please have a look and later finish this PR once you have time? It seems to me that the overall approach is correct but we got stuck on OpenPGP and S/MIME internals that are outside of @limonte's experties and I'm myself unsure what is the best way to proceed.

tomholub (commented):
This looks good but it uncovered a more difficult task to also resolve - I'm marking this as draft, and let's have Roman finalize it, as this is his area.

extension/chrome/elements/passphrase.ts
rrrooommmaaa: S/MIME key only has a single longid -- "Issuer and Serial Number of a certificate" -- I'll add a test for S/MIME decryption.

PR #4003, opened 2021-09-26
issue #3895 Show all fingerprints for recipient

### LINKED ISSUE #3985: show all fingerprints when hovering over recipient
In the past, we used to only have one public key possible per recipient. After that, we refactored some code to allow more than one public key. So now, if I email `joe@example.com`, I may have 3 public keys recorded for them locally, and encrypt the message for all three public keys.

However, when I hover over the recipient in compose box, it will only show one of the public key fingerprints. This is because we didn't refactor all of the code, and somewhere we pull just one public key instead of pulling all as we should to display this.

Instead, it should be showing the primary fingerprint of each public key we have recorded for that person.

Feel free to ask rrrooommmaaa where you need questions asked about the codebase.

IvanPizhenko:
This PR introduces showing all recepient's public key fingerprints

close #3985

- Difficult to test (explain why) - not sure at the moment how to do it.

### To be filled by reviewers


IvanPizhenko:
@tomholub @rrrooommmaaa I think I basically did it, but some test git broken, and from what I see in the CI logs I absolutely don't get any ideas what needs to be fixed. Please have a look on this.

tomholub:
@IvanPizhenko I think you're missing local ci secrets? To run tests locally. See https://github.com/FlowCrypt/flowcrypt-browser/wiki/Writing-and-running-tests meanwhile I'll send them over email.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -196,13 +196,16 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
         if (recipient.status !== RecipientStatus.WRONG) {
           pubkeyLookupRes = await this.view.storageModule.lookupPubkeyFromKeyserversThenOptionallyFetchExpiredByFingerprintAndUpsertDb(recipient.email, undefined);
         } else {
           pubkeyLookupRes = 'wrong';
         }
-        await this.renderPubkeyResult(recipient, pubkeyLookupRes);
+        if (pubkeyLookupRes === 'fail' || pubkeyLookupRes === 'wrong')
+          await this.renderPubkeyResult(recipient, pubkeyLookupRes);
+        else
+          await this.renderPubkeyResult(recipient, pubkeyLookupRes as Contact[]);
```
tomholub, line 208: A detail - please use { } blocks even for single statements.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -798,68 +801,111 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
-      if (dbContact) {
-        recipientEl.element.classList.remove('no_pgp');
-        await this.renderPubkeyResult(recipientEl, dbContact);
+      const dbContacts = await ContactStore.get(undefined, [email]);
+      if (dbContacts && dbContacts.length) {
+        const realDbContacts = dbContacts.filter(contact => contact !== undefined);
+        if (realDbContacts && realDbContacts.length) {
+          recipientEl.element.classList.remove('no_pgp');
+          await this.renderPubkeyResult(recipientEl, realDbContacts as Contact[]);
+        }
```
tomholub, line 813: This code makes me think we should start calling `Contact` something else. Probably `DbPubkey`, as the meaning of the structure has changed over time. That would be for another issue, of course.

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -75,28 +75,36 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
+          this.checkKeyserverForNewerVersionOfKnownPubkeyIfNeeded(storedContact).catch(Catch.reportErr);
+          result.push(storedContact);
+        }
+      }
+      return result;
     }
-    return await this.lookupPubkeyFromKeyserversAndUpsertDb(email, name, storedContact);
+    const res = await this.lookupPubkeyFromKeyserversAndUpsertDb(email, name, undefined);
+    if (res === 'fail') return res;
+    return [res];
```
tomholub, line 107: I know you have to draw the line somewhere, so you don't end up refactoring the whole app :-) And it seems you drew it here - taking a single result and returning it as an array. But there's something wrong about it - can't quite put mi finger on it without investigating further.

Originally, `lookupPubkeyFromKeyserversAndUpsertDb` was working from an existing contact but it no longer does. Maybe this could still be done by running another for loop? Like

```ts
const results: Contact[] = [];
for(const storedContact of storedContacts) {
    const res = await this.lookupPubkeyFromKeyserversAndUpsertDb(email, name, storedContact);
    if (res === 'fail') {
      return res;
    }
    results.append(res);
}
```

It would still need some further thought, and later refactoring, but I think it would be possibly less problematic.

tomholub (commented):
You're right that this is basically it. The broken tests could be that some behavior has changed (maybe for the better?) and the tests will need to be updated. Or a bug caused by this change, or a combination of both - that has to be investigated.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
IvanPizhenko: Posted #4005

extension/chrome/elements/compose-modules/compose-storage-module.ts
IvanPizhenko: ok, will do this

IvanPizhenko:
> @IvanPizhenko I think you're missing local ci secrets? To run tests locally. See https://github.com/FlowCrypt/flowcrypt-browser/wiki/Writing-and-running-tests meanwhile I'll send them over email.

Thanks Tom, yes, I really wasn't even thinking that secrets are needed to run tests locally, but now I see.

IvanPizhenko:
@tomholub I made it passing tests

IvanPizhenko:
Now should I add some UI for this?

IvanPizhenko:
@rrrooommmaaa Ready for review

tomholub:
> Now should I add some UI for this?

I suppose you could modify an existing test. You won't be able to test the hover but you could extract the `title` attribute in a test that has a green recipient (most of them) and in a test that has an expired one. I think that's enough.

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -75,28 +75,39 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
 
   public lookupPubkeyFromKeyserversThenOptionallyFetchExpiredByFingerprintAndUpsertDb = async (
     email: string, name: string | undefined
-  ): Promise<Contact | "fail"> => {
+  ): Promise<Contact[] | "fail"> => {
```
rrrooommmaaa, line 78: Hi @IvanPizhenko 
`Contact` isn't stored in the database
- `test/source/platform/store/contact-store.ts` -- this is a mock database used only in tests, so it is implemented to the extent required for tests to run -- it doesn't fully mimick the real store's behaviour.
- `extension/js/common/platform/store/contact-store.ts` -- this is the real one.
As you can see, only `Pubkey`, `Email` and `Revocation` entities are stored there. `Contact` type is only left for some consumer classes that we haven't refactored yet -- it emulates the older storage by picking the best candidate pubkey for emal and including it in the `Contact`. If consumer class needs more data then the best pubkey, it shouldn't use `Contact` type nor `ContactStore.get`. We will remove these type and method when no more consumers are left (updating the mock `contact-store.ts` in due course).

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -75,28 +75,39 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
-      // This implementation is imperfect in that, if sender didn't pull a particular pubkey
-      //    for a long time and the local pubkey has since expired, and there actually is a
-      //    newer version available on external key server, this may unnecessarily show "bad pubkey",
-      //    until next time user tries to enter recipient in the field again, which will at that point
-      //    get the updated key from db. This could be fixed by:
-      //      - either life fixing the UI after this call finishes, or
-      //      - making this call below synchronous and using the result directly
-      this.checkKeyserverForNewerVersionOfKnownPubkeyIfNeeded(storedContact).catch(Catch.reportErr);
-      return storedContact;
+    const storedContacts = await ContactStore.get(undefined, [email]);
```
rrrooommmaaa, line 85: If only one email is provided, `ContactStore.get` will return a single contact (if email is found) with the best pubkey.
So you should use `ContactStore.getOneWithAllPubkeys` instead. Feel free to add new types if required

extension/js/common/core/crypto/key.ts
```diff
@@ -61,6 +61,15 @@ export type Contact = {
   revoked: boolean;
 };
 
+export class ContactUtil {
+
+public static isExpired = (contact: Contact) => {
+  return contact.pubkey && ((contact.expiresOn || Infinity) <= Date.now()
```
rrrooommmaaa, line 67: this isn't necessary.
As `Contact` isn't stored in the database, its `expiredOn` attribute is calculated from its pubkey in `ContactStore.getKeyAttributes`. Anyway, the `Contact` class will probably be not required for this feature.

rrrooommmaaa (changes requested):
Per my comments in the PR

IvanPizhenko:
@tomholub After last commit I've got failing unit test:
```
[unit-Sks.js] Sks lookup pubkey - from live host [ALL RETRIES FAILED for [unit-Sks.js] Sks lookup pubkey - from live host] 04:07
    ℹ Expected unit test to return "pass" but got: "unit test threw something:Error: Expecting key.id=6BF16EE1ECE7A66C4B6636DF0C9C2E6A4D273C6F but got 8B8A05A2216EE6E4C5EE3D540D5688EBF3102BE7 04:07
      Error: Expecting key.id=6BF16EE1ECE7A66C4B6636DF0C9C2E6A4D273C6F but got 8B8A05A2216EE6E4C5EE3D540D5688EBF3102BE7 04:07
          at __puppeteer_evaluation_script__:11:11 04:07
          at async __puppeteer_evaluation_script__:4:24" 04:07
    ℹ (attempt 1 of 1) Failed:   Error: unit test threw something:Error: Expecting key.id=6BF16EE1ECE7A66C4B6636DF0C9C2E6A4D273C6F but got 8B8A05A2216EE6E4C5EE3D540D5688EBF3102BE7 04:07
          at /home/semaphore/git/flowcrypt-browser/test/source/tests/unit-browser.ts:45:17 04:07
          at /home/semaphore/git/flowcrypt-browser/test/source/test.ts:76:7 04:07
          at /home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:143:61 04:07
          at BrowserPool.withNewBrowserTimeoutAndRetry (/home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:143:11) 04:07
          at /home/semaphore/git/flowcrypt-browser/test/source/test.ts:71:5 04:07
          at <async> <anonymous> (/test/source/tests/unit-browser.ts:25:91) 04:07
          at <async> <anonymous> (../test/source/test.ts:71:53) 04:15
          at <async> <anonymous> (/test/source/browser/browser-pool.ts:143:49)
```
Test receives key for `human at flowcrypt.com` from Attester. Did keys change and test must be updated with new key fingerprint?

tomholub:
here https://github.com/FlowCrypt/flowcrypt-browser/issues/4009

IvanPizhenko:
> here #4009

Thanks. I will update tests.

IvanPizhenko:
Failing test:
```
can lookup public key from WKD directly [ALL RETRIES FAILED for can lookup public key from WKD directly] 03:03
    ℹ (attempt 1 of 3) Retrying: Error: waiting failed: Elements did not appear: .email_address.has_pgp 03:03
    ℹ (attempt 2 of 3) Retrying: Error: waiting failed: Elements did not appear: .email_address.has_pgp 03:03
    ℹ (attempt 3 of 3) Failed:   Error: waiting failed: Elements did not appear: .email_address.has_pgp 03:03
          at ControllablePage.ControllableBase.waitAny (/home/semaphore/git/flowcrypt-browser/test/source/browser/controllable.ts:84:11) 03:03
          at ControllablePage.ControllableBase.waitForContent (/home/semaphore/git/flowcrypt-browser/test/source/browser/controllable.ts:317:5) 03:03
          at /home/semaphore/git/flowcrypt-browser/test/source/tests/compose.ts:1033:7 03:03
          at /home/semaphore/git/flowcrypt-browser/test/source/test.ts:76:7 03:03
          at /home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:143:61 03:03
          at BrowserPool.withNewBrowserTimeoutAndRetry (/home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:143:11) 03:03
          at /home/semaphore/git/flowcrypt-browser/test/source/test.ts:71:5 03:03
          at <async> waitAny (/test/source/browser/controllable.ts:64:19) 03:03
          at <async> waitForContent (/test/source/browser/controllable.ts:316:26) 03:03
          at <async> <anonymous> (/test/source/tests/compose.ts:1030:93) 03:03
          at <async> <anonymous> (../test/source/test.ts:71:53) 03:03
          at <async> <anonymous> (/test/source/browser/browser-pool.ts:143:49)
```

IvanPizhenko:
Browser console shows errors:
```
Failed to load resource: net::ERR_NAME_NOT_RESOLVED
openpgpkey.flowcrypt.test/.well-known/openpgpkey/flowcrypt.test/policy:1

Failed to load resource: net::ERR_NAME_NOT_RESOLVED
flowcrypt.test/.well-known/openpgpkey/policy:1 Failed to load resource: net::ERR_NAME_NOT_RESOLVED
```
But I don't have `flowcrypt.test` in the `/etc/hosts`. Will try to add it.

rrrooommmaaa:
> @tomholub what this could be? Why does it go to that domain?

It tries to search user's pubkey via WKD. Obviously you have user @flowcrypt.test

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -798,58 +807,71 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
-      Xss.sanitizePrepend(el, '<img class="lock-icon" src="/img/svgs/locked-icon.svg" />');
-      $(el).attr('title', 'Does use encryption' + this.recipientKeyIdText(contact));
+    } else if (sortedPubKeyInfos.length) {
+      // New logic
+      // - if there is at least one valid (non-expired, non-revoked) public key, then it's HAS_PGP
+      // - else if there is at least one expired public key, then it's EXPIRED
+      // - else if there is at least one revoked key, then REVOKED
+      // - else it's NO_PGP
+      const firstKeyInfo = sortedPubKeyInfos[0];
+      if (!firstKeyInfo.revoked && !PubKeyInfoUtil.isExpired(firstKeyInfo)) {
```
rrrooommmaaa, line 855: Actually, a valid key is a key that has `!firstKeyInfo.revoked && firstKeyInfo.pubkey.usableForEncryption` (there was already time check at the moment of `Key` object creation).
Ok, if you want to double-check if the key hasn't expired within seconds (minutes) after retrieval, you can additionally `&& !KeyUtil.expired(firstKeyInfo.pubkey)`

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -798,58 +807,71 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
+        recipient.status = RecipientStatus.HAS_PGP;
+        $(el).addClass('has_pgp');
+        Xss.sanitizePrepend(el, '<img class="lock-icon" src="/img/svgs/locked-icon.svg" />');
+        $(el).attr('title', 'Does use encryption\n' + this.publicKeysToRenderedText(sortedPubKeyInfos));
+      } else if (PubKeyInfoUtil.isExpired(firstKeyInfo)) {
```
rrrooommmaaa, line 860: The code will get here if the key is both revoked and expired, right?
So the key will be marked as "EXPIRED" and potentially usable.
While revoked keys should not be used under any circumstances as they were compromised.
So you need to move the `revoked` clause up one level. Only remaining non-revoked keys can be "EXPIRED"
There might be (in rare cases) keys left that are neither usableForEncryption, nor revoked, nor expired (example -- weak keys). So it makes sense to check `KeyUtil.expired(firstKeyInfo.pubkey)` in the one-before-last `else if` clause to label the key as "EXPIRED", all other keys can be classified as NO_PGP. Or, maybe, @tomholub has an idea on how to mark weak keys?

rrrooommmaaa (changes requested):

tomholub:
> Tried to run mentioned above test locally. Browser console shows 2 errors:
> 
> ```
> Failed to load resource: net::ERR_NAME_NOT_RESOLVED
> openpgpkey.flowcrypt.test/.well-known/openpgpkey/flowcrypt.test/policy:1
> 
> Failed to load resource: net::ERR_NAME_NOT_RESOLVED
> flowcrypt.test/.well-known/openpgpkey/policy:1 Failed to load resource: net::ERR_NAME_NOT_RESOLVED
> ```
> 
> But I don't have `flowcrypt.test` in the `/etc/hosts` and instructions in wiki don't mention such host. I also can't find any setup for this domain in the `semaphore.yml`. @tomholub what this could be? Why does it go to that domain?

These are benign errors. It's how WKD works.

I want to email joe@example.com then it tries a WKD server on openpgpkey.example.com. If the server is there, good - I grab the public key. If it's not there, then I could not find the public key.

Therefore it's normal to see these in the browser console, and I suppose they are not the cause of that broken test - something else is breaking it, I think.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
IvanPizhenko: For now I will change order of checks, so check for revoked will be first.

IvanPizhenko:
Now another test failure:
```
x send with several S/MIME certs [ALL RETRIES FAILED for send with several S/MIME certs] 02:16
    ℹ (attempt 1 of 3) Retrying: Error: waiting failed: Elements did not appear: .email_address.no_pgp 02:16
    ℹ (attempt 2 of 3) Retrying: Error: waiting failed: Elements did not appear: .email_address.no_pgp 02:16
    ℹ (attempt 3 of 3) Failed:   Error: waiting failed: Elements did not appear: .email_address.no_pgp
```

IvanPizhenko:
Fixed those tests. Let's see, if something else fails...

IvanPizhenko:
More failing tests:
```
consumer mock group

x compose - load contacts - contacts should be properly ordered [ALL RETRIES FAILED for compose - load contacts - contacts should be properly ordered]
    ℹ (attempt 1 of 3) Retrying: AssertionError: expected 'No Contacts Found' to equal 'testsearchorder1@flowcrypt.com'
    ℹ (attempt 2 of 3) Retrying: AssertionError: expected 'No Contacts Found' to equal 'testsearchorder1@flowcrypt.com'
    ℹ (attempt 3 of 3) Failed:   AssertionError: expected 'No Contacts Found' to equal 'testsearchorder1@flowcrypt.com'
```

```

**Comment by IvanPizhenko** [comment]
@tomholub Please look at latest Semaphore run https://flowcrypt.semaphoreci.com/workflows/c43b9f2b-3f69-4d9d-891a-412dcbc26a0f?pipeline_id=cc23d73b-14ee-42ac-b75c-ccb7bbb1e1df

There 2 failing tests:

**consumer mock - standard test group**
```
x compose - load contacts - contacts should be properly ordered [ALL RETRIES FAILED for compose - load contacts - contacts should be properly ordered]
    ℹ (attempt 1 of 3) Retrying: AssertionError: expected 'No Contacts Found' to equal 'testsearchorder1@flowcrypt.com'
    ℹ (attempt 2 of 3) Retrying: AssertionError: expected 'No Contacts Found' to equal 'testsearchorder1@flowcrypt.com'
    ℹ (attempt 3 of 3) Failed:   AssertionError: expected 'No Contacts Found' to equal 'testsearchorder1@flowcrypt.com'
```

**Live Gmail tests**
```
x mail.google.com - success notif after setup, click hides it, does not re-appear + offers to reauth [ALL RETRIES FAILED for mail.google.com - success notif after setup, click hides it, does not re-appear + offers to reauth] 05:55
    ℹ (attempt 1 of 3) Retrying: AssertionError: expected 'Please reconnect FlowCrypt to your Gmail Account. This is typically needed after a long time of no use, a password change, or similar account changes. Re-connect Account' to include 'Connected successfully. You may need to reload the tab.' 05:55
    ℹ (attempt 2 of 3) Retrying: Error: Protocol error (Runtime.callFunctionOn): Argument should belong to the same JavaScript world as target object 05:55
    ℹ (attempt 3 of 3) Failed:   Error: Protocol error (Runtime.callFunctionOn): Argument should belong to the same JavaScript world as target object
```
"compose - load contacts - contacts should be properly ordered" fails in CI but passes locally when running as the `.only` test.
"mail.google.com - success notif after setup ... " -  **status to be updated**

**Comment by rrrooommmaaa** [comment]
> Any ideas how to deal with this?

`compose -` tests may be run in parallel with other tests. Perhaps, another test is influencing it in a mock backend or something.
Is this failure due to changes in this PR?

**Comment by tomholub** [comment]
Moving the test to flaky group of tests (see CI separate group) will make it run serially. You could verify that way.

**Comment by IvanPizhenko** [comment]
ok, I will try to move it to flaky tests...

**Comment by IvanPizhenko** [comment]
> Moving the test to flaky group of tests (see CI separate group) will make it run serially. You could verify that way.

Yes, works in "flaky", so leaving it there.

**Comment by IvanPizhenko** [comment]
Surprisingly, GMail tests also have passed now.

**Comment by IvanPizhenko** [comment]
@tomholub @rrrooommmaaa Finally, now all tests are passing, so it is ready for review again.

**Comment by IvanPizhenko** [comment]

**Comment by tomholub** [comment]
Thank you Ivan - I won't be generally reviewing browser extension as I got busy on iOS and customer deployments until around November.

**Comment by IvanPizhenko** [comment]
ok, so then waiting for review from @rrrooommmaaa

File: extension/chrome/elements/compose-modules/compose-recipients-module.ts [hunk 9]
```diff
@@ -860,6 +886,48 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
+    for (const pubKeyInfo of pubKeyInfos) {
+      if (pubKeyInfo.revoked) {
+        revoked.push(pubKeyInfo);
+      } else if (KeyUtil.expired(pubKeyInfo.pubkey)) {
+        expired.push(pubKeyInfo);
+      } else {
+        valid.push(pubKeyInfo);
+      }
+    }
+
```
rrrooommmaaa, line 902: @IvanPizhenko you see `linter` complains about these empty lines.
You should either remove them, or place some remarks there

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -75,28 +75,52 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
+          //    newer version available on external key server, this may unnecessarily show "bad pubkey",
+          //    until next time user tries to enter recipient in the field again, which will at that point
+          //    get the updated key from db. This could be fixed by:
+          //      - either life fixing the UI after this call finishes, or
+          //      - making this call below synchronous and using the result directly
+          this.checkKeyserverForNewerVersionOfKnownPubkeyIfNeeded(storedContact.info.email, pubkey)
+            .catch(Catch.reportErr);
+          results.push(pubkey);
+        } else {
+          const res = await this.lookupPubkeyFromKeyserversAndUpsertDb(email, name, pubkey);
```
rrrooommmaaa, line 105: why call `lookup...()` several times for the same `email`? actually `pubkey` parameter is no longer needed as this method no longer needs to form a "contact". And this method should return PubkeyInfo[]. (please check for duplicates with existing set).

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -114,8 +138,8 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
    *    discussion: https://github.com/FlowCrypt/flowcrypt-browser/pull/3898#discussion_r686244628
    */
   public lookupPubkeyFromKeyserversAndUpsertDb = async (
-    email: string, name: string | undefined, existingContact: Contact | undefined
-  ): Promise<Contact | "fail"> => {
+    email: string, name: string | undefined, existingPubKey: PubKeyInfo | undefined
+  ): Promise<PubKeyInfo | "fail" | undefined> => {
```
rrrooommmaaa, line 142: I guess this should be `Promise<PubkeyInfo[]> |  'fail'`?

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -138,7 +162,9 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
             updates.push({ name } as ContactUpdate);
           } else {
             // No public key found. Returning early, nothing to update in local store below.
-            return existingContact ?? await ContactStore.obj({ email });
+            if (existingPubKey) return existingPubKey;
+            const contact = await ContactStore.obj({ email });
+            return ContactUtil.toPubKeyInfo(contact);
```
rrrooommmaaa, line 167: These to lines will simply return `undefined`. Thus `toPubkeyInfo()` is not required

extension/chrome/elements/compose-modules/compose-storage-module.ts
rrrooommmaaa: UPD: it should be `Promise<'fail' | 'success'> or `Promise<boolean>` as re-reading the storage perplexes some of the callers (`lookupPubkeyFromKeyserversThenOptionallyFetchExpiredByFingerprintAndUpsertDb`)

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -147,45 +173,43 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
         if (updates.length) {
           await Promise.all(updates.map(async (update) => await ContactStore.update(undefined, email, update)));
         }
-        const [preferred] = await ContactStore.get(undefined, [email]);
-        return preferred ?? PUBKEY_LOOKUP_RESULT_FAIL;
-      } else {
-        return PUBKEY_LOOKUP_RESULT_FAIL;
+        const emailWithKeys = await ContactStore.getOneWithAllPubkeys(undefined, email);
```
rrrooommmaaa, line 176: Actually, this method is re-reading the storage to provide the "result". I think it should simply return 'fail' | 'success' or boolean

extension/chrome/elements/compose-modules/compose-storage-module.ts
rrrooommmaaa: UPD: as `lookupPubkeyFromKeyserversAndUpsertDb` re-reads the storage, this is all wrong.
We should call `lookupPubkeyFromKeyserversAndUpsertDb(email, name)` and re-read from storage

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -75,28 +75,52 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
-  ): Promise<Contact | "fail"> => {
+  ): Promise<PubKeyInfo[] | "fail"> => {
     // note by Tom 2021-08-10: We are only getting one public key from storage, but should
     //    work with arrays instead, like with `ContactStore.getOneWithAllPubkeys`.
     //    However, this whole fingerprint-base section seems unnecessary, we could likely
     //    remove it and the keys will be updated below using
     //    `lookupPubkeyFromKeyserversAndUpsertDb` anyway.
     //    discussion: https://github.com/FlowCrypt/flowcrypt-browser/pull/3898#discussion_r686229818
-    const [storedContact] = await ContactStore.get(undefined, [email]);
-    if (storedContact && storedContact.hasPgp && storedContact.pubkey && !storedContact.revoked) {
```
rrrooommmaaa, line 86: The point of this condition is to be able to return immediately if we have a valid key (and initiating an async operation for storage update)

extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -75,28 +75,52 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
...
+          //    for a long time and the local pubkey has since expired, and there actually is a
+          //    newer version available on external key server, this may unnecessarily show "bad pubkey",
+          //    until next time user tries to enter recipient in the field again, which will at that point
+          //    get the updated key from db. This could be fixed by:
+          //      - either life fixing the UI after this call finishes, or
+          //      - making this call below synchronous and using the result directly
+          this.checkKeyserverForNewerVersionOfKnownPubkeyIfNeeded(storedContact.info.email, pubkey)
+            .catch(Catch.reportErr);
+          results.push(pubkey);
+        } else {
```
rrrooommmaaa, line 104: Now, with this loop, if have at least one revoked key, this method will wait for lookup completion, thus not providing immediate result. This is not needed if we have at least one valid key. If we have a valid key, we rightaway color the recipient green or yellow (expired but possible to encrypt) and do the `lookup` asynchronously

rrrooommmaaa (changes requested):
@IvanPizhenko I left plenty of comments. `lookup...` methods can be simplified. They re-read the storage to provide return valus, the don't need to. The caller can do this itself.

extension/chrome/elements/compose-modules/compose-storage-module.ts
IvanPizhenko: Why these two lines simply return `undefined`?? Sorry, but I can't understand this at the moment. Can you please provide more detailed explanation explanation? (p.s. I just was trying to port existing code without losing any already implemented functionality.)

extension/chrome/elements/compose-modules/compose-storage-module.ts
IvanPizhenko: Still don't catch how return of public key can be changed to just boolean?

extension/chrome/elements/compose-modules/compose-storage-module.ts
IvanPizhenko: Why? It must return a key. How this can be changed to just boolean?

extension/chrome/elements/compose-modules/compose-storage-module.ts
IvanPizhenko: @rrrooommmaaa I think I understood what is the problem, but I think absolutely don't understand how I should solve it,

IvanPizhenko:
> @IvanPizhenko I left plenty of comments. `lookup...` methods can be simplified. They re-read the storage to provide return values, the don't need to. The caller can do this itself.

@rrrooommmaaa It all good, but I could understood only some of your comments regarding `lookup` methods. I would say I don't catch what should I do now to solve all this, so I think my attempts to do something now will be more like "guessing on the coffee" than programming when I know what I do. So I suggest you to make all these changes to `lookup...` methods and its usages. And I will study what you've done to increase my understanding and use it for some future tasks.

IvanPizhenko:
@rrrooommmaaa Alternatively, we can merge it now as is, and have a separate task (for you) to do all these changes a bit later.

IvanPizhenko:
ok, this refactors method `reRenderRecipientFor`. What about `lookup...` methods?

IvanPizhenko:
> > ok, this refactors method `reRenderRecipientFor`. What about `lookup...` methods?
> 
> Not all callers of `lookupPubkeyFromKeyserversAndUpsertDb` require the results, so we don't need to re-query the storage. We should probably refactor it further to return `Promise<void>` as per Tom's remark and use `catch`
@rrrooommmaaa 
I could find only 2 calls of this methods:
```typescript
    if (!await this.lookupPubkeyFromKeyserversAndUpsertDb(email, name)) {
      return PUBKEY_LOOKUP_RESULT_FAIL;
    }
```
and
```
    await Promise.all(toLookup.map(c => this.view.storageModule.lookupPubkeyFromKeyserversAndUpsertDb(c.email, c.name || undefined).then(lookupRes => {
      if (!lookupRes) {
        this.failedLookupEmails.push(c.email);
      }
    })));
```
Which of them "doesn't require the results" and why? I assumed the second one, but it likely needs to be sure that new contact is in the DB and preferably some valid key is available, but initially it is not, so you have to try to lookup on the remote keyserver.
Is the whole thing about "not require result" is about using boolean return flag vs exception to report an error?

rrrooommmaaa:
> Which of them "doesn't require the results" and why? I assumed the second one,

Yes, the second one only needs to warn in a situation when 1) lookup resulted in network failure or 2) saving to the store caused an exception. These are the situations when `lookupPubkeyFromKeyserversAndUpsertDb` returns `false`. 
We can refactor it to (re-)throw an exception and rewrite the second call with `.catch(...)` as Tom suggested.
If lookup yielded no keys -- that's not an error, that's the NO_PGP situation.

The first call extracts the keys from the storage if no error happens.
```
    // re-query the storage
    return (await ContactStore.getOneWithAllPubkeys(undefined, email))?.sortedPubkeys ?? [];
```
If no keys, it will return an empty set (this is not 'fail').
Previously, re-quering was done in `lookupPubkeyFromKeyserversAndUpsertDb` after all updates to return the sorted keys, but this is not required for the second case, so I made this method lighter, removed re-querying.
And one more thing -- it may look like `lookupPubkeyFromKeyserversAndUpsertDb` can format `PubkeyInfo[]` out of `Key[]` it obtained, right? That's not good, because in the storage we have the table `Revocations` that stores fingerprints of revoked keys that may not exist in the database (`Pubkeys` table) -- that is pre-emptive external revocation. So (in a rare case) the lookup method receives a valid key, saves it to the storage, and after re-querying the storage, this key maybe returned as revoked. This is why `PubkeyInfo` has `revoked` property regardless of the fact that `Key` itself also has it.

IvanPizhenko:
> This is why `PubkeyInfo` has `revoked` property regardless of the fact that `Key` itself also has it.

Maybe this should be documented via some comments in the code.

test/source/tests/util.ts
```diff
@@ -0,0 +1,33 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
```
rrrooommmaaa, line 1: I guess these methods should go to `test/source/tests/page-recipe/compose-page-recipe.ts`

rrrooommmaaa:
> Maybe this should be documented via some comments in the code.

Sure. Where would the appropriate place be, in your opinion?

IvanPizhenko:
@rrrooommmaaa So I've tried to implement all the notes above. Are there any other improvements to be done here?

rrrooommmaaa:
Should look like
```
Valid public key fingerprints:
FF5E C310 4388 CA0F 4014 EBC9 2B6A C469 03E7 16E4 (openpgp)
6FE1 16D2 759F 0FFA C562 3E7E 10D6 E379 41EA A0BB (x509)

Expired public key fingerprints:
whatever (openpgp)

Revoked public key fingerprints:
D666 2C5F B9BD E9DA 01F3 994A AA1E F832 D8CC A4F2 (openpgp)
```

rrrooommmaaa:
I suggest to display `x509` instead of `s/mime` as we do on other pages, e.g. Contacts.

IvanPizhenko:
@rrrooommmaaa Fixed it

PR #4013, opened 2021-09-28
issue 3751 use client-configuration endpoint on FES

### LINKED ISSUE #3751: load Client Configuration from auth-less url on FES
Today we use `GET <fes>/api/v1/account/` which requires authentication with an access token or OIDC.

For now we could keep the access token request during setup as is (in case it's needed for other endpoints later like logging), however when actually grabbing OrgRules from the FES, we should use `GET <fes>/api/v1/client-configuration?domain=example.com` API instead.

The new API is missing `FcAccount$info` which contains `default_message_expire` which has to be thought about further.

tomholub:
This PR uses this endpoint https://fes.flowcrypt.com/api/v1/client-configuration?domain=flowcrypt.com (or similar depending on customer domain) to fetch client configuration (org rules) from FES, instead of the `/api/v1/account/` endpoint which was authenticated.

This is to avoid relying on a non-OIDC access token, which our customer doesn't prefer.

close #3751

- Tests updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/api/account-server.ts
+++ b/extension/js/common/api/account-server.ts
@@ -30,7 +30,18 @@ export class AccountServer extends Api {
   public accountGetAndUpdateLocalStore = async (fcAuth: FcUuidAuth): Promise<BackendRes.FcAccountGet> => {
     if (await this.isFesUsed()) {
       const fes = new EnterpriseServer(this.acctEmail);
-      return await fes.getAccountAndUpdateLocalStore();
+      const fetchedOrgRules = await fes.fetchAndSaveOrgRules();
+      return {
+        domain_org_rules: fetchedOrgRules,
+        // the subscription and default_message_expire below is a refactor relic
+        //  (used to come from a deprecated FES API that was authenticated)
+        // todo - remove this - issue #4012
+        subscription: { level: 'pro', expired: false },
+        // todo - rethink this. On FES, expiration is handled with S3 bucket policy regardless of this number
+        //  which is set to 180 days on buckets we manage. This number below may still be rendered somewhere
+        //  when composing, which should be evaluated.
+        account: { default_message_expire: 180 }
+      };
     } else {
       return await FlowCryptComApi.accountGetAndUpdateLocalStore(fcAuth);
     }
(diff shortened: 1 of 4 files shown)


extension/js/common/api/account-server.ts
```diff
@@ -30,7 +30,18 @@ export class AccountServer extends Api {
...
+      return {
+        domain_org_rules: fetchedOrgRules,
+        // the subscription and default_message_expire below is a refactor relic
+        //  (used to come from a deprecated FES API that was authenticated)
+        // todo - remove this - issue #4012
+        subscription: { level: 'pro', expired: false },
+        // todo - rethink this. On FES, expiration is handled with S3 bucket policy regardless of this number
+        //  which is set to 180 days on buckets we manage. This number below may still be rendered somewhere
+        //  when composing, which should be evaluated.
+        account: { default_message_expire: 180 }
```
tomholub, line 43:
```ts
        // todo - rethink this. On FES, expiration is handled with S3 bucket policy regardless of this number
        //  which is set to 180 days on buckets we manage. This number below may still be rendered somewhere
        //  when composing, which should be evaluated.
        account: { default_message_expire: 180 }
```

check how `default_message_expire` affects the extension when FES is used. I suppose it's used to render a hint to the user. (and I know it's not used to decide the actual expiration)

test/source/mock/fes/fes-endpoints.ts
```diff
@@ -43,15 +43,13 @@ export const mockFesEndpoints: HandlersDefinition = {
     }
     throw new HttpClientErr('Not Found', 404);
   },
-  '/api/v1/account/': async ({ }, req) => {
+  '/api/v1/client-configuration': async ({ }, req) => {
+    // ?domain=fes.localhost:8001
+    // ?domain=fes.standardsubdomainfes.test:8001
+    // ?domain=fes.google.mock.flowcryptlocal.test:8001
```
tomholub, line 49: Need to see if it will work this way in mock tests. Also, should do an additional check on the domain in this mock endpoint.

extension/js/common/api/account-servers/flowcrypt-com-api.ts
```diff
@@ -20,7 +20,7 @@ export type SubscriptionInfo = { level?: SubscriptionLevel; expired?: boolean };
 
 export namespace BackendRes {
   export type FcAccountLogin = { registered: boolean, verified: boolean };
-  export type FcAccount$info = { alias: string, email: string, intro: string, name: string, photo: string, default_message_expire: number };
+  export type FcAccount$info = { alias?: string | null, default_message_expire: number };
```
tomholub, line 23: These were unused, after we moved Encrypted Contact Page setup to flowcrypt.com/account/

tomholub:
I'll self-review this because it's due soon for a security penetration testing that was scheduled.

PR #4015, opened 2021-09-28
issue 4002 DISABLE_FES_ACCESS_TOKEN OrgRule

### LINKED ISSUE #4002: FES message upload to use JWT directly with DISABLE_FES_ACCESS_TOKEN OrgRule
Currently, during setup, the app will send Google JWT to FES, to get back an access token. Then uses the access token to upload password messages.

Customer who uses our FES prefers that it does not use any access tokens. Since they are our primary customer, instead of making this configurable, I'll  follow their preference to not use access tokens when communicating with FES. That means that their users will be using Google JWT to upload password protected messages, with a 1 hour expiration and have to re-authenticate after that.

tomholub:
This PR adds an OrgRule `DISABLE_FES_ACCESS_TOKEN` that will cause all calls to FES to contain IdP ODIC instead of an access token, per enterprise customer request.

close #4002

progress:
 - [x] add the org rule
 - [x] use it during setup
 - [x] use it when sending password protected message (getting reply token, upload)
 - [x] do manual testing of sending
 - [x] do manual testing of renewing token
 - [x] add or adjust automated tests

- Tests added or updated **(soon will be added to this PR)**

### To be filled by reviewers

- [x] is documented clearly and usefully, or doesn't need documentation -> docs updated here: https://github.com/FlowCrypt/flowcrypt-docs/commit/1075ec09c3e87198edbbc1a655020b7414bc2da1


tomholub:
The following flow is not ideal:
1) set up with account that has `DISABLE_FES_ACCESS_TOKEN` (newly all `flowcrypt.com` accounts, or for tests `user@fes.disablefesaccesstoken.test:8001` account
2) restart browser (to forget in-memory ID Token)
3) compose email and choose recipient without pubkey, add password, click send
4) authentication dialog appers (this is correct)
5) go through / re-authenticate. dialog disappears

expected result: there should be no authentication-related prompt on top of Gmail page anymore, and the message sending should be automatically retried (or, a dialog should appear prompting user to retry - maybe better)

actual result: there is a confusing authentication prompt  (yellow notification on top of gmail) plus user now doesn't know they should retry sending

tomholub (commented):
LGTM

PR #4026, opened 2021-09-29
issue #3126 Add support for multiple offline compose drafts

### LINKED ISSUE #3126: more than one offline draft should be possible + should show in drafts list

limonte:
This PR allows users to create several offline compose drafts. It also takes care of migrating the existed compose offline drafts, so nothing will be lost for users.

close #3126

CleanShot 2021-09-30 at 11 20 41@2x

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/conf/tslint.yaml
+++ b/conf/tslint.yaml
@@ -10,7 +10,7 @@ rules: #
     object-literal-sort-keys: false
     arrow-parens: false
     no-console: false
-    prefer-const: true
+    prefer-const: false
     ban-types:
      - true
      - ["Object", "Avoid using the `Object` type. Did you mean `object`?"]
@@ -25,7 +25,7 @@ rules: #
      - true
      - allow-leading-underscore
     member-access: true
-    member-ordering: 
+    member-ordering:
       - true
       - order:
         - "public-static-field"
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -757,7 +757,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
       }
       const offlineComposeDrafts: Dict<LocalDraft> = {};
       for (const draftId in storage.local_drafts) {
-        if (draftId.startsWith('local-draft-compose-')) {
+        if (draftId.startsWith('local-draft-compose-') && storage.local_drafts[draftId].accEmail === this.acctEmail) {
           offlineComposeDrafts[draftId] = storage.local_drafts[draftId];
         }
       }
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -11,7 +11,7 @@ import { ApiErr } from '../../common/api/shared/api-error.js';
 import { Attachment } from '../../common/core/attachment.js';
 import { BrowserMsg } from '../../common/browser/browser-msg.js';
 import { Catch } from '../../common/platform/catch.js';
-import { GlobalStore } from '../../common/platform/store/global-store.js';
+import { GlobalStore, LocalDraft } from '../../common/platform/store/global-store.js';
 import { Gmail } from '../../common/api/email-provider/gmail/gmail.js';
 import { Injector } from '../../common/inject.js';
 import { PubLookup } from '../../common/api/pub-lookup.js';
@@ -755,10 +755,10 @@ export class GmailElementReplacer implements WebmailElementReplacer {
       if (!storage.local_drafts) {
         return;
       }
-      const offlineComposeDrafts: Dict<number> = {};
+      const offlineComposeDrafts: Dict<LocalDraft> = {};
       for (const draftId in storage.local_drafts) {
         if (draftId.startsWith('local-draft-compose-')) {
-          offlineComposeDrafts[draftId] = parseInt(storage.local_drafts[draftId].id);
+          offlineComposeDrafts[draftId] = storage.local_drafts[draftId];
         }
       }
       let offlineDraftsContainer = $(this.sel.draftsList).find('#fc_offline_drafts');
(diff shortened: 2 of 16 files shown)


limonte:
@tomholub the issue description wasn't specific about how to represent offline drafts, so I made a separate section "FlowCrypt offline drafts" with the list of drafts with their saving datetime as textual representation. Drafts are sorted by date desc.

Let me know if I missed something here or you'd like to change anything.

tomholub:
> @tomholub the issue description wasn't specific about how to represent offline drafts, so I made a separate section "FlowCrypt offline drafts" with the list of drafts with their localized datetime as textual representation. Drafts are sorted by date desc.
> 
> Let me know if I missed something here or you'd like to change anything.

Looks good. Haven't looked at code (will be afk) for the rest of the week - will let Roman review

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -746,4 +749,47 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
+          offlineComposeDrafts[draftId] = parseInt(storage.local_drafts[draftId].id);
+        }
+      }
+      let offlineDraftsContainer = $(this.sel.draftsList).find('#fc_offline_drafts');
+      if (!offlineDraftsContainer.length) {
+        offlineDraftsContainer = $('<div id="fc_offline_drafts"></div>');
+        $(this.sel.draftsList).append(offlineDraftsContainer); // xss-safe-value
+      }
+      if (Object.keys(offlineComposeDrafts).length) {
+        const draftsToRender: JQueryEl[] = [];
```
rrrooommmaaa, line 770: why is this array `draftsToRender` keyed differently than `storage.local_drafts`? Is this only to call `sort()` without a sorting function/lambda parameter?
`local-draft-compose-HMhqrZKsUo`  should be a good key, and its uniqueness is in line with the storage implementation. I see the click  handler operates with storage keys. 

The first thing that perplexed me is why `id` inside a draft object differs from the outer `draftId` in storage.
Is the only purpose to make it sortable? I guess this is unnecessary.
Also, it is named `id` as if its meaning is unknown (opaque) to the outside modules, and then on line 771 you apply the knowledge that it is actually a timestamp to format it with `new Date(...).toLocaleString()`.
This knowledge (requirement) should be explicitly reflected in the property name.
So I suggest for drafts to not have `id` field (or matching with the outside id, if required).
Have `timestamp` field instead. It is usually inique but not really required to be.
Do not create any dictionaries with `timestamp` as key. This is unnecessary complication and uniqueness requirement.

test/source/tests/gmail.ts
```diff
@@ -253,6 +253,19 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       await pubkeyPage.waitForContent('@container-pgp-pubkey', 'Fingerprint: 50B7 A032 B5E1 FBAB 24BA B205 B362 45FD AC2F BF3D');
     }));
 
+    ava.default('mail.google.com - saving and rendering a compose draft when offline', testWithBrowser('ci.tests.gmail', async (t, browser) => {
+      const gmailPage = await openGmailPage(t, browser, '/');
+      // create compose draft
+      await gmailPage.waitAndClick('@action-secure-compose', { delay: 1 });
+      await createSecureDraft(t, browser, gmailPage, 'compose draft', { offline: true });
```
rrrooommmaaa, line 260: I would like to see adding more offline drafts here, checking the list of local drafts, opening one of them and checking that the correct one is opened.
Was there a problem with that?
I guess we can make the whole `gmailPage` offline for the period of testing? 
I managed to do this in a broswer, I guess, it can be puppeteered too.

rrrooommmaaa (changes requested):

extension/js/content_scripts/webmail/gmail-element-replacer.ts
limonte: Thank you for the advice @rrrooommmaaa, I followed it and the code is cleaner now 👍

test/source/tests/gmail.ts
limonte: Done 👍

limonte:
Thank you for the review @rrrooommmaaa! I fixed the issues you've mentioned and also added one unrelated commit https://github.com/FlowCrypt/flowcrypt-browser/pull/4026/commits/4dfee3c6a66f1161b81f08f57414b1421aef0051 which moves one rule from tslint to eslint. The reason for that is that eslint would autofix that rule for us and tslint is deprecated.

limonte:
ping @rrrooommmaaa

conf/tslint.yaml
```diff
@@ -10,7 +10,7 @@ rules: #
     object-literal-sort-keys: false
     arrow-parens: false
     no-console: false
-    prefer-const: true
+    prefer-const: false
```
rrrooommmaaa, line 13: why don't we prefer constants?

limonte: I mentioned this change in [my previous comment](https://github.com/FlowCrypt/flowcrypt-browser/pull/4026#issuecomment-932503485):

> also added one unrelated commit 4dfee3c which moves the prefer-const rule from tslint to eslint. The reason for that is that eslint would autofix that rule for us and tslint is deprecated.

test/source/tests/gmail.ts
```diff
@@ -253,6 +251,34 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
...
+      await gmailPage.page.reload();
+      await gmailPage.waitAndClick('[data-tooltip="Drafts"]');
+      await gmailPage.waitForContent('#fc_offline_drafts', 'FlowCrypt offline drafts:');
+      await gmailPage.ensureElementsCount('#fc_offline_drafts a', 2);
+      await gmailPage.waitAndClick('#fc_offline_drafts a');
+      let urls = await gmailPage.getFramesUrls(['/chrome/elements/compose.htm'], { sleep: 1 });
+      // compose draft 2 should be first in list as drafts are sorted by date descending
+      const draft = await pageHasSecureDraft(t, browser, urls[0], 'compose draft 2');
+      await Util.sleep(3); // the draft isn't being saved if start typing without this delay
+      await draft.type('@input-body', 'trigger saving to the clound', true);
```
rrrooommmaaa, line 270: "clound" typo

limonte: fixed 👍

rrrooommmaaa (approved):
Looking much neater now, thank you!

PR #4055, opened 2021-10-10
issue #3925 Allow leading whitespaces generally (do not trim messages) + allow tabs when pasting tab-intended text

### LINKED ISSUE #3925: copying tab-indented text into rich text compose strips tabs
If I copy some text (eg from Geany) to secure compose, it loses leading tabs.

```
line
\tanother
```

(where \t is an actual tab copied over) will become

```
line
another
```

But works with spaces.

limonte:
This PR:

- allows leading and trailing whitespaces in general (users should be able to start and end their messages with spaces or tabs, e.g. for formatting purposes)
- preserves leading tabs when pasting tab-intended text (related PR in Squire repo https://github.com/neilj/Squire/pull/417)

close #3925

- Tests are added

### To be filled by reviewers

### DIFF
--- a/test/source/tests/compose.ts
+++ b/test/source/tests/compose.ts
@@ -801,9 +801,10 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       const appendUrl = 'threadId=16b584ed95837510&skipClickPrompt=___cu_false___&ignoreDraft=___cu_false___&replyMsgId=16b584ed95837510';
       const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility', { appendUrl, hasReplyPrompt: true });
       await composePage.waitAndClick('@encrypted-reply', { delay: 5 });
-      await ComposePageRecipe.fillMsg(composePage, {}, undefined, '\tline 1\n\t\tline 2');
+      const bodyWithLeadingTabs = '\tline 1\n\t\tline 2';
+      await ComposePageRecipe.fillMsg(composePage, {}, undefined, bodyWithLeadingTabs);
       await composePage.click('@action-send');
-      await composePage.waitForContent('@container-reply-msg-successful', '\tline 1\n\t\tline 2');
+      await composePage.waitForContent('@container-reply-msg-successful', bodyWithLeadingTabs);
     }));
 
     ava.default('compose - RTL subject', testWithBrowser('compatibility', async (t, browser) => {
--- a/test/source/tests/compose.ts
+++ b/test/source/tests/compose.ts
@@ -83,7 +83,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       await SettingsPageRecipe.forgetAllPassPhrasesInStorage(settingsPage, k.passphrase);
       const inboxPage = await browser.newPage(t, TestUrls.extensionInbox('ci.tests.gmail@flowcrypt.test'));
       const composeFrame = await InboxPageRecipe.openAndGetComposeFrame(inboxPage);
-      await ComposePageRecipe.fillMsg(composeFrame, { to: 'human@flowcrypt.com' }, 'sign with entered pass phrase', { encrypt: false });
+      await ComposePageRecipe.fillMsg(composeFrame, { to: 'human@flowcrypt.com' }, 'sign with entered pass phrase', '', { encrypt: false });
       await composeFrame.waitAndClick('@action-send');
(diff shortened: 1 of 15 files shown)


test/source/tests/compose.ts
```diff
@@ -844,7 +856,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       const text = 'This message is encrypted with 2 keys of flowcrypt.compatibility';
       const subject = `Test Sending Multi-Encrypted Message With Test Text ${Util.lousyRandom()}`;
       const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
-      await ComposePageRecipe.fillMsg(composePage, { to: 'flowcrypt.compatibility@gmail.com' }, subject, { sign: true, encrypt: true });
```
rrrooommmaaa, line 847: well, the following line used to fill the body (prepending to existing text if I remember correctly).
```
await composePage.waitAndType('@input-body', text, { delay: 1 });
```
I guess since you moved the body text to a parameter of fillMsg, we should get rid of `waitAndType('@input-body'` as well as `read('@input-body')` here.

test/source/tests/compose.ts
```diff
@@ -878,7 +890,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
         { isSavePassphraseChecked: false, isSavePassphraseHidden: false });
       const subject = `Test Sending Message With Test Text and HIDE_ARMOR_META OrgRule ${Util.lousyRandom()}`;
       const composePage = await ComposePageRecipe.openStandalone(t, browser, acct);
-      await ComposePageRecipe.fillMsg(composePage, { to: 'human@flowcrypt.com' }, subject, { sign: true });
+      await ComposePageRecipe.fillMsg(composePage, { to: 'human@flowcrypt.com' }, subject, '', { sign: true });
       await composePage.waitAndType('@input-body', 'any text', { delay: 1 });
```
rrrooommmaaa, line 894: same here
```
await composePage.waitAndType('@input-body'

File: test/source/tests/compose.ts [hunk 3]
```diff
@@ -1312,7 +1324,7 @@ const sendTextAndVerifyPresentInSentMsg = async (t: AvaContext,
 ) => {
   const subject = `Test Sending ${sendingOpt.sign ? 'Signed' : ''} ${sendingOpt.encrypt ? 'Encrypted' : ''} Message With Test Text ${text} ${Util.lousyRandom()}`;
   const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
-  await ComposePageRecipe.fillMsg(composePage, { to: 'human@flowcrypt.com' }, subject, sendingOpt);
```
rrrooommmaaa, line 1315: same here
```
await composePage.waitAndType('@input-body', text, { delay: 1 });
```

rrrooommmaaa (changes requested):

test/source/tests/compose.ts
limonte: I changed the logic in `fillMsg()` method so if the nullable or empty `body` param passed, it will fall back to fill the same way as before: https://github.com/FlowCrypt/flowcrypt-browser/pull/4055/files#diff-3f9f08c331d96dfb3de5f73f315d944f2272176b514691eb0c491b2b7b62cd46R72

Would it be clearer for reading if I replace `''` with `undefined`?

rrrooommmaaa: I'd be happy if you replace `''` with actual values from the tests where we used to `waitAndType` -- there is no need for this separate statement as long as we have the paramater, as I see it. Prepending the text is also not required. Tests mostly search for the substring as I understand.

limonte: Done 👍

rrrooommmaaa: is it possible to pass `text` variable instead of `undefined` in this test and remove `await composePage.waitAndType('@input-body', text, { delay: 1 }); ...`?
Will it break the test?

limonte: Right, sorry I didn't get that from the initial comment. Done 👍

test/source/tests/compose.ts
limonte: Done 👍

test/source/tests/compose.ts
limonte: Done 👍

rrrooommmaaa (approved):

PR #4058, opened 2021-10-15
issue #3992 [Setup] Fix showing error modal when uploading invalid private key

### LINKED ISSUE #3992: Nothing happens when selecting an invalid key file during setting up the FlowCrypt Extension
### Steps to reproduce

1. Go To Settings -> Additional Settings -> Add key
2. Chose "Load from a file" option
3. Select this file [111.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/7202566/111.txt)

### Alternative steps to reproduce

1. Start setting up the FlowCrypt extension for a new account or after resetting the existed one
2. Click "Enter existing key" button
2. Chose "Load from a file" option
3. Select this file [111.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/7202566/111.txt)

### Actual behavior

Nothing happens, console is clear

### Expected behavior

Users should get feedback from their actions (e.g. the warning modal), if the selected file is invalid there should be an error message and optional error description.

limonte:
This PR fixes showing the error modal when user imports invalid key file. The functionality was already there, but it was broken.

close #3992

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/page-recipe/settings-page-recipe.ts
+++ b/test/source/tests/page-recipe/settings-page-recipe.ts
@@ -14,7 +14,6 @@ import { KeyUtil } from '../../core/crypto/key';
 export type SavePassphraseChecks = {
   isSavePassphraseHidden?: boolean | undefined,
   isSavePassphraseChecked?: boolean | undefined
-  isInvalidKey?: boolean | undefined
 };
 
 export class SettingsPageRecipe extends PageRecipe {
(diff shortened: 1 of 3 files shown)


test/source/tests/page-recipe/settings-page-recipe.ts
```diff
@@ -14,6 +14,7 @@ import { KeyUtil } from '../../core/crypto/key';
 export type SavePassphraseChecks = {
   isSavePassphraseHidden?: boolean | undefined,
   isSavePassphraseChecked?: boolean | undefined
+  isInvalidKey?: boolean | undefined
```
rrrooommmaaa, line 17: This check isn't related to "save passphrase", so the type name `SavePassphraseChecks` would no longer be relevant.
Moreover, these checks are supposed to be optional (can be skipped), which is not the case with invalid key -- we HAVE to dismiss the alert to proceed with the recipe, so conceptually this is not the best place for this parameter.
Can we put it somewhere to `ManualEnterOpts` maybe?

rrrooommmaaa (approved):

PR #4062, opened 2021-10-19
More robust `createKey` tests

rrrooommmaaa:
This PR moves all `createKey` tests to `flaky.ts` as per the advisory.
Added `retryErrs: true` to `createBegin` for robustness


test/source/tests/setup.ts
```diff
@@ -490,6 +490,27 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await settingsPage.close();
     }));
 
+    ava.default('setup - create key - choose no backup', testWithBrowser(undefined, async (t, browser) => {
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'flowcrypt.test.key.new.manual@gmail.com');
+      await Util.sleep(5);
```
limonte, line 495: What exactly are we waiting for during this sleep? Is it possible to rely on some element to appear instead? (that'd be more robust)

limonte: I'm unable to see the error both locally and in CI (all 3 tests pass without retries). I will rerun tests 3 times and if there will be no errors I believe those sleeps could be removed.

rrrooommmaaa: Yes, on mine.

I got this strange error like 'element disappeared too quickly'.

rrrooommmaaa: Maybe we should move other `createKey` tests to `flaky` instead because of this comment?
```
    // note - `SetupPageRecipe.createKey` tests are in `defineFlakyTests` - running serially
    // because the keygen CPU spike can cause trouble to other concurrent tests
```

rrrooommmaaa: > That should fix the error for you, and not affect others adversely.

That worked, thank you

tomholub (approved):

limonte:
FYI tests are failing because of https://github.com/FlowCrypt/flowcrypt-browser/issues/4085

PR #4089, opened 2021-11-06
Correctly order uids when parsing OpenPGP key

### LINKED ISSUE #4071: pgp user ids are wrongly in reverse order?
Description:
A key contains multiple user-id rendered the last user id. A user suggests if the userid in the My Keys section and the account email be shown the same.

For example, here I have logged in twosince85 as the account. Then I import the second key with test and twosince85 as its email. 
import:remove key

1. Does it look okay -- should it necessarily match the corresponding account?
2. [question from user] Is it possible to choose which user id gets used?
As far as I know, we have don't have this feature yet. Do you have any thoughts about it?

Thank you!

Reference: https://mail.google.com/mail/u/2/#inbox/FMfcgzGlkZChlnWzzmqHDCVRfXbSdsGV

rrrooommmaaa:
This PR orders uids when parsing keys per isPrimaryUserID flag etc.
As there may be several userids marked as primary, they should be ordered somehow more or less deterministically. This is not strictly documented by RFC4880.
This PR orders the same way as OpenPGP.js: `revoked ASC, isPrimaryUserID DESC, created DESC`.
Those 3 are attributes of a signature packet found by getLatestValidSignature() for this UserID.

close #4071

- Tests added or updated

### To be filled by reviewers


martgil:
Hello sir, @rrrooommmaaa,

>Also, I visually inspected a manually imported multi-aliased key keyMultiAliasedUser from your branch and it looks all right

I also import my private key manually but, the main difference is, I use the master branch.

>How was this first... second... key created? Can you send it to me, please?

I created the key with the GPG keychain from mac. Here is the key that I use. 
[first@example.asc.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/7501229/first%40example.asc.txt)

rrrooommmaaa:
> I created the key with the GPG keychain from mac. Here is the key that I use.
> [first@example.asc.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/7501229/first%40example.asc.txt)

Maybe you used some incorrect command switches to generate it?
I see Thunderbird also shows this order
image

martgil:
Hello @rrrooommmaaa,

Thanks for your reply. I agree. Maybe it was a false positive.

I have try to replicate it and see the difference. 

Keys that is displayed with the correct order has the tag "ultimate" in the key description.
image
image

Where as the keys that are ordered incorrectly don't have this in the description.
image

tomholub (approved):
👍

PR #4103, opened 2021-11-09
Do not show post-it reminder for EKM

### LINKED ISSUE #4080: remove post-it note reminder in setup when setting up browser extension with EKM
When keys are fetched from EKM, there should not be any yellow post-it note prompt to note down pass phrase during setup.

(customer request - they handle pass phrases differently)

limonte:
This PR removes the post-it reminder for when setting up browser extension with EKM.

close #4080

Note for reviewers: add `?w=1` for better review experience: https://github.com/FlowCrypt/flowcrypt-browser/pull/4103/files?w=1

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/page-recipe/setup-page-recipe.ts
+++ b/test/source/tests/page-recipe/setup-page-recipe.ts
@@ -246,7 +246,7 @@ export class SetupPageRecipe extends PageRecipe {
 
   public static autoKeygen = async (settingsPage: ControllablePage, { expectErr, enterPp }: {
     expectErr?: { title: string, text: string },
-    enterPp?: { passphrase: string, expectPostitNoteReminder: boolean, checks?: SavePassphraseChecks }
+    enterPp?: { passphrase: string, checks?: SavePassphraseChecks }
   } = {}): Promise<void> => {
     if (enterPp) {
       await Util.sleep(3);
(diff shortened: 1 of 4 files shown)


test/source/tests/page-recipe/setup-page-recipe.ts
```diff
@@ -260,7 +260,9 @@ export class SetupPageRecipe extends PageRecipe {
         expect(await settingsPage.isChecked('@input-step2ekm-save-passphrase')).to.equal(enterPp.checks.isSavePassphraseChecked);
       }
       await settingsPage.waitAndClick('@input-step2ekm-continue');
-      await settingsPage.waitAndRespondToModal('confirm-checkbox', 'confirm', 'Please write down your pass phrase');
+      if (enterPp.checks?.postitNoteReminder !== false) {
```
rrrooommmaaa, line 263: I wouldn't classify `postitNoteReminder` as an "optional check"  because optional checks are supposed to produce no errors when `undefined` is passed. However, when I run any of the tests you modified with `postitNoteReminder: undefined`, they fail. Hence this parameter should be kept separately from optional checks and better be named `expectPostitNoteReminder`. As it's logically connected to `enterPp`, it can be put inside `enterPp`.

rrrooommmaaa (changes requested):

test/source/tests/page-recipe/setup-page-recipe.ts
limonte: Done, thanks for the suggestion! 👍

rrrooommmaaa: Thank you... but... will we ever need to set `expectPostitNoteReminder: true` for `autoKeygen` recipe? This recipe is for ekm tests only, as we have `createKey` recipe for key generation and `manualEnter` for import etc.
I guess this `if` block can be removed completely. Sorry I haven't noticed it yesterday.

rrrooommmaaa (approved):

PR #4144, opened 2021-11-25
Verify message by sender

### LINKED ISSUE #3884: verify message as long as any of the contact public keys match + autofetch
There was earlier a related issue for encryption: #3769 

I would like to ensure something similar, but for verifying signatures.

If I have 2 public keys imported for `user@example.com`, can we ensure that messages signed with either of the two equivalent private keys will be showing as valid?

There's also a related issue:

image

Since we have more than one public key per recipient, I figure the above situation should no longer happen?

rrrooommmaaa:
This PR refactors decryption functions to use verificationPubs as parameter,
fetches unknown sender's pubkeys with publookup, automatically saving them to ContactStore and using to verify signature.
sender email is always displayed as Signer (primary UID from pubkey is ignored)

close #3884

- Tests added or updated

### To be filled by reviewers


extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -148,80 +143,33 @@ export class MsgUtil {
...
+  public static decryptMessage: PgpMsgMethod.Decrypt = async ({ kisWithPp, encryptedData, msgPwd, verificationPubs }) => {
     const longids: DecryptError$longids = { message: [], matching: [], chosen: [], needPassphrase: [] };
     let prepared: PreparedForDecrypt;
     try {
       prepared = await PgpArmor.cryptoMsgPrepareForDecrypt(encryptedData);
     } catch (formatErr) {
       return { success: false, error: { type: DecryptErrTypes.format, message: String(formatErr) }, longids };
     }
+    if (prepared.isCleartext) {
+      // todo: error if no verificationPubs?
```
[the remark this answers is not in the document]
tomholub: But what if this is a signed-only pkcs7 message, wouldn't it go into this OpenPGP branch and fail? Mabe need to add another if or another condition in this if so that we handle the issue gracefully.

tomholub (commented):
this is roughly it, as a wip

extension/js/common/core/crypto/pgp/msg-util.ts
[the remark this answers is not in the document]
rrrooommmaaa: `isCleartext` and `isPkcs7` are mutually exclusive, so it won't go to this branch

rrrooommmaaa:
@tomholub 
Implemented, though there are some differences with the task description

> You'll need to take a method from compose window and put it somewhere more general (it's comparing received pubkeys by fingerprint to the ones saved for that recipient, and only saving the newer versions / new keys, then saving to storage). 

Why don't we want to send all the keys to ContactStore.update, to update `pubkeyLastCheck` property?

> If the fetch fails, we render "error when fetching public keys to verify this message: $error" in the signature result.

It currently says "Could not look up sender's pubkey due to network error, click to retry." or, on `ApiErr.isSignificant()` -- "Could not load sender's pubkey due to an error." Should I change it and write a unit test?

> after remote public keys are fetched and local storage is updated, we check if any of the keys from local storage now contain the longid that signed the message. If they don't, now we render the "missing pub key" signature result

I implemented asynchronous ContactStore update calls -- no need to wait for completion -- is it ok?

> if one of the keys now contain the longid of the key that signed the message, we re-process the whole message with that public key - in background. 

By background do you mean sending PgpMsgVerify? That is implemented.

> Once we've done re-processing the message, we only look at the result of signature verification and re-render the signature result with the final verification result (valid / bad / error), without re-rendering the rest of the message.

Are you concerned about taking too much time to re-render the message? I need to double-check on that. The whole chain of calls DecryptModule->RenderModule->SignatureModule->DecryptModule is far too complicated. Also, I think we SHOULD re-render the message text when re-fetching it from API -- I filed an issue #4164

extension/js/common/core/crypto/pgp/msg-util.ts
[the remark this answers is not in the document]
tomholub: Ah. no such thing as a signed-only pkcs7 message?

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
```diff
@@ -3,111 +3,101 @@
...
+                ContactStore.update(undefined, signerEmail, { pubkey: fetched, pubkeyLastCheck: Date.now() }).catch(Catch.reportErr);
+              }
+              await this.renderPgpSignatureCheckResult(await retryVerification(fetchedPubkeys), fetchedPubkeys, undefined);
+              return;
+            }
+            if (intersection.length) {
+              this.renderBadSignature();
+            } else {
+              this.renderMissingPubkey(verifyRes.signerLongids[0]);
+            }
```
tomholub, line 63: I don't think you have to look for intersection here. Consider this code from iOS (which is using VerifyRes that comes from TypeScript, originally copied from browser extension, should be very similar code):

```swift
  private func renderPgpSignatureCheckResult(
    signature: VerifyRes?, 
    sender: RecipientWithOrderedPublicKeys
  ) -> ProcessedMessage.MessageSignature {
    if signature is nil then return .unsigned // render "not signed" in red
    if signature.error is not nil then return .error(signature.error) // render the "cannot verify signature: \(error)"
    if signature.signer is nil then return .unsigned // render "not signed" in red
    let signingPubKey = sender.pubkeys.find { $0.longids.contains(signature.signer) }
    if signingPubKey is nil or signature.match is nil then return .missingPubkey(longid: signature.signer) // render "cannot verify signature: no Public Key $longid" in red
    if signature.match is false then return .bad // render "bad signature" in red
    return .good
  }
```

By following the same steps, you should be able to simplify the code and retain correctness.

Please try that, see if that breaks any tests you wrote, in which case let's investigate the delta.

extension/js/common/browser/browser-msg.ts
```diff
@@ -148,8 +143,6 @@ export class BrowserMsg {
         pgpHashChallengeAnswer: (bm: Bm.PgpHashChallengeAnswer) => BrowserMsg.sendAwait(undefined, 'pgpHashChallengeAnswer', bm, true) as Promise<Bm.Res.PgpHashChallengeAnswer>,
         pgpMsgDecrypt: (bm: Bm.PgpMsgDecrypt) => BrowserMsg.sendAwait(undefined, 'pgpMsgDecrypt', bm, true) as Promise<Bm.Res.PgpMsgDecrypt>,
         pgpMsgVerifyDetached: (bm: Bm.PgpMsgVerifyDetached) => BrowserMsg.sendAwait(undefined, 'pgpMsgVerifyDetached', bm, true) as Promise<Bm.Res.PgpMsgVerify>,
-        keyParse: (bm: Bm.KeyParse) => BrowserMsg.sendAwait(undefined, 'keyParse', bm, true) as Promise<Bm.Res.KeyParse>,
-        keyMatch: (bm: Bm.KeyMatch) => BrowserMsg.sendAwait(undefined, 'keyMatch', bm, true) as Promise<Bm.Res.KeyMatch>,
```
tomholub, line 152: I like to see stuff removed 👍

test/source/tests/browser-unit-tests/unit-ContactStore.js
```diff
@@ -371,57 +299,20 @@ BROWSER_UNIT_TEST_NAME(`ContactStore stores postfixed fingerprint internally for
...
-    throw Error(`Failed to extract pubkey ${fingerprint}`);
-  }
   const [contactByEmail] = await ContactStore.get(db, [email]);
   if (contactByEmail.pubkey.id !== fingerprint) {
     throw Error(`Failed to extract pubkey ${fingerprint}`);
   }
   return 'pass';
 })();
 
-BROWSER_UNIT_TEST_NAME(`ContactStore searches S/MIME Certificate by PKCS#7 message recipient`);
```
tomholub, line 395: Is this usecase covered elsewhere, or not needed?

test/source/tests/decrypt.ts
```diff
@@ -371,6 +375,36 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      const msgId = '1754cfc37886899e';
+      const acctEmail = 'flowcrypt.compatibility@gmail.com';
+      {
+        const settingsPage = await browser.newPage(t, TestUrls.extensionSettings(acctEmail));
+        await SettingsPageRecipe.toggleScreen(settingsPage, 'additional');
+        const contactsFrame = await SettingsPageRecipe.awaitNewPageFrame(settingsPage, '@action-open-contacts-page', ['contacts.htm', 'placement=settings']);
+        await contactsFrame.waitAll('@page-contacts');
+        await Util.sleep(1);
+        expect(await contactsFrame.isElementPresent('@action-show-email-flowcryptcompatibilitygmailcom')).to.be.true;
+        expect(await contactsFrame.isElementPresent('@action-show-email-dhartleyverdoncollegeschoolnz')).to.be.false;
```
tomholub, line 388: Prefer not to have 3rd party or user emails in codebase -better craft our own message for this if needed that has no PII in it.

test/source/tests/decrypt.ts
```diff
@@ -371,6 +375,36 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      }
+      const params = `?frameId=none&acctEmail=${acctEmail}&msgId=${msgId}&signature=___cu_true___&senderEmail=dhartley@verdoncollege.school.nz`;
```
tomholub, line 390: And here.

test/source/tests/decrypt.ts
```diff
@@ -394,26 +428,73 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+        'president@forged.com');
+      const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId}`));
+      await inboxPage.waitAll('iframe', { timeout: 2 });
+      const urls = await inboxPage.getFramesUrls(['/chrome/elements/pgp_block.htm'], { sleep: 10, appearIn: 20 });
+      expect(urls.length).to.equal(1);
+      const url = urls[0].split('/chrome/elements/pgp_block.htm')[1];
+      await BrowserRecipe.pgpBlockVerifyDecryptedContent(t, browser, {
+        params: url,
+        content: ['How is my message signed?'],
+        signature: ['President@Forged.Com', 'matching signature']
```
tomholub, line 448: Was this the sender of the message in the mock? A bit of an unfortunate choice for a test because it looks like we are testing for forged signatures succeeding, which is not the case :-)

tomholub, line 448: Also, we should remove the capitalization from the email address rendering in signature result. It was a bad idea.

test/source/tests/decrypt.ts
```diff
@@ -394,26 +428,73 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+        .match(/\-\-\-\-\-BEGIN PGP SIGNATURE\-\-\-\-\-.*\-\-\-\-\-END PGP SIGNATURE\-\-\-\-\-/s)![0];
+      const params = `?frameId=none&account_email=${acctEmail}&senderEmail=${signerEmail}&msgId=${msgId}&message=Some%20corrupted%20message&signature=${encodeURIComponent(signature)}`;
+      // as the verification pubkey is not known, this scenario doesn't trigger message re-fetch
+      await BrowserRecipe.pgpBlockVerifyDecryptedContent(t, browser, { params, content: ['Some corrupted message'], signature: ["Missing pubkey 2864E326A5BE488A"] });
+    }));
+
+    ava.default('decrypt - re-fetch signed-only message from API on non-fatal verification error', testWithBrowser('compatibility', async (t, browser) => {
+      const msgId = '1754cfd1b2f1d6e5';
+      const acctEmail = 'flowcrypt.compatibility@gmail.com';
+      const signerEmail = 'dhartley@verdoncollege.school.nz';
```
tomholub, line 468: Here PII better not include

test/source/tests/decrypt.ts
```diff
@@ -394,26 +428,73 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      const data = await GoogleData.withInitializedData(acctEmail);
+      const msg = data.getMessage(msgId)!;
+      const signature = Buf.fromBase64Str(msg!.raw!).toUtfStr()
+        .match(/\-\-\-\-\-BEGIN PGP SIGNATURE\-\-\-\-\-.*\-\-\-\-\-END PGP SIGNATURE\-\-\-\-\-/s)![0];
+      const params = `?frameId=none&account_email=${acctEmail}&senderEmail=${signerEmail}&msgId=${msgId}&message=Some%20corrupted%20message&signature=${encodeURIComponent(signature)}`;
+      // as the verification pubkey is retrieved from the attester, the incorrect message digest will trigger re-fetching from API
+      await BrowserRecipe.pgpBlockVerifyDecryptedContent(t, browser, {
+        params,
+        content: [], // todo: #4164 I would expect '1234' here
+        signature: ['Dhartley@Verdoncollege.School.Nz', 'matching signature']
```
tomholub, line 478: And here

test/source/tests/decrypt.ts
```diff
@@ -394,26 +428,73 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      const signerEmail = 'dhartley@verdoncollege.school.nz';
+      const data = await GoogleData.withInitializedData(acctEmail);
+      const msg = data.getMessage(msgId)!;
+      const signature = Buf.fromBase64Str(msg!.raw!).toUtfStr()
+        .match(/\-\-\-\-\-BEGIN PGP SIGNATURE\-\-\-\-\-.*\-\-\-\-\-END PGP SIGNATURE\-\-\-\-\-/s)![0];
+      const params = `?frameId=none&account_email=${acctEmail}&senderEmail=${signerEmail}&msgId=${msgId}&message=Some%20corrupted%20message&signature=${encodeURIComponent(signature)}`;
+      // as the verification pubkey is retrieved from the attester, the incorrect message digest will trigger re-fetching from API
+      await BrowserRecipe.pgpBlockVerifyDecryptedContent(t, browser, {
+        params,
+        content: [], // todo: #4164 I would expect '1234' here
```
tomholub, line 477: Highlighting todo. We'll need a new message anyway

test/source/tests/unit-node.ts
```diff
@@ -776,19 +775,40 @@ jLwe8W9IMt765T5x5oux9MmPDXF05xHfm4qfH/BMO3a802x5u2gJjJjuknrFdgXY
...
-      const verifyRes = (decrypted as DecryptSuccess).signature!;
-      expect(verifyRes.match).to.be.true;
-      expect(verifyRes.signer?.primaryUserId).to.equal('A50 Sam <sams50sams50sept@gmail.com>');
+      // actual key the message was signed with
+      const pubkey = testConstants.pubkey2864E326A5BE488A;
+      // better key
+      const betterKey = "-----BEGIN PGP PUBLIC KEY BLOCK-----\r\nVersion: FlowCrypt Email Encryption [BUILD_REPLACEABLE_VERSION]\r\nComment: Seamlessly send and receive encrypted email\r\n\r\nxjMEYZeW2RYJKwYBBAHaRw8BAQdAT5QfLVP3y1yukk3MM/oiuXLNe1f9az5M\r\nBnOlKdF0nKnNJVNvbWVib2R5IDxTYW1zNTBzYW1zNTBzZXB0QEdtYWlsLkNv\r\nbT7CjwQQFgoAIAUCYZeW2QYLCQcIAwIEFQgKAgQWAgEAAhkBAhsDAh4BACEJ\r\nEMrSTYqLk6SUFiEEBP90ux3d6kDwDdzvytJNiouTpJS27QEA7pFlkLfD0KFQ\r\nsH/dwb/NPzn5zCi2L9gjPAC3d8gv1fwA/0FjAy/vKct4D7QH8KwtEGQns5+D\r\nP1WxDr4YI2hp5TkAzjgEYZeW2RIKKwYBBAGXVQEFAQEHQKNLY/bXrhJMWA2+\r\nWTjk3I7KhawyZfLomJ4hovqr7UtOAwEIB8J4BBgWCAAJBQJhl5bZAhsMACEJ\r\nEMrSTYqLk6SUFiEEBP90ux3d6kDwDdzvytJNiouTpJQnpgD/c1CzfS3YzJUx\r\nnFMrhjiE0WVgqOV/3CkfI4m4RA30QUIA/ju8r4AD2h6lu3Mx/6I6PzIRZQty\r\nLvTkcu4UKodZa4kK\r\n=7C4A\r\n-----END PGP PUBLIC KEY BLOCK-----\r\n";
```
tomholub, line 787: What is the difference between the keys, how is the other one better? Would be good to explain in comments.

test/source/tests/unit-node.ts
```diff
@@ -844,15 +862,52 @@ jLwe8W9IMt765T5x5oux9MmPDXF05xHfm4qfH/BMO3a802x5u2gJjJjuknrFdgXY
...
+JZOxQ4jpJJUszz2WsvLNwtKoqwV15Eg3oxZzHWYE8P63xXoE4G762604SIqv/ggyQZTt/Es6Scun
+A1BJflFm+cHzQTW2yQfwCCvlzEZNiNwXfwGfV99K5iG1eW3lv7sMLJnitwTidNIlD5LTNdeUnTXj
+XJvkEQsyTUI4qbzzJbUNYz7lraizC2nPiwFzLv692mS0urtD3mUhOBA9hZwk3l/20GsGia0FeUIS
+E1d8Vh/Ey7IJ8TXbfFrdv5ZP3HqMK0089SooZwx/GN2QIaOYQXsS0u7IFNhU\n=q5Sf
+-----END PGP SIGNATURE-----`);
+      const data = await GoogleData.withInitializedData('flowcrypt.compatibility@gmail.com');
+      const msg = data.getMessage('1754cfc37886899e')!;
+      const msgText = Buf.fromBase64Str(msg!.raw!).toUtfStr();
+      {
+        const dhartleyPubkey = msgText
```
tomholub, line 890: And here PII

test/source/tests/unit-node.ts
```diff
@@ -844,15 +862,52 @@ jLwe8W9IMt765T5x5oux9MmPDXF05xHfm4qfH/BMO3a802x5u2gJjJjuknrFdgXY
...
+          .match(/\-\-\-\-\-BEGIN PGP PUBLIC KEY BLOCK\-\-\-\-\-.*\-\-\-\-\-END PGP PUBLIC KEY BLOCK\-\-\-\-\-/s)![0]
+          .replace(/=\r\n/g, '').replace(/=3D/g, '=');
+        const resultRightKey = await MsgUtil.verifyDetached({
+          plaintext: Buf.fromUtfStr('some irrelevant text'),
+          sigText,
+          verificationPubs: [dhartleyPubkey]
         });
-        await ContactStore.save(undefined, contact);
+        expect(resultRightKey.match).to.be.null;
```
tomholub, line 898: Originally I imagine this would be a `false` ? Have the right pubkey, but didn't verify succcesfully.

tomholub:
Thank you for the PR - I hope to do some manual testing tomorrow 👍

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
rrrooommmaaa: You're right, I actually only need `hasIntersection` method which is implemented here 
`sender.pubkeys.find { $0.longids.contains(signature.signer) }`

test/source/tests/browser-unit-tests/unit-ContactStore.js
rrrooommmaaa: Search by longid is no longer required anywhere in the system. We're now loading all sender's keys and only use them to verify. This should also be the case for S/MIME messages (though verification isn't yet implemented for them)

tomholub: Got it

test/source/tests/decrypt.ts
rrrooommmaaa: Renamed

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
```diff
@@ -3,111 +3,101 @@
...
+          try {
+            const { pubkeys: fetchedPubkeys } = await this.view.pubLookup.lookupEmail(signerEmail);
+            if (fetchedPubkeys.length) {
+              for (const fetched of fetchedPubkeys) { // todo: should we ignore the ones we already have in verificationPubs?
+                // or better yet, create a common method to use both here and in ComposeStorageModule
+                ContactStore.update(undefined, signerEmail, { pubkey: fetched, pubkeyLastCheck: Date.now() }).catch(Catch.reportErr);
+              }
+              await this.renderPgpSignatureCheckResult(await retryVerification(fetchedPubkeys), fetchedPubkeys, undefined);
+              return;
+            }
```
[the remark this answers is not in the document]
rrrooommmaaa: Done. I exported functions directly, should I have used a class with static members instead?

test/source/tests/unit-node.ts
rrrooommmaaa: Fixed

test/source/tests/decrypt.ts
rrrooommmaaa: I constructed a new message containing '1234'. Here my concern is whether the "slightly modified" originally loaded message should be updated (in the end we would like to show signature next to the really signed message, don't we?)

test/source/tests/unit-node.ts
rrrooommmaaa: The newest (the greater date of creation) was supposed to be "better", but currently this field isn't used in sorting. Should we take it into account?

tomholub: Let's name them `olderVersionOfPubkey` and `newerVersionOfPubkey` or similar. Better is unclear

extension/js/common/core/crypto/pgp/msg-util.ts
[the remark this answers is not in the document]
rrrooommmaaa: Added appropriate remarks

test/source/tests/unit-node.ts
rrrooommmaaa: Done. Although they are not really versions -- they are compoletely independent. Is `newlyCreatedPubkey` good enough?

tomholub: > The newer (the greater date of creation) was supposed to be "better", but currently this field isn't used in sorting. Should we take it into account?

Newer key is only better if both have the exact same primary fingerprint, and therefore the newer is an updated version of the old. There is nothing better about a newly created different key. It's just another key.\

Do we have anywhere in our codebase a preference for newer keys that have a different primary fingerprint? If yes, that has to be removed (sounds like a security problem).

For this test, let's call the second key simply `wrongKey`. Since it's a key unrelated to the signature that was made. Whether it's newer or not should not matter.

extension/js/common/core/crypto/pgp/msg-util.ts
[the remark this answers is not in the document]
rrrooommmaaa: Please see the updated comment

extension/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.ts
```diff
@@ -54,39 +54,37 @@ export class PgpBlockViewDecryptModule {
...
     } catch (e) {
       await this.view.errorModule.handleInitializeErr(e);
     }
   };
 
-  private decryptAndRender = async (encryptedData: Buf, optionalPwd?: string, plainSubject?: string) => {
-    if (typeof this.view.signature !== 'string') {
+  public canFetchFromApi = () => this.canReadEmails && this.msgFetchedFromApi !== 'raw';
+
+  private decryptAndRender = async (encryptedData: Buf, verificationPubs: string[], optionalPwd?: string, plainSubject?: string) => {
```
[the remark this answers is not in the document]
rrrooommmaaa: fixed

test/source/tests/unit-node.ts
rrrooommmaaa: > Do we have anywhere in our codebase a preference for newer keys that have a different primary fingerprint? If yes, that has to investigated and discussed.

No, it turns out we don't have such preference.
We sort by expiration date. Both keys in this test are non-expirable, so they are considered equivalent

tomholub: Got it. Let's use `wrongKey` as a name for the second one.

rrrooommmaaa: Done

tomholub (commented):
New changes all look good. I'll do another review for all of the changes hollistically

tomholub:
> This pull request **introduces 1 alert** when merging [2df3953](https://github.com/FlowCrypt/flowcrypt-browser/commit/2df39531610a2cd28a31d4dd1cb49b6e101be2ce) into [c636c10](https://github.com/FlowCrypt/flowcrypt-browser/commit/c636c103cfbef2171c3a22497f0fca6c1b052838) - [view on LGTM.com](https://lgtm.com/projects/g/FlowCrypt/flowcrypt-browser/rev/pr-120585728db0347a98592a13974cc9f22f9a5cde)
> 
> **new alerts:**
> 
>     * 1 for User-controlled bypass of security check

@rrrooommmaaa have you looked into this - is that a false positive?

rrrooommmaaa:
> have you looked into this - is that a false positive?

Looks like this means that `verify` call is skipped if the module is invoked without `signature` query parameter (I guess it was the same way before), if I understood correctly. Anyway, if `verify` isn't called, then `matching signature` status won't appear, so I can't see any problems here.

test/source/tests/decrypt.ts
tomholub: Let's discuss this in main comments below

tomholub:
image

If you are talking about re-fetching the message because the DOM may have mingled it, then yes - we should display the re-fetched message we got from Google API.

If you mean re-fetching public keys for verification, then we don't have to re-render the message, just verification result.

Does this mean this is still actionable to change / update in this PR?

rrrooommmaaa:
> If you are talking about re-fetching the message because the DOM may have mingled it, then yes - we should display the re-fetched message we got from Google API.

yes, in this particular test the message is completely different when re-fetched from API.

tomholub:
> > If you are talking about re-fetching the message because the DOM may have mingled it, then yes - we should display the re-fetched message we got from Google API.
> 
> yes, in this particular test the message is completely different when re-fetched from API.

We should indeed re-render the message body to match what we got from API. Since that's what we are verifying on the second attempt. Could be in another issue - since we were never doing that before, it's not new to this PR. Well noticed.

rrrooommmaaa:
> If you mean re-fetching public keys for verification, then we don't have to re-render the message, just verification result.

Is this for performance reasons. I didn't pay too much attention on what is rendered and re-rendered. I can do this in this or another PR

rrrooommmaaa:
> We should indeed re-render the message body to match what we got from API. Since that's what we are verifying on the second attempt. Could be in another issue - since we were never doing that before, it's not new to this PR. Well noticed.

The issue is already filed (#4164)

tomholub:
> > If you mean re-fetching public keys for verification, then we don't have to re-render the message, just verification result.
> 
> Is this for performance reasons. I didn't pay too much attention on what is rendered and re-rendered. I can do this in this or another PR

It's just for less work reasons, development-wise. Usually re-rendering fewer things is less work. As long as it doesn't create any side effects, it doesn't really matter if the body is re-rendered.

There is one possible side effect - if the user starts wanting to copy and paste part of the message, re-rendering will upset them because their highlight will be gone if the text gets replaced while trying to copy it.

rrrooommmaaa:
> There is one possible side effect - if the user starts wanting to copy and paste part of the message, re-rendering will upset them because their highlight will be gone if the text gets replaced while trying to copy it.

I'm afraid there might be such side effects. I suggest to fix them after (or as part of) #4164 as there is a stack of modules calling each other. Perhaps, a bit of refactoring is in order.

tomholub:
Getting an exception - this scenario is likely not tested? it's on `flowcrypt.compatibility` account.

```
TypeError: Cannot set properties of undefined (setting 'parsedSignature')
    at PgpBlockViewSignatureModule.renderPgpSignatureCheckResult (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block_modules/pgp-block-signature-module.js:17:57)
    at PgpBlockViewRenderModule.decideDecryptedContentFormattingAndRender (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block_modules/pgp-block-render-module.js:70:45)
    at PgpBlockViewDecryptModule.decryptAndRender (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:69:50)
    at async PgpBlockViewDecryptModule.initialize (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:34:21)
    at async PgpBlockView.render (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block.js:44:17)
    at async chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/js/common/view.js:26:17

### Catch.reportErr calling stack ###
#     at Function.Catch.formatExceptionForReport (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/js/common/platform/catch.js:252:90)
#     at Function.Catch.onErrorInternalHandler (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/js/common/platform/catch.js:76:29)
#     at Function.Catch.reportErr (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/js/common/platform/catch.js:92:18)
#     at PgpBlockViewErrorModule.handleInitializeErr (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block_modules/pgp-block-error-module.js:56:23)
#     at PgpBlockViewDecryptModule.initialize (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:55:45)
#     at async PgpBlockView.render (chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block.js:44:17)
#     at async chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/js/common/view.js:26:17
# 
# url: chrome-extension://pljjhfjdhkbkhhkmngllhdmccdflgnbp/chrome/elements/pgp_block.htm?frameId=frame_tXWErApBbb&message=-----BEGIN%20PGP%20SIGNED%20MESSAGE-----%0AHash%3A%20SHA256%0A%0Athis%20message%20is%20signed%20and%20was%20modified%20in%20transit%0A-----BEGIN%20PGP%20SIGNATURE-----%0AVersion%3A%20FlowCrypt%205.0.4%20Gmail%20Encryption%20flowcrypt.com%0AComment%3A%20Seamlessly%20send%2C%20receive%20and%20search%20encrypted%20email%0A%0AwsFcBAEBCAAQBQJZ%2B8VXCRAGylU%2BwkVdcAAAgRwQAMkc%2B2WrIBf%2BX3T%2ByNps%0Ava4xMQEVvBFS4GxbTA1xndmfuo3hr3Vfq%2FsAI2Plo65D0rd9Wc6p88Awl%2B9w%0Aa0l4Yx3isH0jm9gcpypzYJf9leertBcvIxDvqGZc%2FxPxyRzQKgXiNkvaZI8U%0ANNkoAdCves4wx3sb0lJhIKyRhXwKWWroWKfoBYAryKDSE2q8d1z%2BP0mDbop7%0AJ5YLui%2B6VsGJoqZdM8lMGLp2TuyGVxy5moEvrIY8zvYm61Ke%2BwGLbL63Bqdg%0AZb8pMEp1xdW%2FxPgFT6IrtFdHTEiqczkeiHOG7%2BF8zPcKBbvx6luFAvFh103I%0Al06RRefeJkppLJ2S%2BA0iDzmqDknYin2Hm8VBa2ill%2FRfI4c1ThzM5trjrC1b%0ArOMzIMZfDdCDCRjfILNyjnanhsYEYZ30NnmjOrNDthA9VCEUupoRvSmDbA8O%0AVtckZH3YzIi6CsHs8dl2WTQ%2Flj1yXKDqnpoWzndzVrFRtjVW7I75lL%2F%2B%2FMYp%0AhktbFAWIkrqqWapwIUEGefzbrulRRVUU6m6slbKIurOgkEt%2BcoJUNMTsuPPX%0AaAa43ze8hUji5EXALE9I%2B41uz%2BdtnJymaQhsnXL8R8VHAgowrvQgLeuTZNVH%0AAeNkbcng%2BKz3ErVk3Q%2BRDt0i5FL0oIzP%2FAsxhGeOG38j%2BQVe6OgYm7IjfaIA%0Ar498%0A%3D76q6%0A-----END%20PGP%20SIGNATURE-----&msgId=15f7f7c5979b5a26&senderEmail=cryptup.tester%40gmail.com&isOutgoing=___cu_false___&acctEmail=flowcrypt.compatibility%40gmail.com&parentTabId=73%3A0
# 
######################
```

image

tomholub:
Also happens on `[enigmail] signed message inline + attachment`

tomholub:
Separately, it seems to me that the pgp block doesn't finish rendering anything until the decrypt -> try to verify -> aha missing pubkey -> fetch pubkeys -> reverify loop is complete.

I didn't investigate in detail, could it be true? I'd recommend to write a test as follows:
 - add `/attester/pub/this.pubkey.takes.long.to.load@sender.test` to attester mock api. It will include a 5 second sleep before returning the pubkey.
 - write a ui test that checks a message signed by that key
 - wait in ui test of the pgp block for texts in this sequence: 1) message body 2) "verifying signature" 3) good signature

Do you have a similar test that checks for the "verifying signature" step specifically?

tomholub:
> >  if one of the keys now contain the longid of the key that signed the message, we re-process the whole message with that public key - in background.
> By background do you mean sending PgpMsgVerify? That is implemented.

By background I meant that the user should not need to wait for it. Render the text first with "verifying signature..". So user can see message right away, and signature verification when it's done verifying.

tomholub:
> >  Once we've done re-processing the message, we only look at the result of signature verification and re-render the signature result with the final verification result (valid / bad / error), without re-rendering the rest of the message.
>
> Are you concerned about taking too much time to re-render the message? I need to double-check on that. The whole chain of calls DecryptModule->RenderModule->SignatureModule->DecryptModule is far too complicated. Also, I think we SHOULD re-render the message text when re-fetching it from API -- I filed an issue #4164

Here I was concerned that the user would have to wait 10 seconds to read the message if it takes 10 seconds to fetch public keys to verify it. Sorry for not clarifying earlier, I overlooked this comment.

rrrooommmaaa:
> Getting an exception - this scenario is likely not tested? it's on `flowcrypt.compatibility` account.

fixed this -- no longer re-fetching the encrypted message when verification fails

rrrooommmaaa:
> Also happens on `[enigmail] signed message inline + attachment`

This one should now be passing too

tomholub:
This does not close the issue fully yet - I've filed #4197 which should be done right after.

tomholub (approved):

PR #4178, opened 2021-12-08
Fix injecting the Secure Compose button on the new Gmail look

### LINKED ISSUE #4099: secure compose button broken for some users
The new Gmail look breaks the Secure Compose button:

image

limonte:
This PR checks if the new Gmail look is enabled and if so the Secure Compose button will be injected into another `<div>`.

Here's the markup of the new Gmail look:

CleanShot 2021-12-08 at 15 34 18@2x

close #4099

- Difficult to test, I'm still unable to turn on that new look for myself and I can't turn on this new look in the tests. The markup above is from the screensharing session with a customer.

### To be filled by reviewers

### DIFF
--- a/extension/js/common/inject.ts
+++ b/extension/js/common/inject.ts
@@ -26,7 +26,7 @@ export class Injector {
   private S: SelCache;
   private container: { [key: string]: Host } = {
     composeBtnSel: {
-      'gmail': 'div.aeN',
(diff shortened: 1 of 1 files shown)


tomholub (approved):
approving blindly, since don't have access to the updated DOM version

Issue #4086, opened 2021-11-04
Unreadable encrypted message with attachment sent using Microsoft outlook


martgil:
Description:
Encrypted messages with an attachment that is sent using Microsoft Outlook failed to be recognized by the FlowCrypt extension.

The message from Microsoft outlook was separated and turned into two files named `ATT00001` and `ATT00002` without any file extension. But the `ATT00002` has the PGP message. I also tried to tweak some settings to my outlook, installing the latest version of the GpgOL (gpg4win) outlook plugin, but still getting the decrypted message without any problem.

The FlowCrypt extension, however,  has no problem when decrypting a message without an attachment (in my testing case) where these messages were named `openpgp-encrypted-message.asc` instead of the more random one. 

I have searched for the same problem on google, and, some other says this was an issue from the Microsoft outlook where the email with the attachment was turned into `ATT00001` (https://www.google.com/search?q=att00001)

I have replicated it once but never happened it again as it was really tricky.

I was able to decrypt it with my key and have the decrypted mime message. I'm not knowledgeable yet on how to proceed but the mime message looks like it can be rendered and be fully decrypted.

I highly suspect that it could be because the attachment doesn't have a `.asc` extension so the FlowCrypt browser ignored to render it?

Thank you for taking a look. Any ideas are welcomed!

Files:
decrypted mime message => [decrypted.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/7473511/decrypted.txt)
original attachment => [ATT00002 (2).txt](https://github.com/FlowCrypt/flowcrypt-browser/files/7473517/ATT00002.2.txt)

Reference:
https://mail.google.com/mail/u/2/#inbox/FMfcgzGlkXsdKtBnmmkzrqnJcSrPFFTl

tomholub:
Please have the user send us a sample email with subject `Microsoft outlook OpenPGP email not recognized` to our compatibility email. Instructions here https://flowcrypt.com/docs/help/compatibility.html

martgil:
Thanks for your response. I have requested the user to send us a sample email. Thank you!

martgil:
We have successfully received a sample email from the user with the subject `Microsoft outlook OpenPGP email not recognized` in the flowcrypt compatibility account.

tomholub:
Thank you - this is helpful. As a workaround, until we add compatibility, then can view it here in flowcrypt icon -> encrypted inbox:

[...]

PR #4202, opened 2021-12-17
Add support for Microsoft Outlook encrypted emails 

martgil:
This PR will add support for ATTxxxxx emails from Microsoft Outlook and render them to Gmail and FlowCrypt email clients within the browser extension.

close https://github.com/FlowCrypt/flowcrypt-browser/issues/4086

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/core/attachment.ts
+++ b/extension/js/common/core/attachment.ts
@@ -120,6 +120,8 @@ export class Attachment {
       return 'publicKey';
     } else if (this.name.match(/\.asc$/) && this.length < 100000 && !this.inline) {
       return 'encryptedMsg';
+    } else if (this.name.match(/ATT[0-9]{5}$/) || this.type === 'application/pgp-encrypted') {
+      return 'encryptedMsg';
     } else {
       return 'plainFile';
     }


extension/js/common/core/attachment.ts
```diff
@@ -120,6 +120,8 @@ export class Attachment {
       return 'publicKey';
     } else if (this.name.match(/\.asc$/) && this.length < 100000 && !this.inline) {
       return 'encryptedMsg';
+    } else if (this.name.match(/ATT[0-9]{5}$/) || this.type === 'application/pgp-encrypted') {
```
rrrooommmaaa, line 123: Did you mean `&&`? If Outlook sets the attachement type as 'application/pgp-encrypted', do we really need to test attachment file name? Can we "download api export" one of such messages to have it in our "decrypt" tests or does it have PII? Is it encrypted-only or signed too?

martgil: Thanks! I should have used `&&`. 

>If Outlook sets the attachement type as 'application/pgp-encrypted', do we really need to test attachment file name?

I'm not sure yet, because the filename test is dedicated to the Microsoft outlook weird email. Maybe the test helps in cases such as a user is forwarding a PGP message in the attachment and flowcrypt extension will let it be in the attachment section rather than treating it as an encrypted message and rendering it?

>Can we "download api export" one of such messages to have it in our "decrypt" tests or does it have PII? Is it encrypted-only or signed too?

Yes, we can. I have decrypted it, it is only a sample encrypted message with an attachment.

I also noticed that the message is already rendering into the FlowCrypt Inbox as follows:
Screen Shot 2021-12-17 at 3 16 57 PM

But I can't find the reference code used in Gmail for Web so the changes don't take effect in the Gmail Web app.
Screen Shot 2021-12-17 at 12 52 12 PM

martgil:
I am now able to view the encrypted message from both the browser extension and Gmail for Web.
[Gmail]
Screen Shot 2021-12-20 at 7 20 29 PM
[browser extension]
Screen Shot 2021-12-20 at 7 20 49 PM

rrrooommmaaa:
@martgil have you figured out what goes to `ATT00001` file?
Is it the message text?
This is what I see at the current head.
image
No message text and `ATT00001` is still hanging in there.

martgil:
> @martgil have you figured out what goes to the ATT00001 file?
Is it the message text?

@rrrooommmaaa, I have figured out that ATT00001 only contains this text.
Screen Shot 2021-12-21 at 6 01 52 PM

But it was marked as pgp-encrypted in part 0, though we can see ATT00002 part 1 that contains PGP message has a mime type of octet-stream:
Screen Shot 2021-12-21 at 6 05 03 PM

I think the correspondent wrote and send the email without a message body. Would you like me to create a new test email with a message body?

I would also like to write another test for the Gmail view similar to the screenshot you have shared. May I know how could I achieve to log in like on the screenshot? Lastly, is it sufficient for the encrypted message to be decrypted in order to pass the test for this one, like what I did in f053e2f?

rrrooommmaaa:
> But it was marked as pgp-encrypted in part 0, though we can see ATT00002 part 1 that contains PGP message has a mime type of octet-stream

Right, I see, this particular message has no text, only one attachment.
These multipart types you mentioned seemed to be in strict accordance with the protocol https://datatracker.ietf.org/doc/html/rfc3156#section-4
I see the difference with Thunderbird is only that it generates a different filename (`encrypted.asc`), is this correct?
If possible, we shouldn't be dependent on filenames (as they can be changed at any time and not even mandatory to be present in the message).
As I remember, sometimes we're re-fetching messages from Gmail API if possible for better processing.
Please have a look whether we already have pieces of code capable of decoding PGP/MIME message without relying on filenames.
As for gmail live test, yes, I uploaded the redacted copy of this message to our `ci.tests.gmail@flowcrypt.dev` account. The message is accessible with this path `FMfcgzGllVqqBbjHQQRDsSwcZBlMRzDr`.
I also added `mail.google.com - Outlook encrypted message with attachment is recognized` test to `gmail.ts`
(there is one complication -- as the original message was encrypted for flowcrypt.compatibility recipient, I had to enter the corresponding private key first).
We also decided to keep copies of our ci.tests.gmail account's messages in eml format in a special folder to restore them in case they are (accidentally) lost. This file is not immediately required.

martgil:
Thank you, sir!

> Please have a look whether we already have pieces of code capable of decoding PGP/MIME messages without relying on filenames.

sir @rrrooommmaaa, I think we could use the `MsgUtil.decryptMessage()` without requiring the filename at all. For example:
```javascript
const result = await MsgUtil.decryptMessage({
      kisWithPp: await KeyStore.getAllWithOptionalPassPhrase(this.acctEmail),
      encryptedData: encrypted.getData(),
      verificationPubs: []
    });
```

Just to make sure I understand it correctly we may don't want to rely on the filename so we would try to decrypt the message from the attachment? Then check for `result.success` if true to and finally `return 'encryptedMsg';` in `treatAs()` (attachment.ts) - is it correct?

rrrooommmaaa:
> Just to make sure I understand it correctly we may don't want to rely on the filename so we would try to decrypt the message from the attachment?

We are eventually decrypting the message from the attachment treated as `encryptedMsg`, yes. The question is how we recognize that an attachment should be treated as `encryptedMsg`? By recognizing its filename? That's not a good idea, as the PGP/MIME format doesn't require for its attachments to have specific filenames. So the attachment propeties only are not enough to reliably determine the type of attachment, however we can recognize the type of the enveloping Content-Type as I committed in https://github.com/FlowCrypt/flowcrypt-browser/pull/4202/commits/0919ba70f96539e08188ba478b1d8b165982c061.
Can you check if `ATT` pattern recognition is still required and add test statements that ATT..1 isn't visible for both tests?
I suspect we might also need support for `multipart/signed; ... protocol="application/pgp-signature"`, perhaps in a different issue.

martgil:
sir @rrrooommmaaa, I have tested the recent change and I can tell that the `ATT` pattern from both the web Gmail view (`webmailNamePattern`) and FlowCrypt inbox view (name pattern condition in `treatAs`) requires it.

Removing the pattern checks will remain them from being an untouched attachment on the email.

On the other hand, the `ATT...1` becomes hidden for both webs Gmail view and the FlowCrypt inbox view.

> That's not a good idea, as the PGP/MIME format doesn't require its attachments to have specific filenames.

Yes, sir, I agree. The `MsgUtil.decryptMessage()` should work in our case. It is the function that is also use to decrypt standalone files through the experimental functionalities > Manually decrypt encrypted files but for some reason, the `attachmentGetChunk` from `gmail-element-replacer` returns an inconsistent number of bytes of the encrypted message. For example,  the encrypted message's original length is `17717`
Screen Shot 2021-12-23 at 2 22 19 PM

When trying to get the same bytes of the same encrypted message it returns a different number of bytes that leads to a broken message so the `MsgUtil.decryptMessage()` is failing:
Screen Shot 2021-12-23 at 2 22 54 PM

The `MsgUtil.decryptMessage()` has a returned property `boolean` success that we can use to determine if it was successfully decrypted - this signifies that it is an encrypted message and we will no longer need to use filename based as long the message can be decrypted in my opinion

extension/js/common/api/email-provider/gmail/gmail-parser.ts
```diff
@@ -86,24 +86,37 @@ export class GmailParser {
...
+        && contentType.value.includes('protocol="application/pgp-encrypted"'));
+      for (const [i, part] of parts.entries()) {
+        GmailParser.findAttachments(part, internalResults, internalMsgId, { pgpEncryptedIndex: pgpEncrypted ? i : undefined });
       }
     }
     if (msgOrPayloadOrPart.hasOwnProperty('body') && (msgOrPayloadOrPart as GmailRes.GmailMsg$payload$part).body!.hasOwnProperty('attachmentId')) {
+      let treatAs: 'hidden' | 'encryptedMsg' | undefined;
+      if (pgpEncryptedIndex === 0) {
+        treatAs = 'hidden';
+      } else if (pgpEncryptedIndex === 1) {
```
rrrooommmaaa, line 110: we should add "application/octet-stream" check here

extension/js/common/api/email-provider/gmail/gmail-parser.ts
```diff
@@ -86,24 +86,37 @@ export class GmailParser {
...
+      const parts = payload.parts!;
+      const pgpEncrypted = Boolean(parts.length === 2 && contentType?.value?.startsWith('multipart/encrypted;')
+        && contentType.value.includes('protocol="application/pgp-encrypted"'));
+      for (const [i, part] of parts.entries()) {
+        GmailParser.findAttachments(part, internalResults, internalMsgId, { pgpEncryptedIndex: pgpEncrypted ? i : undefined });
       }
     }
     if (msgOrPayloadOrPart.hasOwnProperty('body') && (msgOrPayloadOrPart as GmailRes.GmailMsg$payload$part).body!.hasOwnProperty('attachmentId')) {
+      let treatAs: 'hidden' | 'encryptedMsg' | undefined;
+      if (pgpEncryptedIndex === 0) {
```
rrrooommmaaa, line 108: we should add "application/pgp-encrypted" check here

extension/js/common/core/attachment.ts
```diff
@@ -16,7 +16,8 @@ export type FcAttachmentLinkData = { name: string, type: string, size: number };
 
 export class Attachment {
 
-  public static readonly webmailNamePattern = /^(((cryptup|flowcrypt)-backup-[a-z0-9]+\.(key|asc))|(.+\.pgp)|(.+\.gpg)|(.+\.asc)|(noname)|(message)|(PGPMIME version identification)|())$/m;
+  // eslint-disable-next-line max-len
+  public static readonly webmailNamePattern = /^(((cryptup|flowcrypt)-backup-[a-z0-9]+\.(key|asc))|(.+\.pgp)|(.+\.gpg)|(.+\.asc)|(noname)|(message)|(PGPMIME version identification)|(ATT[0-9]{5})|())$/m;
```
rrrooommmaaa, line 20: Looks like we actually need these patterns only to trigger message download in gmail replacer. This makes the application somewhat unreliable as the PGP/MIME format doesn't require for specific filenames to be present. This means we are prone to encountering similar issues in future.
@tomholub Should we consider trying to download GmailMsg for every message (or based on some other criteria) at least to check for `Content-Type: multipart/encrypted` etc.?

extension/js/common/core/mime.ts
```diff
@@ -376,7 +376,20 @@ export class Mime {
   };
 
   private static getNodeAsAttachment = (node: MimeParserNode): Attachment => {
+    let treatAs: 'hidden' | 'encryptedMsg' | undefined;
+    // are we dealing with a PGP/MIME encrypted message?
```
rrrooommmaaa, line 380: @martgil There are two different code paths for parsing messages on gmail page (GmailParser) and on Flowcrypt inbox (here), so there is some code logic duplication. We can think of re-using some common helper function to minimize code duplication.

test/source/tests/decrypt.ts
```diff
@@ -25,6 +25,23 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
 
   if (testVariant !== 'CONSUMER-LIVE-GMAIL') {
 
+    ava.default(`decrypt - outlook message with ATTxxxx encrypted email doesn't show empty attachment`, testWithBrowser('compatibility', async (t, browser) => {
+      const threadId = '17dbdf2425ac0f29';
+      const acctEmail = 'flowcrypt.compatibility@gmail.com';
+      const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId}`));
+      expect(await inboxPage.isElementPresent('@attachments')).to.equal(false);
```
rrrooommmaaa, line 32: well, it's not that simple. We should await for some readiness indication (please see what's inside InboxPageRecipe.checkDecryptMsg, something like     `await inboxPage.waitAll('iframe');`).
Need to double-check that this test actually fails if we don't mark ATT00001 as 'hidden'

test/source/tests/gmail.ts
```diff
@@ -352,6 +353,16 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       await gmailPage.click('[aria-label^="Discard draft"]');
     }));
 
+    ava.default('mail.google.com - Outlook encrypted message with attachment is recognized', testWithBrowser(undefined, async (t, browser) => {
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'ci.tests.gmail@flowcrypt.dev');
+      await SetupPageRecipe.manualEnter(settingsPage, 'flowcrypt.compatibility.1pp1', { submitPubkey: false, usedPgpBefore: true, },
+        { isSavePassphraseChecked: false, isSavePassphraseHidden: false });
+      const gmailPage = await openGmailPage(t, browser);
+      await gotoGmailPage(gmailPage, '/FMfcgzGllVqqBbjHQQRDsSwcZBlMRzDr');
+      expect(await gmailPage.isElementPresent('@attachments')).to.equal(false);
```
rrrooommmaaa, line 362: We should make sure that we're waiting for the loading to complete.
Need to double-check that this test actually fails if we don't mark ATT00001 as 'hidden'

test/source/tests/decrypt.ts
martgil: sir @rrrooommmaaa, I have added a waiting time for the webpage to complete and also wait for the iframe before testing the element.

>Need to double-check that this test actually fails if we don't mark ATT00001 as 'hidden'

I have checked it manually by reverting all changes on the `mime.ts` and the following tests above failed when reading the message through the FlowCrypt inbox.

test/source/tests/gmail.ts
martgil: Yes, sir.

> Need to double-check that this test actually fails if we don't mark ATT00001 as 'hidden'

I can also confirm that the test above in `gmail.ts` is failing when the changes from `gmail-parser.ts` are reverted to the original - Please correct me if I'm wrong, Is that the right approach to double-check?

test/source/tests/decrypt.ts
rrrooommmaaa: That's good. However using `sleep` is not reliable and should be used as last resort only.
Have you considered waiting for 'ready' test state, as implemented in many elements and their tests?
For example, we have `Ui.setTestState('ready');` as the last statement in `SettingsView.render()` in 
`extension/chrome/settings/index.ts`.
Would it help if we implement the same mechanism for `InboxView` too?

test/source/tests/gmail.ts
rrrooommmaaa: Yes, that's the right approach.
I guess for gmail we have to use `sleep`

martgil: Thank you for the confirmation, sir. I have retained using the`sleep` function on the gmail test.

test/source/tests/decrypt.ts
martgil: Thank you, sir! I have used the `waitForSelTestState('ready')` in replacement for `sleep()` when waiting for the body element's ready state.

extension/js/common/core/mime.ts
martgil: I tried to make a helper function into the `Attachments.ts` based on my understanding. The tests are working fine after code refactoring.

extension/js/common/core/attachment.ts
rrrooommmaaa: @tomholub this triggering mechanism doesn't seem very reliable for PGP/MIME messages

extension/js/common/core/mime.ts
rrrooommmaaa: Yes, thhat would do. Thank you.

extension/js/common/core/attachment.ts
rrrooommmaaa: ping @tomholub

tomholub: Agree, will file an issue - not for this PR.

tomholub:
@martgil what is the state of this PR? It's still marked as a Draft, without updates.

martgil:
Hello @tomholub, I have added all the requested changes by sir @rrrooommmaaa. But maybe there is something that he would like me to add here. I'll be converting this now as ready for review. Thanks!

rrrooommmaaa (approved):

PR #4203, opened 2021-12-17
replaced signature info with badge, also added encryption badge

### LINKED ISSUE #4201: update signature+encryption rendering to use badges
Instead of how we currently render signatures, I'd like to do something more similar to this (FlowCrypt iOS app)

image

Badge styles can be reused from backup screen (there is also a red color background style)

image

There will be one encryption and one signing badge, and the states are:

signing:
 - `signed` (green, only when matching and successful)
 - `not signed` (red)
 - `could not verify signature: unknown public key` (red)
 - `error verifying signature: ...` (red)
 - `verifying signature...` (gray)

encryption:
 - `encrypted` (green, if decrypted succesfully)
 - `decrypt error: key mismatch` (red) for `key_mismatch`
 - `decrypt error: security hazard` (red) - for `bad_mdc` and `no_mdc`
 - `decrypt error: ...` (red) - other decrypt errors similarly
 - `not encrypted` (red)

After this, we'll have to examine the screenshots of what all of these situations look like, and consider what does that mean for the color of the border of the whole message. That would be in another issue.

rrrooommmaaa:
This PR displays signature and encryption badges as per the task

close #4201

- Tests added or updated

### To be filled by reviewers


extension/css/cryptup.css
```diff
@@ -1309,35 +1298,15 @@ td {
   max-width: 300px;
   float: right;
   text-align: right;
-  opacity: 0.5;
+  opacity: 1;
   margin-left: 1em;
 }
 
-#pgp_signature.rtl {
```
rrrooommmaaa, line 1316: do we still need `rtl` class for signature?

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
```diff
@@ -55,9 +55,9 @@ export class PgpBlockViewSignatureModule {
           } catch (e) {
             if (ApiErr.isSignificant(e)) {
               Catch.reportErr(e);
-              $('#pgp_signature').addClass('bad').find('.result').text(`Could not load sender's pubkey due to an error.`);
+              $('#pgp_signature').addClass('red_label').find('.result').text(`error verifying signature: ${e}`);
             } else {
-              $('#pgp_signature').addClass('bad').find('.result').text(`Could not look up sender's pubkey due to network error, click to retry.`).click(
+              $('#pgp_signature').addClass('red_label').find('.result').text('error verifying signature: offline, click to retry').click(
                 this.view.setHandler(() => window.location.reload()));
```
rrrooommmaaa, line 61: @tomholub on retry we reload the whole frame, should we make an effort to preserve the rendered text?

extension/css/cryptup.css
tomholub: I suppose no. Can treat it the same.

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
tomholub: It would be nicer, but not really necessary. 

In this case, user has actively clicked on something, so they should expect things to change. 

That's different than when an automated process changes things when they thought it's already done.

rrrooommmaaa:
I have implemented and tested these "encryption" 3 cases when we actually have message text:
- encrypted (green, if decrypted succesfully)
- decrypt error: security hazard (red) - for bad_mdc and no_mdc
- not encrypted (red)

In other cases we can leave the error message as it was implemented before: as we don't have message text, we don't need badges, do we?

All signature badges are implemented given the decryption succeeds. if decryption returns "security hazard", then "not signed" badge is shown.

rrrooommmaaa:
@limonte Can you please review my CSS, as this "click to retry" message isn't multi-lined too nicely
image

extension/chrome/elements/pgp_block.htm
```diff
@@ -13,8 +13,10 @@
 
 <body style="background-color: transparent;">
   <div id="pgp_background" class="pgp_secure">
-    <div id="pgp_signature" data-test="pgp-signature">
-      <div class="cursive"><span></span></div>
+    <div id="pgp_signature" data-test="pgp-signature" class="pgp_badge short">
+      <div class="result"></div>
```
rrrooommmaaa, line 17: also not sure whether we need `result` divs

tomholub: we probably don't

tomholub:
> I have implemented and tested these "encryption" 3 cases when we actually have message text:
> 
>     * encrypted (green, if decrypted succesfully)
> 
>     * decrypt error: security hazard (red) - for bad_mdc and no_mdc
> 
>     * not encrypted (red)
> 
> 
> In other cases we can leave the error message as it was implemented before: as we don't have message text, we don't need badges, do we?
> 
> All signature badges are implemented given the decryption succeeds. if decryption returns "security hazard", then "not signed" badge is shown.

I guess so.

tomholub:
@rrrooommmaaa limonte will be moving on to a role elsewhere starting next year, not sure if he still has time left this year.

Please try to remove all of the css that make it float on the right. Instead, there should be position:relative and the badges should be on the left size, as if forming the first line of the text. That should fix your issue too.

rrrooommmaaa:
> Please try to remove all of the css that make it float on the right. Instead, there should be position:relative and the badges should be on the left size, as if forming the first line of the text. Then the width of the container they are in should be full line (standard div with no css). That should fix your issue too.

Is it looking better now?
image

tomholub:
Somewhat, not really. I'll take a look

tomholub:
It's the `max-width: 300px;` causing it.

An easy way to debug that is a right click on the badge and "inspect element". 

image

There you can uncheck or change individual items until you are happy with them.

I'll do some fiddling with the css and submit a commit.

tomholub:
Tried the functionality - looks good. After testing it, here below, and for all decrypt errors, I would still add a "decrypt error" red badge.

image

I would like user to get trained to always expect some sort of a badge there and look for it (in case they care).

I'll file another issue for that.

tomholub (approved):
excellent

PR #4215, opened 2021-12-25
extend searchable entries till the end of email/name

### LINKED ISSUE #3735: cannot find contacts in settings when searching by larger string

rrrooommmaaa:
This PR extends each entry in searchable till the end of the original full string,
making it possible to search whole emails etc.

close #3735 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -94,9 +94,10 @@ export class Str {
     return id;
   };
 
-  public static splitExtended = (str: string): string[] => {
+  // splits the string to matches,
+  // each match is extended till the end of the original string
+  public static splitExtended = (str: string, regexp: RegExp): string[] => {
     const result: string[] = [];
-    const regexp = /[a-z0-9]+/g;
     for (; ;) {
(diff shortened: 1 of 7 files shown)


extension/js/common/core/common.ts
```diff
@@ -94,6 +94,26 @@ export class Str {
     return id;
   };
 
+  // splits the string to matches,
+  // each match is extended till the end of the original string
+  public static splitExtended = (str: string, regexp: RegExp): string[] => {
+    const result: string[] = [];
+    for (; ;) {
```
tomholub, line 101: Isn't this a `while (true) {` ? I think a `while` is more standard.

rrrooommmaaa: It's arguable, as `while (true)` involves extra constant, but if you say so -- corrected.

tomholub (approved):

PR #4220, opened 2021-12-28
Remove 'Download all attachments' and 'Add all to Drive' buttons for encrypted messages

### LINKED ISSUE #4200: hide drive+download all buttons for encrypted conversations
image

drive: no point storing encrypted attachments to drive - not readable
download all: no point getting a zip of encrypted files

@limonte do you still have time for this one this month?

limonte:
This PR hides 'Download all attachments' and 'Add all to Drive' buttons for encrypted messages:

CleanShot 2021-12-28 at 11 08 59@2x

drive: no point storing encrypted attachments to drive - not readable
download all: no point getting a zip of encrypted files

close #4200

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -67,6 +67,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     standardComposeRecipient: 'div.az9 span[email][data-hovercard-id]',
     numberOfAttachments: '.aVW',
     numberOfAttachmentsDigit: '.aVW span',
+    attachmentsButtons: '.aZi',
     draftsList: '.ae4',
   };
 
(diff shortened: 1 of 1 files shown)


extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -372,6 +373,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     let msgEl = this.getMsgBodyEl(msgId); // not a constant because sometimes elements get replaced, then returned by the function that replaced them
     const senderEmail = this.getSenderEmail(msgEl);
     const isOutgoing = !!this.sendAs[senderEmail];
+    $(this.sel.attachmentsButtons).hide();
```
rrrooommmaaa, line 376: So these buttons are hidden if `processAttachments` is called at least once?
and it is called when an attachment file name fits the pattern of `webmailNamePattern`
However, in PR#4202 we're adding a filename masked as `ATTNNNNN` to support Outlook email encrypted attachments, my vision is that `webmailNamePattern` is a rough recognition mechanism, meaning that not necessarily  every file named `ATTNNNNN` is encrypted. Maybe we should be thinking of disabling these buttons if a file is actually recognized as encrypted (or signed) by treatAs (and the like), not by triggering the recognbition process..

rrrooommmaaa: ping @tomholub

tomholub: Yes, that is very sensible to only remove these when we actually recognize an encrypted message. 

Roman - could you please update the logic accordingly? Limonte moved on to another company at the end of last year.

tomholub (approved):
Filed another issue for improvement - thank you for the review.

PR #4231, opened 2022-01-01
Contact prompt is now based on `isFesUsed` or `!!fesUrl`

### LINKED ISSUE #4212: change prompts for support when FES is used
Some of our enterprise customers do their own end-user support internally, so we'll want to point the user to them instead of us.

--------------------------

> [fom: tom, to:customer]  there are various places in the browser extension that prompt the user to email human@flowcrypt.com. This particular user correctly contacted your own support, but others may end up writing flowcrypt directly. I suppose we should modify such message to guide the user to the right place. What should it ideally say? Something like "If this happens repeatedly, please contact your ... ... ...".  

> [from: customer, to: tom] Indeed, we should modify such kind of error messages. So, if there are other messages of this type, could you please replace the sentence by « If this happens repeatedly, please contact your Help Desk for support ».

-----------------------------

can be discerned by `fesUrl` in AcctStorage - if present, use the above wording. Please search for the email `human@flowcrypt.com` and replace these places appropriately by using the phrase `please contact your Help Desk for support` when FES is used.

rrrooommmaaa:
This PR make the extension display "contact your Help Desk" (in various forms) instead of "write at human@flowcrypt.com"
when FES is used for the account

close #4212

- Difficult to test (scattered across the codebase)

### To be filled by reviewers


extension/js/common/lang.ts
```diff
@@ -2,8 +2,14 @@
 
 'use strict';
 
+import { isFesUsed } from './shared.js';
```
rrrooommmaaa, line 5: looks like referencing AcctStore from Lang causes cyclic dependencies (on settings/inbox page etc.)?
There is this strange error: `ReferenceError: Cannot access 'Api' before initialization`
Perhaps, it's too heavy to reference `AcctStore` from a util class like lang.ts?
We'd better introduce `isFesUsed` function as always returning `false` first, and then inject the correct function upon some initialization?
Or, better yet, and more efficient, keep a cache dictionary of boolean values per acctEmail, and add entries to this dictionary when appropriate, @tomholub?

tomholub: Too heavy, you're right.

Let's have the methods accept `isFesUsed` directly as argument, instead of acctEmail. Then settle this problem at callsite? Wherever we call this, `AcctStore` will be available. Since we already pull information from `AcctStore` in the beginning of every `View`, it's easy to add `fesUrl` to the list of fetched items and maybe store `isFesUsed` as a private property on that `View` class.

rrrooommmaaa:
`extension/chrome/settings/fatal.ts` is left with hard-coded `human@flowcrypt.com` as well as some `.htm` files

tomholub:
> `extension/chrome/settings/fatal.ts` is left with hard-coded `human@flowcrypt.com` as well as some `.htm` files

Good compromise for the fatal.ts. For the .htm files, It would be good to have a second look. Could be another issue.

tomholub (approved):
This looks good to me. Let me know if you still wanted to do some work on this - currently marked as Draft

tomholub:
image

tomholub (dismissed):

tomholub (approved):

PR #4242, opened 2022-01-11
added signerFingerprints to VerifyRes

### LINKED ISSUE #3873: possible to sometimes retrieve full fingerprint from signed message?
In OpenPGP, it's possible that a signed message may include full fingerprint of the signing key. But on this codebase, we keep only pulling longid, even if fingerprint was included. This is slightly less secure then using full fingerprint.

> Apparently there are the following methods of denoting which key made a particular signature:
> 
>     * Issuer Key-ID: the short key-id of a subkey (eg. `Issuer: 0935E20FDC99D986`). MUST NOT be used for V5+ keys.
> 
>     * Issuer Fingerprint: The long representation of a key identifier. This is versioned, so there are different representations possible:
>       
>       * V4 fingerprint: 20 octets of the keys fingerprint (eg. `Issuer Fingerprint: 8DE64CAD1758FC194560CF590935E20FDC99D986`)
>       * V5 fingerprint: 32 octets of the keys fingerprint
> 
> 
> PGPainless (or rather BC) by default only includes an issuer key-id packet, but it is possible to add a V4 issuer fingerprint packet as well. OpenPGP.js apparently does this by default. Maybe we should upstream a change to BC to include issuer fingerprint by default?

 
https://github.com/pgpainless/pgpainless/issues/163#issuecomment-890507470

Does OpenPGP.js allow pulling fingerprint from signed messages, when available, instead of just longid?
 - if yes, we should start using it (whenever the fingerprint is in fact encoded in the message)
 - if no, we should file an issue on OpenPGP.js to add that method, so that we could start using it

rrrooommmaaa:
This PR extracts signer fingerprints from signature subpackets to `VerifyRes`

close #3873

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/core/buf.ts
+++ b/extension/js/common/core/buf.ts
@@ -170,6 +170,14 @@ export class Buf extends Uint8Array {
     return chars.join('');
   };
 
+  public toHexStr = (): string => {
+    const chars: string[] = [];
+    for (const v of this.values()) {
+      chars.push((('00' + v.toString(16)).slice(-2)).toUpperCase());
+    }
+    return chars.join('');
+  };
+
   public toBase64Str = (): string => {
     return base64encode(this.toRawBytesStr());
   };
--- a/extension/js/common/core/crypto/pgp/msg-util.ts
+++ b/extension/js/common/core/crypto/pgp/msg-util.ts
@@ -62,6 +62,7 @@ export type DecryptError = { success: false; error: DecryptError$error; longids:
 
 export type VerifyRes = {
   signerLongids: string[]; // signers longids from the message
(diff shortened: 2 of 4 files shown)


tomholub (approved):
This looks good. However, it seems my original issue definition may have been dated - we no longer pay attention to longids (or fingerprints) when verifying messages, now that we fetch pubkeys by sender email. Nevertheless, let's merge this as it may come handy in the future.

tomholub:
@rrrooommmaaa it's marked a s a draft, if it's ready for merging, you can mark it as ready & merge.

PR #4267, opened 2022-01-25
issue #4112 Display unique emails in the copy from contact

### LINKED ISSUE #4112: "Copy from Contact" option in `add_pubkey.htm` shows list with duplicate emails
This is how it looks when invoked in `'compose - settings - manually copied pubkey'` test

image

Should there be fingerprints?

_Originally posted by @rrrooommmaaa in https://github.com/FlowCrypt/flowcrypt-browser/issues/4110#issuecomment-966205474_

IvanPizhenko:
This PR makes option "copy from contact" on the page `add_pubkey.htm` displaying unique emails.

close #4112

Not sure how to make automated test for this (please advise if such test is needed)

### To be filled by reviewers


IvanPizhenko:

tomholub:
Thank you - will let Roman review.

extension/chrome/elements/add_pubkey.ts
```diff
@@ -35,10 +35,16 @@ View.run(class AddPubkeyView extends View {
...
+      Xss.sanitizeAppend('select.email', `<option value="${escapedMissingPubkeyEmail}">${escapedMissingPubkeyEmail}</option>`);
     }
+    const uniqueEmails = new Set<string>();
     for (const contact of await ContactStore.search(undefined, { hasPgp: true })) {
-      Xss.sanitizeAppend('select.copy_from_email', `<option value="${Xss.escape(contact.email)}">${Xss.escape(contact.email)}</option>`);
+      uniqueEmails.add(contact.email);
+    }
+    for (const email of Array.from(uniqueEmails).sort()) {
+      const escapedEmail = Xss.escape(email);
+      Xss.sanitizeAppend('select.copy_from_email', `<option value="${escapedEmail}">${escapedEmail}</option>`);
```
rrrooommmaaa, line 47: That's the first part, all right.
And where is the second part, as specified by Tom's comment?
```
Then all public keys of that recipient should be copied over for the other recipient.
```
I suggest to not copy revoked keys, `ContactStore.getEncryptionKeys` would be enough.
Also, `<textarea class="pubkey" ` should be invisible when an email from "Copy From Contact" is selected, as it isn't appropriate for multiple pubkeys (we discussed that we'll no longer use `ContactStore.get` here)

rrrooommmaaa (changes requested):
Need to implement and test copying multiple pubkeys from the selected email

test/source/browser/controllable.ts
```diff
@@ -683,6 +683,10 @@ export class ControllablePage extends ControllableBase {
     return html;
   };
 
+  public sleep = async (seconds: number) => {
+    return await Util.sleep(seconds);
+  };
```
tomholub, line 688: Why not use `Util.sleep` directly? Having another method doesn't add value.

tomholub: Got it, but that doesn't justify duplicating functionality. Like this, the code will bloat. It's a small method here and a small method there, and before we know it the codebase is triple the size.

If the method is in the wrong place, it should be moved, not duplicated. It would be technically incorrect to have it on a controllable (because it's not just one page that's sleeping, it's the test itself sleeping). But it would be marginally more convenient.

@rrrooommmaaa maybe you have some thoughts

extension/chrome/elements/add_pubkey.ts
IvanPizhenko: > That's the first part, all right. And where is the second part, as specified by Tom's comment?
> 
> ```
> Then all public keys of that recipient should be copied over for the other recipient.
> ```
> 
> I suggest to not copy revoked keys, `ContactStore.getEncryptionKeys` would be enough. Also, `<textarea class="pubkey" ` should be invisible when an email from "Copy From Contact" is selected, as it isn't appropriate for multiple pubkeys (we discussed that we'll no longer use `ContactStore.get` here)

@rrrooommmaaa Please recheck now

test/source/tests/compose.ts
```diff
@@ -269,25 +270,37 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
-      await ComposePageRecipe.fillMsg(composeFrame, { to: 'human+manualcopypgp@flowcrypt.com' }, 'manual copied key');
+      const composeFrame = await InboxPageRecipe.openAndGetComposeFrame(inboxPage);
+      await ComposePageRecipe.fillMsg(composeFrame, { to: 'manualcopypgp@flowcrypt.com' }, 'manual copied key');
       await composeFrame.waitAndClick('@action-open-add-pubkey-dialog', { delay: 1 });
       await inboxPage.waitAll('@dialog-add-pubkey');
       const addPubkeyDialog = await inboxPage.getFrame(['add_pubkey.htm']);
       await addPubkeyDialog.waitAll('@input-select-copy-from');
       await Util.sleep(1);
-      await addPubkeyDialog.selectOption('@input-select-copy-from', 'human@flowcrypt.com');
+      await addPubkeyDialog.selectOption('@input-select-copy-from', 'tocopyfrom@example.test');
```
rrrooommmaaa, line 289: @IvanPizhenko I added filling the contact store source contact with pubkeys 
and testing the resulting contact in the contact store via `dbPage`.
It would also be good to test that `@input-pubkey` textarea is visible before selecting the option and invisible right after.
And I guess "Manually importing Public Keys can be dangerous" only applies by pasting a pubkey text, but that's more a question for @tomholub

IvanPizhenko: @rrrooommmaaa I've updated test to check appearance/disapperance of the `@input-pubkey `.

test/source/browser/controllable.ts
rrrooommmaaa: I usually just type `Util.sleep` and `vscode` import `Util`. So adding `sleep` to `Controllable` in my opinion isn't justified

IvanPizhenko: Yes, that's it, but when you then comment out it or remove, and it is no longer used, import is not auto-removed (unfortunately), and this causes build error due to linters, so you have to do it manually, and it's quite annoying if you have to do it multiple times in a row, which happens often during debugging. Ok, I will remove, but it is then reasonable to keep a dummy function or at least one call to it that it is not necessary to remove import manually.

IvanPizhenko:
@rrrooommmaaa Please recheck.

rrrooommmaaa (approved):

PR #4271, opened 2022-01-26
validate private keys

rrrooommmaaa:
This PR validates all private key packets on `decrypt` or on `parse` (if they are unencrypted)

close #0000 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/unit-node.ts
+++ b/test/source/tests/unit-node.ts
@@ -1703,8 +1703,7 @@ j+w8FyoMKOrmOAyFTWjJVyVEruMl2a7QDO/CjaWV4sAUt0LMcRdZdTM=
       t.pass();
     });
 
-    ava.default('[unit][KeyUtil.parse] validates the private key if it is not encrypted', async t => {
-      const corruptedRsaKey = `-----BEGIN PGP PRIVATE KEY BLOCK-----
+    const unencryptedCorruptedRsaKey = `-----BEGIN PGP PRIVATE KEY BLOCK-----
 Version: FlowCrypt Email Encryption [BUILD_REPLACEABLE_VERSION]
 Comment: Seamlessly send and receive encrypted email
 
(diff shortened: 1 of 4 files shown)


test/source/tests/unit-node.ts
```diff
@@ -1634,6 +1634,142 @@ jA==
       t.pass();
     });
 
+    ava.default('[unit][KeyUtil.decrypt] validates the private key', async t => {
```
tomholub, line 1637: This looks good. Does this also fix the issue as experienced when private key is loaded:
 - from inbox backup
 - from EKM

tomholub (commented):
:+1:

test/source/tests/unit-node.ts
tomholub: In other words, is this a full / verified fix for the problem?

tomholub:
@martgil I think you had issue reproducing the problem. You can ask here. Did you try to put the broken key into inbox and then load it? If it succeeded, then with the above PR it should fail - please triage.

test/source/tests/unit-node.ts
rrrooommmaaa: Let me add validation for `KeyUtil.readBinary()` too...

martgil:
> Did you try to put the broken key into inbox and then load it?

Yes, that's exactly what I did.

>  If it succeeded, then with the above PR it should fail - please triage.

Yes @tomholub, the private key is blocked and no longer accepted when a user tried to import/add it.

Screen Shot 2022-01-27 at 2 02 52 PM
Screen Shot 2022-01-27 at 2 03 37 PM

test/source/tests/unit-node.ts
rrrooommmaaa: Done

tomholub:
@martgil yes, please file the issues there. For Android, ping @IvanPizhenko so that he can maybe create a test case for Paul to look at.

tomholub (approved):

PR #4284, opened 2022-02-13
Individual message per recipient

### LINKED ISSUE #4269: send MessageID to FES for easier threading of conversations on Gmail
Roman - on FES we allowed two methods to supply message ID (the mime email header).

Option 1: pre-generate message id before constructing mime / uploading the message to Gmail, and include that in FES /upload call

Option 2: let Gmail generate message id (as it does now) and then retrieve it & send that to FES in another API request

Option 1 is preferred if it can be made to work (if Gmail respects message id constructed by us when we're sending the message) as it requires fewer API calls.

Pre-generate option:
- create a valid Message-ID - these should include the angled brackets.
- Send it along in the MessageUploadRequest class:
```kotlin
data class MessageUploadRequest(
   val associateReplyToken: String,
   val expiration: Instant?,
   val from: String,
   val to: List<String>,
   val cc: List<String> = emptyList(),
   val bcc: List<String> = emptyList(),
   val emailGatewayMessageId: String? = null
)
```

Separate API request option:
- after sending the email retrieve the Message-ID.
- Send it to the new POST
/api/$VERSION_CLIENT/message/{externalId}/gateway endpoint.
Authorization is required and the class is:
```kotlin
data class MessageGatewayUpdateRequest(
   val emailGatewayMessageId: String,
)
```

### LINKED ISSUE #4348: send one message per web portal recipient, each with different url
> This PR now sends a separate message per recipient.

This was planned eventually - sending a separate email message per recipient, each with a separate download link. But that was for a different issue. I don't think we even have any issue filed for that yet on browser extension.

As for this issue as we are tackling now, I think we should still be sending single one message, and then make another API call that updates FES with regards to the message ID that it was sent with.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4284#discussion_r834403077_

rrrooommmaaa:
close #4348

also this PR correctly stores `threadId` when saving sent messages

- Tests added or updated

### To be filled by reviewers


rrrooommmaaa:
I see FES still returns { url } only, so it hasn't been updated yet?

rrrooommmaaa:
ping @tomholub

rrrooommmaaa:
Should we merge now or wait until FES is updated to test it again manually?

tomholub:
I'll take a look

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -197,21 +199,26 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
+        } else {
+          this.isSendMessageInProgress = false;
+          throw e;
+        }
+      }
+      if (msg.externalId) {
+        this.view.acctServer.messageGatewayUpdate(msg.externalId, msgSentRes.id).catch(Catch.reportErr);
       }
     }
     BrowserMsg.send.notificationShow(this.view.parentTabId, { notification: `Your ${this.view.isReplyBox ? 'reply' : 'message'} has been sent.` });
```
rrrooommmaaa, line 224: This PR now sends a separate message per recipient.
However, the problem is now -- how to show this message in inbox-active-thread? We don't want to show multiple copies, do we? So we need to think of a mechanism to collapse multiple messages into one. Right, @tomholub ?

tomholub:
Sorry for lack of info. Indeed, I suppose we didn't update FES yet. I'll let you know once updated.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
tomholub: > This PR now sends a separate message per recipient.

This was planned eventually - sending a separate email message per recipient, each with a separate download link. But that was for a different issue. I don't think we even have any issue filed for that yet on browser extension.

As for this issue as we are tackling now, I think we should still be sending single one message, and then make another API call that updates FES with regards to the message ID that it was sent with.

tomholub: I think you are doing two issues in one here.

Issue 1: (this issue) make another API call after a message was sent, binding the sent message email id to the FES uploaded message

Issue 2: (not this issue) send one message per recipient, each with a separate url

Let's stick only to issue one in this PR, which should be a lot easier to do and review. Then we'll discuss and do issue 2 separately.

rrrooommmaaa: ok, created PR #4353

tomholub:
We can discuss this one once rebased onto master. Thanks!

tomholub:
Let's not worry about it in this PR, and then evaluate it in another issue/PR.

for new messages: If Gmail classifies it as many threads of one message each, then that's what it is

for replies, since we are adding thread ids, several messages will all show up in that one thread, and I suppose they'll look identical except for recipients.

In another issue, we could only actually render the first one, and then for all that directly follow and belong to the same "group" we may collapse them somehow.

Mobile apps could be more clever about this, since we have full control.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -197,21 +199,26 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
         msgSentRes = await this.view.emailProvider.msgSend(msg, (p) => this.renderUploadProgress(p, progressRepresents));
-      } else {
-        this.isSendMessageInProgress = false;
-        throw e;
+      } catch (e) {
+        if (msg.thread && ApiErr.isNotFound(e) && this.view.threadId) { // cannot send msg because threadId not found - eg user since deleted it
+          msg.thread = undefined;
+          msgSentRes = await this.view.emailProvider.msgSend(msg, (p) => this.renderUploadProgress(p, progressRepresents));
+        } else {
+          this.isSendMessageInProgress = false;
```
tomholub, line 216: should investigate what is `isSendMessageInProgress` used for. Maybe there should be one before the whole loop and one after.

Error handling will be a real pain, too. what to do if a few messages were sent and then there was an error. At a minimum, we'll have to communicate it to the user such as `Messages to some recipients were sent successfully, while messages to others encountered an error: ...`

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,28 +22,41 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
+      const uploadedMessageData = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
       newMsg.pwd = undefined;
-      return await this.sendablePwdMsg(newMsg, pubkeys, { msgUrl, externalId }, signingPrv); // encrypted for pubkeys only, pwd ignored
+      const entries = Object.entries(uploadedMessageData.emailToExternalIdAndUrl ?? {});
+      if (entries.length) {
+        const msgs: SendableMsg[] = [];
+        for (const entry of entries) {
+          const recipientEmail = entry[0];
+          const msgUrl = entry[1].url;
+          const externalId = entry[1].externalId;
```
tomholub, line 43: I would say a more handy way to loop over dict is `for recipientEmail of Object.keys(uploadedMessageData.emailToExternalIdAndUrl ?? {})` because then you can avoid the `[0]` and `[1]` which is not very descriptive

tomholub (commented):
Just some thoughts.

One thing to think about is format of the sent message. Till now, we were sending both link and encrypted data in attachment in the same message (encrypted data attached directly for any pgp recipients and the sender, and link for non-pgp). But it's not realistic to send that to 20 recipients.. so we will have to rethink that.

tomholub (commented):
Looks good so far

Sent messages should be split in following way:
1) single message to all recipients who don't use FES
   - this message will have attachments and encrypted message text sent in the email
2) one message per each recipient who uses FES
   - these messages will have only FES link. That means, these messages will be very light.

That way, if I send a 10mb attachment to 20 recipients, I don't have to upload it 20 times.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
rrrooommmaaa: Done

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -195,41 +197,45 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
-        } else {
-          Catch.report('Failed to extract Message-ID of sent message');
-        }
-      })(msg.externalId, msgSentRes.id));
-    }
     await Promise.all(operations);
     this.isSendMessageInProgress = false;
     if (this.view.isReplyBox) {
-      this.view.renderModule.renderReplySuccess(msg, msgSentRes.id);
+      this.view.renderModule.renderReplySuccess(msgObj.attachments, msgObj.recipients, sentIds[0]);
```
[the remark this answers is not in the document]
tomholub: I still hope we will somehow avoid having to implement collapsing :-) as that is too much magic to do on the Gmail UI.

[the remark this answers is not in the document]
rrrooommmaaa: Maybe, we can play with `threadId`? Putting pwd-protected messages in a different thread?

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,49 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
 
-  public sendableMsg = async (newMsg: NewMsgData, pubkeys: PubkeyResult[], signingPrv?: Key): Promise<SendableMsg> => {
+  public sendableMsgs = async (newMsg: NewMsgData, pubkeys: PubkeyResult[], signingKey?: ParsedKeyInfo): Promise<MultipleMessages> => {
     if (newMsg.pwd && !this.isDraft) {
       // password-protected message, temporarily uploaded (already encrypted) to:
       //    - flowcrypt.com/api (consumers and customers without on-prem setup), or
       //    - FlowCrypt Enterprise Server (enterprise customers with on-prem setup)
       //    It will be served to recipient through web
-      const { url: msgUrl, externalId } = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
+      const uploadedMessageData = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
```
rrrooommmaaa, line 37: @tomholub Do I understand correctly that we need to send all recipients to FES (including those who have pubkeys and won't be sent individual messages to)? FES meanwhile generates and returns  personalized `externalId`s for every recipient, some of them will not be used.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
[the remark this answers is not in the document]
tomholub: I'd leave them in the same thread. Even if it shows duplicates. If it ends up excessively disturbing the ux, then we'll figure something out.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
tomholub: That's a design mistake, I think. @jerbell any thoughts?

For now, yes - please send all recipients, but only use some of the results that actually need it.

Jerbell: It feels like it has been a while since I've looked at this.  I'll try to give the correct answer.

I assume we are talking about the initial message send/post.

It is true that an external id is created for each To/Cc/Bcc recipient.  The overhead isn't that great, and the difference is an empty or filled field in the DB.  I can see how it might be misleading though.  If you didn't want this we would need to change the `MessageUploadRequest` object - perhaps I considered this the least disruptive course at the time.

We need to send all recipients though in case there is a reply all through the web portal.

If you see this as a design flaw then we could add the field in a list of `passwordEmails` so I would know who to generate & save external ids for.  The response would then filter out the `null` external ids (this is happening already).

tomholub: Thanks - I've filed issue to improve that later. @rrrooommmaaa for now, let's use the API as is, and do your own filtering on the result as you said.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,49 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
       newMsg.pwd = undefined;
-      return await this.sendablePwdMsg(newMsg, pubkeys, { msgUrl, externalId }, signingPrv); // encrypted for pubkeys only, pwd ignored
-    } else if (this.richtext) { // rich text: PGP/MIME - https://tools.ietf.org/html/rfc3156#section-4
+      const collectedAttachments = await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubkeys);
+      // we always have at least one pubkey, this is the main message, encrypted for pubkeys
+      const msgs: SendableMsg[] = [
+        await this.sendablePwdMsg(
+          newMsg,
```
rrrooommmaaa, line 43: @tomholub For recipients with a valid pubkey we're currently sending full To/Cc/Bcc headers (including pwd recipients) -- this leads to sort of duplicate messages to pwd recipients (each pwd recipient receives a copy of pubkey-encrypted message with the legacy link and also the pwd-only message without attachments with the personalized link).
If we exclude pwd recipients from the pubkey-encrypted message, then pubkey receivers won't be able to reply-all.
Do you have any particular solution for this in mind?

rrrooommmaaa: Reply-all from pwd recipient is working ok, but I'm getting this artifact at the bottom of pubkey-encrypted reply. I guess we should file it as a bug?
image

rrrooommmaaa: > @tomholub For recipients with a valid pubkey we're currently sending full To/Cc/Bcc headers (including pwd recipients) -- this leads to sort of duplicate messages to pwd recipients (each pwd recipient receives a copy of pubkey-encrypted message with the legacy link and also the pwd-only message without attachments with the personalized link).
> If we exclude pwd recipients from the pubkey-encrypted message, then pubkey receivers won't be able to reply-all.
> Do you have any particular solution for this in mind?

ping @tomholub

tomholub: not easy

tomholub: To clarify, don't duplicate emails to send them more than once per recipient. The pubkey email would only have pubkey recipients, and other emails only other recipients.

But on the pubkey email, add a `reply-to` header so that the replying client will have other recipients autofilled when they reply. The header accepts a list of recipients.

Jerbell: I've been thinking too - without a good solution.

In FES - when I didn't know who was replying (legacy) I set the `Reply-to` to be an amalgamation of `From`, `To` and `Cc` users.  So `To` and `Cc` works.  

Could there be a custom email header `X-reply-all` and put all the emails there?  This assumes all people are using a FlowCrypt extension/plugin though.

tomholub: We could think about custom headers later, once we've done a reasonably working solution with regular headers

rrrooommmaaa: I tested `reply-to` -- it indeed arrives in a message, but neither GMail nor Flowcrypt picks it up to enable "Reply-All" button. I guess, we should file a separate issue for this? Also (or instead), we can possibly include a text like `This message was also sent to xxx@yyy.com, zzz@xyz.com, don't forget to include them in reply recipients if you'd like to keep them in the loop` etc. This text will be hidden by Flowcrypt in a similar way as the legacy link is currently hidden for pubkey recipients.

tomholub: Ok, let's file this as a separate issue. For me to understand this better:
 - What exactly happens on such message when I click reply?
 - What exactly happens on such message when I click reply all?

rrrooommmaaa: in Gmail (inside "see original window"), there are 2 possible cases:
1) if the message has 1 pubkey and 1 pwd recipient, https://mail.google.com/mail/u/0/#inbox/FMfcgzGpFqPWLCWgPVsfxGRssLlZQQzn 
image
GMail doesn't provide "Reply all" button, but pressing "Reply" actually adds 2 recipients (from "Reply-To" header) -- might be a bit misleading for the user.
Flowcrypt doesn't provide "Secure Reply to All" option. pressing "Secure Reply" adds a reply box with empty recipient list.
2) if the message has 2 pubkey and 1 pwd recipient,
https://mail.google.com/mail/u/0/#inbox/FMfcgzGpFgzrRHJVXtgrlxgDgLCZXxGX
Gmail provides "Reply all" button, pressing it adds all 3 recipients to the reply (good).
Flowcrypt doesn't provide "Secure Reply to All" button, pressing "Secure Reply" adds a reply box with empty recipient list.

tomholub: Sounds ok to me for now. In a separate issue, we should ensure that FlowCrypt / secure reply understands reply-to header that has more than one address in it, and fills out the recipients appropriately.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,49 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
       //    - flowcrypt.com/api (consumers and customers without on-prem setup), or
       //    - FlowCrypt Enterprise Server (enterprise customers with on-prem setup)
       //    It will be served to recipient through web
-      const { url: msgUrl, externalId } = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
+      const uploadedMessageData = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
       newMsg.pwd = undefined;
-      return await this.sendablePwdMsg(newMsg, pubkeys, { msgUrl, externalId }, signingPrv); // encrypted for pubkeys only, pwd ignored
-    } else if (this.richtext) { // rich text: PGP/MIME - https://tools.ietf.org/html/rfc3156#section-4
+      const collectedAttachments = await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubkeys);
+      // we always have at least one pubkey, this is the main message, encrypted for pubkeys
```
rrrooommmaaa, line 40: @tomholub Do I get it right? If all the recipients are pwd recipients, I should send pubkey-encrypted copy to myself? This is to be able to see the message content in sent messages and to not lose attachments (as attachments are still not uploaded to FES)?

rrrooommmaaa: ping @tomholub

tomholub: Sorry - the more complicated the issue, the slower I'm to respond :-( I'll take a look

tomholub: > Do I get it right? If all the recipients are pwd recipients, I should send pubkey-encrypted copy to myself? 

It does sound odd. Let's not send a message to ourselves if there are no pubkey recipients.

A better fix should be done in the future. Maybe we will allow reading of FES messages from browser extension - whenever the browser extension notices a FES link in the message, it could load the message from it and render it.

> This is to be able to see the message content in sent messages

We'll have to fix that by loading and rendering FES messages, later

> and to not lose attachments (as attachments are still not uploaded to FES)?

Attachments are definitely uploaded to FES. That's why it takes so long to upload and that's why we cannot make 1 request to FES per FES recipient. We would end up uploading the attachments to FES 5 times. That's why we had to edit FES API to define several recipients in one request.

The attachments are encoded into a pgp/mime message and uploaded as a whole. It may appear there are no attachments in the code after it was encoded as pgp/mime, but try it out (on master): add attachments and you'll notice you can download them from FES. Or not?

rrrooommmaaa: You're right. The attachments are indeed there.
What seems strange, why we have this <div> tag in the `text/plain` section, is this a bug? This also leads to an artifact when replying....
image

tomholub: The cryptup_reply is there on purpose, because FES and also legacy backend need it.

We will move away from it, and use mime headers instead, which I should have done from the beginning. Tech debt.

You can use `MsgBlockParser.stripFcTeplyToken` (lol - supposed to be called `MsgBlockParser.stripFcReplyToken`) to strip it away when you don't want it in the result.

rrrooommmaaa: My question was actually: isn't it enough to have this `div` in `text/html` part of the multipart MIME message and strip it completely in `text/plain`? However, if you're going to scrap it anyway, no need to spend time on this.

tomholub: I guess you may be right, I'm not entirely sure (would have to be tested, and verified reply still works on both FES and flowcrypt.com/api). But as you say, it will get scrapped anyway.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -194,65 +200,63 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
-    } catch (e) {
-      if (msg.thread && ApiErr.isNotFound(e) && this.view.threadId) { // cannot send msg because threadId not found - eg user since deleted it
-        msg.thread = undefined;
-        msgSentRes = await this.view.emailProvider.msgSend(msg, (p) => this.renderUploadProgress(p, progressRepresents));
-      } else {
-        this.isSendMessageInProgress = false;
-        throw e;
-      }
+    const sentIds: string[] = [];
+    const operations = [this.view.draftModule.draftDelete()]; // todo: don't delete on error
```
rrrooommmaaa, line 208: I think we only should start `draftDelete` operation after all sends are successful?

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
+          collectedAttachments,
+          signingKey?.key)
+        );
+      }
+      // adding individual messages for each recipient that doesn't have a pubkey
+      for (const recipientEmail of pwdRecipients) {
+        const { url, externalId } = uploadedMessageData.emailToExternalIdAndUrl![recipientEmail];
+        const foundParsedRecipient = (newMsg.recipients.to ?? []).concat(newMsg.recipients.cc ?? []).concat(newMsg.recipients.bcc ?? []).
+          find(r => r.email.toLowerCase() === recipientEmail.toLowerCase());
+        // todo: since a message is allowed to have only `cc` or `bcc` without `to`, should we preserve the original placement(s) of the recipient?
```
rrrooommmaaa, line 78: should I preserve the placement of the individual recipient (to, cc or bcc) when sending an individual message. It works fine with Gmail -- a message can have only `bcc` filled, for example.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
rrrooommmaaa: ping @tomholub

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
rrrooommmaaa: ping @tomholub

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
rrrooommmaaa: ping @tomholub

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
tomholub: looks good

>  Also a question -- how to display non-critical errors (binding gateway ids of the sent messages) -- I called them supplementaryOperationError

You could toast the error using `Ui.toast`. I think that would be ok enough.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
tomholub: I think so

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
tomholub: Unsure. I'll let you decide, you can do whatever is simpler.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
tomholub: Unsure. I'll let you decide, you can do whatever is simpler.

tomholub:
@rrrooommmaaa could you please list things that should still be worked on within this PR? I'd like to move this along :-)

rrrooommmaaa:
> could you please list things that should still be worked on within this PR? I'd like to move this along :-)

The only thing left is the toast for supplementary (non-critical) error, like failure to bind gateway Message-ID.
Debugging it now.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -194,65 +214,62 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
-          Catch.report('Failed to extract Message-ID of sent message');
+    const sentIds: string[] = [];
+    const supplementaryOperations: Promise<void>[] = [];
+    const supplementaryOperationsErrors: any[] = []; // tslint:disable-line:no-unsafe-any
+    const success: EmailParts[] = [];
+    const failures: { recipient: EmailParts, e: any }[] = [];
+    for (const msg of msgObj.msgs) {
+      let msgSentRes: GmailRes.GmailMsgSend;
+      const msgRecipients = msg.getAllRecipients();
+      while (true) { // first try with msg.thread, and then possibly try again without it
```
tomholub, line 229: Could we use some different approach here than `while(true)` ?  It functions correctly here but doesn't seem like what the `while` statement was meant for.

To be honest, I'm not entirely sure what the different way could be. Maybe a bit of recursion. The contents of `try { } catch { }` could be its own async method, and the catch could include a conditional call to itself. I figured you wanted to avoid that because you'd have to pass many arrays to it for count-keeping, which you can pass as one object as follows:

```ts
const sendMsgsResults: SendMsgsResult = {
  sentIds: [],
  supplementaryOperations: [],
  supplementaryOperationsErrors: [],
  success: [],
  failures: []
}  
```

Which could be passed into method, passed internally during recursion, then directly returned below.

I think that will make the code more readable.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -194,65 +214,62 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
...
+          sentIds.push(msgSentRes.id);
+          if (msg.externalId) {
+            const operation = (async (externalId, id) => {
+              const gmailMsg = await this.view.emailProvider.msgGet(id, 'metadata');
+              const messageId = GmailParser.findHeader(gmailMsg, 'message-id');
+              if (messageId) {
+                await this.view.acctServer.messageGatewayUpdate(externalId, messageId);
+              } else {
+                throw new Error('Failed to extract Message-ID of sent message');
+              }
```
tomholub, line 242: The contents of `operation` could be their own, appropriately named method. Leading to a bit less clutter here.

tomholub (commented):
first round review - I've seen 9 of 25 files so far

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
+          pubkeyRecipients[sendingType] = value?.filter(emailPart => pubkeys.some(p => p.email === emailPart.email)
+            // pwd recipients that have no personal links go to legacy message
+            || (uploadedMessageData.emailToExternalIdAndUrl || {})[emailPart.email] === undefined);
+        }
+      }
+      const uniquePubkeyRecipientToAndCCs = Value.arr.unique((pubkeyRecipients.to || []).concat(pubkeyRecipients.cc || [])
+        .map(recipient => recipient.email.toLowerCase()));
+      // pubkey recipients should be able to reply to "to" and "cc" pwd recipients
+      const replyTo = (newMsg.recipients.to ?? []).concat(newMsg.recipients.cc ?? [])
+        .filter(recipient => !uniquePubkeyRecipientToAndCCs.includes(recipient.email.toLowerCase()));
```
tomholub, line 56: Let's call this `replyToForMessageSentToPubkeyRecipients` :-)

It seems it contains all non-pubkey recipients in the reply-to. Why not all recipients, including the ones with pubkey?

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
+      const msgs: SendableMsg[] = [];
+      if (pubkeyRecipients.to?.length || pubkeyRecipients.cc?.length || pubkeyRecipients.bcc?.length) {
```
tomholub, line 58: Let's add a comment: `pubkey recipients get one combined message. If there are not pubkey recpients, only password-protected messages will be sent`

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
tomholub: Also should be moved inside the `if` just below

tomholub: same for `uniquePubkeyRecipientToAndCCs` since it's only used there

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
 
-  public sendableMsg = async (newMsg: NewMsgData, pubkeys: PubkeyResult[], signingPrv?: Key): Promise<SendableMsg> => {
+  public sendableMsgs = async (newMsg: NewMsgData, pubkeys: PubkeyResult[], signingKey?: ParsedKeyInfo): Promise<MultipleMessages> => {
     if (newMsg.pwd && !this.isDraft) {
       // password-protected message, temporarily uploaded (already encrypted) to:
       //    - flowcrypt.com/api (consumers and customers without on-prem setup), or
       //    - FlowCrypt Enterprise Server (enterprise customers with on-prem setup)
       //    It will be served to recipient through web
-      const { url: msgUrl, externalId } = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
+      const uploadedMessageData = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
```
tomholub, line 38: The whole content of this if should be its own private method like maybe `formatSendablePwdMsgs`

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -77,26 +135,31 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
...
     const pgpMimeNoAttachments = await Mime.encode(msgBody, { Subject: newMsg.subject }, []); // no attachments, attached to email separately
     const { data: pubEncryptedNoAttachments } = await this.encryptDataArmor(Buf.fromUtfStr(pgpMimeNoAttachments), undefined, pubs, signingPrv); // encrypted only for pubs
-    const attachments = this.createPgpMimeAttachments(pubEncryptedNoAttachments).
-      concat(await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubs)); // encrypted only for pubs
     const emailIntroAndLinkBody = await this.formatPwdEncryptedMsgBodyLink(msgUrl);
-    return await SendableMsg.createPwdMsg(this.acctEmail, this.headers(newMsg), emailIntroAndLinkBody, attachments, { isDraft: this.isDraft, externalId });
+    const headers = this.headers(newMsg);
+    return await SendableMsg.createPwdMsg(this.acctEmail, headers, emailIntroAndLinkBody,
+      this.createPgpMimeAttachments(pubEncryptedNoAttachments).concat(attachments),
+      { isDraft: this.isDraft, externalId });
```
tomholub, line 162: This is confusing - I imagine it may be wrong, but not sure.

When sending password-protected message, attachments should only be uploaded one time, to FES (or backend) as one big pgp/mime string that includes the message and the attachment.

That means that password-protected message that goes only to password recipients should no longer have any attachments on it at all.

Here, `sendablePwdMsg` seems to be adding attachments to each of these messages, and it's called in a loop so if I have 10 password recipints I may be uploading my attachments 10 times to google as implemented. I think? Which would be incorrect.

I didn't run it, just commenting from reading the code.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
-      return await this.sendablePwdMsg(newMsg, pubkeys, { msgUrl, externalId }, signingPrv); // encrypted for pubkeys only, pwd ignored
-    } else if (this.richtext) { // rich text: PGP/MIME - https://tools.ietf.org/html/rfc3156#section-4
+      const collectedAttachments = await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubkeys);
+      const pubkeyRecipients: { [type in RecipientType]?: EmailParts[] } = {};
+      for (const [key, value] of Object.entries(newMsg.recipients)) {
+        if (['to', 'cc', 'bcc'].includes(key)) {
+          const sendingType = key as RecipientType;
+          pubkeyRecipients[sendingType] = value?.filter(emailPart => pubkeys.some(p => p.email === emailPart.email)
+            // pwd recipients that have no personal links go to legacy message
+            || (uploadedMessageData.emailToExternalIdAndUrl || {})[emailPart.email] === undefined);
```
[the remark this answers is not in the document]
rrrooommmaaa: Yes, this situation was discovered when running a test with flocrypt.com/api. (The previous logic was trying to send a message with 0 recipients, causing an unexpected error, so this solution looks neat and safe for me).

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
+      const uploadedMessageData = await this.prepareAndUploadPwdEncryptedMsg(newMsg); // encrypted for pwd only, pubkeys ignored
+      // pwdRecipients that have their personal link
+      const pwdRecipients = Object.keys(uploadedMessageData.emailToExternalIdAndUrl ?? {}).filter(email => !pubkeys.some(p => p.email === email));
       newMsg.pwd = undefined;
-      return await this.sendablePwdMsg(newMsg, pubkeys, { msgUrl, externalId }, signingPrv); // encrypted for pubkeys only, pwd ignored
-    } else if (this.richtext) { // rich text: PGP/MIME - https://tools.ietf.org/html/rfc3156#section-4
+      const collectedAttachments = await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubkeys);
+      const pubkeyRecipients: { [type in RecipientType]?: EmailParts[] } = {};
+      for (const [key, value] of Object.entries(newMsg.recipients)) {
+        if (['to', 'cc', 'bcc'].includes(key)) {
```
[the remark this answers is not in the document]
rrrooommmaaa: Just a fancy type coercion: `string` to `"to" | "cc" | "bcc"`

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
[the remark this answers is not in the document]
tomholub: Got it, let's please add a code comment, somewhat similar to my comment above, to explain it more specifically.

(eventually we'll deprecate flowcrypt.com/api and instead have another FES instance that is for all consumers. So eventually the functionality will somewhat merge when using consumer or enterprise, just enterprise will be on a dedicated instance per comapny)

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
rrrooommmaaa: > It seems it contains all non-pubkey recipients in the reply-to. Why not all recipients, including the ones with pubkey?

Because the pubkey recipients are seen in `to` and `cc` headers. And we should be careful to not show `bcc`.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
rrrooommmaaa: `attachments` were probably `[]`, but, you're right, the code is not obvious. I refactored it considerably to be more straightforward

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
+        .filter(recipient => !uniquePubkeyRecipientToAndCCs.includes(recipient.email.toLowerCase()));
+      const msgs: SendableMsg[] = [];
+      if (pubkeyRecipients.to?.length || pubkeyRecipients.cc?.length || pubkeyRecipients.bcc?.length) {
+        const pubkeyMsgData = {
+          ...newMsg,
+          recipients: pubkeyRecipients,
+          // brackets are required for test emails like '@test:8001'
+          replyTo: replyTo.length ? `${Str.formatEmailList([newMsg.from, ...replyTo], true)}` : undefined
+        };
+        msgs.push(await this.sendablePwdMsg(
```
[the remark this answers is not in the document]
rrrooommmaaa: I introduced a method named `sendablePubkeyMsgWithPwdLink` to preserve the original behaviour.
It format message body slightly differently (in case of `!richtext`) and also embeds the legacy link.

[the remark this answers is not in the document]
rrrooommmaaa: Now, `sendableNonPwdMsg`, `sendablePwdMsg` and `sendablePubkeyMsgWithPwdLink` are the main points, 
`sendableSimpleTextMsg` and `sendableRichTextMsg` are deleted.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -22,19 +22,77 @@ import { AcctStore } from '../../../../js/common/platform/store/acct-store.js';
...
+        msgs.push(await this.sendablePwdMsg(individualMsgData, pubkeys, { msgUrl: url, externalId }, [], signingKey?.key));
+      }
+      return { senderKi: signingKey?.keyInfo, msgs, recipients: newMsg.recipients, attachments: collectedAttachments };
+    } else {
+      const msg = await this.sendableNonPwdMsg(newMsg, pubkeys, signingKey?.key);
+      return {
+        senderKi: signingKey?.keyInfo,
+        msgs: [msg],
+        recipients: msg.recipients,
+        attachments: msg.attachments // todo: perhaps, we should hide technical attachments, like `encrypted.asc` and use collectedAttachments too?
```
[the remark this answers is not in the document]
rrrooommmaaa: > `sendableNonPwdMsg` decides how to format it based on rich text or simple text email, and I guess adding `msg.attachments` preserves that.

The top-level properties of `MultipleMessages` like `recipients` and `attachments` are only used to render the message after sending in the inbox reply thread (so we don't have to collect attachments and recipients across multiple messages).
Actual attachments to be sent are per-message inside `msgs`.

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -194,65 +214,69 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
     return { mimeType, data };
   };
 
-
-  private doSendMsg = async (msg: SendableMsg) => {
+  private attemptSendMsg = async (msg: SendableMsg): Promise<GmailRes.GmailMsgSend> => {
     // if this is a password-encrypted message, then we've already shown progress for uploading to backend
     // and this requests represents second half of uploadable effort. Else this represents all (no previous heavy requests)
+    // todo: this isn't correct when we're sending multiple messages
```
rrrooommmaaa, line 220: @tomholub Looks like this concept is totally broken. Should I file a separate issue for it?

rrrooommmaaa:
@tomholub updated according to your requests, please re-review

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
tomholub: Sure - we'll adjust that, in separate issue, after we collect initial customer feedback.

tomholub (commented):
A like all these changes following the PR - that looks good. I skipped [encrypted-mail-msg-formatter.ts](https://github.com/FlowCrypt/flowcrypt-browser/pull/4284/files#diff-6a41fcb6a316f4f1624b03980cd802ea91685ec695fcce50377cb0152b2a03c8) which I still have to look into. Plus need to do some manual testing.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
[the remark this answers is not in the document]
tomholub: I see. Let's do this `renderSentMessage: {recipients, attachments}` instead of 

```ts
        recipients: msg.recipients,
        attachments: msg.attachments // todo: perhaps, we should hide technical attachments, like `encrypted.asc` and use collectedAttachments too?
```

And then hide technical attachments as you say, in another issue.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -77,67 +171,42 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
...
     const pgpMimeNoAttachments = await Mime.encode(msgBody, { Subject: newMsg.subject }, []); // no attachments, attached to email separately
     const { data: pubEncryptedNoAttachments } = await this.encryptDataArmor(Buf.fromUtfStr(pgpMimeNoAttachments), undefined, pubs, signingPrv); // encrypted only for pubs
-    const attachments = this.createPgpMimeAttachments(pubEncryptedNoAttachments).
-      concat(await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubs)); // encrypted only for pubs
     const emailIntroAndLinkBody = await this.formatPwdEncryptedMsgBodyLink(msgUrl);
-    return await SendableMsg.createPwdMsg(this.acctEmail, this.headers(newMsg), emailIntroAndLinkBody, attachments, { isDraft: this.isDraft, externalId });
+    return await SendableMsg.createPwdMsg(this.acctEmail, this.headers(newMsg), emailIntroAndLinkBody,
+      this.createPgpMimeAttachments(pubEncryptedNoAttachments),
+      { isDraft: this.isDraft, externalId });
   };
```
tomholub, line 197: For a refactor in another issue:
 - this `sendablePwdMsg` method could be removed
 - `formatSendablePwdMsgs` could call `const emailIntroAndLinkBody = await this.formatPwdEncryptedMsgBodyLink(msgUrl);` and `    return await SendableMsg.createPwdMsg(this.acctEmail, this.headers(newMsg), emailIntroAndLinkBody, messageBodyAsPgpMimeAttachmentsEncryptedForSender,
      { isDraft: this.isDraft, externalId });`
 - `messageBodyAsPgpMimeAttachmentsEncryptedForSender` could be created only once, and reused in each cycle of the loop in `formatSendablePwdMsgs`
 - to create `messageBodyAsPgpMimeMetaAttachmentsEncryptedForSender` you could reuse most of existing `sendablePwdMsg` that relates to creating this pgpmime attachments, and this new method could be called `encryptMessageBodyAsPgpMimeMetaAttachments` and only sender pubkeys would be passed into it to create `messageBodyAsPgpMimeMetaAttachmentsEncryptedForSender`

The point is that this encryption only needs to happen once per batch, not once per each pwd msg, and it only needs to use pubkeys of sender, not everybody's pubkeys.

tomholub (commented):
I'm filing some comments for "later" and resolving them here in this review.

The ones I leave unresolved should be done still in this PR now.

tomholub (approved):
Looks good. I still need to do some manual tests before merging.

extension/chrome/elements/compose-modules/compose-render-module.ts
```diff
@@ -125,13 +124,13 @@ export class ComposeRenderModule extends ViewModule<ComposeView> {
     this.view.S.cached('replied_body').css('width', ($('table#compose').width() || 500) - 30);
     this.view.S.cached('compose_table').css('display', 'none');
     this.view.S.cached('reply_msg_successful').find('div.replied_from').text(this.view.senderModule.getSender());
-    this.view.S.cached('reply_msg_successful').find('div.replied_to span').text(msg.headers.To.replace(/,/g, ', '));
-    if (msg.recipients.cc !== undefined && msg.recipients.cc.length > 0) {
-      this.view.S.cached('reply_msg_successful').find('div.replied_cc span').text(msg.recipients.cc.join(', '));
+    this.view.S.cached('reply_msg_successful').find('div.replied_to span').text(Str.formatEmailList(recipients.to || []));
+    if (recipients.cc !== undefined && recipients.cc.length > 0) {
+      this.view.S.cached('reply_msg_successful').find('div.replied_cc span').text(Str.formatEmailList(recipients.cc));
```
rrrooommmaaa, line 129: this fixes #4497

Issue #4322, opened 2022-03-07
frozen setup screen when EKM not reachable?


tomholub:
Customer report. Will have to see this in action to be able to debug it.

Normally, any network error should result in a modal on this page, after a 30 second timeout at the latest.

----------

![image](https://user-images.githubusercontent.com/6306961/157087269-a289689e-8e09-4974-a123-c0f01fd184bd.png)

martgil:
Hello sir @tomholub, I remembered someone reported something similar to this - with the help of the user sending the network log, I'm able to spot that EKM for their org is not reachable. This results in login failure to any of the users trying to set up the extension for the first time.

martgil:
By configuring a proxy to pointing to a non-working http server, i think we are able to replicate the problem: <img width="1267" alt="Screen Shot 2022-03-09 at 3 21 29 PM" src="https://user-images.githubusercontent.com/46025304/157392343-a30114b1-bd58-4091-a79a-08a1c7ffd132.png">

I can also quickly identify that the request to pull private keys is not wrapped in a try & catch so networks errors are happening in the background silently.
```typescript
public getPrivateKeys = async (idToken: string): Promise<LoadPrvRes> => {
    return await this.request('GET', '/keys/private', undefined, idToken) as LoadPrvRes;
  };
```

tomholub:
fixed finally in https://github.com/FlowCrypt/flowcrypt-browser/pull/4334 and then there is a followup in https://github.com/FlowCrypt/flowcrypt-browser/issues/4335

PR #4331, opened 2022-03-09
patch request private method and catch error in a form of modal

martgil:
This PR will  patch on a general function `request` to throw an error in a modal

<video src=https://user-images.githubusercontent.com/46025304/157395710-19e4ee29-bfc8-45ea-a41d-51c02dbd5ac5.mov>

close https://github.com/FlowCrypt/flowcrypt-browser/issues/4322

- Not worth testing
- Tests added or updated

### To be filled by reviewers


extension/js/common/api/key-server/key-manager.ts
```diff
@@ -30,7 +31,12 @@ export class KeyManager extends Api {
...
-
-}
+    let res;
+    try {
+      res = await Api.apiCall(this.url, path, vals, vals ? 'JSON' : undefined, undefined, idToken ? { Authorization: `Bearer ${idToken}` } : undefined, undefined, method);
+    } catch (e) {
+      await Ui.modal.error(`Error while communicating with the key manager: ${e}`);
+    }
+    return res as RT;
+  }
```
tomholub, line 41: This is a wrong place to do handling of errors. The approach in handling errors in low-level method (like API method) is that they should be allowed to throw upwards.

Then upwards again and again, until they finally reach a place where we deal with UI (some `View`). Somewhere there, some user action has triggered the chain, and that's where we should be catching the error, once it already bubbled up.

Don't catch errors in low level methods this way. If we did that, you would have to write this code 100 times, for each low level method there is, which doesn't make sense.

martgil: I tried to rewrite the `getPrivateKeys()` to return both keys and error then process the returned data where the function call has been made.

tomholub: why? either return result, or have it throw error. catch the error on the highest-level method that initiated the chain of calls.

rrrooommmaaa (approved):

extension/chrome/settings/setup/setup-key-manager-autogen.ts
```diff
@@ -54,6 +54,7 @@ export class SetupWithEmailKeyManagerModule {
       if (ApiErr.isNetErr(e) && await Api.isInternetAccessible()) { // frendly message when key manager is down, helpful during initial infrastructure setup
         e.message = `FlowCrypt Email Key Manager at ${this.view.orgRules.getKeyManagerUrlForPrivateKeys()} is down, please inform your network admin.`;
       }
+      await Ui.modal.error((e as Error).message);
       throw e;
     }
   };
```
tomholub, line 60: We should either handle an error (with a modal) or throw it. It's weird to be doing both. Plus this change wasn't tested. I'll revert this PR.

rrrooommmaaa: there is a test `fail@key-manager-server-offline.flowcrypt.test - shows friendly KM not reachable error` that covers this new modal

tomholub: That test was already there before. Since functionality was broken, I don't think it covered it.

tomholub: The lines below, already present before this PR, already expected a modal to show up:

image

And yet, the error was still present.

The existing test only covers situation with autogen (ekm not available during automatic setup) but the bug we encountered is during user-involved setup. Most likely, that uses different code.

rrrooommmaaa: Oh I see. However, the original bug is not referenced here

tomholub: Let's have a look at #4334 once I'm ready

tomholub: It's a bit more complicated, even - the added line above checks for a modal, while the lines below are testing for a "view-rendered error" that is rendered into the page instead.

To my knowledge, the this PR:
 - in case of autogen, it would show both a modal and then also a whole-page error (that's why the test above is now testing for both, which is anyway wrong)
 - in case of import + require user to enter pass phrase (like flowcrypt.com has it), it would have showed a modal. But this case wasn't tested at all.

Cleaned all of this up in the alternative PR.

Issue #4332, opened 2022-03-11
browser extension trying to decrypt invalid pgp message with string `-----BEGIN PGP MESSAGE-----`


martgil:
Description:

A user reported a problem to us where the extension is trying to decrypt an invalid PGP message if the email body contains `-----BEGIN PGP MESSAGE-----`. 

In our android app, the same problem does not exist :
<img width="774" alt="Screen Shot 2022-03-11 at 2 02 21 PM" src="https://user-images.githubusercontent.com/46025304/157812090-2c338eb0-4ce9-4c45-8956-a568acf6349c.png">
(browser extension on the left, android emulator on the right)

Steps to reproduce:
1. Open the `Bogus PGP Message` test message using the compatibility account.

PR #4333, opened 2022-03-11
handle not signed + not encrypted message (false positive PGP message)

martgil:
This PR will add a conditional test to check for bogus PGP message (false positive PGP message) inside the email body

changes (same for gmail for web & flowcrypt extension page):
Screen Shot 2022-03-11 at 3 40 00 PM

close https://github.com/FlowCrypt/flowcrypt-browser/issues/4332

- Tests added or updated

### To be filled by reviewers


rrrooommmaaa (changes requested):
Looking good.
Please add a unit test for this case

martgil:
sir @rrrooommmaaa - I've added a test for this PR. thanks!

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -277,14 +278,20 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
...
       if (fromTextBody) {
         return { armored: fromTextBody, subject, isPwdMsg };
       }
-      const fromHtmlBody = PgpArmor.clip(Xss.htmlSanitizeAndStripAllTags(Buf.fromBase64UrlStr(bodies['text/html'] || '').toUtfStr(), '\n'));
+      const fromHtmlBody = PgpArmor.clip(htmlBody);
       if (fromHtmlBody) {
         return { armored: fromHtmlBody, subject, isPwdMsg };
       }
+      if ((fromTextBody === undefined || fromHtmlBody === undefined) && attachments.length === 0) {
+        const clippedText = (textBody || htmlBody).substring((textBody || htmlBody).search('----BEGIN PGP MESSAGE----'));
```
rrrooommmaaa, line 292: Not quite clear why we need to strip characters preceding `-----BEGIN PGP MESSAGE-----`.
What would happen if we have a message with the following text:
```
An OpenPGP message starts with this header:
-----BEGIN PGP MESSAGE-----

...
```
?
Also, what would/should happen with a message that reads:
```
An OpenPGP messages starts with this header:
-----BEGIN PGP MESSAGE-----

and ends with this footer
-----END PGP MESSAGE-----
```

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -277,14 +278,20 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
     if (format === 'full') {
```
rrrooommmaaa, line 278: @martgil the changes only involve messages loaded in `full` format.
I guess, the clause below `} else { // format === raw` should behave similarly.
Please don't copy/paste code, extract common processing statements into a helper function.

extension/js/common/api/email-provider/gmail/gmail.ts
martgil: >Not quite clear why we need to strip characters preceding -----BEGIN PGP MESSAGE-----.
this is the propose change does:
Screen Shot 2022-03-16 at 4 35 29 PM
Without stripping the preceding chars before the pgp begin header, it will repeat the entire message in the pgpblock.

Also, does the gray border looks inappropriate?

> Also, what would/should happen with a message that reads:

It does return a generic message says : "Error: Error during parsing. This message / key probably does not conform to a valid OpenPGP format.", it also shows generic error on the android version. not sure what's best for it.

Screen Shot 2022-03-16 at 4 33 26 PM

extension/js/common/api/email-provider/gmail/gmail.ts
martgil: ok sir - I looked into the block of code inside of ` else { // format === raw` and don't find a reliable way to "detect" for a bogus message so I had to rely on different error message to check based on `result.error.message`.

I then add a conditional checks for the error message `This message / key probably does not conform to a valid OpenPGP format.` - this gives me a closest way to detect potential plain text inside of pgp header/footer

Screen Shot 2022-03-16 at 6 20 07 PM

martgil: > isn't this the place where we could add the .search(...) condition?

Sorry, I run out of an idea of what to search for?

I started thinking if I can add a condition in the `} else { // raw` and can't find a suitable condition checks since the `armoredMsg` always returned a good looking armored message eg. `Header + content + Footer` wherein the content seems not suitable for checking so I went to rely on a specific error - in this case the format error that says the message/key does not conform to a valid pgp format.

Moreover, here is what's happening with the given scenario:

assume that there was a message:
```
----BEGIN PGP MESSAGE-----
Version: FlowCrypt Email Encryption 8.2.6

wV4DgCPkBPUgCC4SAQdAK4uPXBhuG9iaUDoXs3LqKrq2AHjEByD3F7UhqrCj
aaslakjdlaksjdlkajsdlkajsdlkjaslkjaslkdjalksdjlaksdjlaksjlkasdlaskdlakjsdlkasadklasd
```

this message will fall under the condition where `(format === 'full')` and will show:
image

the following message will fall under the `else` condition `(raw)`
```
-----BEGIN PGP MESSAGE-----
Version: FlowCrypt Email Encryption 8.2.6

wV4DgCPkBPUgCC4SAQdAK4uPXBhuG9iaUDoXs3LqKrq2AHjEByD3F7UhqrCj
aaslakjdlaksjdlkajsdlkajsdlkjaslkjaslkdjalksdjlaksdjlaksjlkasdlaskdlakjsdlkasadklasd

-----END PGP MESSAGE-----
```

I would add a test and add further changes if needed.

test/source/tests/browser-unit-tests/unit-Gmail.js
```diff
@@ -0,0 +1,35 @@
...
+
+BROWSER_UNIT_TEST_NAME(`Gmail.extractArmoredBlock helps detect bogus PGP message`).acct(`compatibility`).only;
+(async () => {
+  const gmail = new Gmail('flowcrypt.compatibility@gmail.com');
+  const extractedFull = await gmail.extractArmoredBlock('17d7a337b7b87eb9', 'full', undefined);
+  if (extractedFull.plaintext !== '----BEGIN PGP MESSAGE-----\r\n\r\nThis is not a valid PGP message\r\n') {
+    throw Error(`extractedFull.plaintext unexpectedly equals ${extractedFull.plaintext}`);
+  }
+  const extractedRaw = await gmail.extractArmoredBlock('17d7a337b7b87eb9', 'raw', undefined);
+  if (extractedRaw.plaintext !== '----BEGIN PGP MESSAGE-----\r\n\r\nThis is not a valid PGP message\r\n') {
```
rrrooommmaaa, line 31: @martgil my point is:
`extractArmoredBlock` behaves differently based on extraction format.
with `full` it returns plaintext,
with `raw` it throws exception though it could have extracted plaintext in a similar way.

Perhaps, your point is that `extractArmoredBlock` is only called with `raw` for detached signatures, and with `full` for armored messages, that's not obvious and happens on an upper-level (`PgpBlockViewDecryptModule.initialize`).

The method `Gmail.extractArmoredBlock` should be either logically consistent or well documented. Better logically consistent. Right now it's not obvious, why plaintext should not be extracted from a message in `raw` format.

extension/js/common/api/email-provider/gmail/gmail.ts
rrrooommmaaa: I set up a (failing) unit test to test Gmail.extractArmoredBlock directly and added a comment.
Please have a look.

martgil: thanks sir, @rrrooommmaaa - What I did is replaced the `throw FormatError...` within the `raw === format` else codeblock and return the plaintext instead.

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -277,14 +278,20 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
...
       if (fromTextBody) {
         return { armored: fromTextBody, subject, isPwdMsg };
       }
-      const fromHtmlBody = PgpArmor.clip(Xss.htmlSanitizeAndStripAllTags(Buf.fromBase64UrlStr(bodies['text/html'] || '').toUtfStr(), '\n'));
+      const fromHtmlBody = PgpArmor.clip(htmlBody);
       if (fromHtmlBody) {
         return { armored: fromHtmlBody, subject, isPwdMsg };
       }
+      if ((fromTextBody === undefined || fromHtmlBody === undefined) && attachments.length === 0) {
+        const clippedText = (textBody || htmlBody).substring((textBody || htmlBody).search('-----BEGIN PGP MESSAGE-----'));
```
rrrooommmaaa, line 292: perhaps, `indexOf` is better than `search` since we're not dealing with regexes.
Also, instead of copying the string literal, I'd suggest to reference it via
`PgpArmor.headers('encryptedMsg').begin`

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -307,6 +314,8 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
         const armoredMsg = PgpArmor.clip(decoded.text); // todo - the message might be in attachments
         if (armoredMsg) {
           return { armored: armoredMsg, subject, isPwdMsg };
+        } else if (armoredMsg?.indexOf('-----END PGP MESSAGE-----') !== -1) {
```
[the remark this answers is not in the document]
rrrooommmaaa: Also, instead of copying the string literal, I'd suggest to reference it via
`PgpArmor.headers('encryptedMsg').end`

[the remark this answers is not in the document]
martgil: > why is `indexOf` better than `includes`?

I actually look upon it through the internet and some recommend indexOf over includes though maybe it was not correct.

I check the execution time for includes vs indexOf and looks like indexOf is slightly slower: 
Screen Shot 2022-03-24 at 3 57 12 PM

I'm now using the includes() and use applied the said suggestion. thank you, sir!

extension/js/common/api/email-provider/gmail/gmail.ts
martgil: thanks, I applied these changes now.

extension/js/common/api/email-provider/gmail/gmail.ts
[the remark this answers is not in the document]
rrrooommmaaa: in this context code readability is more important than performance.
Performance makes a difference if this operation is performed a million of times in a row (or, at least a thousand)

extension/js/common/api/email-provider/gmail/gmail.ts
rrrooommmaaa: @martgil, can you please add the message we previously discussed
```
An OpenPGP message starts with this header:
-----BEGIN PGP MESSAGE-----

example
...
into exported-messages too?
And also, test calling Gmail.extractArmoredBlock() on this message in both raw and full formats to make sure they return the same plaintext.

martgil: yes, sir, I can add it and add a similar test to check both raw and full formats.

martgil: sir @rrrooommmaaa - I've added the new test for the following message above and updated the code a little bit. thank you, sir!

rrrooommmaaa (approved):

PR #4334, opened 2022-03-11
issue 4322 fix ekm not reachable error

tomholub:
Reverts FlowCrypt/flowcrypt-browser#4331 and then introduces a more appropriate fix


tomholub:
There are still some things left to adjust, but that will be for another issue:

 - following click of the setup button, there is no visual indication - no loader / spinner. We should make the button `gray` color with text `Loading..` and disable click events on it. Then the try/catch I just added should have a `finally` that re-enables the button
 - the EKM timeout is 30 seconds which is too long. Should make it 15 seconds for the EKM call

As it is now, user has to sit there and wait 30 seconds to see any UI change in some situations, even though I have fixed the issue. Only after that does the modal appear.

tomholub:
@rrrooommmaaa please review

test/source/mock/backend/backend-data.ts
```diff
@@ -130,8 +130,14 @@ export class BackendData {
           'FORBID_STORING_PASS_PHRASE']
       };
     }
-    if (domain === 'key-manager-server-offline.flowcrypt.test') {
-      return { ...keyManagerAutogenRules, "key_manager_url": "https://localhost:1230/intentionally-wrong", };
+    if (domain === 'key-manager-server-offline.flowcrypt.test') { // EKM offline during key autogen / upload flow
+      return { ...keyManagerAutogenRules, "key_manager_url": "https://localhost:1230/intentionally-wrong" };
+    }
+    if (domain === 'ekm-offline-retrieve.flowcrypt.test') {
+      return { // EKM offline during gey retrieval flow
```
rrrooommmaaa, line 137: typo: should be "key retrieval..."

tomholub: 👍

rrrooommmaaa (approved):

PR #4336, opened 2022-03-13
conditionally increase WKD timeout

### LINKED ISSUE #4320: conditionally increase WKD lookup timeout
Actually, I may have a simpler and better fix to do first, not involving UI.

We have a pretty aggressive timeout for WKD lookups - probably 4 seconds. This may be the root cause of the flakiness. 4 seconds timeout is because for most users most of the time, we don't expect WKD to be there (most consumer users or other recipients don't have it).

But for customers that run Email Key Manager, we know they have very likely WKD running too. So we could increase WKD request timeout in these situations to 10 seconds.

In the following situations, there will be a timeout of 10 seconds on WKD lookups:
 - if my own account is using email key manager, AND
 - I'm looking up a WKD email on the same domain as my own account

In remaining situations let's leave the timeout at 4 seconds as before.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4319#issuecomment-1060938464_

rrrooommmaaa:
This PR sets a timeout of 10 seconds on WKD lookups:

- if my own account is using email key manager, AND
- I'm looking up a WKD email on the same domain as my own account

close #4320 

- Difficult to test (deals with internal timeouts, I tested it manually)

### To be filled by reviewers

### DIFF
--- a/extension/js/common/api/key-server/wkd.ts
+++ b/extension/js/common/api/key-server/wkd.ts
@@ -19,6 +19,13 @@ export class Wkd extends Api {
 
   public port: number | undefined;
 
+  constructor(
+    private domainName: string,
+    private usesKeyManager: boolean
+  ) {
+    super();
+  }
+
   // returns all the received keys
   public rawLookupEmail = async (email: string): Promise<{ keys: Key[], errs: Error[] }> => {
     // todo: should we return errs on network failures etc.?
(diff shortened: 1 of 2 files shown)


tomholub (approved):
Thanks!

PR #4337, opened 2022-03-13
issue #3679 Stricter contact matching

### LINKED ISSUE #3679: Search contacts matching should be stricter, do not suggest `name.surname@domain.cc` for `zzzz@zzzzz.cc`
I think the matching top domain zone should not suggest contacts:

image

IvanPizhenko:
This PR improves contact matching.

close #3679

- Difficult to test (explain why) - not sure how to write the test

### To be filled by reviewers


IvanPizhenko:
@rrrooommmaaa I'm trying to implement here what we were talking about in the #3679 discussion, please check and let me know what do you think.

extension/js/common/platform/store/contact-store.ts
```diff
@@ -380,12 +380,940 @@ export class ContactStore extends AbstractStore {
     });
   };
 
+
+  // taken from https://github.com/datasets/top-level-domain-names/blob/master/top-level-domain-names.csv
+  private static tlds = [
```
rrrooommmaaa, line 385: if we don't want to index top-level domains, we don't have to list them.
as a e-mail (example: `com@example.com`) now breaks into these index entries:
1) `com@example.com`
2) `example.com`
3) `com`
You see, the top-level domain is the last entry of the resulting array, so it is simple to exclude it:
`Str.splitAlphanumericExtended(email).slice(0, -1)`
As for _common domains_. we can create and use a list like this the way you did.

IvanPizhenko: @rrrooommmaaa Fixed it.

IvanPizhenko: Anything else should be done here? Where to put this list of common domains and how to use it for filtering?

extension/js/common/platform/store/contact-store.ts
```diff
@@ -368,7 +368,14 @@ export class ContactStore extends AbstractStore {
     const name = emailEntity.name ? emailEntity.name.toLowerCase() : '';
     // we only need the longest word if it starts with a shorter one,
     // e.g. we don't need "flowcrypt" if we have "flowcryptcompatibility"
-    const sortedNormalized = [...Str.splitAlphanumericExtended(email), ...Str.splitAlphanumericExtended(name)].filter(p => !!p)
+    // also, filter out top level domains from emails
+    const emailTokens = Str.splitAlphanumericExtended(email);
+    // handle correctly email like user@host
```
rrrooommmaaa, line 373: how is it possible to get an email address like `user@host`?

extension/js/common/platform/store/contact-store.ts
rrrooommmaaa: The way you did previously:
const commonDomains = [
'gmail.com'
...
]

```
.filter(s => !ContactStore.commonDomains.includes(ContactStore.normalizeString(s)));
```

Also, need to update this procedure
https://github.com/FlowCrypt/flowcrypt-browser/blob/42d87e826decfe663940972ad60a518bacbf00df/extension/js/background_page/migrations.ts#L143
It currently updates all searchables (and sets the flag `contact_store_searchable_extended` in order to not be run twice).
As we now need to run the update even if this flag is set, we need to add a new flag like `contact_store_searchable_pruned`. The old flag `contact_store_searchable_extended` should be removed because we no longer need it (if `searchable` wasn't extended, it will be during this migration -- `updateSearchable()` fully re-generates the `searchable`).
And the method itself should be renamed to `updateSearchables`.

extension/js/common/platform/store/contact-store.ts
IvanPizhenko: I don't know, just potentially possible. Maybe in some test. But ok, will rewrite code.

IvanPizhenko:
@rrrooommmaaa Added filtering by common domains, please recheck

test/source/tests/browser-unit-tests/unit-ContactStore.js
```diff
@@ -83,12 +83,6 @@ BROWSER_UNIT_TEST_NAME(`ContactStore is able to search by a chunk spanning acros
       throw Error(`Expected 3 contacts to match "abcdef." but got "${test.length}"`);
     }
   }
-  {
```
rrrooommmaaa, line 86: So actually instead of deleting this test, we can test that `.length === 0`.
We'd better put this into a separate test (`ContactStore.search ignores top-level domains and common domains`).
With this setup 
```
  await ContactStore.update(undefined, 'abcdef.com@abcdef.com', { pubkey: testConstants.abcdefTestComPubkey });
  await ContactStore.update(undefined, 'abcdef@test.com', { pubkey: testConstants.abcdefTestComPubkey });
  await ContactStore.update(undefined, 'test@abcdef.com', { pubkey: testConstants.abcdefTestComPubkey });
```
searching for `test` should yield one result (`test@abcdef.com`)
searching for `com` should yield one result (`abcdef.com@abcdef.com`)
searching for `test.com` should yield no results

IvanPizhenko:
@rrrooommmaaa I have added the test. Please recheck.

rrrooommmaaa (approved):

PR #4353, opened 2022-03-26
bind message gateway id

### LINKED ISSUE #4269: send MessageID to FES for easier threading of conversations on Gmail

rrrooommmaaa:
This PR binds message gateway id for pwd-encrypted messages

close #4269

- Tests added or updated

### To be filled by reviewers


rrrooommmaaa:
> Sorry for lack of info. Indeed, I suppose we didn't update FES yet. I'll let you know once updated.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4284#issuecomment-1077716906_

Should I keep this PR draft meanwhile?

tomholub:
yup 👍

tomholub:
Actually, the whole time we had 2022-05 running, which is the latest version.

Have you tried this smaller PR against a live instance, is it giving you any trouble?

tomholub:
If you test it against a live instance and it works as intended, you can mark it as ready for review. Thanks! 👍

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -214,6 +214,9 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
         throw e;
       }
     }
+    if (msg.externalId) {
+      this.view.acctServer.messageGatewayUpdate(msg.externalId, msgSentRes.id).catch(Catch.reportErr);
```
tomholub, line 218: I think we can `await` this, and put it just after `draftDelete`. I suppose we could make the two calls in parallel if we want to speed it up a little bit.

tomholub (commented):
Looks good - see comment below

rrrooommmaaa:
> Actually, the whole time we had FES 2022-05 running, which is the latest version.
> 
> Have you tried this smaller PR against a live instance, is it giving you any trouble?

I tried that in early February. I don't quite get it -- should I log in as `@flowcrypt.com` to connect to the right instance?

tomholub:
> I tried that in early February. 

ok - that was likely on older instance

> I don't quite get it -- should I log in as @flowcrypt.com to connect to the right instance?

correct - fes.flowcrypt.com will be used on `@flowcrypt.com` account. fes.example.com will be used on example.com account if it's running there, etc.

tomholub (approved):

Issue #3937, opened 2021-08-29
Save S/MIME public key (certificate) to a file with a reasonably long filename


rrrooommmaaa:
Currently, `keyinfoAsPubkeyAttachment` is using `longid` to generate a file name for saving a public key.
Thus, for X.509 S/MIME certificates we have filenames of 200+ characters

PR #4392, opened 2022-04-11
issue #4387 expect exact file names when using awaitDownloadTriggeredByClicking() in tests

### LINKED ISSUE #4387: expect exact file names when using`awaitDownloadTriggeredByClicking` in tests
For all calls to `awaitDownloadTriggeredByClicking` in tests, we should start checking for the exact filename. Also need to rename `fileText` to `downloadedFiles` or similar.

Something like

```diff
- expect(Object.values(fileText).pop()!.toString()).to.equal(`small text file\nnot much here\nthis worked\n`);
+ expect(downloadedFiles["some-specific-file-name.txt"].toString()).to.equal('tested content')
```

and so on.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4372#discussion_r847157102_

IvanPizhenko:
This PR adds exact file names when using `awaitDownloadTriggeredByClicking()` in tests

close #4387

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/flaky.ts
+++ b/test/source/tests/flaky.ts
@@ -108,9 +108,9 @@ export const defineFlakyTests = (testVariant: TestVariant, testWithBrowser: Test
       const fingerprint = (await settingsPage.read('.good', true)).split(' ').join('');
       const myKeyFrame = await browser.newPage(t, `chrome/settings/modules/my_key.htm?placement=settings&parentTabId=60%3A0&acctEmail=${acctEmail}&fingerprint=${fingerprint}`);
       const downloadedFiles = await myKeyFrame.awaitDownloadTriggeredByClicking('@action-download-prv');
-      // Key ID is last 16 characters of the fingerprint
-      const keyID = fingerprint.substring(fingerprint.length - 16);
-      const fileName = `flowcrypt-backup-usernosubmitorgruleflowcrypttest-0x${keyID}.asc`;
+      // const longid = OpenPGPKey.fingerprintToLongid(fingerprint);
+      const longid = fingerprint.substring(fingerprint.length - 16);
+      const fileName = `flowcrypt-backup-usernosubmitorgruleflowcrypttest-0x${longid}.asc`;
       const key = await KeyUtil.parse(downloadedFiles[fileName]!.toString());
       expect(key.algo.bits).to.equal(3072);
       expect(key.algo.algorithm).to.equal('rsa_encrypt_sign');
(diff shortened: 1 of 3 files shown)


IvanPizhenko:
@tomholub Ready for review

rrrooommmaaa (approved):

test/source/tests/flaky.ts
```diff
@@ -107,8 +107,10 @@ export const defineFlakyTests = (testVariant: TestVariant, testWithBrowser: Test
       await SettingsPageRecipe.toggleScreen(settingsPage, 'additional');
       const fingerprint = (await settingsPage.read('.good', true)).split(' ').join('');
       const myKeyFrame = await browser.newPage(t, `chrome/settings/modules/my_key.htm?placement=settings&parentTabId=60%3A0&acctEmail=${acctEmail}&fingerprint=${fingerprint}`);
-      const files = await myKeyFrame.awaitDownloadTriggeredByClicking('@action-download-prv');
-      const key = await KeyUtil.parse(Object.values(files).pop()!.toString());
+      const downloadedFiles = await myKeyFrame.awaitDownloadTriggeredByClicking('@action-download-prv');
+      // It is not possible to have predictable file name here, because key is generated
```
rrrooommmaaa, line 111: however, we have the new key's fingerprint in the `fingerprint` variable?

rrrooommmaaa (changes requested):

test/source/tests/flaky.ts
IvanPizhenko: file name contains not fingerprint, but key id. is it possible to compute key id from the fingerprint?

tomholub: they are the same thing

IvanPizhenko: fingerprint is longer then key id.

tomholub: In this codebase we refer to primary key id (fingerprint) as `.id` eg `key.id` which is what I used for the filename. It should be the same as the fingerprint above.

IvanPizhenko:
@tomholub @rrrooommmaaa Fixed it, please recheck.

test/source/tests/flaky.ts
rrrooommmaaa: > fingerprint is longer than key id. is key id part of fingerprint?

there may be spacing between groups of 4 characters. Why don't you print the results of the test? They should show both actual and expected values.

test/source/tests/flaky.ts
```diff
@@ -107,8 +107,11 @@ export const defineFlakyTests = (testVariant: TestVariant, testWithBrowser: Test
       await SettingsPageRecipe.toggleScreen(settingsPage, 'additional');
       const fingerprint = (await settingsPage.read('.good', true)).split(' ').join('');
       const myKeyFrame = await browser.newPage(t, `chrome/settings/modules/my_key.htm?placement=settings&parentTabId=60%3A0&acctEmail=${acctEmail}&fingerprint=${fingerprint}`);
-      const files = await myKeyFrame.awaitDownloadTriggeredByClicking('@action-download-prv');
-      const key = await KeyUtil.parse(Object.values(files).pop()!.toString());
+      const downloadedFiles = await myKeyFrame.awaitDownloadTriggeredByClicking('@action-download-prv');
+      // Key ID is last 16 characters of the fingerprint
```
rrrooommmaaa, line 111: This is actually named `longid` and it is calculated this way for `OpenPGP` keys only. For S/MIME keys we use another approach that helps identify the key from the message data. Note that it may be too long and I filed this issue #3937 previously

rrrooommmaaa (approved):

PR #4394, opened 2022-04-12
show original message derived from generic error message

### LINKED ISSUE #4386: "show original message" functionality shows JSON representation of the mime message
Description:
A concerned user reported to us that the "show original message" functionality that is usually shown when a decrypt error occurs shows a JSON representation of a raw mime message that confuses the user.

I wasn't able to replicate the issue for the time being so I'm taking down some notes so anyone would be aware of it and can provide more information if anyone encountered the said issue. Likewise, I'll provide more info whenever I have.

Reference: https://mail.google.com/mail/u/2/#inbox/FMfcgzGmvpBZtPqzLZGdBHFjQkGvHDNs

martgil:
This PR will improve how the original message was shown from JSON data from native error to parsing JSON data and show actual original message

preview:
image

steps to reproduce:
1. send a signed message (detached signature) with random message to trigger the error.
2. clicking 'show original message' will now show the email body instead of the json error #4386

close #4386 // if this PR closes an issue

### To be filled by reviewers


rrrooommmaaa:
@martgil can you provide a message that produces this behaviour?

martgil:
Here you go - [Sample.eml.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/8501364/Sample.eml.txt). It typically occurs as a bad format error showing the JSON data. For me, I just reloaded the extension and choose the consumer build.

rrrooommmaaa:
> Here you go - [Sample.eml.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/8501364/Sample.eml.txt). 

Well, this particular message is displayed correctly with current `master`. It doesn't even fall into `handleInitializeErr()` method
image

Also, I concerned about type casts. Maybe you wanted to fix the `data` property of the exception at the moment of its generation?
https://github.com/FlowCrypt/flowcrypt-browser/blob/77751e9ac1f9437a2f2f3e8c350e66d62a92ca83/extension/js/common/api/email-provider/gmail/gmail.ts#L306

rrrooommmaaa:
If we want to report this error with `reportErr`, we may implement 2 properties in `FormatError` -- one for report and the other for display. If we don't want to report it, only text for display is enough.

martgil:
> Well, this particular message is displayed correctly with current master. It doesn't even fall into handleInitializeErr() method

You're right, I have seen this to be working properly though it's strange that I'd encounter the error by loading the other build with the same branch (master).

> Also, I'm concerned about type casts. Maybe you wanted to fix the data property of the exception at the moment of its generation?

thanks for sharing the reference. your idea is more correct - I'll follow your suggestion.

> 
If we want to report this error with reportErr, we may implement 2 properties in FormatError -- one for report and the other for display. If we don't want to report it, only text for display is enough.

I'm not entirely sure what's best for it. may we ask Tom's suggestions here?

rrrooommmaaa:
> 
> If we want to report this error with reportErr, we may implement 2 properties in FormatError -- one for report and the other for display. If we don't want to report it, only text for display is enough.
> 
> I'm not entirely sure what's best for it. may we ask Tom's suggestions here?

Yes, please. ping @tomholub

martgil:
sir @tomholub - Would you be interested in reporting the error with `reportErr` or showing the plain text (display) is sufficient? Thanks!

tomholub:
No report here. Thanks

martgil:
sir @rrrooommmaaa ready for review. thank you!

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -304,7 +305,7 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
       if (plaintext) {
         return { armored: '', plaintext, subject, isPwdMsg };
       }
-      throw new FormatError('Armored message not found', JSON.stringify(gmailMsg.payload, undefined, 2));
+      throw new FormatError('Armored message not found', base64decode(gmailMsg.payload!.parts![0].parts![0].body!.data as string));
```
rrrooommmaaa, line 308: please explain the structure of `gmailMsg.payload`.
Why is it guaranteed to have `parts[0].parts[0]`?

rrrooommmaaa (changes requested):

extension/js/common/api/email-provider/gmail/gmail.ts
martgil: Hello @rrrooommmaaa,

1.) As far as I understand, no - I realize what if I wrap the throw formatError block in an if condition, for example:
```typescript
const bodies = GmailParser.findBodies(gmailMsg);
...
...
if (typeof bodies['text/plain'] === 'string') {
        throw new FormatError('Armored message not found', base64decode(bodies['text/plain']));
}
```
or something similar to ensure that every property has a value that is not undefined.


Another problem is that - I'm unable to reproduce the issue even on an older version 8.2.4. However, I'm worrying that leaving this JSON.stringify in the code would not be good? 

2.) I have prepared the test message within the compatibility test account. It has the subject name of `Bad Format produces json output`. However, I'm no longer able to replicate it now - Tbh, there are some users that are experiencing this but reinstalling the extension helps them resolve the issue. 

Here it is from the email (support): https://mail.google.com/mail/u/6/#inbox/FMfcgzGpFgmLZvDsTLCKhgRRDVLxhWhQ

martgil: The error pop up once again:

I'm using firefox browser 99.0.1 in ubuntu and a flowcrypt extension copy from flowcrypt.com and updated to 8.2.8. I'm not experiencing the error when using firefox on a mac. this is really strage.

rrrooommmaaa: We already have `textBody` and `htmlBody` variables, why not reuse them?
`throw new FormatError('Armored message not found', textBody || htmlBody);`

martgil: I applied the changes - thanks for your patience. I will try to be more attentive next time.

rrrooommmaaa (approved):

PR #4400, opened 2022-04-17
issue #3520 Properly show ISO-2022-JP encoded content

### LINKED ISSUE #3520: iso-2022-jp signed message garbled
ref: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgxwLtGlbzXGTcSBkfbffLLgFwttB

It appears we assume UTF-8,  most explicitly in the `fromUtfStr ` method here:

https://github.com/FlowCrypt/flowcrypt-browser/blob/53ce61e2228a55bc624a09d2f354e6f0be90aaea/extension/chrome/elements/pgp_block.ts#L58

But perhaps in other ways as well. 

This may not seem particularly important given the dominance of Unicode, even in Japanese writing. But email is an exception:

> ISO-2022-JP is the 7-bit JIS encoding, that e-mails are usually sent in. It is rarely used in other contexts.

Source: https://www.sljfaq.org/afaq/encodings.html#encodings-Overview-of-the-encoding-schemes

So by not being compatible with ISO-2022-JP, FlowCrypt is incompatible with the dominant form of encoding used for email in Japan!

IvanPizhenko:
This PR fixes showing ISO-2022-JP encoded content

close #3520

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/lib/emailjs/emailjs-mime-parser.js
+++ b/extension/lib/emailjs/emailjs-mime-parser.js
@@ -810,13 +810,13 @@
             }
 
             // override charset for text nodes
-            // issue #3520 - do not override ISO-2022-JP
-            if (this.charset?.toUpperCase() !== 'ISO-2022-JP') {
+            // issue #3520 - do not override 7bit econdings and ISO-2022-JPxxx
+            if (!((this.headers['Content-Transfer-Encoding'] && this.headers['Content-Transfer-Encoding'].toLowerCase() === '7bit')
+                || (this.charset && this.charset.toUpperCase().startsWith('ISO-2022-JP')))) {
                 this.charset = this.contentType.params.charset = 'utf-8';
             }
         }
         this._bodyBuffer = '';
-
         this._parser.onbody(this, this.content);
     };
 
--- a/extension/lib/emailjs/emailjs-mime-parser.js
+++ b/extension/lib/emailjs/emailjs-mime-parser.js
@@ -811,7 +811,7 @@
 
             // override charset for text nodes
             // issue #3520 - do not override 7bit econdings and ISO-2022-JPxxx
-            if (!((this.headers['Content-Transfer-Encoding'] && this.headers['Content-Transfer-Encoding'].toLowerCase() === '7bit')
+            if (!((this.contentTransferEncoding && this.contentTransferEncoding.value && this.contentTransferEncoding.value === '7bit')
                 || (this.charset && this.charset.toUpperCase().startsWith('ISO-2022-JP')))) {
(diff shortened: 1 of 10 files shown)


IvanPizhenko:

IvanPizhenko:

extension/lib/emailjs/emailjs-mime-parser.js
```diff
@@ -810,7 +810,10 @@
             }
 
             // override charset for text nodes
-            this.charset = this.contentType.params.charset = 'utf-8';
+            // issue #3520 - do not override ISO-2022-JP
+            if (this.charset?.toUpperCase() !== 'ISO-2022-JP') {
```
rrrooommmaaa, line 814: I suggest to exclude based on `Content-Transfer-Encoding: 7bit`
As I mentioned before, 7bit is guaranteed to be not compatible with UTF-8.

rrrooommmaaa (changes requested):

extension/lib/emailjs/emailjs-mime-parser.js
rrrooommmaaa: However, this header may be missing...
So we'd better check that either 7bit or it starts with (ISO-2022-JP...) (there is also ISO-2022-JP-2)

IvanPizhenko: @rrrooommmaaa How do I get value of `Content-Transfer-Encoding`?

IvanPizhenko:
@rrrooommmaaa Please recheck

extension/lib/emailjs/emailjs-mime-parser.js
```diff
@@ -810,10 +810,13 @@
             }
 
             // override charset for text nodes
-            this.charset = this.contentType.params.charset = 'utf-8';
+            // issue #3520 - do not override 7bit econdings and ISO-2022-JPxxx
+            if (!((this.headers['Content-Transfer-Encoding'] && this.headers['Content-Transfer-Encoding'].toLowerCase() === '7bit')
+                || (this.charset && this.charset.toUpperCase().startsWith('ISO-2022-JP')))) {
+                this.charset = this.contentType.params.charset = 'utf-8';
+            }
```
rrrooommmaaa, line 817: have you checked if this method was run and the property `contentTransferEncoding` was set?
image
Also, headers are probably keyed lower-cased

IvanPizhenko:
@rrrooommmaaa Please recheck

rrrooommmaaa (approved):

PR #4416, opened 2022-04-27
use options_ui in manifest.json file

### LINKED ISSUE #4415: Usage of deprecated options_page in manifest.json
affected file: https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/manifest.json#L72

https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/options_page

solution: update the manifest file using options_ui as an alternative.

image

An error message like I mentioned on https://github.com/FlowCrypt/flowcrypt-browser/pull/4394 started to work more appropriately after the update. I tried testing it locally. eg. From 'Bad Format...' to 'Error: could not properly parse signed message...' (might be coincidence 😮‍💨).

It's best when we are using non-deprecated property.

reference: https://mail.google.com/mail/u/2/#inbox/FMfcgzGpFgmKwsFhjdgvVmGqDThsdGbb

martgil:
This PR will update the manifest.json file to use options_ui

reference: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/options_ui

close #4415  // if this PR closes an issue

### To be filled by reviewers


martgil:
@tomholub 

> How about chrome - could you check the documentation there too?

based on https://developer.chrome.com/docs/extensions/reference/browserAction/, the properties were using is not yet deprecated.

I have faced some issue that needs helps - https://github.com/FlowCrypt/flowcrypt-workspace-key-manager/issues/338

tomholub:
I was talking about the "URI filtering" email.

tomholub:
I mean, is `options_ui` supported in chrome? If you are replacing `options_page` with `options_ui` then you'll want to double check it's not breaking existing functionality on chrome.

martgil:
@tomholub works fine on both Chrome and Firefox browsers after adding another property "open_in_tab: true".

tomholub:
You have to be careful with this. Properly checking that your update works across all the needed contexts (in this case chrome vs firefox) is what differentiates a junior developer from others.

In this case, it was recognized in the review process, but you have to have your own initiative to check for that first.

tomholub:
@rrrooommmaaa please see if this looks ok

rrrooommmaaa (approved):

PR #4419, opened 2022-04-28
use Google API subdomains alternatives

martgil:
This PR will use Google API specific subdomains as an alternative to www.googleapis.com

close 4417

- Tests added or updated

### To be filled by reviewers


tomholub (changes requested):
I'm adding a "request changes" just to make sure we don't merge this.

For now this will work as a WIP.

To finalize it, I will later have to:
 - change the build steps to add one more build type, `chrome-enterprise-google-api-subdomains` that has the desired subdomains
 - appropriately change the API URL based on which one is used (www.googleapis.com or oauth2.googleapis.com) probably with another build-time value
 - double check functionality

tomholub:
Here is the dev build that uses these urls

[DEV-chrome-enterprise-google-api-subdomains.zip](https://github.com/FlowCrypt/flowcrypt-browser/files/8584057/DEV-chrome-enterprise-google-api-subdomains.zip)

tomholub (approved):

tomholub (approved):

PR #4427, opened 2022-05-03
Remove extra new line on signed message with detached signature

### LINKED ISSUE #4342: signed message occasionally throws an error 'Message digest did not match'
Description:

A user reported an issue where a signed message occasionally throws an error 'Message digest did not match'. I asked the responded to send a couple of signed messages and those worked well without any issue.

I looked closer into the signed message that throws 'Message digest did not match' by opening/sending it again with myself and it works well (needs modifying signerEmail not to trigger the missing pubkey error checks as its a different error with a different root cause).

I'm having trouble spotting the problem for this one as it works after moving opening it in my inbox but it shouldn't be the case. 

will try to move the .eml file to my inbox to dig into this issue.

martgil:
This PR will fix the issue for signed messages with a detached signature 

close #4342  // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -167,7 +167,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
       }
       const senderEmail = this.getSenderEmail(emailContainer);
       const isOutgoing = !!this.sendAs[senderEmail];
-      const replacementXssSafe = XssSafeFactory.replaceRenderableMsgBlocks(this.factory, emailContainer.innerText, this.determineMsgId(emailContainer), senderEmail, isOutgoing);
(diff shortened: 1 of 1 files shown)


extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -167,7 +167,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
       }
       const senderEmail = this.getSenderEmail(emailContainer);
       const isOutgoing = !!this.sendAs[senderEmail];
-      const replacementXssSafe = XssSafeFactory.replaceRenderableMsgBlocks(this.factory, emailContainer.innerText, this.determineMsgId(emailContainer), senderEmail, isOutgoing);
+      const replacementXssSafe = XssSafeFactory.replaceRenderableMsgBlocks(this.factory, emailContainer.innerText.replace('\n\n\n','\n\n'), this.determineMsgId(emailContainer), senderEmail, isOutgoing);
```
tomholub, line 170: This will cause more problems then it will fix. You shouldn't be working on signature verification, it's not trivial.

PR #4479, opened 2022-06-07
#4476 added allow_attester_search_only_for_domains org rule

### LINKED ISSUE #4476: implement allow_attester_search_only_for_domains client configuration
Regex would probably be the most flexible but also complicated.

Let's go with what Roman suggests.

> What about adding `allow_attester_search_only_for_domains` option?
> 
> If it's set then app will ignore `disallow_attester_search_for_domains` option and search only in listed domains.

Plus add an appropriate test.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1668#issuecomment-1146221302_

ioanmo226:
This PR added allow_attester_search_only_for_domains org rule

close #4476 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/org-rules.ts
+++ b/extension/js/common/org-rules.ts
@@ -166,7 +166,6 @@ export class OrgRules {
     if (!userDomain) {
       throw new Error(`Not a valid email ${emailAddr}`);
     }
-
     // When allow_attester_search_only_for_domains is set, ignore disallow_attester_search_for_domains rule
     if (this.domainRules.allow_attester_search_only_for_domains) {
       return this.domainRules.allow_attester_search_only_for_domains.includes(userDomain);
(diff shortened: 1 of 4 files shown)


extension/js/common/org-rules.ts
```diff
@@ -161,14 +162,18 @@ export class OrgRules {
...
     if (!userDomain) {
       throw new Error(`Not a valid email ${emailAddr}`);
     }
+    // When allow_attester_search_only_for_domains is set, ignore disallow_attester_search_for_domains rule
+    if (this.domainRules.allow_attester_search_only_for_domains) {
+      return this.domainRules.allow_attester_search_only_for_domains.includes(userDomain);
+    }
+    if (this.disallowLookupOnAttester()) {
+      return false;
+    }
```
tomholub, line 175: I'll file another issue. This exists for looking up by fingerprint, which we may want to later deprecate.

tomholub (approved):
👍 that went smoothly for first contribution

extension/js/common/org-rules.ts
ioanmo226: ok

PR #4508, opened 2022-06-24
Convert type any to type unknown

### LINKED ISSUE #4490: no-explicit-any warnings - prefer unknown
> It seems we have delayed after-effects after this merge. #4284 https://flowcrypt.semaphoreci.com/jobs/47dbcb8e-5bc4-437a-9812-e70699e7d855

They are all warnings, except for 

```
   148:4   error    Missing semicolon                         @typescript-eslint/semi 00:09 
```

In either case, we should try to address these warnings - I'll file an issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4484#issuecomment-1152617605_

martgil:
This PR's goal is to convert type any to type unknown. 

To check the warnings on master branch:
`npm run test_eslint`

close #4490 // if this PR closes an issue

- Not worth testing
- Tests added or updated

### To be filled by reviewers


martgil:
Hello sir, @ioanmo226 - I initially started with converting 5 of these... there is some type of checking in common places like catching an error... if this looks okay maybe I can start converting type any to unknown from there?

extension/chrome/elements/compose-modules/compose-send-btn-module.ts
```diff
@@ -249,9 +249,9 @@ export class ComposeSendBtnModule extends ViewModule<ComposeView> {
   private doSendMsgs = async (msgObj: MultipleMessages): Promise<SendMsgsResult> => {
     const sentIds: string[] = [];
     const supplementaryOperations: Promise<void>[] = [];
-    const supplementaryOperationsErrors: any[] = []; // tslint:disable-line:no-unsafe-any
+    const supplementaryOperationsErrors: unknown[] = []; // tslint:disable-line:no-unsafe-any
```
ioanmo226, line 252: you may remove `// tslint:disable-line:no-unsafe-any` comment if you already converted `any` to `unknown`
There are several places which contains this comment

martgil: oh - ok, thanks... I'll be removing them as well. thank you!

extension/types/squire.d.ts
```diff
@@ -1,9 +1,12 @@
...
+
 export as namespace SquireClass;
 
 declare global {
   interface Window {
     // @ts-ignore
     Squire: SquireClass;
+    forge: typeof forge;
   }
 }
```
tomholub, line 12: Also this is wrong to add forge into squire definitions. Let's remove it and file an issue for Ioan to have a look later.

martgil: okay sir, I'll remove it - I just converted most of it from any to unknown but there are still some that need to have proper types but I can't convert them just yet.

tomholub: That's ok. You can leave these for another PR in the future.

extension/js/common/core/const.ts
```diff
@@ -3,7 +3,7 @@
 'use strict';
 
 export const VERSION = '[BUILD_REPLACEABLE_VERSION]';
-export const FLAVOR: 'consumer' | 'enterprise' = '[BUILD_REPLACEABLE_FLAVOR]' as any;
+export const FLAVOR: 'consumer' | 'enterprise' = '[BUILD_REPLACEABLE_FLAVOR]' as unknown as 'consumer' | 'enterprise';
```
ioanmo226, line 6: I think we don't need `as unknown` here.
`as 'consumer' | 'enterprise'` will be enough

tomholub: It wouldn't work because `'[BUILD_REPLACEABLE_FLAVOR]'` is not of that type. It's a bit of a special case.

ioanmo226: It works...
Tslint and eslint doesn't return any error and vscode displays fine result.

image

ioanmo226: This way works because left type is string union type and right type is string.

tomholub: Ah - good. Can be done separately. Thanks for noticing.

PR #4514, opened 2022-07-01
#4513 fix: PUT rename /v1/end-user/key-pairs field

### LINKED ISSUE #4513: PUT /v1/end-user/key-pairs field renamed
used to be `decryptedPrivateKey` and now it's just `privateKey`. Need to update mocks too. Happened here https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/pull/3043/files

I think that change was unintentional in that PR for end-user API (I overlooked it in my review, anyway), but since we don't have any customer currently using that API, we can leave it on on-prem and update it here, even if that causes incompatibility between previous extension versions and the new on-prem version.

ioanmo226:
This PR renamed PUT /v1/end-user/key-pairs field

close #4513 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/setup.ts
+++ b/test/source/tests/setup.ts
@@ -658,7 +658,7 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await settingsPage.waitAll('@container-overlay-details');
       await Util.sleep(0.5);
       const details = await settingsPage.read('@container-overlay-details');
-      expect(details).to.contain('500 when PUT-ing https://localhost:8001/flowcrypt-email-key-manager/v1/keys/private string: decryptedPrivateKey,publicKey -> Intentional error for put.error user to test client behavior');
+      expect(details).to.contain('500 when PUT-ing https://localhost:8001/flowcrypt-email-key-manager/v1/keys/private string: privateKey -> Intentional error for put.error user to test client behavior');
       expect(details).to.not.contain('PRIVATE KEY');
       expect(details).to.not.contain('<REDACTED:');
     }));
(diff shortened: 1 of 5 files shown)


tomholub:
fixed

test/source/tests/setup.ts
```diff
@@ -625,8 +625,6 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await myKeyFrame.waitAll('@content-fingerprint');
       const fromKm = MOCK_KM_LAST_INSERTED_KEY[acct];
       expect(fromKm).to.exist;
-      const k = await KeyUtil.parse(fromKm.publicKey);
-      expect(await myKeyFrame.read('@content-fingerprint')).to.equal(Str.spaced(k.id));
```
tomholub, line 629: You can bring these two back, but use `fromKm.privateKey`. Public and private keys in a key pair have the same id (fingerprint). So you can use either to test the UI.

test/source/tests/setup.ts
```diff
@@ -695,8 +693,6 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
         await myKeyFrame.waitAll('@content-fingerprint');
         const fromKm = MOCK_KM_LAST_INSERTED_KEY[acct];
         expect(fromKm).to.exist;
-        const k = await KeyUtil.parse(fromKm.publicKey);
-        expect(await myKeyFrame.read('@content-fingerprint')).to.equal(Str.spaced(k.id));
```
tomholub, line 699: Same here

tomholub (commented):
👍 see comments

tomholub (dismissed):

tomholub:
Not sure why different id now, maybe I made a new key for demo instead of recovering.

This should be it now 09977F6F512CA5AD76F0C210248B60EB6D044DF8 you'll just need to space each 4 chars.

tomholub (approved):

PR #4519, opened 2022-07-04
#4335 Improve EKM import key screen response

### LINKED ISSUE #4335: improve EKM import key screen response
There are still some things left to adjust, but that will be for another issue:

 - following click of the setup button, there is no visual indication - no loader / spinner. We should make the button `gray` color with text `Loading..` and disable click events on it. Then the try/catch I just added should have a `finally` that re-enables the button
 - the EKM timeout is 30 seconds which is too long. Should make it 15 seconds for the EKM call

As it is now, user has to sit there and wait 30 seconds to see any UI change in some situations, even though I have fixed the issue. Only after that does the modal appear.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4334#issuecomment-1065320946_

ioanmo226:
This PR improved EKM import key screen response

close #4335 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/chrome/settings/setup/setup-key-manager-autogen.ts
+++ b/extension/chrome/settings/setup/setup-key-manager-autogen.ts
@@ -23,6 +23,7 @@ export class SetupWithEmailKeyManagerModule {
   public continueEkmSetupHandler = async () => {
     const submitButtonSelector = '#step_2_ekm_choose_pass_phrase .action_proceed_private';
     const submitButton = $(submitButtonSelector);
+    const submitButtonText = submitButton.text();
     const setBtnColor = (type: 'gray' | 'green') => {
       submitButton.addClass(type === 'gray' ? 'gray' : 'green');
       submitButton.removeClass(type === 'gray' ? 'green' : 'gray');
@@ -39,7 +40,7 @@ export class SetupWithEmailKeyManagerModule {
       await Ui.modal.error(String(e));
     } finally {
       setBtnColor('green');
-      submitButton.text('CONTINUE');
+      submitButton.text(submitButtonText);
     }
   };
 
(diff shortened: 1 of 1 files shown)


extension/chrome/settings/setup/setup-key-manager-autogen.ts
```diff
@@ -13,21 +13,33 @@ import { Settings } from '../../../js/common/settings.js';
...
 import { Lang } from '../../../js/common/lang.js';
+import { Xss } from '../../../js/common/platform/xss.js';
 
 export class SetupWithEmailKeyManagerModule {
 
   constructor(private view: SetupView) {
   }
 
   public continueEkmSetupHandler = async () => {
+    const submitButtonSelelctor = '#step_2_ekm_choose_pass_phrase .action_proceed_private';
```
rrrooommmaaa, line 24: a typo, should be `submitButtonSelector`

extension/chrome/settings/setup/setup-key-manager-autogen.ts
```diff
@@ -13,21 +13,33 @@ import { Settings } from '../../../js/common/settings.js';
...
-      }
+      Xss.sanitizeRender(submitButtonSelector, Ui.spinner('white') + 'Loading...');
+      setBtnColor('gray');
       const passphrase = $('#step_2_ekm_choose_pass_phrase .input_password').val();
       await this.setupWithEkmThenRenderSetupDone(typeof passphrase === 'string' ? passphrase : '');
     } catch (e) {
       await Ui.modal.error(String(e));
+    } finally {
+      setBtnColor('green');
+      submitButton.text('CONTINUE');
```
rrrooommmaaa, line 42: We essentially define the label 'CONTINUE' in 2 places: here and in `.htm` -- not perfect.
It'd be better to make it visible/invisible by class. But we'll need a separate spinner then, similar to `spinner_container` in `add_key.htm`...

rrrooommmaaa (approved):

extension/chrome/settings/setup/setup-key-manager-autogen.ts
tomholub: > Well, i think better to save button text temporary before we set spinner and then recover it after spinner is finished.

I've done that in several places in this codebase - save at the beginning, restore in catch or finally depending on context. It works well and text is defined in a single place.

PR #4535, opened 2022-07-10
Issue 4498 reply box recipients preview width fix

### LINKED ISSUE #4498: encyrpted reply frame is initially too wide if there are many recipients
Ther ecipients were already on the previous message, and I clicked `secure reply all`

image

There are many more recipients (about 10) but notice the horizontal scroller on the bottom - the acrual frame is about 3-4x wider than the visible part, I scrolled slightly to the right to demostrate - notice the send button slightly cut off on the left when I do that.

It self-adjusts when I click the recipients and click away:

image

Could add a test that would be quite simple: render reply box with 10 recipients (stand-alone) and ensure that it shows the "and X more" as soon as it renders, without me having to focus/blur the recipients.

rrrooommmaaa:
This PR is going to resize reply box on opening
(Also correctly parses multiple emails from `Reply-To` header)
(Also small debug output fix)
close #4498

- Tests added or updated

### To be filled by reviewers


tomholub:
I will be traveling, once ready, @ioanmo226 please have a look

ioanmo226 (approved):
Looks good 👍

PR #4536, opened 2022-07-12
Issue 206 add input fields character limit

martgil:
This PR adds a character limit on input fields, including the secure compose message box

close https://github.com/FlowCrypt/flowcrypt-security/issues/206

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/compose.ts
```diff
@@ -199,6 +199,11 @@ export class ComposeView extends View {
     this.S.cached('body').on('focusin', setActiveWindow);
     this.S.cached('body').on('click', setActiveWindow);
     this.S.cached('icon_help').click(this.setHandler(async () => await this.renderModule.openSettingsWithDialog('help'), this.errModule.handle(`help dialog`)));
+    this.S.cached('input_text').on('keypress paste',this.setHandler((el, ev) => {
+      if (this.S.cached('input_text').children('div').text().length >= 10) { // setting 10 as a limit for ease of testing; 50kB is tested to work with low spec pc;
+        ev.preventDefault();
+      }
+    }));
```
martgil, line 206: Good afternoon, sir, @rrrooommmaaa. I'm trying to add an input limit on the secure compose message box and it doesn't work as expected; 

What happens is that onpaste seems not working and doesn't prevent the user from pasting text/contents even though the keypress works just fine.

I tried to try the whole concept at https://jsfiddle.net/ftye9xrn/ starting on the minimal code. The output is much more acceptable there. Do you have any suggestions here?

Furthermore, the goal of this PR is to prevent user's from populating input fields with large texts that cause an unresponsive browser tab (not the entire browser)  - related PR https://github.com/FlowCrypt/flowcrypt-security/issues/206. If by any chance you have any feedback or suggestions, please let me know and I'll be happy to share them with Tom.

Thank you!

rrrooommmaaa: > I tried to try the whole concept at https://jsfiddle.net/ftye9xrn/ starting on the minimal code. The output is much more acceptable there. Do you have any suggestions here?

Hi @martgil. Sorry for the late answer, I somehow missed the notification.
The `paste` event handler obviously takes the value of the input _before_ the modification, this is why there is no reaction.
The handler for paste event should look somthing like this:
```
$("#test").on("paste", function (e) {
  const selectedLength = window.getSelection().toString().length;
  const curLength = $("#test").text().length;
  const clipboardLength = e.originalEvent.clipboardData.getData('text').length;
  const resultingLength = curLength - selectedLength + clipboardLength;
  if (resultingLength > 10) {
  	e.preventDefault();
  }
})
```
However the [Selection](https://developer.mozilla.org/en-US/docs/Web/API/Selection) is a complex object, consisting of possibly more than one range, maybe we need to do something more sophisticated than `toString().length` on it, I'm encouraging you to further investigate this, @martgil

rrrooommmaaa: Probably it's better to use these properties:
- `input.selectionStart` – position of selection start (writeable),
- `input.selectionEnd` – position of selection end (writeable),
as described [here](https://javascript.info/selection-range#selection-in-form-controls)

martgil: thank you for providing such detailed feedback, sir. I will get back to this after i accomplished some tasks with higher priority. thanks again.

martgil:
@rrrooommmaaa do you think it's reasonable to move the tasks for adding a length constraint to compose window (contenteditable) to a separate issue so non-complex changes will prioritize to get merged?
Also, should we add a length constraint to the armored key inputs (I can add it in this PR)?

rrrooommmaaa:
> @rrrooommmaaa do you think it's reasonable to move the tasks for adding a length constraint to compose window (contenteditable) to a separate issue so non-complex changes will prioritize to get merged?

Sure, feel free to create a new issue for this sub-task. Please don't forget to include/reference my instructions there.

> Also, should we add a length constraint to the armored key inputs (I can add it in this PR)?

I'm not sure about what kind of vulnerability we're trying to protect against here. If we're supposed to accept only a single key there, then it makes sense to limit the length. If multiple keys are allowed, well... depends on performance, I guess. Unless you know of a particular attack.

martgil:
> Sure, feel free to create a new issue for this sub-task. Please don't forget to include/reference my instructions there.

Thanks!

> I'm not sure about what kind of vulnerability we're trying to protect against here.

We are trying to mitigate web browser freezing & browser crash when populating input fields with extremely long strings (client-side) - reference: https://github.com/FlowCrypt/flowcrypt-security/issues/206. 

To reproduce the issue,  download and paste the payload content to any input fields of the browser extension and the browser may start to freeze/crash.

[payload.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/9999188/payload.1.txt)

>  If we're supposed to accept only a single key there, then it makes sense to limit the length. If multiple keys are allowed, well... depends on performance, I guess. Unless you know of a particular attack.

Sorry, I'm just referring with the text length rather than the key length.

rrrooommmaaa (approved):

PR #4537, opened 2022-07-12
#3505 Fix: decode attachment file name

### LINKED ISSUE #3505: Special characters in attachment filenames don't work
### Steps to reproduce:

1. Create a file whose name contains a character not allowed in URL's, like `what's_up?.txt`
2. Send the file via Secure Compose
3. Download the file from the recipient inbox
4. Viola, the filename on your machine will be `what&#39;s_up_.txt`

My guess would be that this filename is decoded from a URL, but a cursory glance at the relevant code didn't immediately reveal anything, so I'm still unsure.

I'll look deeper into this but in the meantime any ideas/suggestions are appreciated :)

ioanmo226:
This PR decodes attachment file name

close #3505 // if this PR closes an issue

- Tests added

### To be filled by reviewers


rrrooommmaaa (commented):
why doesn't it need a test? We could add a message (to `exported-messages`) that was wrongly rendered previously and correctly rendered now.

test/source/tests/browser-unit-tests/unit-Mime.js
```diff
@@ -338,3 +338,24 @@ BROWSER_UNIT_TEST_NAME(`Mime attachment file name issue 3352`);
   }
   return 'pass';
 })();
+
+BROWSER_UNIT_TEST_NAME(`Mime attachment file name issue #3505`);
+(async () => {
+  const originalName = 'what\'s_up?.txt';
```
ioanmo226, line 344: @rrrooommmaaa FYI I tried to combine those duplicated code (With above logic) but then I realized I couldn't add custom javascript function in this file as this file is parsed in unit-browser.ts file.

rrrooommmaaa: To re-use library code in unit browser tests we should add import into `extension/chrome/dev/ci_unit_test.ts` file and also add the imported instance to `libs` array there, if that's what you're asking.

ioanmo226: Aha, got it
Thanks

test/source/tests/browser-unit-tests/unit-Mime.js
```diff
@@ -320,21 +320,10 @@ BROWSER_UNIT_TEST_NAME(`Mime attachment file names`);
...
-  if (extractedName !== originalName) {
-    throw Error(`extractedName unexpectedly ${extractedName}, expecting ${originalName}`);
-  }
-  return 'pass';
+  return await Mime.checkAttachmentDecode('XX J 1 IT E (P 4) p_c.pdf');
+})();
+
+BROWSER_UNIT_TEST_NAME(`Mime attachment file name issue #3505`);
+(async () => {
+  return await Mime.checkAttachmentDecode('what\'s_up?.txt');
```
rrrooommmaaa, line 328: not quite clear how this test covers `XssSafeFactory.srcPgpAttachmentIframe` method.
Can you please provide a `eml`-formatted message with this attachment? We can then convert it to our `json` format for testing.

rrrooommmaaa (changes requested):

test/source/tests/browser-unit-tests/unit-Mime.js
rrrooommmaaa: if you have it in a gmail inbox, it's even easier. EML is not needed. You need to have Flowcrypt extension installed for this account and...
https://github.com/FlowCrypt/flowcrypt-browser/blob/fae2bb7d00555a53713cb13c2300b54ff29c3aa6/test/source/mock/google/google-data.ts#L102:L106

rrrooommmaaa:
```
```

ioanmo226: Got it. Thank you

extension/js/common/browser/browser.ts
```diff
@@ -27,7 +26,7 @@ export class Browser {
     const blob = new Blob([attachment.getData()], { type: attachment.type });
     const a = window.document.createElement('a');
     a.href = window.URL.createObjectURL(blob);
-    a.download = Xss.escape(attachment.name);
+    a.download = attachment.name;
```
ioanmo226, line 29: @rrrooommmaaa FYI This escape function converted special characters to something like `&quot;` and as I checked it's not needed for `download` a tag.

extension/chrome/elements/attachment.ts
```diff
@@ -79,9 +79,10 @@ export class AttachmentDownloadView extends View {
       return;
     }
     $('#type').text(this.type || 'unknown type');
-    $('#name').text(this.name || 'noname');
+    const attachmentName = decodeURIComponent(this.name || 'noname');
```
rrrooommmaaa, line 82: Are you sure we need this extra decoding? 
I sent a message with an attachment name `what's_up%253F.txt`
When downloaded, the file name became `what's_up%3F.txt`
(I committed the message 182263bf9f105adf to the repo)
Please add a test that the downloaded file name matches the initial one.

test/source/tests/settings.ts
```diff
@@ -506,6 +506,20 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
       Util.deleteFileIfExists(downloadedAttachmentFilename);
     }));
 
+    ava.default('settings - check if downloaded attachment name is correct', testWithBrowser('compatibility', async (t, browser) => {
+      // `what's up?.txt` becomes `what's_up_.txt` and this is native way and we can't change this logic
+      // https://github.com/FlowCrypt/flowcrypt-browser/issues/3505#issuecomment-812269422
+      const downloadedAttachmentFilename = `${__dirname}/what\'s_up_.txt`;
+      const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=flowcrypt.compatibility@gmail.com&threadId=1821bf879a6f71e0`));
+      // @ts-ignore
+      await (inboxPage.target as Page)._client.send('Page.setDownloadBehavior', { behavior: 'allow', downloadPath: __dirname });
```
rrrooommmaaa, line 515: I'd very much like to re-use the method `awaitDownloadTriggeredByClicking` that we have in `controllable.ts`, rather than making copies this low-level code.
We could make `awaitDownloadTriggeredByClicking` return filenames as well and probably an option to include the file contents (so we can turn it off when it's not required)

extension/js/common/xss-safe-factory.ts
```diff
@@ -134,7 +134,7 @@ export class XssSafeFactory {
       a.url = Browser.objUrlCreate(a.getData());
     }
     return this.frameSrc(this.extUrl(iframeUrl), {
-      frameId: this.newId(), msgId: a.msgId, name: a.name, type: a.type, size: a.length, attachmentId: a.id, url: a.url, isEncrypted, errorDetailsOpened, initiatorFrameId
+      frameId: this.newId(), msgId: a.msgId, name: decodeURIComponent(a.name), type: a.type, size: a.length, attachmentId: a.id, url: a.url, isEncrypted, errorDetailsOpened, initiatorFrameId
```
rrrooommmaaa, line 137: or this decoding?

extension/chrome/elements/attachment.ts
ioanmo226: You are right. We actually don't need this encoding.

extension/js/common/xss-safe-factory.ts
ioanmo226: No. Thanks for your notice

test/source/tests/settings.ts
ioanmo226: Thank you for your suggestion, That method seems what I was looking for.

test/source/tests/settings.ts
```diff
@@ -506,6 +508,23 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
       Util.deleteFileIfExists(downloadedAttachmentFilename);
     }));
 
+    const checkIfFileDownloadCorrectly = async (t: AvaContext, browser: BrowserHandle, threadId: string, fileName: string) => {
```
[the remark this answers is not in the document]
ioanmo226: Yeah. sure. it sounds better.

rrrooommmaaa (approved):

test/source/mock/google/exported-messages/message-export-1821bf879a6f71e0.json
```diff
@@ -0,0 +1,122 @@
...
+          }
+        },
+        {
+          "partId": "1",
+          "mimeType": "application/octet-stream",
+          "filename": "\"what\"'s_up%3F.txt.pgp",
```
rrrooommmaaa, line 70: Looks like google doesn't like this apostrophe in the middle of the file name (between `what` and `s_up`
```
 filename*0*=utf-8''what's_up%25253F.txt.pgp
```
Probably this is why extra double quotes appear
image
```
        {
          "partId": "1",
          "mimeType": "application/octet-stream",
          "filename": "\"what\"'s_up%3F.txt.pgp",
```

According to https://datatracker.ietf.org/doc/html/rfc2231#section-3
the apostrophe has special meaning like in this example
```
 title*0*=us-ascii'en'This%20is%20even%20more%20
```
we have the 2 required apostrophes right after `utf-8` denoting unspecified language, that's ok, but we should probably escape the apostrophe with %27, so we can take the filename directly from Gmail API (when decoding).
If the filename was encoded like this by the browser extension, we should also fix the encoding part.
@sosnovsky ?

sosnovsky: @rrrooommmaaa I just checked this message in `flowcrypt.compatibility` account and attachment has correct name in default Gmail view and in decoded FlowCrypt view:

Screenshot 2023-05-12 at 21 39 04

Screenshot 2023-05-12 at 21 40 31

Not clear, what should be fixed here

rrrooommmaaa: There aren't double quotes around "what" in the MIME message
```
------sinikael-?=_1-16583266435440.5429396972265759
Content-Type: application/octet-stream; name="what's_up?.txt.pgp"
Content-Disposition: attachment; filename*0*=utf-8''what's_up%3F.txt.pgp
```
My guess is that Gmail added them because the apostrophe isn't allowed in the `Content-Disposition: ...` line
We should try `Content-Disposition: attachment; filename*0*=utf-8''what%27s_up%3F.txt.pgp` instead.
Gmail also didn't decode %3F, probably because of the same reason.

sosnovsky: Ah, now I see that original file name is `what's_up?.txt`, initially I thought quotes were included in original file name.
Thanks for explanation, filing an issue for it

PR #4551, opened 2022-07-20
make smart replies available for non-encrypted messages

### LINKED ISSUE #4549: FlowCrypt hides smart replies even in unencrypted messages
Description:
When opening an unencrypted message with FlowCrypt installed/enabled. When Gmail with a smart reply feature is enabled, the FlowCrypt browser extension tries to hide/remove it from the web UI in a couple/split second(s) when loading an unencrypted message, and the only way to make the smart reply suggestion re-appear is to click the reply button and then close the reply message box.

test email to use: https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGpGwrFHNCnNScHJzWqKwTBZJWV

To replicate the issue:
1. Open the test email present in our test account while the flowcrypt extension is installed. (may need to refresh the webpage because it is not noticeable when opening the message for the first time.)
2. (optional) disable flowcrypt browser extension at chrome://extensions or about:addons (firefox) to see that smart reply works initially when opening a message that is smart reply compatible.

reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGpGwpvXXpDtdbGscnkKkpxNfTh

martgil:
This PR will make smart replies available for non-encrypted messages

screenshots:
(before)
Screenshot from 2022-07-20 14-38-35
(after)
Screenshot from 2022-07-20 14-38-15

close #4549 // if this PR closes an issue

- Tests will be added later

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -196,7 +196,9 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     // reply buttons
     const visibleReplyBtns = $('td.acX:visible');
     if (visibleReplyBtns.not('.replaced, .inserted').length) { // last reply button in convo gets replaced
-      $(this.sel.autoReplies).remove();
+      if (this.isEncrypted()) {
+        $(this.sel.autoReplies).remove(); // make smart replies available for non-encrypted conversations
+      }
       const convoReplyBtnsToReplace = visibleReplyBtns.not('.replaced, .inserted');
       const convoReplyBtnsArr = convoReplyBtnsToReplace.get();
       // only replace the last one FlowCrypt reply button if does not have any buttons replaced yet, and only replace the last one
--- a/test/source/mock/google/live-gmail-test-emails/Plain message triggering a smart reply (FMfcgzGpHHKCrKRLptBSNwkpMxzkdcQc).eml
+++ b/test/source/mock/google/live-gmail-test-emails/Plain message triggering a smart reply (FMfcgzGpHHKCrKRLptBSNwkpMxzkdcQc).eml
@@ -1,0 +1,77 @@
+Delivered-To: ci.tests.gmail@flowcrypt.dev
+Received: by 2002:a05:6408:330f:b0:19d:62d6:6df8 with SMTP id ck15csp390243egb;
+        Fri, 22 Jul 2022 06:12:25 -0700 (PDT)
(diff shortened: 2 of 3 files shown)


martgil:
@rrrooommmaaa ready for review. Thank you!

rrrooommmaaa:
I wouldn't say that this change is `refactor only, docs or internal changes`.
Maybe, it's not very easy to test, but we can wire a `gmail` test to check the presence of smart replies. It would also be helpful to detect if gmail changes the classes for them.
I tried to send a message from `ci.tests.gmail@flowcrypt.dev` to `ci.tests.gmail@flowcrypt.dev` but they didn't appear.
image
Maybe it's because I sent a message to myself. Can you try to send it from `human` or `compatibility` email, please? I wouldn't recommend to use a personal email as we usually keep a copy of the message.

martgil:
Thank you for the reminder. We can use this test email from flowcrypt.compatibility account:
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGpGwrFHNCnNScHJzWqKwTBZJWV

> I tried to send a message from ci.tests.gmail@flowcrypt.dev to ci.tests.gmail@flowcrypt.dev but they didn't appear.

I had some trouble making a test email at first. But once it appears one time, the smart replies option would be available for that email.

> Maybe, it's not very easy to test, but we can wire a gmail test to check the presence of smart replies. 

I haven't done this before. Does that mean the test should be in the live Gmail test?

> It would also be helpful to detect if gmail changes the classes for them.

Okay. If I understand it correctly, if the class is not visible/rendered, then, there's no point in removing them?

rrrooommmaaa:
> Thank you for the reminder. We can use this test email from flowcrypt.compatibility account: https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGpGwrFHNCnNScHJzWqKwTBZJWV

We have live gmail tests configured for ci.tests.gmail@flowcrypt.dev only.
I sent this message from `flowcrypt.compatibility` to `ci.tests.gmail` and now we're ok.
image

> I had some trouble making a test email at first. But once it appears one time, the smart replies option would be available for that email.

It's important that the sender differs from the recipient.
I'll wire a live gmail test for it.

> Okay. If I understand it correctly, if the class is not visible/rendered, then, there's no point in removing them?

Removing non-existing elements won't crash the app. I'm more worried about the situation when Gmail changes the class for smart replies (currently `brb`, correct?) that are expected by the test. In such a case, the test will start to fail and we can update the code quickly for a different class etc.

rrrooommmaaa (approved):

PR #4562, opened 2022-07-29
#4541 Fix: any in ci_unit_test

### LINKED ISSUE #4541: fix forge definitions usage to not use any
Also this is wrong to add forge into squire definitions. Let's remove it and file an issue for Ioan to have a look later.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4508#discussion_r920921712_

ioanmo226:
This PR fixed any in ci_unit_test

close #4541 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/chrome/dev/ci_unit_test.ts
+++ b/extension/chrome/dev/ci_unit_test.ts
@@ -38,9 +38,10 @@ const libs: unknown[] = [
   Catch,
   Gmail,
 ];
-
+/* eslint-disable @typescript-eslint/no-explicit-any */
 // add them to global scope so ci can use them
 for (const lib of libs) {
   (window as any)[(lib as any).name] = lib;
 }
 (window as any).forge = forge;
+/* eslint-enable */


ioanmo226:
FYI, I also think it's safe to use `any` in window variable declaration(like Tom mentioned) and we should use `as any` for this line as we can't know which lib we are going to define.

`(window as any)[(lib as any).name] = lib`

https://github.com/FlowCrypt/flowcrypt-browser/issues/4541#issuecomment-1184192647

extension/chrome/dev/ci_unit_test.ts
```diff
@@ -38,9 +38,10 @@ const libs: unknown[] = [
   Catch,
   Gmail,
 ];
-
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
rrrooommmaaa, line 41: Do you think it may be worth trying to go this direction (adding properties to `Window` interface)?
https://stackoverflow.com/a/12709880

ioanmo226: yeah, it's worth it.
But for this line, we can't implement like that.

> (window as any)[(lib as any).name] = lib

ioanmo226: I think (window as any) is enough for this logic

rrrooommmaaa (approved):

PR #4585, opened 2022-08-08
#4532 fix: cache found contacts locally as soon as they are fetched issue

### LINKED ISSUE #4532: cache found contacts locally as soon as they are fetched
ok. On browser extension, we actually do keep a cache and it should be a lot faster the second time. Is it not? Maybe we don't currently cache all, but only cache items that the user actually used. If that's the case, we could start caching all on browser extension too.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-ios/issues/1741#issuecomment-1177831585_

ioanmo226:
This PR fixed cache found contacts locally as soon as they are fetched issue

close #4532 // if this PR closes an issue

- Test added

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -702,7 +702,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       const offsetTop = input.outerHeight()! + offset.top; // both are in the template
       const bottomGap = 10;
       this.view.S.cached('contacts').css({
-        display: 'block',
+        display: renderableContacts.length > 0 ? 'block' : 'none', // show popup only when there are renderableContacts 
         left: leftOffset,
         top: offsetTop,
         maxHeight: `calc(100% - ${offsetTop + bottomGap}px)`,
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -567,6 +567,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       this.view.errModule.debug(`searchContacts 3`);
       const foundOnGoogle = await this.searchContactsOnGoogle(substring, contacts);
       await this.addApiLoadedContactsToDb(foundOnGoogle);
+      this.view.errModule.debug(`searchContacts foundOnGoogle, count: ${foundOnGoogle.length}`);
       contacts.push(...foundOnGoogle.map(c => ContactStore.previewObj({ email: c.email, name: c.name })));
       this.renderSearchRes(input, contacts, { substring });
       if (contacts.length >= this.MAX_CONTACTS_LENGTH) {
@@ -578,6 +579,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
         this.view.errModule.debug(`searchContacts (Gmail Sent Messages) 6.b`);
         await this.guessContactsFromSentEmails(substring, contacts, async guessed => {
           await this.addApiLoadedContactsToDb(guessed.new);
+          this.view.errModule.debug(`searchContacts (Gmail Sent Messages), count: ${guessed.new.length}`);
           contacts.push(...guessed.new.map(c => ContactStore.previewObj({ email: c.email, name: c.name })));
           this.renderSearchRes(input, contacts, { substring });
         });
@@ -706,7 +708,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       this.setContactPopupStyle(input);
     } else {
       this.setContactPopupStyle(input);
-      contactEl.find('ul').html('<li>No Contacts Found</li>'); // xss-direct
+      contactEl.find('ul').html('<li data-test="no-contact-found">No Contacts Found</li>'); // xss-direct
       if (!this.googleContactsSearchEnabled) {
         this.addBtnToAllowSearchContactsFromGoogle(input);
       }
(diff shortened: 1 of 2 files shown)


ioanmo226:
Please review this PR and [issue comments](https://github.com/FlowCrypt/flowcrypt-browser/issues/4532).

ioanmo226:
Also I find that in Gmail app, debounce time is less than 50ms. It shows results as soon as i finish typing.
But in our app, debounce time is set as 500ms. Maybe we can decrease it to 100 or 200ms?

ioanmo226:
@rrrooommmaaa Please check above comment too

rrrooommmaaa:
> But in our app, debounce time is set as 500ms. Maybe we can decrease it to 100 or 200ms?

Maybe we can. I'm not an expert in this.

ioanmo226:
I see. Maybe @tomholub can confirm.
And he confirms, we may do it in another PR.

rrrooommmaaa:
But we need to make sure the new search hasn't started until the previous one has completed. Or that we handle this situation gracefully.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -647,7 +644,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       return 0;
     });
     const renderableContacts = sortedContacts.slice(0, this.MAX_CONTACTS_LENGTH);
-    if ((renderableContacts.length > 0 || this.contactSearchInProgress) || !this.googleContactsSearchEnabled) {
+    if (renderableContacts.length > 0 || !this.googleContactsSearchEnabled) {
```
rrrooommmaaa, line 647: I meant to remove the `if`, make this code run each time. We do need to show `No Contacts Found` at the end of the search, don't we?

rrrooommmaaa: Oh we don't. Ok, then.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -702,11 +693,16 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
-        display: 'none',
+        display: renderableContacts.length > 0 ? 'block' : 'none', // show popup only when there are renderableContacts
         left: leftOffset,
         top: offsetTop,
         maxHeight: `calc(100% - ${offsetTop + bottomGap}px)`,
       });
+    } else {
+      this.view.S.cached('contacts').find('ul').html('<li>No Contacts Found</li>'); // xss-direct
+      if (!this.googleContactsSearchEnabled) {
+        this.addBtnToAllowSearchContactsFromGoogle(input);
```
ioanmo226, line 704: I changed this logic because with previous logic, when there's search result for something like `e2e` and then user continues type like `e2eaaaaaaaa`, it still displays `e2e` result.

ioanmo226:
@rrrooommmaaa Please review this PR

extension/chrome/elements/compose-modules/compose-recipients-module.ts
rrrooommmaaa: I have tested it and it's looking ok. So, if `googleContactsSearchEnabled` is `true` and I'm typing a non-existing name `aaaaaa`, nothing is displayed (`No Contacts Found` isn't displayed).
However if I first type an existing name like `flow`, wait until the contacts appear and then continue typing to `flowwww`, then `No Contacts Found` is displayed replacing the found contacts. Is this the desired behaviour?

ioanmo226: I don't think so. I just fixed and pushed. Please check

rrrooommmaaa: So how should it behave? Can you add a unit test, please? Something like `compose - load contacts through API` but simpler. If we run such a test only for `testVariant === 'ENTERPRISE-MOCK'`, `googleContactsSearchEnabled` will be `true`. You can always find the profile's own email address in contacts, that is `ci.tests.gmail@flowcrypt.test` for `ci.tests.gmail` profile

ioanmo226: Current behavior is to show `no contact founds` for all cases when there are no contacts found

ioanmo226: This is current implemented behavior.
@tomholub Not sure who can confirm this. Can you confirm if this is what you are looking for? I'll add ui test (as Roman suggested) after your confirmation

chrome-capture-2022-7-9

tomholub: behavior-wise, the screen recording looks ok to me. I haven't read the rest of the conversation

ioanmo226: @rrrooommmaaa Ready for review

rrrooommmaaa: The element added by `addBtnToAllowSearchContactsFromGoogle` remains as a leftover after the "No Contacts Found" case, is this intentional?
The look when a contact is found right away:
image
The look when a contact is found after the "No Contacts Found" case:
image

ioanmo226: Thanks for your notice
> The element added by addBtnToAllowSearchContactsFromGoogle remains as a leftover after the "No Contacts Found" case

I don't think so. Just fixed
> is this intentional?

rrrooommmaaa (approved):

PR #4589, opened 2022-08-09
#4543 Live reload extension

### LINKED ISSUE #4543: Consider implementing live-reload for development
Currently when I want to see changes (while in development), I have to build project and reload extension frame.
This doesn't seem to be a problem when you want to see changes for so many files.
But it makes me annoying when I want to see few changes (something like changing ui, change font color, margin, etc or just change one line of typescript code)

Here are references.

https://stackoverflow.com/questions/2963260/how-do-i-auto-reload-a-chrome-extension-im-developing
https://github.com/vitejs/vite

Ping: @tomholub

ioanmo226:
This PR implemented live reload extension feature

close #4543 // if this PR closes an issue

### To be filled by reviewers


tomholub:
Can it already be used? Or it's just a concept and there more work required that is not so easy?

ioanmo226:
It's still POC and need more work (but not that difficult though)

ioanmo226:
Had/have to look into other PRs because this issue is defined as 2nd priority

tomholub:
Got it. I'll move it. I think this would be handy for your development

package.json
```diff
@@ -52,10 +53,14 @@
...
     "squire-rte": "1.11.3",
     "sweetalert2": "11.4.33",
+    "tsc-watch": "^5.0.3",
     "zxcvbn": "4.4.2"
   },
   "scripts": {
     "build": "./build.sh",
+    "sync": "./sync.sh",
+    "watch": "tsc-watch -p ./tsconfig.json --onSuccess 'npm run sync'",
+    "start": "concurrently \"npm run watch\" \"npm run run_chromium\"",
```
ioanmo226, line 63: @tomholub I think we'd better stick with current implementation.
tsc-watch only watches ts files and doesn't watch changes for .html, manifest.json, etc.
More over when we run `watch` with `run_chromium`, chromium active page(s) gets removed when user changes something as we remove whole build directory and re-build it again.

With current build mechanism, it seems hard/impossible to implement hot-reload.

ioanmo226: Let me know what you think and what you have in mind.

ioanmo226: ok

Issue #4552, opened 2022-07-20
Browser extension is experiencing problems with Brave Browser v1.41.96


martgil:
Looks like the same issue was re-introduced again on Brave Browser v1.41.96 (latest version)  - https://github.com/FlowCrypt/flowcrypt-browser/issues/4472

The message was all readable if they are viewed through the web extension pages. The error message says "Could not load message due to network error. retry"
![image](https://user-images.githubusercontent.com/46025304/179928793-4f7ad1ce-41bd-487f-a196-e6a7c8b6680b.png)

Steps to reproduce
- pre-requisite:
Flowcrypt browser extension installed on brave browser (latest version)

1. open an encrypted message and the error "Could not load message due to network error. retry" will appear.
2. (optional) messages are readable through web extension pages (click on the FlowCrypt browser extension icon  -> Encrypted Inbox -> open an encrypted email -> encrypted message is readable)

<details><summary>stack traces:</summary>

```
[callerStack]
    at chrome-extension://mimchhddcelobciniokjnakmkcbdggof/js/common/browser/browser-msg.js:289:124
    at new Promise (<anonymous>)
    at BrowserMsg.sendAwait (chrome-extension://mimchhddcelobciniokjnakmkcbdggof/js/common/browser/browser-msg.js:287:22)
    at Object.pgpMsgDecrypt (chrome-extension://mimchhddcelobciniokjnakmkcbdggof/js/common/browser/browser-msg.js:41:47)
    at decrypt (chrome-extension://mimchhddcelobciniokjnakmkcbdggof/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:76:67)
    at PgpBlockViewDecryptModule.decryptAndRender (chrome-extension://mimchhddcelobciniokjnakmkcbdggof/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:82:42)
    at async PgpBlockViewDecryptModule.initialize (chrome-extension://mimchhddcelobciniokjnakmkcbdggof/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:35:25)
    at async PgpBlockView.render (chrome-extension://mimchhddcelobciniokjnakmkcbdggof/chrome/elements/pgp_block.js:46:21)

[...]

PR #4597, opened 2022-08-11
#992 add pdf print feature

### LINKED ISSUE #992: button to export email content as pdf (print/for printing)
ref https://mail.google.com/mail/u/0/#inbox/1616a7dd632ac39d

ioanmo226:
This PR added pdf print feature

close #992 // if this PR closes an issue

- Tests added

### To be filled by reviewers


ioanmo226:
@tomholub Please check
image
image

ioanmo226:
@tomholub You mean like this?
image

tomholub:
yes, that looks good

rrrooommmaaa:
> Difficult to test (Puppeteer won't give us access to print dialog)

Are we willing to test exactly, the print dialog or the saved PDF file?
It'd be great to see at least the test that checks for existence of this element `@action-print`, and then clicking it.
The print dialog, as I understand, can be skipped by the `--kiosk-printing` command line option (in `BrowserPool.newBrowserHandle()`), have you tried that?

ioanmo226:
@rrrooommmaaa You are right. We need to indeed test print dialog and can be done with `--kiosk-printing` flag.
Just implemented ui test.
Please check

extension/chrome/elements/pgp_block.ts
```diff
@@ -106,7 +113,117 @@ export class PgpBlockView extends View {
   };
 
   public setHandlers = () => {
-    // defined as needed, depending on what rendered
+    $('.pgp_print_button').click(this.setHandler(() => this.printPGPBlock()));
+  };
+
+  private initPrintView = async () => {
```
tomholub, line 119: These two new methods could go into `PgpBlockViewRenderModule` since they mostly deal with rendering things. If the render module has `setHandlers` there as well, the handler could probably go there to stay close to the moved code.

extension/chrome/elements/pgp_block.ts
```diff
@@ -106,7 +113,117 @@ export class PgpBlockView extends View {
...
+          }
+          #attachments a.preview-attachment .download-attachment {
+            display: none;
+          }
+        </style>
+      </head>
+      <body>
+        ${$('#print-header').html()}
+        <br>
+        ${this.printMailInfoHtml}
```
tomholub, line 213: Should sanitize this value here. I think it could potentially contain some security-nasty content, causing HTML injection, maybe.

extension/chrome/elements/pgp_block.ts
```diff
@@ -106,7 +113,117 @@ export class PgpBlockView extends View {
...
+        <br>
+        <div data-test="print-content">
+          ${$("#pgp_block").html()}
```
tomholub, line 216: Maybe it wouldn't hurt to re-sanitize this one as well, as long as it doesn't break anything.

tomholub (commented):
Thanks! See comments. It's looking good.

test/source/tests/decrypt.ts
```diff
@@ -514,6 +515,26 @@ XZ8r4OC6sguP/yozWlkG+7dDxsgKQVBENeG6Lw==
...
 
+    ava.default('decrypt - print feature in pgp block', testWithBrowser('compatibility', async (t, browser) => {
+      const threadId = '182917712be838e1';
+      const acctEmail = 'flowcrypt.compatibility@gmail.com';
+      const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId}`));
+      await inboxPage.waitAll('iframe');
+      const pgpBlock = await inboxPage.getFrame(['pgp_block.htm']);
+      await pgpBlock.waitForSelTestState('ready');
+      expect(await pgpBlock.isElementPresent('@action-print')).to.be.true;
+      await pgpBlock.click('@action-print');
```
rrrooommmaaa, line 526: It's be nice to use
```
const printPage = await browser.newPageTriggeredBy(t, () => pgpBlock.click('@action-print'));
```
instead of these 4 lines, but we'd need `<body data-test-view-state="loaded">` in `printPGPBlock`

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -14,15 +14,128 @@ import { Ui } from '../../../js/common/browser/ui.js';
...
+  private printMailInfoHtml!: string;
 
   constructor(private view: PgpBlockView) {
   }
 
+  public initPrintView = async () => {
+    const fullName = await AcctStore.get(this.view.acctEmail, ['full_name']);
+    Xss.sanitizeRender('.print_user_email', `<b>${fullName.full_name}</b> &lt;${this.view.acctEmail}&gt;`);
+    try {
+      const gmailMsg = await this.view.gmail.msgGet(this.view.msgId!, 'full', undefined);
```
rrrooommmaaa, line 33: I have several concerns about this call. Perhaps, some of them can be addressed in a separate issue.
1) Do we want to ignore the `canReadEmails` "setting" (calculated in `PgpBlockViewDecryptModule`)? @tomholub Can you give us some background why this setting had appeared? Was it to prevent fetching messages when an Outlook mail site is opened or for some limited permission installations?
2) If we only need headers, wouldn't it be enough to fetch `metadata` instead of `full` message?
3) It'd be better to do the fetching in `PgpBlockViewDecryptModule`, as it sometimes fetches the message by `msgId`, sometimes just takes it from a URL object (`encryptedMsgUrlParam`), and it's probably going to be more complicated as with https://github.com/FlowCrypt/flowcrypt-browser/issues/4552#issuecomment-1207852190 we may need to try and fetch the message if reading the URL object fails. It seems the `PgpBlockViewDecryptModule` is more competent about what message data is already fetched and what is missing. And `PgpBlockViewRenderModule` would be limited to rendering only.

tomholub: `canReadEmails` can be ignored, see https://github.com/FlowCrypt/flowcrypt-browser/issues/4600

tomholub: indeed `metadata` should work. (btw `full` still skips attachments so it's not the worst)

tomholub: Agree with third point. Sometimes PgpBlockViewDecryptModule already downloads the message so no need to do it twice. We could make it always download the message, for example (but in a way that offline viewing of encrypted messages still works)

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -14,15 +14,128 @@ import { Ui } from '../../../js/common/browser/ui.js';
...
+      const sentDate = new Date(GmailParser.findHeader(gmailMsg, 'date') ?? '');
+      const sentDateStr = Str.fromDate(sentDate).replace(' ', ' at ');
+      const from = Str.parseEmail(GmailParser.findHeader(gmailMsg, 'from') ?? '');
+      const fromHtml = from.name ? `<b>${from.name}</b> &lt;${from.email}&gt;` : from.email;
+      const ccString = GmailParser.findHeader(gmailMsg, 'cc') ? `Cc: <span data-test="print-cc">${Xss.escape(GmailParser.findHeader(gmailMsg, 'cc')!)}</span><br/>` : '';
+      const bccString = GmailParser.findHeader(gmailMsg, 'bcc') ? `Bcc: <span>${Xss.escape(GmailParser.findHeader(gmailMsg, 'bcc')!)}</span><br/>` : '';
+      this.printMailInfoHtml = `
+      <hr>
+      <p class="subject-label" data-test="print-subject">${GmailParser.findHeader(gmailMsg, 'subject')}</p>
```
martgil, line 42: I think we should sanitize the `subject` and the `from.name` as Google previously allowed special chars within the name field such as this one: https://www.youtube.com/user/securityteamvictim

Someone could change the UI by injecting HTML within the subject (limited tags + attributes):
image

ioanmo226: Thank you for your notice. Just sanitized the above elements

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
ioanmo226: @rrrooommmaaa Thank you for your notice.
Just changed `full` to `meta`
Will fix getting message twice issue(which we do in both render and decrypt module) in #4600

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
martgil: Thank you, too. sir

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -49,9 +162,11 @@ export class PgpBlockViewRenderModule {
       await AcctStore.set(this.view.acctEmail, { successfully_received_at_leat_one_message: true });
     }
     if (!isErr) { // rendering message content
+      $('.pgp_print_button').show();
       const pgpBlock = $('#pgp_block').html(Xss.htmlSanitizeKeepBasicTags(htmlContent, 'IMG-TO-LINK')); // xss-sanitized
       pgpBlock.find('a.image_src_link').one('click', this.view.setHandler((el, ev) => this.displayImageSrcLinkAsImg(el as HTMLAnchorElement, ev as JQuery.Event<HTMLAnchorElement, null>)));
     } else { // rendering our own ui
+      $('.pgp_print_button').hide();
```
rrrooommmaaa, line 169: Do we need this statement? Is there really a scenario when we first render the message content and then overwrite it with an error?

ioanmo226: Yes. Like when passphrase is expired(or not exist) or when user taps `end passphrase` section

rrrooommmaaa: Meaning, when passphrase isn't known, we have `Print` button? I don't quite understand.

ioanmo226: Sorry I misunderstood what you mean.
Yeah. We indeed don't need that hide statement as initial print button visibility is hidden.

rrrooommmaaa (approved):

PR #4614, opened 2022-08-18
Client configuration is updated on opening the settings or loading the content script

### LINKED ISSUE #4510: update client configuration each time settings page is refreshed or gmail content script loaded
.

rrrooommmaaa:
This PR makes sure that client configuration is updated when settings page or content script is loaded.

close #4510

- Tests added or updated

### To be filled by reviewers


extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -285,6 +286,11 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
       const acctEmail = await waitForAcctEmail();
       const { tabId, notifications, factory, inject } = await initInternalVars(acctEmail);
       await showNotificationsAndWaitTilAcctSetUp(acctEmail, notifications);
+      const authInfo = await AcctStore.authInfo(acctEmail);
+      if (authInfo.uuid) { // have auth email set
+        // todo: catch? ignore?
+        await (new AccountServer(acctEmail)).accountGetAndUpdateLocalStore(authInfo);
```
rrrooommmaaa, line 292: @tomholub Do we need some special error handling when
1) uuid is not set
2) accountGetAndUpdateLocalStore throws an exception?
Should we toast a message and continue?

test/source/mock/backend/backend-data.ts
```diff
@@ -7,9 +7,35 @@ import { OauthMock } from '../lib/oauth';
...
+    "PASS_PHRASE_QUIET_AUTOGEN",
+    "DEFAULT_REMEMBER_PASS_PHRASE",
+  ],
+  "key_manager_url": "https://localhost:8001/flowcrypt-email-key-manager",
+  "enforce_keygen_algo": "rsa2048",
+  "disallow_attester_search_for_domains": []
+};
+
+export type ClientConfiguration = {
+  // todo: should we somehow import the type from `client-configuration.ts`?
```
rrrooommmaaa, line 24: `client-configuration.ts` is not under `core`, so it's not reachable from tests...
Should we do something about it?

tomholub: i'm not sure. will leave up to you

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
rrrooommmaaa: ping @tomholub

test/source/tests/settings.ts
```diff
@@ -967,6 +969,88 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+      setup.flags = [
+        'NO_PRV_BACKUP',
+        'ENFORCE_ATTESTER_SUBMIT',
+        'PRV_AUTOIMPORT_OR_AUTOGEN',
+        'PASS_PHRASE_QUIET_AUTOGEN',
+        'DEFAULT_REMEMBER_PASS_PHRASE'];
+      setup.disallow_attester_search_for_domains = ['disallowed_domain5.test', 'disallowed_domain6.test'];
+      setup.enforce_keygen_algo = 'rsa2048';
+      const extraAuthHeaders = { Authorization: `Bearer ${accessToken}` };
+      const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
```
rrrooommmaaa, line 1034: @tomholub These tests confirm that the updated `rules` get their way into the storage on every load of `settings` page or web content script. Is this sufficient? Should we worry about other possibly open tabs with older versions of `ClientConfiguration`?
Also, should we consider some scenarios when user interaction or warning is necessary?
For example, if `FORBID_STORING_PASS_PHRASE` rule has appeared and the passphrase is currently stored, should we clear it? What if it was previously generated in the silent mode?
Also, what if `enforce_keygen_algo` has changed? Should we make the user re-generate the keys?
Perhaps, you know much more about such possible scenarios.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
tomholub: There are two types of endpoints to get client configuration from:
 - flowcrypt.com/api - authenticated, which uses uuid
 - fes.customer.com - unauthenticated

If you look inside of `AccountServer.accountGetAndUpdateLocalStore`, you'll notice that it's only using uuid for one of the cases.

I would remove the `if` in your code and instead try/catch:
  - catching `BackendAuthErr` and ignoring it, with a comment that user will see a prompt to log in during some other actions that involve backend, and therefore the value will eventually be filled, at which point it will work successfully (no need to wait and retry, it will just get redone next time user loads the page).
  - catching ApiErr.isNetErr and ignoring it
  - remaining errors should be toasted: `Failed to update FlowCrypt Client Configuration: e.message (e.type)`

test/source/tests/settings.ts
tomholub: > @tomholub These tests confirm that the updated rules get their way into the storage on every load of settings page or web content script. Is this sufficient? Should we worry about other possibly open tabs with older versions of ClientConfiguration?

That's enough. Not worried about already open tabs elsewhere

> Also, should we consider some scenarios when user interaction or warning is necessary?
> For example, if FORBID_STORING_PASS_PHRASE rule has appeared and the passphrase is currently stored, should we clear it? What if it was previously generated in the silent mode?
> Also, what if enforce_keygen_algo has changed? Should we make the user re-generate the keys?
> Perhaps, you know much more about such possible scenarios.

We'll ignore such cases for now, but we'll file an issue. Once a customer actually needs to change a particular client configuration, we'll implement a mechanism for whatever they need to change. The most likely one to get implemented will probably be a chance of pass phrase storage, either from in-memory to stored or stored to in-memory.

Nobody is currently using auto-generated pass phrases in silent mode.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
rrrooommmaaa: not sure what you mean by `e.type` -- we don't have this in this project.
Other changes have been applied.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -280,11 +282,30 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
...
+    }
+
+  };
+
   const entrypoint = async () => {
     try {
       const acctEmail = await waitForAcctEmail();
       const { tabId, notifications, factory, inject } = await initInternalVars(acctEmail);
       await showNotificationsAndWaitTilAcctSetUp(acctEmail, notifications);
+      await updateClientConfiguration(acctEmail);
```
rrrooommmaaa, line 308: @tomholub I'm afraid, in some cases (on a network error) it may considerably delay the gmail replacer so that it'll be annoying to the user?

tomholub: Yes, I'd like it to happen async, like EKM refetch.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
tomholub: Oh, I meant the error instance. In case it says something meaningful like `NetworkError` or whatnot.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
tomholub: Sure. I'm ok with that. I don't necessary need right-now immediate change, what I need is that changes do eventually get propagated.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
tomholub: Looks good to me 👍

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
rrrooommmaaa: Done

tomholub (approved):
:+1:

PR #4615, opened 2022-08-18
#4610 allow multiple pub keys in attester

### LINKED ISSUE #4610: `Attester` class should understand more than one returned pubkeys
1) remove unused code + refactor

unrelated to this issue, but I just noticed

```ts
      // when requested from the content script, `getResponseHeader` will be missing because it's not a real XMLHttpRequest we are getting back
      // because it had to go through background scripts, and objects are serialized when this happens
      // the proper fix would be to send back headers from bg along with response text, and parse it here
      if (!r.getResponseHeader) {
        return { pubkey: r.responseText };
      }
```

This can be removed, no longer fulfills any function. Also `Attester.doLookup` attribute should be `email` instead of `emailOrFingerprint`

2) `lookupEmail` (and `doLookup`) should return a list of public keys

the response from API may contain more than one armored public key block. They are concatenated by `\n`. 

Since more than one armored block can be returned, you need to parse them apart with `MsgBlockParser.detectBlocks` if you get a http 200, then filter only public key blocks from the parsed blocks, and then return the resulting contents as an array of strings in `PubkeysSearchResult`

For a test, add another email address to attester-endpoints where it returns `[pubkey1, pubkey2].join('\n')`, then look that up with browser extension compose, and check that both pubkeys were imported.

ioanmo226:
This PR allows multiple pub keys in attester

close #4610 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/chrome/settings/modules/keyserver.ts
```diff
@@ -128,18 +128,9 @@ View.run(class KeyserverView extends View {
...
-      } else {
-        const pub = await KeyUtil.parse(pubkeySearchResult.pubkey);
-        let match = true;
-        if (!storedKeysIds.includes(pub.id)) {
-          diagnosis.hasPubkeyMismatch = true;
-          match = false;
-        }
-        diagnosis.results[email] = { pubkey: pubkeySearchResult.pubkey, match };
-      }
+      const hasMatchingKey = pubkeySearchResult.pubkeys.some(async (pubkey) => storedKeysIds.includes((await KeyUtil.parse(pubkey)).id));
```
rrrooommmaaa, line 131: I'm afraid you can't put an async function into `some` -- it will simply return `Promise<boolean>` and always be considered as truthy value -- `some` won't wait for fulfillment. So you should do it either in a loop (parsing a key is a fast operation anyway), or do this fancy `race` thing https://stackoverflow.com/a/56553349 -- I think we can add it to our `extension/js/common/core/common.ts` utils.
So this case is not covered by the unit test? Can you add a test for it? As I understand, this page is invoked with `@action-open-attester-page` click?

ioanmo226: Nice catch. Thanks
You are right. `some` doesn't support `async`. Let me implement it.

extension/js/common/core/common.ts
```diff
@@ -353,3 +353,10 @@ export class Url {
 export const emailKeyIndex = (scope: string, key: string): string => {
   return `${scope.replace(/[^A-Za-z0-9]+/g, '').toLowerCase()}_${key}`;
 };
+
+export const asyncSome = async (arr: Array<unknown>, predicate: (e: unknown) => Promise<boolean>) => {
```
rrrooommmaaa, line 357: good. Can we use (Generics)[https://www.typescriptlang.org/docs/handbook/2/generics.html] instead of `unknown`

ioanmo226: Sounds good. Just changed it

ioanmo226:
@rrrooommmaaa Ready for review. Thanks

test/source/tests/compose.ts
```diff
@@ -1485,6 +1485,29 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       await composePage.waitAny('@password-or-pubkey-container', { visible: false });
     }));
 
+    ava.default('attester should understand more than one pub key', testWithBrowser('ci.tests.gmail', async (t, browser) => {
```
rrrooommmaaa, line 1488: The title should read something like `attester client should...` -- we're testing reading multiple keys from the attester.

test/source/tests/settings.ts
```diff
@@ -56,6 +56,26 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
       await SettingsPageRecipe.toggleScreen(settingsPage, 'basic');
     }));
 
+    ava.default('settings - attester shows mismatch information correctly', testWithBrowser(undefined, async (t, browser) => {
```
rrrooommmaaa, line 59: `Attester diagnostics page shows....`

extension/chrome/settings/modules/my_key.ts
```diff
@@ -76,7 +76,8 @@ View.run(class MyKeyView extends View {
     try {
       const result = await this.pubLookup.attester.lookupEmail(this.acctEmail);
       const url = FlowCryptWebsite.url('pubkey', this.acctEmail);
-      if (result.pubkey && (await KeyUtil.parse(result.pubkey)).id === this.keyInfo.fingerprints[0]) {
+      const doesContainKey = await asyncSome(result.pubkeys, async (pubkey) => (await KeyUtil.parse(pubkey)).id === this.keyInfo.fingerprints[0]);
```
rrrooommmaaa, line 79: We should compare keys using the `identityEquals` helper method, as ids (or fingerprints) are not guaranteed to be unique across key families. Here this line would fit good:
```
const doesContainKey = await asyncSome(result.pubkeys, async (pubkey) => KeyUtil.identityEquals(await KeyUtil.parse(pubkey), this.keyInfo));
```

rrrooommmaaa (approved):

Issue #3346, opened 2021-01-26
custom SKS server client configuration test


tomholub:
I'll make an issue for this

> The bug at #3066 was caused by the Sks class not covered by existing tests.
> 
> Currently Sks class is tested independently in unit tests but is not tested hollistically with the rest of the code based on the OrgRule that activates it.

- [ ] add a test on a new domain like `custom-sks.flowcrypt.test` with client configuration `custom_keyserver_url: (some mock url pretending to be SKS)`
- [ ] add one key to the SKS mock `mockSksEndpoints` that is not present elsewhere in the mock
- [ ] test that the pubkey will be fetched from there with this rule enabled
- [ ] test that the SKS integration is mentioned in FlowCrypt Settings -> Additional Settings -> Contacts (when this client configuration is used, there is supposed to be a notice there)

Ask @rrrooommmaaa when unsure

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/3070#issuecomment-767362114_

PR #4623, opened 2022-08-22
issue 4622 add WKD as source of public keys when verifying messages

### LINKED ISSUE #4622: include WKD in pubkey search when verifying messages
currently skipped, often causing:

image

tomholub:
This PR removes a limitation that caused WKD to be skipped when searching for public keys to verify a message with.

The limitation caused some messages to show as not verified if the user didn't load the public key previously in secure compose.

close #4622

**Tests*:
- Test added

### To be filled by reviewers


test/source/tests/browser-unit-tests/unit-Sks.js
```diff
@@ -69,20 +69,3 @@ BROWSER_UNIT_TEST_NAME(`Sks lookup pubkey - not found`);
...
-  const key = await KeyUtil.parse(pubkey);
-  const expectedKeyIds = [
-    '6BF16EE1ECE7A66C4B6636DF0C9C2E6A4D273C6F',
-    '8B8A05A2216EE6E4C5EE3D540D5688EBF3102BE7'
-  ];
-  if (!expectedKeyIds.includes(key.id)) {
-    throw Error(`Expecting key.id in [${expectedKeyIds}] but got ${key.id}`);
-  }
-  return 'pass';
-})();
```
tomholub, line 88: https://github.com/FlowCrypt/flowcrypt-browser/issues/3346

tomholub:
@rrrooommmaaa this is now ready. There was some wrestling with tests and also lack of openpgp.js in decrypt frame.

extension/js/common/api/key-server/wkd.ts
```diff
@@ -27,59 +29,60 @@ export class Wkd extends Api {
...
-    } catch (e) {
-      return { pubkeys: [] };
+    const all = await this.rawLookupEmail(email);
+    console.log(all);
+    const filtered = all.filter(key => key.emails.some(e => this.pubkeyUidFilter(e, email)));
+    return { pubkeys: filtered.map(pubkey => pubkey.armored) };
+  };
+
+  private pubkeyUidFilter = (uidEmail: string, expectedEmail: string) => {
+    if (!IS_MOCK_TEST_ENVIRONMENT) { // lgtm [js/trivial-conditional]
```
rrrooommmaaa, line 79: I'd suggest to use normal email addresses in Wkd queries, but use this `IS_MOCK_TEST_ENVIRONMENT` flag in `rawLookupEmail` method instead to redirect queries to `localhost:8001` for certain emails (or for all emails except for the only one when we're testing the live Wkd?) -- line `const directHost = ...` .
Please also note this property `public port: number | undefined;`.
I created it earlier to hack into Wkd's behaviour from `test/source/tests/browser-unit-tests/unit-Wkd.js` tests. Now (if we change this line `const directHost = ...` accordingly) we'll be able to rely on `IS_MOCK_TEST_ENVIRONMENT` flag for these tests too and remove the `port` property.

extension/js/common/core/crypto/key.ts
```diff
@@ -100,6 +100,8 @@ export type PrvPacket = (OpenPGP.packet.SecretKey | OpenPGP.packet.SecretSubkey)
 
 export class UnexpectedKeyTypeError extends Error { }
 
+export type ArmoredKeyWithEmailsAndId = { id: string, emails: string[], armored: string };
```
rrrooommmaaa, line 103: We can extend this type (interface) from `KeyIdentity` and set `family` too, so we can later use `KeyUtil.identityEquals` in `ComposeMyPubkeyModule.reevaluateShouldAttachOrNot` instead of comparing fingerprints.

extension/js/common/api/key-server/wkd.ts
tomholub: I'll take this suggestion but I'll do it a bit differently. Instead of `IS_MOCK_TEST_ENVIRONMENT` I'll use a property like `FORCE_WKD_HOST = ''` (prod) or `FORCE_WKD_HOST = 'http://localhost:8001'` (mock). It will be a bit more flexible, to be able to have one mock API per test (which I'd like to, eventually)

tomholub (commented):
thank you for the suggestions

extension/js/common/api/key-server/wkd.ts
```diff
@@ -27,60 +27,47 @@ export class Wkd extends Api {
...
-    const directDomain = lowerCaseRecipientDomain;
     const timeout = (this.usesKeyManager && lowerCaseRecipientDomain === this.domainName) ? 10 : 4;
-    const advancedDomainPrefix = (directDomain === 'localhost') ? '' : 'openpgpkey.';
-    const hu = opgp.util.encodeZBase32(await opgp.crypto.hash.digest(opgp.enums.hash.sha1, Buf.fromUtfStr(user.toLowerCase())));
-    const directHost = (typeof this.port === 'undefined') ? directDomain : `${directDomain}:${this.port}`;
-    const advancedHost = `${advancedDomainPrefix}${directHost}`;
+    const hashed = await window.crypto.subtle.digest('SHA-1', Buf.fromUtfStr(user.toLowerCase()));
+    const hu = this.encodeZBase32(new Uint8Array(hashed));
+    const directHost = WKD_API_HOST || `https://${lowerCaseRecipientDomain}`; // lgtm [js/trivial-conditional]
+    const advancedHost = WKD_API_HOST || `https://openpgpkey.${lowerCaseRecipientDomain}`; // lgtm [js/trivial-conditional]
```
rrrooommmaaa, line 45: we can add subdomains of `localhost` -- not a problem --  I tried that with `gmail.localhost` -- it works perfectly well.
Maybe we don't even need to set up fes domain in `hosts` and use `fes.localhost`?

extension/js/common/core/common.ts
```diff
@@ -84,8 +84,9 @@ export class Str {
     if (email.indexOf(' ') !== -1) {
       return false;
     }
-    // for MOCK tests, we need emails like me@domain.com:8001 to pass
+    // for MOCK tests, we need emails like me@domain.com:8001 or me@localhost:8001to pass
```
rrrooommmaaa, line 87: do we still need these updates?

extension/js/common/api/key-server/wkd.ts
rrrooommmaaa: Unless you want to explicitly specify `https://` in `WKD_API_HOST`...

tomholub: If I didn't know that! That means we can throw away the requirement to set up domains with a hosts file. Let's look into this in another issue, it will be a larger change to the tests.

extension/js/common/core/common.ts
tomholub: indeed, should not be needed

tomholub: We still need the original `:8001` removal because many account emails for tests are written that way. I've filed #4631

extension/js/common/api/key-server/wkd.ts
tomholub: #4631

rrrooommmaaa (approved):

Issue #4593, opened 2022-08-10
Fix semaphoreci ui test


ioanmo226:
Sometimes semaphoreci ui test fails with timeout issue.
Need to be investigated.
https://flowcrypt.semaphoreci.com/jobs/cdbf4d6b-e900-494c-9481-c595677f9481

rrrooommmaaa:
I have also seen periodic failures of this test `user4@standardsubdomainfes.test:8001 - PWD encrypted message with FES web portal - some sends fail with BadRequest error` -- maybe related.

ioanmo226:
Right

PR #4627, opened 2022-08-22
#4607 look up users on LDAP explicitly using a ldap-relay API

### LINKED ISSUE #4607: look up users on LDAP explicitly using a ldap-relay API
should first do #4610

Until now, when the browser extension looked up public key on https://flowcrypt.com/attester/pub/user@example.test, the server would:
1) check if there is a public key in internal database, if yes return it
2) check if there is a public key in LDAP at `ldap://keys.[domain]`, if yes save & return
3) check WKD at recipient domain, if yes save & return

We we are updating attester to be purely pulling public keys from its own database and ignore other sources. Browser extension already knows how to pull from WKD, and now we'll also teach it how to pull from LDAP. Since the browser cannot do arbitrary TCP connections, there is now a proxy for this: https://flowcrypt.com/attester/ldap-relay?server=keys.example.test&search=user@example.test .  When a key was found, it will return status 200 and the armored key as text. Else it's 404. The format is the same as https://flowcrypt.com/attester/pub/user@example.test except that this ldap-relay endpoint can return more than one public key, concatenated with `\n`. 

Steps, approximately:

1) add a public method similar to `doLookup` maybe called `doLookupLdap` (the public method will also later be needed for https://github.com/FlowCrypt/flowcrypt-browser/issues/4609 )

The method accepts `email` and optional `server`. When server is not provided, it will derive it as follows: assume I'm looking up `user@example.test` then the server is `keys.example.test`. There is similar code in WKD and FES API code to extract the domain.

Since more than one armored block can be returned, you need to parse them apart with `MsgBlockParser.detectBlocks` if you get an error 200, then filter public keys from the parsed blocks, and then return the result as an array.

2) in `Attester.lookupEmail`, after checking that lookup is enabled, instead of calling `await this.doLookup(email)` please call the following:

```ts
const results = await Promise.all([
  this.doLookup(email),  // get from flowcrypt.com public keyserver database
  this.doLookupLdap(email),  // get from recipient-specific LDAP server, if any, relayed through flowcrypt.com
  this.doLookupLdap(email, 'keyserver.pgp.com'), // get from keyserver.pgp.com, relayed through flowcrypt.com
])
``` 

Then please choose and return only one set of results, with the following priority, starting from highest:
 - if at least one public key was returned from customer-specific LDAP, return an array of these public keys and nothing else
 - else, if flowcrypt.com public keyserver returned any public key, return only array of these
 - else, return array of keys returned from keyserver.pgp.com (or empty array)

And return that from `lookupEmail` like before.

For tests, you can add attester mock endpoints, to write tests ensuring that the priority is followed and that ldap searches are also disabled when other attester searches are disabled.

Also test that it can parse and recognize more than one public key (eg respond with two public keys for one of the looked up addresses, and then check that both were imported in settings)

ioanmo226:
This PR looks up users on LDAP explicitly using a ldap-relay API

close #4607 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


ioanmo226:
@tomholub With current implementation, Settings/attester page shows this error.
Do I have to ignore errors from ldap search results?

image

tomholub:
buggy endpoint, that was supposed to be a 404

Please treat error 500 as error 404 on this particular endpoint - meaning no pubkeys found.

Meanwhile I've filed an issue to improve the endpoint https://github.com/FlowCrypt/flowcrypt-attester-ts/issues/264 (you won't see it but it will eventually get fixed)

test/source/tests/page-recipe/compose-page-recipe.ts
```diff
@@ -111,7 +111,7 @@ export class ComposePageRecipe extends PageRecipe {
         if (sendingType !== 'to') { // input-to is always visible
           await composePageOrFrame.waitAndClick(`@action-show-${sendingType}`);
         }
-        await composePageOrFrame.waitAndType(`@input-${sendingType}`, email);
+        await composePageOrFrame.waitAndType(`@input-${sendingType}`, email + '\n');
```
ioanmo226, line 114: Seems like this also fixed https://github.com/FlowCrypt/flowcrypt-browser/issues/4593
Let's see if we see new same semaphoreci test after this merge

extension/js/common/api/key-server/attester.ts
```diff
@@ -25,7 +25,40 @@ export class Attester extends Api {
       console.info(`Skipping attester lookup of ${email} because attester search on this domain is disabled.`);
       return { pubkeys: [] };
     }
-    return await this.doLookup(email);
+    // first get from recipient-specific LDAP server, if any, relayed through flowcrypt.com
+    const customerLdapRes = await this.doLookupLdap(email);
```
rrrooommmaaa, line 29: The issue description suggested to start all 3 requests simultaneously to reduce response time.
You can, of course, not wait for all of them to complete,
popping and awaiting each promise in a way like:
```
const requests = [
// in reverse order for easier popping
  this.doLookupLdap(email, 'keyserver.pgp.com'), // get from keyserver.pgp.com, relayed through flowcrypt.com,
  this.doLookup(email),  // get from flowcrypt.com public keyserver database
  this.doLookupLdap(email),  // get from recipient-specific LDAP server, if any, relayed through flowcrypt.com
];
while (requests.length) {
 const result = await requests.pop();
 if (result.pubkeys.length) {
     // but you need to set `.catch` handlers for the remaining promises to avoid possible `unhandled rejection`
    return result;
 }
}

File: extension/js/common/api/key-server/attester.ts [hunk 3]
```diff
@@ -25,7 +25,40 @@ export class Attester extends Api {
...
+    const flowcryptRes = await this.doLookup(email);
+    if (flowcryptRes.pubkeys.length) {
+      return flowcryptRes;
+    }
+    // get from keyserver.pgp.com, relayed through flowcrypt.com
+    return await this.doLookupLdap(email, 'keyserver.pgp.com');
+  };
+
+  public doLookupLdap = async (email: string, server?: string): Promise<PubkeysSearchResult> => {
+    const parts = email.split('@');
```
rrrooommmaaa, line 43: we have a helper function `Str.getDomainFromEmailAddress`, please use it.

extension/js/common/api/key-server/attester.ts
rrrooommmaaa: Perhaps, we can add something like this to the `common.ts` utils

ioanmo226: Problem is with this requirement.
We should first return customer-specific LDAP result (if there are any) and then flowcrypt server pub key results and then keyserver.pgp.com result

```
Then please choose and return only one set of results, with the following priority, starting from highest:

if at least one public key was returned from customer-specific LDAP, return an array of these public keys and nothing else
else, if flowcrypt.com public keyserver returned any public key, return only array of these
else, return array of keys returned from keyserver.pgp.com (or empty array)

```

rrrooommmaaa: > but you need to set `.catch` handlers for the remaining promises to avoid possible `unhandled rejection`

in `finally` block, apparently

ioanmo226: Also with above way, we have to call 3 apis at once when we can call 1-2 apis.

rrrooommmaaa: or set them up from the beginning, but bearing in mind, that they can be ignored in certain circumstances

ioanmo226: Problem with my implementation is user have to wait all 3 requests if there are no customer specific LDAP result, flowcrypt pub key server results.
But all requests only take less than 1 sec and this doesn't seem to be a problem I think

extension/js/common/api/key-server/attester.ts
```diff
@@ -25,7 +25,40 @@ export class Attester extends Api {
...
+    if (parts.length !== 2) {
+      return { pubkeys: [] };
+    }
+    const [, recipientDomain] = parts;
+    const ldapServer = server ?? `keys.${recipientDomain}`;
+    try {
+      const r = await this.pubCall(`ldap-relay?server=${ldapServer}&search=${email}`);
+      const { blocks } = MsgBlockParser.detectBlocks(r.responseText);
```
rrrooommmaaa, line 51: This 3 lines 
```
const { blocks }
...
return { pubkeys}
```
can be re-used as a private method

test/source/mock/attester/attester-endpoints.ts
```diff
@@ -98,6 +104,24 @@ export const mockAttesterEndpoints: HandlersDefinition = {
...
+      if (emailOrLongid === 'test.ldap.priority@gmail.com') {
+        return protonMailCompatKey;
+      }
+      if (emailOrLongid === 'test.flowcrypt.pubkeyserver.priority@gmail.com') {
+        return protonMailCompatKey;
+      }
+      if (emailOrLongid === 'test.ldap.keyserver.pgp@gmail.com' && server === 'keyserver.pgp.com') {
+        return [protonMailCompatKey, testMatchPubKey].join('\n');
+      }
+      throw new HttpClientErr('No OpenPGP LDAP server on this address.', 404);
```
rrrooommmaaa, line 120: Please use constant `Status.NOT_FOUND` instead of 404.

I suggest to also add cases that emulate request timeout
( with `throw new HttpClientErr('RequestTimeout', Status.BAD_REQUEST);`)
before and after the successful fetch.

rrrooommmaaa: That is,
1) the priority (customer-specific LDAP) source timed out -- others returned something.
2) the priority source returned something, another source timed out.

extension/js/common/api/key-server/attester.ts
rrrooommmaaa: They take less than 1 sec to localhost?

rrrooommmaaa: > Problem is with this requirement. We should first return customer-specific LDAP result (if there are any) and then flowcrypt server pub key results and then keyserver.pgp.com result

I can't see what the problem is: `await Promise.all` returns an array of results in the same order as the order of promises in its argument array

rrrooommmaaa: > Problem with my implementation is user have to wait all 3 requests if there are no customer specific LDAP result, flowcrypt pub key server results. But all requests only take less than 1 sec and this doesn't seem to be a problem I think

You can start them all, and then wait one-by-one by popping and `await`ing them individually (as I showed above) until the rest can be skipped

ioanmo226: No. To live api.

> They take less than 1 sec to localhost?

By the way, Sorry i misunderstood what you mean.
You are right. Will implement this way.

tomholub: > Also with above way [as suggested by Roman], we have to call 3 apis at once when we can call 1-2 apis.

I'm ok to call 3 APIs at once even if their results are not always used, as long as that makes the method sometimes finish faster. Don't want to let the user wait.

test/source/mock/attester/attester-endpoints.ts
ioanmo226: Yeah. that test is must to have.
Just added test.

extension/js/common/api/key-server/attester.ts
```diff
@@ -25,7 +25,34 @@ export class Attester extends Api {
       console.info(`Skipping attester lookup of ${email} because attester search on this domain is disabled.`);
       return { pubkeys: [] };
     }
-    return await this.doLookup(email);
+    // Can't use Promise.allSettled because we are using ES2018 (Available from ES2020)
```
tomholub, line 28: You can set `minimum_chrome_version` in manifest to `96`, and start using `ES2020`. It's about time :-) then you won't need this shim.

extension/js/common/api/key-server/attester.ts
```diff
@@ -25,7 +25,34 @@ export class Attester extends Api {
...
+      this.doLookupLdap(email),  // get from recipient-specific LDAP server, if any, relayed through flowcrypt.com
+      this.doLookup(email),  // get from flowcrypt.com public keyserver database
+      this.doLookupLdap(email, 'keyserver.pgp.com'), // get from keyserver.pgp.com, relayed through flowcrypt.com
+    ]);
+    const validResults = results.filter(result => result.status === 'fulfilled' && result.value);
+    for (const result of validResults) {
+      if (result.value!.pubkeys.length) {
+        return result.value!;
+      }
+    }
```
tomholub, line 39: If there was an error searching `flowcrypt.com` (non-ldap) and no key was found elsewhere, than that error should be surfaced (re-thrown). Including network errors, 500s, timeouts etc. We expect that endpoint to respond normally.

I don't want to treat responses from `flowcrypt.com/attester/pub/xxx` as not-found when the server cannot be reached or returns error.

I think already on current codebase, doing that should cause a red color recipient that could be retried.

tomholub (commented):
Thank you - see comments

extension/js/common/api/key-server/attester.ts
tomholub: It's odd we don't have that scenario tested.

Please add a test `allows to retry public key search when attester returns error` that would search `attester.return.error@flowcrypt.test` and add `attester-endpoints` code that will return 500 for that address (or find existing email that already does)

Searching that in secure compose should render it as search error, instead of a no-pubkey result.

extension/js/common/api/key-server/attester.ts
```diff
@@ -25,7 +25,37 @@ export class Attester extends Api {
       console.info(`Skipping attester lookup of ${email} because attester search on this domain is disabled.`);
       return { pubkeys: [] };
     }
-    return await this.doLookup(email);
+    const results = await Promise.allSettled([
+      this.doLookupLdap(email),  // get from recipient-specific LDAP server, if any, relayed through flowcrypt.com
+      this.doLookup(email),  // get from flowcrypt.com public keyserver database
+      this.doLookupLdap(email, 'keyserver.pgp.com'), // get from keyserver.pgp.com, relayed through flowcrypt.com
+    ]);
+    const validResults = results.filter(result => result.status === 'fulfilled');
```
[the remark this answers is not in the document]
tomholub: It could theoretically happen on a lousy connection, where one request got through and the other one didn't. I think it's fine enough as is.

rrrooommmaaa (approved):

PR #4639, opened 2022-08-25
#4621 fix responding with plain message is not possible issue when using triple dot

### LINKED ISSUE #4621: responding with plain message is not possible when using triple dot
### This works:

1) open encrypted message
2) click default gmail reply icon above

image

3) I can write a plain reply, and also get a handy warning

image

### this doesn't work

1) open encrypted message
2) click more -> reply (plain)

image

3) gmail reply box shows
4) it gets replaced by encrypted reply box

------------------

The second case should work same as the first one. I'm unsure why is there any difference.

ioanmo226:
This PR fixes responding with plain message is not possible issue when using triple dot

close #4621 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


ioanmo226:
Thank you :)

PR #4646, opened 2022-08-26
Removing keys if absent on EKM

### LINKED ISSUE #4596: add support for removing private keys during EKM sync (and possibly recovering when syncing from 0 keys)
I forgot to submit my review earlier, so here are some unsubmitted comments from before. From the look of it, I could probably merge it before these are resolved. I'll take one more look and potentially merge it now.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4477#pullrequestreview-1054568449_

rrrooommmaaa:
This PR removes keys not present on EKM, redirects to setup page in case no keys are in the storage,
or silently generates a new key in case `PASS_PHRASE_QUIET_AUTOGEN` flag is present.
Also, correctly sets up passphrases for newly added keys (based on the location of an existing key's pass phrase)

close #4596

- Tests added or updated

### To be filled by reviewers


extension/js/common/helpers.ts
```diff
@@ -16,11 +16,18 @@ export const isFesUsed = async (acctEmail: string) => {
...
+    await KeyStore.set(acctEmail, await Promise.all(prvs.map(KeyUtil.keyInfoObj))); // todo: duplicate identities
+    // todo: should we delete passphrases matching the deleted keys?
+  }
   for (const prv of prvs) {
-    await KeyStore.add(acctEmail, prv);
+    if (!replaceKeys) {
+      await KeyStore.add(acctEmail, prv);
+    }
     if (ppOptions !== undefined) {
+      // todo: perhaps it's easier just to store a set of passphrases without specifying longids?
```
rrrooommmaaa, line 30: @tomholub 
I have a suspicion that we don't necessarily need to store passphrase connected to a particular key, since we already implemented the algo to search for keys that can be unlocked by a passphrase. We can simply store a set (array) of passphrases under `..._passphrases` key -- it would be easier to delete all the passphrases for the account.

tomholub: In principle, but if you have many keys (with many pass phrases) and you'd like to decrypt a key, you'll need to try all of them, which takes time (about 50ms per attempt last time I checked, but may have been improved since).

tomholub: That situation wouldn't happen for people who use EKM (we enforce same pass phrase for all), but could happen for other users.

test/source/tests/setup.ts
```diff
@@ -715,33 +725,108 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+          enterPp: { passphrase, checks: { isSavePassphraseChecked: false, isSavePassphraseHidden: true } }
+        });
+        const accessToken = await BrowserRecipe.getGoogleAccessToken(settingsPage, acct);
+        const extraAuthHeaders = { Authorization: `Bearer ${accessToken}` };
+        const set1 = await retrieveAndCheckKeys(settingsPage, acct, 1, passphrase);
+        MOCK_KM_UPDATING_KEY[acct].response = { privateKeys: [] };
+        // 1. EKM returns the empty set, forcing to auto-generate
+        let gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
+        // The new settingsPage is loaded in place of the existing settings tab (this is by design)
+        // However, after a second the newly-activated (old) settings tab loses focus in favour of the gmailPage, why is that?
```
rrrooommmaaa, line 763: Looks like this behaviour happens in Puppeteer only

test/source/tests/setup.ts
```diff
@@ -715,33 +725,108 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        // Is this only in Puppetteer?
+        await PageRecipe.noToastAppears(gmailPage);
+        await gmailPage.notPresent('@dialog-passphrase');
+        await gmailPage.close();
+        await retrieveAndCheckKeys(settingsPage, acct, 0, passphrase); // no keys, auto-generation
+        // todo: check that passphrase(s) are deleted from session?
```
rrrooommmaaa, line 769: @tomholub when a key gets removed by the EKM, how important is that the passphrase for the removed key gets from the session storage? I can test thoroughly this scenario, if required.

test/source/tests/setup.ts
```diff
@@ -715,33 +725,108 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        const accessToken = await BrowserRecipe.getGoogleAccessToken(settingsPage, acct);
+        const extraAuthHeaders = { Authorization: `Bearer ${accessToken}` };
+        const set1 = await retrieveAndCheckKeys(settingsPage, acct, 1);
+        MOCK_KM_UPDATING_KEY[acct].response = { privateKeys: [] };
+        // 1. EKM returns the empty set, auto-generation is not allowed, hence the error modal
+        let gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
+        await gmailPage.waitAndRespondToModal('error', 'confirm', 'Keys for your account were not set up yet - please ask your systems administrator'); await PageRecipe.noToastAppears(gmailPage);
+        await gmailPage.close();
+        await retrieveAndCheckKeys(settingsPage, acct, 0); // no keys
+        // todo: check that the passphrase is deleted from storage?
```
rrrooommmaaa, line 814: @tomholub In case of local storage, is it more important to make sure that the passphrase for the removed key is deleted as well?

test/source/tests/setup.ts
```diff
@@ -715,33 +725,108 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, 'Account keys updated');
+        await gmailPage.close();
+        const dbPage = await browser.newPage(t, TestUrls.extension('chrome/dev/ci_unit_test.htm'));
+        const set2 = await retrieveAndCheckKeys(dbPage, acct, 1);
+        expect(set2[0].id).to.equal(set1[0].id); // the key was received from the EKM
+        await dbPage.close();
+      }));
+
+    ava.default.todo('DEFAULT_REMEMBER_PASS_PHRASE with auto-generation when all keys are removed by EKM');
+    // should we re-use the known passphrase or delete it from the storage in this scenario?
```
rrrooommmaaa, line 828: @tomholub is there a possible setup where a private key creation is allowed together with PASS_PHRASE_QUIET_AUTOGEN? In this case we don't need to redirect to the settings page when all keys are removed by the EKM, as we can generate both a key and passphrase silently, can't we?

rrrooommmaaa:
Most of the scenarios are covered now.
It'd be interesting to emulate a situation when EKM supplies 2 instances of a key with the same identity --
two versions of the same key at once. We should handle this situation gracefully, right? It makes sense to store only one copy of the key -- but which one?

rrrooommmaaa:
Ping @tomholub 
Can you comment on what strategy on removing passphrases I should adhere to?

test/source/tests/setup.ts
tomholub: I think it is not so important. The pass phrase is the same for all user's keys anyway for EKM deployments, so if we remove one key (along with its pass phrase), the same pass phrase will still be available for the other key.

We should still do it, but we don't need to test for it.

test/source/tests/setup.ts
tomholub: In the interest of not leaving unused items in the storage, I think yes. But it doesn't have to be tested.

test/source/tests/setup.ts
tomholub: We don't have such deployment, and I'm not sure it would work if we actually configured it that way. Let's ignore this scenario.

test/source/tests/setup.ts
tomholub: I'm not sure why this would be. But it didn't seem to cause trouble for testing?

extension/js/common/helpers.ts
```diff
@@ -63,38 +71,49 @@ const parseAndCheckPrivateKeys = async (decryptedPrivateKeys: string[]) => {
...
+  }
   let passphrase = ppOptions?.passphrase;
   if (passphrase === undefined && !existingKeys.length) {
-    return { needPassphrase: false, updateCount: 0 }; // return success as we can't possibly validate a passphrase
-    // this can only happen on misconfiguration
-    // todo: or should we throw?
+    return { needPassphrase: true, noKeysSetup: true };
   }
-  let unencryptedKeysToSave = await filterKeysToSave(unencryptedPrvs, existingKeys);
   let encryptedKeys: Key[] = [];
```
tomholub, line 117: Let's call these `newEncryptedKeysToSave`

extension/chrome/settings/setup/setup-key-manager-autogen.ts
```diff
@@ -102,7 +102,7 @@ export class SetupWithEmailKeyManagerModule {
     }
     const storePrvOnKm = async () => this.view.keyManager!.storePrivateKey(this.view.idToken!, KeyUtil.armor(decryptablePrv));
     await Settings.retryUntilSuccessful(storePrvOnKm, 'Failed to store newly generated key on FlowCrypt Email Key Manager', Lang.general.contactIfNeedAssistance(this.view.isFesUsed()));
-    await saveKeysAndPassPhrase(this.view.acctEmail, [await KeyUtil.parse(generated.private)], setupOptions); // store encrypted key + pass phrase locally
+    await saveKeysAndPassPhrase(this.view.acctEmail, [await KeyUtil.parse(generated.private)], setupOptions, true); // store encrypted key + pass phrase locally
```
tomholub, line 105: I suppose here doesn't need the `replace: true`?

extension/js/common/helpers.ts
```diff
@@ -11,16 +11,24 @@ import { ContactStore } from './platform/store/contact-store.js';
...
   const { fesUrl } = await AcctStore.get(acctEmail, ['fesUrl']);
   return Boolean(fesUrl);
 };
 
-export const saveKeysAndPassPhrase = async (acctEmail: string, prvs: Key[], ppOptions?: PassphraseOptions) => {
+export const saveKeysAndPassPhrase = async (acctEmail: string, prvs: Key[], ppOptions?: PassphraseOptions, replaceKeys: boolean = false) => {
   const clientConfiguration = await ClientConfiguration.newInstance(acctEmail);
+  if (replaceKeys) {
+    await KeyStore.set(acctEmail, await Promise.all(prvs.map(KeyUtil.keyInfoObj))); // todo: duplicate identities
+    // todo: should we delete passphrases matching the deleted keys?
```
tomholub, line 24: Yes, especially for stored ones. But anyway also removing the memory ones is the same amount of work.

extension/js/common/helpers.ts
```diff
@@ -11,16 +11,24 @@ import { ContactStore } from './platform/store/contact-store.js';
...
+  }
   for (const prv of prvs) {
-    await KeyStore.add(acctEmail, prv);
+    if (!replaceKeys) {
+      await KeyStore.add(acctEmail, prv);
+    }
     if (ppOptions !== undefined) {
+      // todo: perhaps it's easier just to store a set of passphrases without specifying longids?
```
tomholub, line 31: In situations with many different keys with different pass phrases (manually imported) we'd have to loop-try through pass phrases, which is computationally expensive. Say 3 keys, 3 pass phrases, could be close to 10 attempts (times amount of subkeys to decrypt per key, in a way). So it makes sense to track the associated longid of each pass phrase.

extension/js/common/helpers.ts
```diff
@@ -63,38 +71,49 @@ const parseAndCheckPrivateKeys = async (decryptedPrivateKeys: string[]) => {
...
+  return { keysToRetain, unencryptedKeysToSave };
 };
 
 export const processAndStoreKeysFromEkmLocally = async (
-  { acctEmail, decryptedPrivateKeys, ppOptions }: { acctEmail: string, decryptedPrivateKeys: string[], ppOptions?: PassphraseOptions }
+  { acctEmail, decryptedPrivateKeys, ppOptions: originalOptions }: Bm.ProcessAndStoreKeysFromEkmLocally & { ppOptions?: PassphraseOptions }
 ): Promise<Bm.Res.ProcessAndStoreKeysFromEkmLocally> => {
   const { unencryptedPrvs } = await parseAndCheckPrivateKeys(decryptedPrivateKeys);
   const existingKeys = await KeyStore.get(acctEmail);
+  let { keysToRetain, unencryptedKeysToSave } = await filterKeysToSave(unencryptedPrvs, existingKeys);
```
tomholub, line 102: let's call the second ones `newUnencryptedKeysToSave`

extension/js/common/helpers.ts
```diff
@@ -63,38 +71,49 @@ const parseAndCheckPrivateKeys = async (decryptedPrivateKeys: string[]) => {
...
+    return { keysToRetain: [], unencryptedKeysToSave: candidateKeys };
   }
-  const result: Key[] = [];
+  const keysToRetain: Key[] = [];
+  const unencryptedKeysToSave: Key[] = [];
   for (const candidate of candidateKeys) {
     const longid = KeyUtil.getPrimaryLongid(candidate);
     const keyToUpdate = existingKeys.filter(ki => ki.longid === longid && ki.family === candidate.family);
     if (keyToUpdate.length === 1) {
       const oldKey = await KeyUtil.parse(keyToUpdate[0].private);
```
tomholub, line 84: Let's call this `parsedExistingKey`. We don't know if it's `old` or not (with respect to lastModified) so it can get confusing.

extension/js/common/helpers.ts
```diff
@@ -63,38 +71,49 @@ const parseAndCheckPrivateKeys = async (decryptedPrivateKeys: string[]) => {
...
 const filterKeysToSave = async (candidateKeys: Key[], existingKeys: KeyInfoWithIdentity[]) => {
+  // todo: check for uniqueness of candidateKeys identities here?
   if (!existingKeys.length) {
-    return candidateKeys;
+    return { keysToRetain: [], unencryptedKeysToSave: candidateKeys };
   }
-  const result: Key[] = [];
+  const keysToRetain: Key[] = [];
+  const unencryptedKeysToSave: Key[] = [];
   for (const candidate of candidateKeys) {
```
tomholub, line 80: Let's call it `candidateKey` also here

extension/js/common/helpers.ts
```diff
@@ -63,38 +71,49 @@ const parseAndCheckPrivateKeys = async (decryptedPrivateKeys: string[]) => {
...
     const longid = KeyUtil.getPrimaryLongid(candidate);
     const keyToUpdate = existingKeys.filter(ki => ki.longid === longid && ki.family === candidate.family);
```
tomholub, line 82: Call this `existingKey`, and instead of a `filter` let's do `find` or similar that yields undefined or the result, not an array. We don't need the `length > 1` check here, since this comes from our own storage.

test/source/tests/setup.ts
```diff
@@ -715,33 +725,108 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
-      const mainKey14 = KeyUtil.filterKeysByIdentity(set14.map(ki => ki), [{ family: 'openpgp', id: '392FB1E9FF4184659AB6A246835C0141B9ECF536' }]);
+      const set14 = await retrieveAndCheckKeys(settingsPage, acct, 1, passphrase);
+      expect(set14.map(entry => entry.id)).to.eql(['277D1ADA213881F4ABE0415395E783DC0289E2E2']);
+      const mainKey14 = KeyUtil.filterKeysByIdentity(set14.map(ki => ki), [{ family: 'openpgp', id: '277D1ADA213881F4ABE0415395E783DC0289E2E2' }]);
       expect(mainKey14.length).to.equal(1);
       expect(mainKey14[0].lastModified).to.equal(mainKey13[0].lastModified); // no update
       await gmailPage.close();
     }));
 
+    ava.default('put.updating.key@key-manager-choose-passphrase-forbid-storing.flowcrypt.test - updates of key found on key manager via setup page (with passphrase)',
```
tomholub, line 747: Do any of the new tests check that keys wouldn't be duplicated in storage when there was an older version of the key previously present? (that we are not storing both original version and updated version of the key alongside each other)

test/source/tests/setup.ts
```diff
@@ -643,42 +654,42 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
       expect(set8[0].lastModified).to.equal(set7[0].lastModified); // no update
       await gmailPage.close();
       // 8. EKM returns an older version of the existing key, and a new key, toast, new key gets added encrypted with the same passphrase
-      MOCK_KM_UPDATING_KEY.response = { privateKeys: [{ decryptedPrivateKey: someOlderVersion }, { decryptedPrivateKey: testConstants.existingPrv }] };
+      MOCK_KM_UPDATING_KEY[acct].response = { privateKeys: [{ decryptedPrivateKey: someOlderVersion }, { decryptedPrivateKey: testConstants.existingPrv }] };
       gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
       await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, 'Account keys updated');
       await gmailPage.notPresent('@dialog-passphrase');
-      const set9 = await retrieveAndCheckKeys(2);
+      const set9 = await retrieveAndCheckKeys(settingsPage, acct, 2);
```
rrrooommmaaa, line 673: @tomholub 
I didn't test earlier that the newly-added keys (with a found or entered passphrase) are immediately ready-to-use ("unlocked" in a sense that their passphrase is stored in either cache or local storage). So this test is failing now.
What would be our strategy on saving passphrases for new keys (we don't provide the checkbox for the user unless the setup page scenario is used)? Should the passphrase for the newly-added key be saved to storage if the existing key (that was used to provide the passphrase) had its passphrase in the storage (settings inheritance)? Or some straightforward algo based on EKM flags (PASS_PHRASE_QUIET_AUTOGEN, DEFAULT_REMEMBER_PASS_PHRASE, FORBID_STORING_PASS_PHRASE...)?
Also, updating keys from EKM may silently extend passphrase cache expiration (4 hours) in session memory if we simply make a copy of the passphrase for the new longid. Should we care about it?

test/source/tests/setup.ts
rrrooommmaaa: All the tests check the total expected number of keys in the storage after each step. If there is an unexpected "extra" key, any of the tests will fail.

rrrooommmaaa: The check is inside `retrieveAndCheckKeys` that has `expectedKeyCount` as a mandatory parameter.
https://github.com/FlowCrypt/flowcrypt-browser/blob/12a5fb78bda305ae028bd25d2b8aaebecf1cc9af/test/source/tests/setup.ts#L583

test/source/tests/setup.ts
rrrooommmaaa: After closing `gmail` tab, the `settings` tab becomes active again and the test resumes normally.

extension/chrome/settings/setup/setup-key-manager-autogen.ts
rrrooommmaaa: Yes, I removed it and it didn't break any tests

test/source/tests/setup.ts
rrrooommmaaa: ping @tomholub 
I can fix this in this PR, should I? If so, please confirm the strategy.

tomholub: Well noticed.

> Should the passphrase for the newly-added key be saved to storage if the existing key (that was used to provide the passphrase) had its passphrase in the storage (settings inheritance)?

Yes, that is the most reasonable

>  Also, updating keys from EKM may silently extend passphrase cache expiration (4 hours) in session memory if we simply make a copy of the passphrase for the new longid. Should we care about it?

I find this ok, as long as it only happens when there is an actual change of the keys. Wouldn't want this to happen each time EKM is checked for new keys with no updates to process.

Let's do that in this PR. Thanks!

test/source/tests/setup.ts
tomholub: Ok. Unfortunately, I don't have any feedback, not sure if I encountered it before.

extension/chrome/settings/modules/add_key.ts
```diff
@@ -100,9 +98,13 @@ View.run(class AddKeyView extends View {
     try {
       const checked = await this.keyImportUi.checkPrv(this.acctEmail, String($('.input_private_key').val()), String($('.input_passphrase').val()));
       if (checked) {
-        await KeyStore.add(this.acctEmail, checked.encrypted); // resulting new_key checked above
-        const storageType: StorageType = ($('.input_passphrase_save').prop('checked') && !this.clientConfiguration.forbidStoringPassPhrase()) ? 'local' : 'session';
-        await PassphraseStore.set(storageType, this.acctEmail, { longid: KeyUtil.getPrimaryLongid(checked.encrypted) }, checked.passphrase);
+        await saveKeysAndPassPhrase(this.acctEmail, [checked.encrypted]); // resulting new_key checked above
```
rrrooommmaaa, line 101: @tomholub `saveKeysAndPassPhrase` differs from plain `KeyStore.add` in a way that it saves the corresponding public keys in ContactStore for all the user's own email aliases.
Should we plan to use this helper function instead of `KeysTore.add` in all the remaining places in a separate task?

test/source/tests/setup.ts
rrrooommmaaa: Done

extension/chrome/settings/modules/add_key.ts
tomholub: Yes, we should consider doing that in a separate task, and also renaming that method. If it does that, it's not too appropriately named. Maybe `savePrvAndPubKeysAndPassPhrases`

tomholub:
@rrrooommmaaa is this ready for final review? There's something still going on with tests, and it looks like a git merge is in order.

extension/js/common/platform/store/in-memory-store.ts
```diff
@@ -4,7 +4,8 @@ import { AbstractStore } from './abstract-store.js';
 import { BrowserMsg } from '../../browser/browser-msg.js';
 
 /**
- * Temporary In-Memory store for sensitive values, expiring after 4 hours
+<<<<<<< HEAD
```
tomholub, line 7: merge artifact

tomholub (commented):
So far good, still some reviewing to do.

rrrooommmaaa:
> There's something still going on with tests

Some tests are acting flaky indeed

test/source/tests/setup.ts
```diff
@@ -738,33 +749,108 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        expect(await getPassphrase(settingsPage, acct, KeyUtil.getPrimaryLongid(set1[0]))).to.be.an.undefined; // the passphrase for the old key was deleted
+        delete MOCK_KM_LAST_INSERTED_KEY[acct];
+        await SetupPageRecipe.autoSetupWithEKM(settingsPage, {
+          enterPp: { passphrase, checks: { isSavePassphraseChecked: false, isSavePassphraseHidden: true } }
+        });
+        expect(MOCK_KM_LAST_INSERTED_KEY[acct]).to.exist;
+        const set2 = await retrieveAndCheckKeys(settingsPage, acct, 1);
+        expect(set2[0].id).to.not.equal(set1[0].id); // entirely new key was generated
+        // 2. Adding a new key from the key manager when there is none in the storage
+        // First, erase the keys
```
tomholub, line 802: Where in the code below are keys being erased? Not so obvious.

tomholub (approved):
With a comment - please see below

test/source/tests/setup.ts
rrrooommmaaa: They are erase by address the EKM again which has an empty set of keys set up. I added a comment and made the configuration a bit more obvious

tomholub (approved):
All looks good to me, good to merge once tests are passing

tomholub:
Ah, GitGuardian

PR #4657, opened 2022-08-31
#2649 Google OAuth OOB migration (Using chrome.identity)

### LINKED ISSUE #2649: stop using urn:ietf:wg:oauth:2.0:oob:auto before Oct 3

ioanmo226:
This PR implemented Google OAuth OOB migration

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/background_page/background_page.ts
+++ b/extension/js/background_page/background_page.ts
@@ -73,7 +73,6 @@ opgp.initWorker({ path: '/lib/openpgp.worker.js' });
   BrowserMsg.bgAddListener('settings', BgHandlers.openSettingsPageHandler);
   BrowserMsg.bgAddListener('update_uninstall_url', BgHandlers.updateUninstallUrl);
   BrowserMsg.bgAddListener('get_active_tab_info', BgHandlers.getActiveTabInfo);
-  BrowserMsg.bgAddListener('get_auth_token', BgHandlers.getAuthToken);
   BrowserMsg.bgAddListener('reconnect_acct_auth_popup', (r: Bm.ReconnectAcctAuthPopup) => GoogleAuth.newAuthPopup(r));
   BrowserMsg.bgAddListener('_tab_', BgHandlers.respondWithSenderTabId);
   BrowserMsg.bgListen();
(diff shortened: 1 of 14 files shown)


tomholub:
This is great, you have basically done it. The rest is cleanup work :-)

> New OAuth popup is not recognized by puppeteer. and all puppeteer tests fail.

Got it, that should be fixable

> With our new OAuth popup, google displays optional checkbox in consent screen like below.
> We might check if permission is allowed in each screen and ask user to allow permission but that might be a huge change.

Indeed that would be too much work, for now. However, we already do this for `Contacts`. Therefore you can ignore missing contacts permission, it's optional during setup.

> So my suggestion is to display error(info) popup when there are missing scopes user didn't check in login screen.
Something like, we need xxx permissions to allow you to do xxx. would you like to try again & allow consent?
And then show consent screen again (this time, checkboxes are not showing as user previously allowed basic scopes)
What do you think?

We've had this situation on iOS. I think the fix should be the same. It's pretty much what you said. Following a successful but incomplete authorization, you can implement it with a modal that says the same thing as we did in iOS, it was something like "This worked, now please also  ..." and when user clicks "ok", show another authorization. Just copy the text and buttons from iOS.

> Also as we save & restore scopes through AcctStore, contact search scope was set as true(even though user doesn't check contact search consent) and contact search api returns 403 error.

Maybe it wasn't such a good idea to keep this in `AcctStore`. It can get out of sync. Is there a Google endpoint where we can figure out current authorizations? Otherwise, any ideas?

ioanmo226:
I totally agree. I think there might be functionality to check authorized scopes in web too as we did successfully in iOS app.
By the way, I think better to do this in another task as this one is big enough task.
> Maybe it wasn't such a good idea to keep this in AcctStore. It can get out of sync. Is there a Google endpoint where we can figure out current authorizations? Otherwise, any ideas?

tomholub:
> I totally agree. User might remove it manually from google account page and it can get out of sync easily. I think there must be functionality to check authorized scopes in web too as we did successfully in iOS app. By the way, I think better to do this in another task as this one is big enough task.
> 
> > Maybe it wasn't such a good idea to keep this in AcctStore. It can get out of sync. Is there a Google endpoint where we can figure out current authorizations? Otherwise, any ideas?

On iOS, the client library does that. It's possible they investigate the response from the authorization endpoint, or maybe it's a separate call. You could investigate the library source to find out.

ioanmo226:
Yeah, sure :-)

tomholub:
> I totally agree. User might remove it manually from google account page and it can get out of sync easily. I think there must be functionality to check authorized scopes in web too as we did successfully in iOS app. By the way, I think better to do this in another task as this one is big enough task.
> 
> > Maybe it wasn't such a good idea to keep this in AcctStore. It can get out of sync. Is there a Google endpoint where we can figure out current authorizations? Otherwise, any ideas?

Sure, if a workable solution can be achieved without removing this, then we should indeed leave that for another day.

tomholub:
closing in favor of https://github.com/FlowCrypt/flowcrypt-browser/pull/4665

PR #4659, opened 2022-09-01
issue 4348 only send sender email in replyInfoDiv

### LINKED ISSUE #4348: send one message per web portal recipient, each with different url

tomholub:
This PR fixes `replyInfoDiv` format. It was sending `{email: string, name: string}` instead of `string`.

close #4348

- I decided not to test this since I'd like to send out update soon and it's not easy to test. I've refactored the code to make it easier to notice this regression in PR code reviews by adding `type ReplyInfoRaw` which must remain stable to work with FES UI. With this in place, the original problem would have been noticed and corrected during development or review.

### To be filled by reviewers

### DIFF
--- a/test/source/tests/compose.ts
+++ b/test/source/tests/compose.ts
@@ -599,13 +599,13 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       const correctButtonStatusTxt = 'Encrypt, Sign and Send';
       await ComposePageRecipe.showRecipientInput(composePage);
       await composePage.waitAll('@container-cc-bcc-buttons');
-      expect(await composePage.read('@action-send')).to.eq(correctButtonStatusTxt);
+      await composePage.waitForContent('@action-send', correctButtonStatusTxt);
       await composePage.waitAndType(`@input-to`, 'aaaaa\n'); // First enter invalid recipient
-      expect(await composePage.read('@action-send')).to.eq('Re-enter recipient..');
+      await composePage.waitForContent('@action-send', 'Re-enter recipient..');
       await composePage.press('Backspace'); // Delete invalid recipient
-      expect(await composePage.read('@action-send')).to.eq(correctButtonStatusTxt); // check if sent button status is correct
+      await composePage.waitForContent('@action-send', correctButtonStatusTxt); // check if sent button status is correct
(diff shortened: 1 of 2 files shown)


test/source/tests/compose.ts
```diff
@@ -599,13 +599,13 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
       await composePage.waitAndType(`@input-to`, 'aaaaa\n'); // First enter invalid recipient
-      expect(await composePage.read('@action-send')).to.eq('Re-enter recipient..');
+      await composePage.waitForContent('@action-send', 'Re-enter recipient..');
       await composePage.press('Backspace'); // Delete invalid recipient
-      expect(await composePage.read('@action-send')).to.eq(correctButtonStatusTxt); // check if sent button status is correct
+      await composePage.waitForContent('@action-send', correctButtonStatusTxt); // check if sent button status is correct
       await composePage.waitAndType(`@input-to`, 'mock.only.pubkey@flowcrypt.com\n'); // Now enter correct recipient and check if send button status is correct.
-      expect(await composePage.read('@action-send')).to.eq(correctButtonStatusTxt);
+      await composePage.waitForContent('@action-send', correctButtonStatusTxt);
     }));
```
tomholub, line 609: @ioanmo226 reading a value "at a point in time" results in flaky tests, because sometimes the value didn't have a chance to change yet if the test code is faster than the browser extension code in that moment. It could be that 5ms later the right value would have appeared, but the test will fail anyway if written that way.

Usually `composePage.waitForContent` is the more stable way to do it. The test will keep testing the value in a loop repeatedly. If it eventually has the right value, it will succeed, else if it had to wait too long (longer than timeout value) then it will fail.

I've updated it accordingly

ioanmo226: Got it. Thanks

tomholub:
I'll do a release

PR #4660, opened 2022-09-02
#2649 Google OAuth OOB migration(Using Web request and webRequestBlocking)

### LINKED ISSUE #2649: stop using urn:ietf:wg:oauth:2.0:oob:auto before Oct 3

ioanmo226:
This PR implemented OOB migration

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -28,6 +28,11 @@
   "permissions": [
     "storage",
     "tabs",
+    "webRequest",
+    "webRequestBlocking",
+    "webNavigation",
+    "https://bnjglocicdkmhmoohhfkfkbbkejdhdgc.chromiumapp.org/",
+    "https://localhost:8001/*",
     "https://flowcrypt.com/*",
     "https://*.google.com/*",
     "https://outlook.live.com/*",
--- a/extension/js/common/api/email-provider/gmail/google-auth.ts
+++ b/extension/js/common/api/email-provider/gmail/google-auth.ts
@@ -5,23 +5,22 @@
 // tslint:disable:no-direct-ajax
 // tslint:disable:oneliner-object-literal
 
-import { OAUTH_GOOGLE_API_HOST, GOOGLE_OAUTH_SCREEN_HOST, FLAVOR } from '../../../core/const.js';
+import { windowsCreate } from '../../../browser/chrome.js';
 import { Str, Url, Value } from '../../../core/common.js';
-import { tabsQuery, windowsCreate } from '../../../browser/chrome.js';
-import { Api } from './../../shared/api.js';
+import { FLAVOR, GOOGLE_OAUTH_SCREEN_HOST, OAUTH_GOOGLE_API_HOST } from '../../../core/const.js';
 import { ApiErr } from '../../shared/api-error.js';
+import { Api } from './../../shared/api.js';
 
-import { Buf } from '../../../core/buf.js';
-import { Catch } from '../../../platform/catch.js';
-import { GmailRes } from './gmail-parser';
-import { GoogleAuthErr } from '../../shared/api-error.js';
 import { GoogleAuthWindowResult$result } from '../../../browser/browser-msg.js';
-import { Ui } from '../../../browser/ui.js';
+import { Buf } from '../../../core/buf.js';
+import { InMemoryStoreKeys } from '../../../core/const.js';
+import { Catch } from '../../../platform/catch.js';
 import { AcctStore, AcctStoreDict } from '../../../platform/store/acct-store.js';
+import { InMemoryStore } from '../../../platform/store/in-memory-store.js';
 import { AccountServer } from '../../account-server.js';
 import { EnterpriseServer } from '../../account-servers/enterprise-server.js';
-import { InMemoryStore } from '../../../platform/store/in-memory-store.js';
-import { InMemoryStoreKeys } from '../../../core/const.js';
+import { GoogleAuthErr } from '../../shared/api-error.js';
+import { GmailRes } from './gmail-parser';
 
 type GoogleAuthTokensResponse = { access_token: string, expires_in: number, refresh_token?: string, id_token: string, token_type: 'Bearer' };
 type AuthResultSuccess = { result: 'Success', acctEmail: string, id_token: string, error?: undefined };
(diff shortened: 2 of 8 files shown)


extension/manifest.json
```diff
@@ -28,6 +28,11 @@
   "permissions": [
     "storage",
     "tabs",
+    "webRequest",
+    "webRequestBlocking",
+    "webNavigation",
+    "https://bnjglocicdkmhmoohhfkfkbbkejdhdgc.chromiumapp.org/",
+    "https://localhost:8001/*",
```
ioanmo226, line 35: @tomholub Rather than using chrome.identity.launchWebAuthFlow (Like in [#4657](https://github.com/FlowCrypt/flowcrypt-browser/pull/4657)) I decided to make my own launchWebAuthFlow function.
The reason is identity.launchWebAuthFlow documentation is really poor, lack of customization, CORS issue after authorization, etc.
Because of above CORS issue/lack of customization, I wasted so much time and decided to use my own method.

With new method, we need above new permissions and I don't think it will hurt us.

This PR still needs cleanup (80-90% completed) but I think much better than original [PR](https://github.com/FlowCrypt/flowcrypt-browser/pull/4657)

Please let me know your thought.

tomholub: But the above permission is just for tests right? Not for production?

ioanmo226: It's also needed for production.
The reason is browser watches oauth window web request & if url matches redirect_uri, then code treats it as oauth2 login completed.

ioanmo226: This logic
```ts
chrome.webRequest.onBeforeRequest.addListener((details) => {
  if (details.frameId || details.tabId !== tabId) return;
  if (!details.url.startsWith(GoogleAuth.OAUTH.redirect_uri)) return;
  resolve(details.url);
  return { cancel: true };
}, {
  urls: ["*://*/*"],
  tabId
}, ["blocking"]);
chrome.tabs.onRemoved.addListener((removedTabId) => {
  if (removedTabId === tabId) {
    reject(new Error("Canceled by user"));
  }
});
```

tomholub: I don't understand. Browser extension can't just run a server at localhost:8001, can it?

ioanmo226: Oh.. of course, `"https://localhost:8001/*",` permission can be removed.
I meant `webNavigation`, `webRequestBlocking`, `webRequest` permissions

tomholub: These are problematic permissions though. Adding them will cause all already installed extension to get disabled, and users will have to re-enable them.

Besides I believe at least one of those is no longer supported in Chrome extension manifest v3. You should check that, too.

Finally, they may cause some scary warnings when installing the extension, as compared to current permissions (that would have to be tested)

tomholub: As an alternative, could we not have a more simple permission for the `redirect_uri` and run a content script there? Similar to what we did before? Or that no longer works? It seems the only thing you need is the URL, which a content script could provide.

ioanmo226: As you can see, using web request api is the best option to catch redirect_url in chrome extension.

This also applies when we add `identity` permission I think.
> These are problematic permissions though. Adding them will cause all already installed extension to get disabled, and users will have to re-enable them.


image

tomholub: they say it's best, but don't say why. it may not be best for us. have you tried the content script option?

is declarativeWebRequest already available in v2 extensions?

ioanmo226: It mentioned it's best because it is activated before a page is even requested.
> they say it's best, but don't say why

Not yet.
> have you tried the content script option?

It's deprecated and need to use `declarativeNetRequest`
image

> is declarativeWebRequest already available in v2 extensions?

tomholub: I'll need to see what kind of warnings will be displayed upon installation when `declarativeWebRequest` is used.

ioanmo226: You mean `declarativeNetRequest`?

tomholub: Yes, I meant `declarativeNetRequest`. Would need to see what permission warnings it triggers, if any.

tomholub: > Use chrome.identity api like this https://github.com/FlowCrypt/flowcrypt-browser/pull/4657 (I couldn't achieve to fix ui test till yet though. Couldn't get access to oauth dialog, and CORS issue after getting authorization code..)

The biggest problem with this is that you can only log into the account that are signed into chrome, right? That wouldn't work well, and wouldn't work on Firefox at all.

> Use webRequest and declarativeNetRequest permissions like in this PR.

Sounds most reliable, but will require a new permission. That disables everybody's extensions until they re-confirm the new permission, which is not convenient. And the permission is `Block page content` which doesn't sound the worst, but at the same time it's confusing to the user as to why exactly this extension needs to block user content. 

I'd say this should be our backup option if we can't get the third option to work.

> Use content script like above library.

I've done this in the past. If we set the redirect url to be `http://www.google.com/robots.txt` like the library did, then we don't have to add a new permission, either, since we already have `https://*.google.com/*` permission. Let's take this option as the first choice.

Thanks!

ioanmo226: No. If we use `chrome.identity.launchWebAuthFlow`, this problem doesn't happen at all.
> The biggest problem with this is that you can only log into the account that are signed into chrome
I've done this in the past. If we set the redirect url to be http://www.google.com/robots.txt like the library did, then we don't have to add a new permission, either, since we already have https://*.google.com/* permission. Let's take this option as the first choice.

tomholub: > Use chrome.identity api like this https://github.com/FlowCrypt/flowcrypt-browser/pull/4657 (I couldn't achieve to fix ui test till yet though. Couldn't get access to oauth dialog, and CORS issue after getting authorization code..)

> No. If we use chrome.identity.launchWebAuthFlow, this [choice of user to log in with] problem doesn't happen at all.

Sorry. Got it. Was the issue you described only applicable to tests?
 - Couldn't get access to oauth dialog: that may be difficult to fix, depending how chrome implements this
 - cors issue: this can be fixed probably easily by adding another permission for the url you are trying to CORS-access, only for the tests

I'll just leave this here for future reference, in case we do have to come back to it.

ioanmo226: Yes.
> Sorry. Got it. Was the issue you described only applicable to tests?

I tried every possible solutions (including below one) but none of them succeed.
Couldn't find where it references google oauth (using identity api) ui test too.
> cors issue: this can be fixed probably easily by adding another permission for the url you are trying to CORS-access, only for the tests

ioanmo226: Better if you run & build & test (manual testing + puppeteer test[it would fail at first stage for PR which used chrome.identity api though]) both 2 PRs

ioanmo226: Did you register `http://www.google.com/robots.txt`  as redirect uri?
I get this error.`Error 400: redirect_uri_mismatch`

tomholub: I just updated it to `https://www.google.com/robots.txt` (notice https vs http). It may take a few minutes to reflect

tomholub:
closing in favor of https://github.com/FlowCrypt/flowcrypt-browser/pull/4665

Issue #4594, opened 2022-08-10
Fix broken live gmail test


ioanmo226:
It seems like this PR broke live gmail test.
https://flowcrypt.semaphoreci.com/jobs/866f120d-0c1d-4cda-ac01-c61325804d2d

_Originally posted by @ioanmo226 in https://github.com/FlowCrypt/flowcrypt-browser/issues/4563#issuecomment-1209617849_

rrrooommmaaa:
Fixed by #4595, thanks for noticing.

ioanmo226:
You are welcome.
Hmm... I can see that live test still fails though

rrrooommmaaa:
Are the live tests on `master` ok now? I'm having some troubles searching them, perhaps missing some privileges...

ioanmo226:
I don't think so.
Live test sometimes fails on master.
Ex: https://flowcrypt.semaphoreci.com/workflows/dbf0f8d0-ea66-400a-acff-f05a6e906bee?pipeline_id=8637312d-0fac-4db0-b246-9809f12466c1

ioanmo226:
Whole live tests are now broken. Maybe because of gmail behavior change.
Need to investigate and fix

tomholub:
@ioanmo226 do you think you may be able to fix this, after the OOB PR?

ioanmo226:
OK

tomholub:
We could maybe go through each test and consider if it could be written as a mocked tests. Some of the tests that don't rely on Gmail HTML explicitly could be moved to mock.

For example there is a live test to check the initial notification user gets when they first install the extension, in Gmail it says "please go through setup" on top with a link to do it or dismiss or don't show again. That test could be moved to mocked tests now that we also have a mock of the Gmail page available. There may be a few more tests like that, which don't use nearly a single HTML tag from the Gmail webpage, and are therefore safe to mock instead. You could look into that in another issue, or if exactly it's such a test that is giving you trouble now, you could look into that in this issue.

ioanmo226:
@tomholub `https://flowcrypt.com/api/account/login` api returns this error for this request.
I think something needs to be fixed in backend side?
```js
account: "ci.tests.gmail@flowcrypt.dev"
token: null
uuid: "c021c7d1eb1be867fb225797bb630ea0838882d6"
```

[...]

PR #4665, opened 2022-09-06
#2649 Google OAuth OOB migration (Using Content Script)

### LINKED ISSUE #2649: stop using urn:ietf:wg:oauth:2.0:oob:auto before Oct 3

ioanmo226:
This PR implemented google oauth oob migration.

- Tests added or updated

### To be filled by reviewers


ioanmo226:
All finished. Now ready for review.
@tomholub Please check GitGuardian

tomholub:
@ioanmo226 could you please manually test this change in Firefox - does it still work?

extension/js/common/oauth2/oauth2_inject.ts
```diff
@@ -0,0 +1,10 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
+
+// Decleared win variable to avoid `Type 'string' is not assignable to type 'Location | (string & Location)'.ts(2322)` error
+const win: Window = window;
+
+// Redirect back to the extension itself so that we have priveledged
+// access again
```
tomholub, line 7: Please update this comment to explain what type of privileged access this is, and what was it needed for that can't be done directly from here.

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -200,72 +189,39 @@ export class GoogleAuth {
...
+    const match = RegExp('[?&]' + name + '=([^&]*)').exec(search);
+    return match && decodeURIComponent(match[1].replace(/\+/g, ' '));
   };
 
-  private static processOauthResTitle = (title: string): { result: GoogleAuthWindowResult$result, code?: string, error?: string, csrf?: string } => {
-    const parts = title.split(' ', 2);
-    const result = parts[0] as GoogleAuthWindowResult$result;
-    const params = Url.parse(['code', 'state', 'error'], parts[1]);
-    let authReq: AuthReq;
+  private static getOAuthResult = async ({ acctEmail, scopes, save }: { acctEmail?: string, scopes: string[], save: boolean }): Promise<AuthRes> => {
```
tomholub, line 197: Method `getOAuthResult` should be called `renderOAuthWindowAndGetResult` to reflect what it actually does.

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -184,6 +161,18 @@ export class GoogleAuth {
...
+    return await new Promise((resolve, reject) => {
+      void OAuth2.launchWebAuthFlow(url, (redirectUrl: string) => {
+        if (chrome.runtime.lastError) {
+          reject(chrome.runtime.lastError.message);
+          return;
+        }
+        resolve(redirectUrl);
+      });
+    });
+  };
```
tomholub, line 174: This `if (chrome.runtime.lastError) {` is rather indirect, but I suppose you had some reason for it? As opposed to just throwing the error where it happens and then catching it normally? (or in async lingo, rejecting the promise where it happens, and then catching that rejection normally in async try/catch.)

When this error happens, exactly where does it happen, which code generates it? It would be better to catch and handle the error right there.

test/source/mock/lib/api.ts
```diff
@@ -225,12 +225,17 @@ export class Api<REQ, RES> {
...
+    if (/^https:\/\/www\.google.com\/robots\.txt/.test(data.toString())) {
+      response.writeHead(302, { 'Location': data.toString() });
+    } else {
+      const chunkSize = 100 * 1024;
+      for (let i = 0; i < data.length; i += chunkSize) {
+        const chunk = data.slice(i, i + chunkSize);
+        response.write(chunk);
+        if (i > 0) {
+          await Util.sleep(this.throttleChunkMsDownload / 1000);
+        }
```
tomholub, line 238: We can probably come up with a cleaner solution here. I suppose, in mocked API, we shouldn't need to deal with the `robots.txt` location at all. We already have a different destination for oauth2 dialog in mock, and we can also have a different place instead of google/robots.txt for the redirect, something that maybe wouldn't need this code here.

However, I haven't fully examined how this is used, therefore I'm not 100% sure if my suggestion is appropriate.

extension/js/common/oauth2/oauth2_finish.ts
```diff
@@ -0,0 +1,4 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
+import { OAuth2 } from './oauth2.js';
+
+OAuth2.finishAuth(window.location.href);
```
tomholub, line 4: I suppose this is where the `lastError` may happen, as showing in

```ts
  public static finishAuth = (url: string) => {
    const views = chrome.extension.getViews();
    // Loop through existing extension views and excute any stored callbacks.
    for (const view of views) {
      if ((view as any)['oauth-callback']) {
        (view as any)['oauth-callback'](url); // tslint:disable-line:no-unsafe-any
        (view as any)['oauth-login-finished'] = true;
      }
    }
    chrome.runtime.lastError = undefined;
    window.open('', '_self', '');
    window.close();
  };
```

It's a rather creative solution :-) but you don't have to do this. To communicate between views, you can use browser messages (`BrowserMsg` class). In this case, you can send a broadcast message `BrowserMsg.send.authWindowResult('broadcast', { any JSON serializable object you need })`. You'll need to add this `authWindowResult` to `BrowserMsg`, and then also add appropriate listeners in the pages that need this, instead of setting 

```ts
    (window.top as any)['oauth-login-finished'] = false;
    (window.top as any)['oauth-callback'] = callback;
```

That will make it more consistent with the rest of the codebase.

Now that you did it, I wonder if there is some advantage to what you've done compared to what we're doing, but that would be a discussion for another time.

ioanmo226:
It still works fine except oauth popup window doesn't close and show some javascript error in console.

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -200,72 +189,39 @@ export class GoogleAuth {
...
-            }
-            return { acctEmail: email, result: 'Success', id_token };
-          }
-          return { acctEmail, result: 'Error', error: `Google auth result was 'Success' but no auth code`, id_token: undefined };
+      const responseUrl = await GoogleAuth.oauthLogin(authUrl);
+      if (!responseUrl) {
+        return { acctEmail, result: 'Denied', error: 'Invalid response url', id_token: undefined };
+      }
+      const url = new URL(responseUrl!);
+      const allowedScopes = this.getParameterByName(url.search, 'scope');
```
tomholub, line 206: Nice, so scopes can be obtained directly from the result.

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -200,72 +189,39 @@ export class GoogleAuth {
...
+          // Try to get auth token to let login authorization be granted
+          await GoogleAuth.googleAuthGetTokens(code);
         }
-        return { acctEmail, result, error: error ? error : '(no error provided)', id_token: undefined };
       }
-      await Ui.time.sleep(250);
+      if (!code) {
+        return { acctEmail, result: 'Denied', error: "Google auth result was 'Success' but no auth code", id_token: undefined };
+      }
+      const { id_token } = save ? await GoogleAuth.retrieveAndSaveAuthToken(code, allowedScopes?.split(' ') ?? []) : await GoogleAuth.googleAuthGetTokens(code);
```
tomholub, line 217: In how many situations is this `save` used vs where is it not used? To try to have method with fewer disparate responsibilities, it may be worthwhile to remove it from here and explicitly save it where needed. But this depends how many places are we talking about. Since this is a private method, maybe it's just one place it's being called from? Then you could add the `if / save` there.

tomholub (changes requested):
Looking good, and very happy we have a functioning solution.

Have you tried Firefox, and also Gmail live tests?

We just need to beat this code into submission through the PR review process in an attempt to simplify or increase clarity. I'll wait for @rrrooommmaaa review as well to also get his perspective.

tomholub:
> It still works fine except oauth popup window doesn't close and show some javascript error in console.

Got it - that would also need to be addressed before merging.

ioanmo226:
Gmail live tests are already broken in master branch and need to be fixed
https://github.com/FlowCrypt/flowcrypt-browser/issues/4594

> Have you tried Gmail live tests?

tomholub:
> Gmail live tests are already broken in master branch and need to be fixed

I think we could do that in a separate PR after this PR merges. So you don't have to fix that here, since it was already broken prior.

extension/js/common/oauth2/oauth2_inject.ts
```diff
@@ -0,0 +1,10 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
+
+// Decleared win variable to avoid `Type 'string' is not assignable to type 'Location | (string & Location)'.ts(2322)` error
```
rrrooommmaaa, line 3: typo. Should be `declared`

extension/js/common/oauth2/oauth2_finish.ts
tomholub: @rrrooommmaaa wdyt?

test/source/mock/lib/api.ts
rrrooommmaaa: Yes, looks like `https://www.google.com/robots.txt` is actually accessed (queried) in mock tests which we try to keep isolated.
This is not necessary as we have the code to replace URLs in mock manifests.
We're replacing `https://mail.google.com` with `https://gmail.localhost:8001` in https://github.com/FlowCrypt/flowcrypt-browser/blob/59bc0b3890294f05bd41f1bcf0eef6dbd25294ca/tooling/build-types-and-manifests.ts#L128
We can also replace `https://www.google.com` with `https://google.localhost:8001` (different host but still resolved to 127.0.0.1)

extension/js/common/oauth2/oauth2_inject.ts
```diff
@@ -0,0 +1,10 @@
...
+const win: Window = window;
+
+// Redirect back to the extension itself so that we have priveledged
```
rrrooommmaaa, line 6: Have you tried to add `frame-ancestors https://www.google.com/robots.txt 'self'` (or the like) to manifest's `"content_security_policy"`. I guess it'll let you oauth2_finish right away.

We have this line `frame-ancestors https://mail.google.com 'self'` already and it is also overriden for mocks with https://github.com/FlowCrypt/flowcrypt-browser/blob/59bc0b3890294f05bd41f1bcf0eef6dbd25294ca/tooling/build-types-and-manifests.ts#L128

extension/js/common/oauth2/oauth2_finish.ts
rrrooommmaaa: > If we use `BrowserMsg.send` in `oauth2.ts` and add listener in `GoogleAuth` we won't benefit callback(promise) feature.

Why, can't the message handler resolve a promise?

ioanmo226: Let's talk about above code and let's say we use `BrowserMsg.send` to communicate between views. 
We might have `GoogleAuth.didGetOAuthResult` listener or something like that.
In `getOAuthResult` function, we launch oauth2 dialog by calling `GoogleAuth.oauthLogin` and we can't continue logic from here as we don't have success/failure callback.
Instead we have to add below logic into new listener method(GoogleAuth.didGetOAuthResult).

This way, google login logic will be seperate into `getOAuthResult` and `didGetOAuthResult` and it seems hard to read (at least for me)

```ts
if (!responseUrl) {
  return { acctEmail, result: 'Denied', error: 'Invalid response url', id_token: undefined };
}
const url = new URL(responseUrl!);
const allowedScopes = this.getParameterByName(url.search, 'scope');
const code = this.getParameterByName(url.search, 'code') ?? '';
if (!allowedScopes?.includes(this.OAUTH.scopes.compose) || !allowedScopes?.includes(this.OAUTH.scopes.modify)) {
...
```

extension/js/common/oauth2/oauth2_inject.ts
ioanmo226: Sorry but not sure exactly what you mean.
AFAIK, frame-ancestors directive specifies valid parents that may embed a page(using iframe or something like that).
I'm not sure how it would let oauth2_finish right away.

rrrooommmaaa: What was the error preventing finishing OAuth2 directly from `robots.txt`?
There is a number of security options, maybe enabling any of them can help.

extension/js/common/oauth2/oauth2_finish.ts
rrrooommmaaa: @ioanmo226 You're setting up a listener inside `launchWebAuthFlow`. Can you also set up a listener to a message that would be sent by `robots.txt`?

extension/js/common/oauth2/oauth2_inject.ts
rrrooommmaaa: What I am trying to understand: what prevents us sending a message to close the oauth2 dialog directly from `robots.txt`?

rrrooommmaaa: Like here:
```
// Even accepts arrays of patterns!
chrome.tabs.query({url: "*://*.example.com/*"}, function(tabs) {
  tabs.forEach(function(tab) {
    chrome.tabs.sendMessage(tab.id, message);
  });
});
```
https://stackoverflow.com/a/28832962

ioanmo226: `robots.txt` page url contains redirect_url which contains `authorization_code`, `scopes`, etc.
Our final goal is to get this url when we call `GoogleAuth.oauthLogin`.
But without help of content script injection (`oauth2_inject.ts` for our case), we can't get that url.

rrrooommmaaa: Yes, we get this data in `oauth2_inject.ts`. Can we send it to the tab that awaits it (has set up the listener for it?)

ioanmo226: Can `robots.txt` page access `chrome.tabs` or `chrome.extension.getViews`?
I don't think it would.
> Like here:

ioanmo226: image

extension/js/common/oauth2/oauth2_finish.ts
ioanmo226: @tomholub , @rrrooommmaaa I tried to migrate to `BrowserMsg.send` and here's a commit. (Below code is temporary, like using localStorage to save datas)
https://github.com/FlowCrypt/flowcrypt-browser/pull/4665/commits/b682f9af7338628724bce4c4068fd200c9f4b290

To me, updated code looks super messy and hard to understand.
I prefer previous way using callback method and want to revert change.

Let me know what you think.

extension/js/common/oauth2/oauth2.ts
```diff
@@ -0,0 +1,52 @@
...
+    const leftOffset = Math.floor((screenWidth / 2) - (adaptiveWidth / 2) + (availLeft || 0));
+    const topOffset = Math.floor((screenHeight / 2) - (adaptiveHeight / 2) + (availTop || 0));
+    const oauthWin = await windowsCreate({ url, left: leftOffset, top: topOffset, height: adaptiveHeight, width: adaptiveWidth, type: 'popup' });
+    if (!oauthWin || !oauthWin.tabs || !oauthWin.tabs.length || !oauthWin.id) {
+      return { error: 'No oauth window returned after initiating it' };
+    }
+    const tabId = oauthWin?.tabs && oauthWin.tabs[0].id;
+    return await new Promise((resolve) => {
+      BrowserMsg.addListener('auth_window_result', async (result: Bm.AuthWindowResult) => {
+        resolve(result);
```
rrrooommmaaa, line 27: I meant something like this. Resolve the promise when the message is received. Wouldn't this work?

ioanmo226: Ohh.. Yeah.
It looks great. It's exactly what I wanted to achieve.
Thank you

extension/js/common/oauth2/oauth2_finish.ts
ioanmo226: Will mark this as resolved as @rrrooommmaaa pushed fix

extension/js/common/oauth2/oauth2.ts
ioanmo226: This method has problem though which will only accept `auth_window_result` event once.
Like when user wants to change email in `Force Google Account email change`(in experiemental page), event doesn't get fired.
Finding solution now.

rrrooommmaaa: Yes, we need to remove both listeners when resolving the promise

ioanmo226: I thought experiemental `Force Google Account email change` feature doesn't work because event only fires once but that seems wrong.
It seems like event doesn't get fired in experimental page because it's wrapped in Iframe.

extension/js/common/api/email-provider/gmail/google-auth.ts
ioanmo226: IMHO I think better to leave this as it is for now.
Only one place has `save` as false and as we also  save `scopes` in `googleAuthSaveTokens`, better to refactor this once we migrate to `load allowed scopes from google api` feature.

https://github.com/FlowCrypt/flowcrypt-browser/blob/37d466bf471bde88eae19a2ccd17f1bf74ce87d2/extension/js/common/api/email-provider/gmail/google-auth.ts#L295

image

extension/js/common/oauth2/oauth2.ts
ioanmo226: @tomholub , @rrrooommmaaa Does anyone know why we use `BrowserMsg.send.bg.await.reconnectAcctAuthPopup` instead of using `GoogleAuth.newAuthPopup` directly?
Seems like UI test (enable google contact search test) fails because of this reason.

https://github.com/FlowCrypt/flowcrypt-browser/blob/37d466bf471bde88eae19a2ccd17f1bf74ce87d2/extension/chrome/elements/compose-modules/compose-recipients-module.ts#L725

tomholub: I don't see any good reason. Maybe some reason used to exist, but we refactored it away and it no longer makes sense. Feel free to remove `reconnectAcctAuthPopup`

ioanmo226: ok

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -119,32 +120,12 @@ export class GoogleAuth {
     if (save || !scopes) { // if tokens will be saved (meaning also scopes should be pulled from storage) or if no scopes supplied
       scopes = await GoogleAuth.apiGoogleAuthPopupPrepareAuthReqScopes(acctEmail, scopes || GoogleAuth.defaultScopes());
     }
+    // below codes are all temporary codes and need cleanup if we want to proceed this way
```
tomholub, line 123: highlighting

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -200,72 +181,40 @@ export class GoogleAuth {
...
-      const [oauth] = await tabsQuery({ windowId });
-      if (oauth?.title && oauth.title.includes(GoogleAuth.OAUTH.state_header) && !GoogleAuth.isUrl(oauth.title) && !GoogleAuth.isForwarding(oauth.title)) {
-        const { result, error, code, csrf } = GoogleAuth.processOauthResTitle(oauth.title);
-        if (error === 'access_denied') {
-          return { acctEmail, result: 'Denied', error, id_token: undefined }; // sometimes it was coming in as {"result":"Error","error":"access_denied"}
-        }
-        if (result === 'Success') {
-          if (!csrf || csrf !== csrfToken) {
-            return { acctEmail, result: 'Error', error: `Wrong oauth CSRF token. Please try again.`, id_token: undefined };
-          }
```
tomholub, line 255: Has the CSRF check moved elsewhere? I see CSRF token getting created with the request but don't see the check anymore, it may be vulnerable to CSRF attacks.

tomholub (commented):
This looks good to me. Will wait for Roman's approval / comments after the update.

extension/js/common/api/email-provider/gmail/google-auth.ts
tomholub: Ok, please link this there to discuss further. Now I wonder, do we really need to do `save: false` in that situation? Maybe it wouldn't hurt to save there also?

tomholub:
Is firefox behavior fixed now?

ioanmo226:
Yes. It's working fine now

extension/js/common/api/email-provider/gmail/google-auth.ts
tomholub: Would be good to have a test for this, too. Since we use mocked tests, it should be straight forward. Could leave the test for another issue, but the checking needs to be fixed.

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -120,11 +121,10 @@ export class GoogleAuth {
     if (save || !scopes) { // if tokens will be saved (meaning also scopes should be pulled from storage) or if no scopes supplied
       scopes = await GoogleAuth.apiGoogleAuthPopupPrepareAuthReqScopes(acctEmail, scopes || GoogleAuth.defaultScopes());
     }
-    // below codes are all temporary codes and need cleanup if we want to proceed this way
     const authRequest: AuthReq = { acctEmail, scopes, csrfToken: `csrf-${Api.randomFortyHexChars()}` };
     const authUrl = GoogleAuth.apiGoogleAuthCodeUrl(authRequest);
     const authWindowResult = await OAuth2.webAuthFlow(authUrl);
-    const authRes = await GoogleAuth.getAuthRes({ acctEmail, save, authWindowResult });
+    const authRes = await GoogleAuth.getAuthRes({ acctEmail, save, state: GoogleAuth.apiGoogleAuthStatePack(authRequest), authWindowResult });
```
tomholub, line 127: It seems you are probably creating `state` twice, once for request and once for verification?

It would be more natural to have a method `GoogleAuth.newAuthRequest(...) and then `state` would be one of the values in `AuthReq`. Then `getAuthRes` would have `expectedState: authRequest.state`.

It would better communicate the intetion for csrf protection, which is "I decide the value in the beginning and it must come back the same". Computing it twice obscures that intention.

extension/js/common/api/email-provider/gmail/google-auth.ts
```diff
@@ -181,33 +181,32 @@ export class GoogleAuth {
...
       if (authWindowResult.error) {
         return { acctEmail, result: 'Denied', error: authWindowResult.error, id_token: undefined };
       }
-      const url = new URL(authWindowResult.url!);
-      const allowedScopes = this.getParameterByName(url.search, 'scope');
-      const code = this.getParameterByName(url.search, 'code') ?? '';
+      const uncheckedUrlParams = Url.parse(['scope', 'code', 'state'], authWindowResult.url!);
+      const allowedScopes = Assert.urlParamRequire.string(uncheckedUrlParams, 'scope');
+      const code = Assert.urlParamRequire.optionalString(uncheckedUrlParams, 'code');
+      const receivedState = Assert.urlParamRequire.string(uncheckedUrlParams, 'state');
```
tomholub, line 196: `Assert.urlParamRequire` throws error text that could be confusing for users, as it was meant for parsing URLs that we created ourselves. You'll notice this when reading the error text in `abortAndRenderErrOnUrlParamValMismatch`. It's not well documented.

But it's good to check it here. You could put these 4 lines in a try/catch, and if they throw error, return `        return { acctEmail, result: 'Error', error: `Could not parse URL returned from Google`, id_token: undefined };`

tomholub (commented):
Looking good, see last few comments from my side :+1:

extension/js/common/api/email-provider/gmail/google-auth.ts
ioanmo226: UI test added

extension/js/common/api/email-provider/gmail/google-auth.ts
ioanmo226: Ok, Done.
I agree with you. but not sure exactly.

ioanmo226:
By the way, @tomholub How are you gonna handle 2 other PRs(branches)?
Just close them and leave branches or what?

tomholub:
yeah we'll just close them. once in a few years i clean up stale branches

extension/js/common/assert.ts
```diff
@@ -2,18 +2,19 @@
...
+import { KeyInfoWithIdentity, KeyUtil } from './core/crypto/key.js';
 import { isFesUsed } from './helpers.js';
+import { Lang } from './lang.js';
+import { Catch } from './platform/catch.js';
+import { AcctStore } from './platform/store/acct-store.js';
+import { KeyStore } from './platform/store/key-store.js';
+import { Xss } from './platform/xss.js';
+import { Settings } from './settings.js';
 
+export class AssertError extends Error { }
```
tomholub, line 17: `AssertError` should inherit from `UnreportableError` so we don't end up having these reported to our error collection endpoint.

test/source/test.ts
```diff
@@ -129,6 +129,8 @@ ava.default.after.always('evaluate Catch.reportErr errors', async t => {
     ].includes(e.message))
     // below for test "user4@standardsubdomainfes.test:8001 - PWD encrypted message with FES web portal - a send fails with gateway update error"
     .filter(e => !e.message.includes('Test error'))
+    // below for test "setup - invalid csrf token returns error on gmail login"
+    .filter(e => !e.message.includes('failed to log into google in newGoogleAcctAuthPromptThenAlertOrForward'))
```
tomholub, line 133: That's what I meant. I don't want this reported, therefore this should be removed.

tomholub:
Thanks for the added test. After my last round of comments I expect to merge this. Since this is important functionality, if @rrrooommmaaa has any comments after that, we should still file issues for them and resolve them in the same version before we do a release.

test/source/test.ts
ioanmo226: Even though, `AssertError` inherits from `UnreportableError`, above code is necessary because of this line.
It returns result as error and so javascript error is occurred
```ts
if (receivedState !== expectedState) {
  return { acctEmail, result: 'Error', error: `Wrong oauth CSRF token. Please try again.`, id_token: undefined };
}
```

tomholub: But it affects other code, as that error is thrown in many other places. We are not testing that, but there are other places using `Assert.` that will now newly be throwing reportable errors, which is not intended.

tomholub: Anyway, please help me remove these two lines in test.ts. Rest looks good.

ioanmo226: Yeah. already inherited `AssertError` from `UnreportableError`.
I think you misunderstood me.
Those above 2 lines are not needed because of AssertError not inherit from UnreportableError
It's needed because we check `crsftoken`(state) and return error if it doesn't match.

ioanmo226: This line
```ts
return { acctEmail, result: 'Error', error: `Wrong oauth CSRF token. Please try again.`, id_token: undefined };
```

tomholub: I see. I would not want that reported, either. It can sometimes happen maybe if user uses the back and forward button on the host page and then re-confirms the old oauth form, that kind of thing. Could this be selectively not reported?

ioanmo226: Done

tomholub (approved):

tomholub:
@ioanmo226 thank you :1st_place_medal: this was pending for years, but now it has become a real problem when Google decided to pull the plug on OOB.

ioanmo226:
@tomholub Thank you for your help too :)

PR #4672, opened 2022-09-08
consider the build type when referring enterprise users to flowcrypt support

### LINKED ISSUE #4669: don't ever refer users to human@flowcrypt.com for enterprise build
In enterprise deployments, users are supposed to be directed to internal company Help Desk instead of our support. We mostly do this, when client configuration is available, we will direct user to internal help desk:

image

But in some situations, we have to prompt user to write someone before they logged in, and so we don't know the client configuration.

Here's an example:

image

Currently it's telling enterprise users to write to our support which is wrong.

I'd like to do the following change: on top of current logic, I would like the user to _never_ be prompted to write to our support email (and instead be directed to their Help Desk) if this is an `enterprise` build. You can use the `const FLAVOR` value in `const.ts` to differentiate this.

martgil:
This PR will add checking if a build version is for enterprise users. That will help the software recognize whether to refer the user to their respective helpdesk or to human@flowcrypt.com.

close #4669

- Tests added or updated

### To be filled by reviewers


test/source/tests/setup.ts
```diff
@@ -430,6 +431,16 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
         { isSavePassphraseChecked: false, isSavePassphraseHidden: false });
     }));
 
+    ava.default('setup - enterprise users should be redirected to their help desk when an error occured', testWithBrowser(undefined, async (t, browser) => {
+      if (testVariant === 'ENTERPRISE-MOCK') {
+        const acctEmail = 'flowcrypt.compatibility@gmail.com';
+        const settingsPage = await browser.newPage(t, TestUrls.extensionSettings());
+        const oauthPopup = await browser.newPageTriggeredBy(t, () => settingsPage.waitAndClick('@action-connect-to-gmail'));
+        await OauthPageRecipe.mock(t, oauthPopup, acctEmail, 'login_with_invalid_state');
+        settingsPage.waitForContent('@container-error-modal-text', 'If this happens again, please contact your Help Desk');
```
tomholub, line 440: You can have if/else only on this statement. For enterprise, test for this sentence, for consumer, test the other one. Plus, missing an `await`.

tomholub:
Also here:

```html
  <p>The code running FlowCrypt is publicly available and anyone can verify the underlying security. Visit <a href="https://github.com/FlowCrypt/flowcrypt-browser"
      target="_blank">FlowCrypt Sources Repository</a> to learn more or inspect the code. Alternatively, please ask us
    directly at human@flowcrypt.com</p>
```

```html
<p>Email human@flowcrypt.com if you need any help.</p>
```

```ts
} else {
  details.text('Unknown reason. Write human@flowcrypt.com if you need help.');
}
```

```html
      <div class="wide line">
        Email human@flowcrypt.com if you have any questions.
      </div>
```

```html
<div class="line">Please write human@flowcrypt.com if you have any questions or concerns.</div>
```

```ts
        await Ui.modal.error('There was an error sending message. Our direct email is human@flowcrypt.com');
```

```ts
      await Ui.modal.error(`There was an error sending message. Our direct email is human@flowcrypt.com\n\n${ApiErr.eli5(e)}`);
```

```ts
$('body').text('Thank you. Feel free to reach out to human@flowcrypt.com in you need assistance.');
```

You don't have to test all of these. Please restructure the sentence so you can reuse Lang that already makes the decision.

martgil:
not ready yet

test/source/tests/setup.ts
martgil: Yes, sir... Also, I finished updating each sentence dynamically to what is appropriate. Thank you!

extension/chrome/settings/modules/auth_denied.htm
```diff
@@ -10,8 +30,14 @@ <h1>Explaining FlowCrypt webmail permissions</h1>
...
 <div class="line">
   Nobody, FlowCrypt developers included, is able to access these permissions, they are stored privately in your
-  browser. Learn more about <a href="#" data-swal-page="/chrome/texts/email_provider_auth_security.htm">email
+  browser. Learn more about <a href="/chrome/texts/email_provider_auth_security.htm">email
     provider access and security</a>.
 </div>
-<div class="line">Please write human@flowcrypt.com if you have any questions or concerns.</div>
-<div class="line">Would you like to proceed to the permissions screen again?</div>
\ No newline at end of file
+<div class="line ask_support_assistance_container">Please write human@flowcrypt.com if you have any questions or concerns.</div>
```
tomholub, line 36: If it's being filled with javascript, then the tag should be empty - it will be less confusing that way.

extension/chrome/settings/modules/auth_denied.ts
```diff
@@ -0,0 +1,12 @@
...
+'use strict';
+
+import { Lang } from '../../../js/common/lang.js';
+import { isEnterpriseBuildUsed } from '../../../js/common/helpers.js';
+
+const contactForSupportContainer = $('.ask_support_assistance_container');
+const htmlBody = $('body');
+
+contactForSupportContainer.text(Lang.general.contactIfNeedAssistance(isEnterpriseBuildUsed()));
+htmlBody.css('display', 'block');
```
tomholub, line 12: Please write it as a `View` like other pages, `View.run(class AuthDeniedView extends View {` for consistency

extension/chrome/settings/setup.htm
```diff
@@ -280,7 +280,7 @@ <h1>Set Up FlowCrypt</h1>
       <div class="wide line">
         <button class="button longer green action_diagnose_browser">test my browser</button>
       </div>
-      <div class="wide line">
+      <div class="wide line ask_support_assistance_container">
         Email human@flowcrypt.com if you have any questions.
       </div>
```
tomholub, line 285: Also here to be empty

extension/js/common/helpers.ts
```diff
@@ -11,6 +11,10 @@ import { ContactStore } from './platform/store/contact-store.js';
 import { KeyStore } from './platform/store/key-store.js';
 import { PassphraseStore } from './platform/store/passphrase-store.js';
 import { Bm } from './browser/browser-msg.js';
+import { FLAVOR } from './core/const.js';
+export const isEnterpriseBuildUsed = () => {
+  return FLAVOR === 'enterprise';
+};
```
tomholub, line 17: In a lot of the code you are mixing concepts: `FLAVOR === 'enterprise'` is being submitted into `isFesUsed`. While the result works, it's really confusing. These are two different things.

`isFesUsed` is the type of property that depends on user setup, and therefore it makes sense to pass it into methods like `contactIfNeedAssistance`.

`isEnterpriseBuildUsed` is however completely constant for the whole build, and there is not much benefit in passing it down. It will always be the same for the build, and therefore can be imported directly in `Lang`. You can just do `import { FLAVOR } from './core/const.js';` and `const isEnterpriseBuild = FLAVOR === 'enterprise';`. There is no need for it to be a method, either.

Then these methods like `contactIfNeedAssistance` will have `isFesUsed` which is only be set **if fes is actually used**. Inside of the method, there will be `isFesUsed || isEnterpriseBuild` and then everything is clear.

For lang sentences where we don't have the context to know if `isFesUsed`, only `isEnterpriseBuild` will be used, and then these sentences don't need any input arguments, it will decided directly in `lang` file.

extension/chrome/texts/email_provider_auth_security.ts
```diff
@@ -0,0 +1,12 @@
...
+'use strict';
+
+import { Lang } from '../../js/common/lang.js';
+import { isEnterpriseBuildUsed } from '../../js/common/helpers.js';
+
+const contactForSupportContainer = $('.ask_support_assistance_container');
+const htmlBody = $('body');
+
+contactForSupportContainer.text(Lang.general.contactMinimalSubsentence(isEnterpriseBuildUsed()));
+htmlBody.css('display', 'block');
```
tomholub, line 12: Write as a View

extension/chrome/texts/email_provider_auth_security.htm
```diff
@@ -1,12 +1,33 @@
...
+</head>
+
+<body style="display: none;">
+  <div class="line">To work properly, your FlowCrypt browser extension needs to connect to your email provider (Gmail or Outlook).
+    This allows you to encrypt and decrypt messages when you choose to.</div>
+  <div class="line">The plugin will receive so called API Tokens. These tokens, and all related privileges (to send and read messages)
+    will strictly stay within your own browser, stored locally in your computer for maximum security.</div>
+  <div class="line"><b>By design, the tokens cannot and will never be exposed to FlowCrypt servers or its developers.</b></div>
+  <div class="line">The code running FlowCrypt is publicly available and anyone can verify the underlying security. Visit <a href="https://github.com/FlowCrypt/flowcrypt-browser"
+      target="_blank">FlowCrypt Sources Repository</a> to learn more or inspect the code. Alternatively, please <span class="ask_support_assistance_container">ask us directly at human@flowcrypt.com</span>.</div>
```
tomholub, line 26: Let's make `ask_support_assistance_container` emtpy

tomholub (commented):
It gets the job done but it's very messy: arguments should be only used for their literal meaning, and not for other meanings.

extension/chrome/settings/modules/auth_denied.htm
martgil: understand sir... thank you

extension/chrome/settings/modules/auth_denied.ts
martgil: understand sir... thank you

extension/chrome/settings/setup.htm
martgil: understand sir... thank you

extension/chrome/texts/email_provider_auth_security.ts
martgil: Done -- thank you!

extension/chrome/texts/email_provider_auth_security.htm
martgil: understand sir... I won't be worrying about default values from now on since I realized extensions won't work without javascript. thank you

extension/js/common/settings.ts
```diff
@@ -305,10 +304,7 @@ export class Settings {
           window.location.href = Url.create('/chrome/settings/setup.htm', { acctEmail: response.acctEmail, idToken: response.id_token });
         }
       } else if (response.result === 'Denied' || response.result === 'Closed') {
-        const authDeniedHtml = await Api.ajax({ url: '/chrome/settings/modules/auth_denied.htm' }, Catch.stackTrace()) as string; // tslint:disable-line:no-direct-ajax
-        if (await Ui.modal.confirm(authDeniedHtml, true)) {
-          await Settings.newGoogleAcctAuthPromptThenAlertOrForward(settingsTabId, acctEmail, scopes);
-        }
+        await this.renderSubPage(acctEmail, settingsTabId!, '/chrome/settings/modules/auth_denied.htm', undefined, 450);
```
tomholub, line 307: You could simplify this further. `auth_denied` could be reverted back to a HTML snippet instead of a HTML page. You could remove the last sentence that is prompting user to write support could be appended here in this code, and then you can render it as a modal. So all of this code will be:

```ts
const authDeniedHtml = await Api.ajax({ url: '/chrome/settings/modules/auth_denied.htm' }, Catch.stackTrace()) as string; // tslint:disable-line:no-direct-ajax
await Ui.modal.info(`${authDeniedHtml}\n<div class="line">${Lang.general.contactIfNeedAssistance(), true)}`)
```

The current one uses more code and is a bit broken looking in comparison:

image

tomholub (approved):

extension/js/common/lang.ts
```diff
@@ -2,11 +2,15 @@
...
+import { FLAVOR } from './core/const.js';
+
 /* eslint-disable max-len */
 
+const isEnterpriseBuildUsed = FLAVOR === 'enterprise';
+
 const contactMinimalSubsentence = (isFesUsed: boolean) => isFesUsed ? 'contact your Help Desk' : 'write us at human@flowcrypt.com';
-const contactIfHappensAgain = (isFesUsed: boolean) => `If this happens again, please ${contactMinimalSubsentence(isFesUsed)}. `;
-const contactForSupportSubsentence = (isFesUsed: boolean) => isFesUsed ? 'please contact your Help Desk for support' : 'please write us at human@flowcrypt.com to fix it';
+const contactIfHappensAgain = (isFesUsed: boolean) => `If this happens again, please ${contactMinimalSubsentence(isFesUsed || isEnterpriseBuildUsed)}. `;
```
tomholub, line 12: Then wouldn't need to be here

extension/js/common/lang.ts
```diff
@@ -2,11 +2,15 @@
 
 'use strict';
 
+import { FLAVOR } from './core/const.js';
+
 /* eslint-disable max-len */
 
+const isEnterpriseBuildUsed = FLAVOR === 'enterprise';
+
 const contactMinimalSubsentence = (isFesUsed: boolean) => isFesUsed ? 'contact your Help Desk' : 'write us at human@flowcrypt.com';
```
tomholub, line 11: Here since it's the base method actually doing the work, `isFesUsed || isEnterpriseBuildUsed` would be appropriate.

It's always best to put the needed conditions in whatever method is the "choke point" of the implementation. A choke point is the place that will always be executed. In this implementation, `contactMinimalSubsentence` and `contactForSupportSubsentence` are the choke points, so you can put the logic there, and then it's not needed elsewhere.

extension/js/common/lang.ts
```diff
@@ -89,11 +93,11 @@ export const Lang = { // tslint:disable-line:variable-name
...
+    contactIfNeedAssistance: (isFesUsed?: boolean) => contactForSupportSubsentence(isFesUsed || isEnterpriseBuildUsed, 'if you need an assistance.'),
     somethingWentWrongTryAgain: (isFesUsed: boolean) => `Something went wrong, please try again. ${contactIfHappensAgain(isFesUsed)}`,
     contactForSupportSubsentence,
-    contactForSupportSentence: (isFesUsed: boolean) => isFesUsed ? 'Please contact your Help Desk for support.' : 'Please write us at human@flowcrypt.com to fix it.',
-    writeMeToFixIt: (isFesUsed: boolean) => isFesUsed ? 'Please contact your Help Desk for support.' : 'Please write us at human@flowcrypt.com to get this resolved.',
+    contactForSupportSentence: (isFesUsed?: boolean) => contactForSupportSubsentence(isFesUsed || isEnterpriseBuildUsed, 'for support.'),
+    writeMeToFixIt: (isFesUsed: boolean) => contactForSupportSubsentence(isFesUsed, 'to fix it.'),
     restartBrowserAndTryAgain: (isFesUsed: boolean) => `Unexpected error occured. Please restart your browser and try again. If this persists after a restart, ${contactForSupportSubsentence(isFesUsed)}.`,
```
tomholub, line 101: Then here you wouldn't need the `isEnterpriseBuildUsed` at all, if it's already implemented in the choke points. That would mean less repetition.

tomholub (changes requested):
It's looking much better, see a hopefully last round of comments.

martgil:
I had to add a flag 'swal-page-allow-js' for swal-pages to execute js because `swal.fire()` doesn't execute js unless they are in an iframe.

extension/js/common/helpers.ts
martgil: Thank you for these - it really helps.  I will make sure to pass an argument that is really for a specific parameter and tried not to mix the two concepts.

extension/js/common/lang.ts
martgil: Thank you - I will keep this and the other helpful comments in mind.

extension/js/common/lang.ts
martgil: thank you, sir. I think I got the main idea.

tomholub (approved):

PR #4682, opened 2022-09-13
#4613 Configure passphrase session length with in_memory_pass_phrase_session_length 

### LINKED ISSUE #4613: pass phrase session length should be configurable with client configuraiton
The returned JSON may contain `"in_memory_pass_phrase_session_length": 3600` for example. The number is an integer and to be interpreted as amount of seconds a pass phrase session will last. Currently hard-coded as 4 hours.

This mechanism applies only when browser extension is storing pass phrase in-memory. In other words, it does not affect the decision as to where the pass phrase will be stored, that was decided either by user or by `FORBID_STORING_PASS_PHRASE`. If the pass phrase is stored in storage, whatever the reason for it, this new client configuration will have no effect.

ioanmo226:
This PR configures passphrase session length with in_memory_pass_phrase_session_length 

close #4613 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/client-configuration.ts
+++ b/extension/js/common/client-configuration.ts
@@ -18,6 +18,7 @@ export type ClientConfigurationJson = {
   disallow_attester_search_for_domains?: string[],
   enforce_keygen_algo?: string,
   enforce_keygen_expire_months?: number,
+  in_memory_pass_phrase_session_length?: number;
 };
 
 /**
@@ -81,6 +82,18 @@ export class ClientConfiguration {
    */
   public getEnforcedKeygenExpirationMonths = (): number | undefined => {
     return this.clientConfigurationJson.enforce_keygen_expire_months;
+  };
+
+  /**
+   * pass phrase session length to be configurable with client configuraiton
+   * default 4 hours
+   */
+  public getInMemoryPassPhraseSessionLength = (): number | undefined => {
+    // in_memory_pass_phrase_session_length min: 1, max: Int max value
+    if (this.clientConfigurationJson.in_memory_pass_phrase_session_length) {
+      return Math.max(1, Math.min(this.clientConfigurationJson.in_memory_pass_phrase_session_length!, Number.MAX_VALUE)) * 1000;
+    }
+    return undefined;
   };
 
   // bools
--- a/extension/js/common/client-configuration.ts
+++ b/extension/js/common/client-configuration.ts
@@ -88,12 +88,13 @@ export class ClientConfiguration {
    * pass phrase session length to be configurable with client configuraiton
    * default 4 hours
    */
-  public getInMemoryPassPhraseSessionLength = (): number | undefined => {
-    // in_memory_pass_phrase_session_length min: 1, max: Int max value
+  public getInMemoryPassPhraseSessionExpiration = (): number => {
+    let expireIn = 4 * 60 * 60;
     if (this.clientConfigurationJson.in_memory_pass_phrase_session_length) {
-      return Math.max(1, Math.min(this.clientConfigurationJson.in_memory_pass_phrase_session_length!, Number.MAX_VALUE)) * 1000;
+      // in_memory_pass_phrase_session_length min: 1, max: Int max value
+      expireIn = Math.max(1, Math.min(this.clientConfigurationJson.in_memory_pass_phrase_session_length, Number.MAX_VALUE));
     }
-    return undefined;
+    return Date.now() + expireIn * 1000;
   };
 
   // bools
--- a/extension/js/common/client-configuration.ts
+++ b/extension/js/common/client-configuration.ts
@@ -88,13 +88,13 @@ export class ClientConfiguration {
    * pass phrase session length to be configurable with client configuraiton
    * default 4 hours
    */
-  public getInMemoryPassPhraseSessionExpiration = (): number => {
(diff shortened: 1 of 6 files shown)


test/source/tests/tooling/browser-recipe.ts
```diff
@@ -100,6 +100,15 @@ export class BrowserRecipe {
     return (result as { result: string }).result;
   };
 
+  public static getSavedPassPhrase = async (controllable: Controllable, acctEmail: string, longid: string): Promise<string> => {
```
rrrooommmaaa, line 103: Good.
However, `getSavedPassPhrase`  is abmiguous, as a pass phrase can be stored in the local storage too.
So this method should be named in a way that indicates the in-memory store is queried,
something like `getPassphraseFromInMemoryStore`

test/source/tests/setup.ts
```diff
@@ -787,6 +787,34 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+          passphrase,
+          longid,
+        }
+      });
+      const { cryptup_userpassphrasesessionlengthclientconfigurationflowcrypttest_rules: rules
+      } = await settingsPage.getFromLocalStorage([
+        'cryptup_userpassphrasesessionlengthclientconfigurationflowcrypttest_rules'
+      ]);
+      let savedPassphrase = await BrowserRecipe.getSavedPassPhrase(settingsPage, acctEmail, longid);
+      expect((rules as { in_memory_pass_phrase_session_length: number }).in_memory_pass_phrase_session_length).to.be.equal(60);
```
rrrooommmaaa, line 810: The test is correct.
Should we really wait for 60 seconds?
Can we set up expiration to 5 or 10 seconds, since you defined `min: 1` second?

extension/js/common/client-configuration.ts
```diff
@@ -83,6 +84,19 @@ export class ClientConfiguration {
...
+   * default 4 hours
+   */
+  public getInMemoryPassPhraseSessionExpiration = (): number => {
+    let expireIn = 4 * 60 * 60;
+    if (this.clientConfigurationJson.in_memory_pass_phrase_session_length) {
+      // in_memory_pass_phrase_session_length min: 1, max: Int max value
+      expireIn = Math.max(1, Math.min(this.clientConfigurationJson.in_memory_pass_phrase_session_length, Number.MAX_VALUE));
+    }
+    return Date.now() + expireIn * 1000;
+  };
```
tomholub, line 98: My feedback here is just a hunch. I feel a little uneasy about this method returning the actual resulting date. I don't know exactly what makes me worry about it, since obviously we are not storing this information anywhere, so time passed between when it was queried, and when it was used, is no problem.

I guess it's a real nit-pick, but would it be a big change / would the code be messier if this method returned just amount of seconds, and the `Date.now() + expirationSeconds` would happen elsewhere?

test/source/tests/setup.ts
tomholub: I was also thinking this

tomholub: Or in addition to the 60 second timer (which would be a fallback) we could have a `Date.now() + expirationSeconds` timeout - just a single run, that will clear it immediately. That way, if you set 1 second timeout, you could test two seconds later.

extension/js/common/client-configuration.ts
rrrooommmaaa: Looks like we (I) implemented `ExpirationCache.set` in a way that it accepts the date, not an offset, that's the reason

tomholub: I'd be quite content if this method was called as `Date.now() + clientConfiguration.getInMemoryPassPhraseSessionExpiration()` wherever it's used. Then it will be a bit more obvious.

test/source/tests/setup.ts
rrrooommmaaa: > Maybe current app code has a 60 second timer (setInterval) for removing pass phrases? We could lower the interval, and have a shorter test.

For removing -- maybe. But there is also an on-demand (post-search) removal in `ExpirationCache.get`.
And from the caller's point of view these two cases can't be distinguished.

tomholub: Could that be made pre-search instead? Then the test could test immediately post expiry.

rrrooommmaaa: I meant -- after search but before (validating and) returning the result.
So we can test immediately after expiry, no changes should be needed.

extension/js/common/client-configuration.ts
ioanmo226: Hmm.. then we need to also accept email parameter in `ExpirationCache` constructor (or in `ExpirationCache.set ` method).
That seems a bit annoying to me.

tomholub: I meant just like this

image

And then also rename `getInMemoryPassPhraseSessionExpiration` to `getInMemoryPassPhraseSessionExpirationMs` for clarity.

ioanmo226: Aha, got it
Just changed

tomholub (commented):
careful - the `* 1000` was lost in this commit

ioanmo226:
Noticed it right after commit & added another commit :)

tomholub (approved):
:+1:

Issue #4626, opened 2022-08-22
BrowserMsg pgpHashChallengeAnswer can be removed


tomholub:
this method

```ts
  public static sha256UtfStr = async (string: string) => {
    return opgp.util.Uint8Array_to_hex(await opgp.crypto.hash.digest(opgp.enums.hash.sha256, Buf.fromUtfStr(string)));
  };
```

can be updated to use `window.crypto.subtle.digest('SHA-256', ...));` instead of `hash.sha256`, and `Buf.toHexStr` instead of `Uint8Array_to_hex`.

Then all `BrowserMsg` parts related to `pgpHashChallengeAnswer` can be removed, and instead of making this call through background page, code cal just call `PgpHash.challengeAnswer` directly. Which will no longer have `opgp` dependency.

ioanmo226:
@tomholub 
Hmmm.... I don't see any `pgpHashChallengeAnswer` calls.
I can only see `pgpHashChallengeAnswer` listeners but can't see any triggers in project.
Could you confirm if it has been removed or what?

tomholub:
these listeners can be removed. 

does nothing call these listeners?

ioanmo226:
I couldn't find any.

tomholub:
You're right. Just remove the listeners then 👍

PR #4689, opened 2022-09-15
#4594 fix live test

### LINKED ISSUE #4594: Fix broken live gmail test

### LINKED ISSUE #4688: broken mocked test
```
  ✖ compose - reply all - from === acctEmail [ALL RETRIES FAILED for compose - reply all - from === acctEmail] 04:40
    ℹ (attempt 1 of 3) Retrying: Error: waiting failed: Elements did not appear: [data-test="replied-to"] 04:40
    ℹ (attempt 2 of 3) Retrying: Error: waiting failed: Elements did not appear: [data-test="replied-to"] 04:40
    ℹ (attempt 3 of 3) Failed:   /home/semaphore/git/flowcrypt-browser/test/source/browser/controllable.ts:91 04:40
          throw Error(`waiting failed: Elements did not appear: ${selectors.join(',')}`); 04:40
                ^ 04:40
      Error: waiting failed: Elements did not appear: [data-test="replied-to"] 04:40
          at ControllablePage.ControllableBase.waitAny (/home/semaphore/git/flowcrypt-browser/test/source/browser/controllable.ts:91:11) 04:40
          at ControllablePage.ControllableBase.waitForContent (/home/semaphore/git/flowcrypt-browser/test/source/browser/controllable.ts:328:5) 04:40
          at <anonymous> (/home/semaphore/git/flowcrypt-browser/test/source/tests/compose.ts:1692:7) 04:40
          at <anonymous> (/home/semaphore/git/flowcrypt-browser/test/source/test.ts:76:7) 04:40
          at <anonymous> (/home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:144:61) 04:40
          at BrowserPool.BrowserPool.withNewBrowserTimeoutAndRetry (/home/semaphore/git/flowcrypt-browser/test/source/browser/browser-pool.ts:144:11) 04:40
          at <anonymous> (/home/semaphore/git/flowcrypt-browser/test/source/test.ts:71:5) 04:40
          at <async> waitAny (/test/source/browser/controllable.ts:71:19) 04:40
          at <async> waitForContent (/test/source/browser/controllable.ts:327:26) 04:40
          at <async> <anonymous> (/test/source/tests/compose.ts:1680:93) 04:40
          at <async> <anonymous> (../test/source/test.ts:71:53) 04:40
          at <async> <anonymous> (/test/source/browser/browser-pool.ts:144:49)
```

example https://flowcrypt.semaphoreci.com/jobs/0af49698-7502-499a-ac53-ff80d59192a5

Likely clicking the previous click a little too early.

ioanmo226:
This PR fixed live test

close #4594 // if this PR closes an issue
close #4688 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -722,7 +722,8 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       .append('<div class="allow-google-contact-search" data-test="action-auth-with-contacts-scope"><img src="/img/svgs/gmail.svg" />Enable Google Contact Search</div>') // xss-direct
       .find('.allow-google-contact-search')
       .click(this.view.setHandler(async () => {
-        const authResult = await GoogleAuth.newAuthPopup({ acctEmail: this.view.acctEmail, scopes: GoogleAuth.defaultScopes('contacts') });
+        // Need to use BrowserMsg.send.bg because chrome.windows is undefined in gmail page
+        const authResult = await BrowserMsg.send.bg.await.reconnectAcctAuthPopup({ acctEmail: this.view.acctEmail, scopes: GoogleAuth.defaultScopes('contacts') });
         if (authResult.result === 'Success') {
           this.googleContactsSearchEnabled = true;
           this.hideContacts();
--- a/extension/js/common/core/crypto/pgp/pgp-hash.ts
+++ b/extension/js/common/core/crypto/pgp/pgp-hash.ts
@@ -15,13 +15,10 @@ export class PgpHash {
     return await PgpHash.cryptoHashSha256Loop(answer);
   };
 
-  private static cryptoHashSha256Loop = async (string: string, times = 100) => {
-    console.log(`cryptoHashSha256Loop start`);
-    console.log(`first try ${(await PgpHash.sha256UtfStr(string))}`);
+  private static cryptoHashSha256Loop = async (string: string, times = 10000) => {
     for (let i = 0; i < times; i++) {
       string = await PgpHash.sha256UtfStr(string);
     }
-    console.log(`cryptoHashSha256Loop end ${string}`);
     return string;
   };
 
--- a/extension/js/common/core/crypto/pgp/pgp-hash.ts
+++ b/extension/js/common/core/crypto/pgp/pgp-hash.ts
@@ -15,9 +15,9 @@ export class PgpHash {
     return await PgpHash.cryptoHashSha256Loop(answer);
   };
 
-  private static cryptoHashSha256Loop = async (string: string, times = 100000) => {
+  private static cryptoHashSha256Loop = async (string: string, times = 100) => {
     console.log(`cryptoHashSha256Loop start`);
-    console.log(`first try (await PgpHash.sha256UtfStr(string)`);
+    console.log(`first try ${(await PgpHash.sha256UtfStr(string))}`);
     for (let i = 0; i < times; i++) {
       string = await PgpHash.sha256UtfStr(string);
     }
--- a/extension/js/common/core/crypto/pgp/pgp-hash.ts
+++ b/extension/js/common/core/crypto/pgp/pgp-hash.ts
@@ -8,7 +8,6 @@ export class PgpHash {
 
   public static sha256UtfStr = async (string: string) => {
     const digest = await window.crypto.subtle.digest('SHA-256', Buf.fromUtfStr(string));
-    console.log(`digest: ${digest}`);
     return Buf.fromUint8(new Uint8Array(digest)).toHexStr(false);
   };
 
@@ -18,10 +17,11 @@ export class PgpHash {
 
   private static cryptoHashSha256Loop = async (string: string, times = 100000) => {
     console.log(`cryptoHashSha256Loop start`);
+    console.log(`first try (await PgpHash.sha256UtfStr(string)`);
     for (let i = 0; i < times; i++) {
       string = await PgpHash.sha256UtfStr(string);
     }
-    console.log(`cryptoHashSha256Loop start ${string}`);
+    console.log(`cryptoHashSha256Loop end ${string}`);
     return string;
   };
 
--- a/extension/js/common/core/crypto/pgp/pgp-hash.ts
+++ b/extension/js/common/core/crypto/pgp/pgp-hash.ts
@@ -15,7 +15,7 @@ export class PgpHash {
     return await PgpHash.cryptoHashSha256Loop(answer);
   };
 
-  private static cryptoHashSha256Loop = async (string: string, times = 10000) => {
+  private static cryptoHashSha256Loop = async (string: string, times = 100000) => {
     for (let i = 0; i < times; i++) {
       string = await PgpHash.sha256UtfStr(string);
     }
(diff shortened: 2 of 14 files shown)


extension/js/common/core/crypto/pgp/pgp-hash.ts
```diff
@@ -16,9 +17,11 @@ export class PgpHash {
   };
 
   private static cryptoHashSha256Loop = async (string: string, times = 100000) => {
+    console.log(`cryptoHashSha256Loop start`);
     for (let i = 0; i < times; i++) {
       string = await PgpHash.sha256UtfStr(string);
     }
+    console.log(`cryptoHashSha256Loop start ${string}`);
```
ioanmo226: Regarding #4688, why do we loop 100000 times and try to get 100000th PgpHashed string? Is it because of security?
It seems too heavy operation.

@tomholub Why to we loop

ioanmo226: As I checked deeply, #4688 issue happens because it takes too long time (maybe 30-60secs in semaphoreci) to get `cryptoHashSha256Loop` value as it tries to run `PgpHash.sha256UtfStr` function 100000 times.

tomholub: It's called iterated hashing, and makes it somewhat harder to brute-force passwords in case they are not so strong. But what you found is interesting. In normal use, it takes about a second or so. You're saying that when run in puppeteer, the result is MUCH slower. Which is a problem for testing.

tomholub: Ah - looks like we run into this following / due to https://github.com/FlowCrypt/flowcrypt-browser/issues/4626 , haven't we?

ioanmo226: It's not slow in puppeteer. if i run in local it runs in a few seconds.
But in semaphoreci, it took longer than 20 secs.

ioanmo226: Correct

tomholub: Can you reproduce this slowness when testing manually without puppeteer? (you will need a non-enterprise account to test on)

ioanmo226: maybe window.crypto is slower than opgp.crypto in semaphoreci

tomholub: Anyway, given that we'd like to do a release, we could consider reverting #4626, which should fix it, and then having a closer look after a release.

ioanmo226: It takes about 3-5 seconds to send password protected emails (which uses this method)
> Can you reproduce this slowness when testing manually without puppeteer? (you will need a non-enterprise account to test on)

ioanmo226: maybe we can increase delay (waiting time) specially for this failed test?

tomholub: Ok. So it's slow in semaphore ci, but not horribly slow locally, even when using puppeteer. And it only happens after our change. Sounds indeed our code is less efficient than what was there before.

You can try to move the test to `flaky` test group. These tests run in series instead of parallel, so the CPU will be available for it. (normal tests run 20 different tests in parallel)

Later we'll still have to investigate further.

ioanmo226: ok

extension/js/common/oauth2/oauth2.ts
```diff
@@ -23,10 +23,14 @@ export class OAuth2 {
...
-        resolve(result);
-        return false;
+      // need to use chrome.runtime.onMessage because BrowserMsg.addListener doesn't work
+      // In gmail page reconnect auth popup, it sends event to background page (BrowserMsg.send.bg.await.reconnectAcctAuthPopup)
+      // thefore BrowserMsg.addListener doesn't work
+      chrome.runtime.onMessage.addListener((message: Bm.Raw) => {
+        if (message.name === 'auth_window_result') {
+          void chrome.tabs.remove(tabId!);
+          resolve(message.data.bm as Bm.AuthWindowResult);
+        }
```
tomholub, line 33: I'd like to understand the situation a little better, maybe there is another solution. Is it the following?

1) user needs to authenticate from Gmail content script, which doesn't have `chrome.windows` access
2) the content script calls background page
3) background page calls `newAuthPopup` which calls this `webAuthFlow` which produces an auth window
4) we have two listeners here in `webAuthFlow`
    - `chrome.tabs.onRemoved` which waits for the window to be closed by user
    - `auth_window_result` listener which waits for the `auth_window_result` message sent from the oauth window content script
5) user approves the auth, and the content script broadcasts `auth_window_result` message
6) the message is intercepted above which will resolve in this method
7) the backround page `reconnect_acct_auth_popup` call will resolve back into the gmail content script, which will now know that user was successfully authenticated

--------------------

On step 5, since it's a broadcast message, it's supposed to reach every frame. But you're saying it only reached the backround page and nowhere else? Sound like our `broadcast` mechanism is broken.

test/source/tests/gmail.ts
```diff
@@ -301,6 +302,7 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       await createSecureDraft(t, browser, gmailPage, 'reply draft');
       await createSecureDraft(t, browser, gmailPage, 'offline reply draft', { offline: true });
       await gmailPage.page.reload({ waitUntil: 'networkidle2' });
+      await Util.sleep(30);
```
tomholub, line 305: That's quite long, but I supposed you tried shorter and it didn't work?

tomholub (approved):
I'd like to discuss the details, but in the meantime I think we're good to use this. Thank you for the fixes.

test/source/tests/gmail.ts
ioanmo226: In unstable connections, it didn't work.
And had to use 30 seconds for unstable internet

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -722,7 +722,8 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       .append('<div class="allow-google-contact-search" data-test="action-auth-with-contacts-scope"><img src="/img/svgs/gmail.svg" />Enable Google Contact Search</div>') // xss-direct
       .find('.allow-google-contact-search')
       .click(this.view.setHandler(async () => {
-        const authResult = await GoogleAuth.newAuthPopup({ acctEmail: this.view.acctEmail, scopes: GoogleAuth.defaultScopes('contacts') });
+        // Need to use BrowserMsg.send.bg because chrome.windows is undefined in gmail page
+        const authResult = await BrowserMsg.send.bg.await.reconnectAcctAuthPopup({ acctEmail: this.view.acctEmail, scopes: GoogleAuth.defaultScopes('contacts') });
```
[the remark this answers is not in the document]
ioanmo226: Ohh, you are right.
It's fine to call `GoogleAuth.newAuthPopup` directly here.
Thakns.

extension/js/common/oauth2/oauth2.ts
ioanmo226: Exactly!
chrome.runtime.onMessage works but BrowserMsg.addListener doesn't work here.
I also think `BrowserMsg.addListener` has some issues for this case.

PR #4704, opened 2022-09-21
#4609 stop using initialLegacySubmit

### LINKED ISSUE #4609: stop using initialLegacySubmit
should first do #4610 and #4607

we're removing method `initialLegacySubmit` as follows:

## in `keyserver.ts` `submitPublicKeyHandler`

 you can instead use `submitPrimaryEmailPubkey` with id_token if this is a primary email that is being updated, or `replacePubkey` for aliases.

## rename client configuration 

`useLegacyAttesterSubmit` to `setupEnsureImportedPrvMatchLdapPub` and rename `USE_LEGACY_ATTESTER_SUBMIT` to `SETUP_ENSURE_IMPORTED_PRV_MATCH_LDAP_PUB`

##  in `setup.ts` `submitPubkeys`, on top

instead of `initialLegacySubmit` use `Attester.doLookupLdap` without specifying server. 
 - If there is no result, show warning modal `Your organization requires public keys to be present on company LDAP server, but no public key was found. Please ask your internal help desk.`
 - If you did get back a public key,  compare the received public keys primary fingerprints to the primary fingerprint of the private key that the extension is being set up with. If at least one primary fingerprint of the prvs matches at least one privmary fingerprint of the received pubkeys, do nothing else. If there is no intersection, show a warning modal with `Imported private key with ids {prv.map(prv => prv.id).join(', ')} does not match public keys on company LDAP server with ids {pubs.map(pub => pub.id).join(', ')}. Please ask your help desk.` 

## in `setup.ts` `submitPubkeys`, for aliases

where aliases are being submitted on the bottom, use `replacePubkey`

ioanmo226:
This PR stops using initialLegacySubmit and implemented new org rule `SETUP_ENSURE_IMPORTED_PRV_MATCH_LDAP_PUB`

close #4609 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/chrome/settings/modules/keyserver.ts
+++ b/extension/chrome/settings/modules/keyserver.ts
@@ -104,7 +104,7 @@ View.run(class KeyserverView extends View {
         const idToken = await InMemoryStore.get(this.acctEmail, InMemoryStoreKeys.ID_TOKEN);
         await this.pubLookup.attester.submitPrimaryEmailPubkey(email, mostUsefulPrv.keyInfo.public, idToken!);
       } else { // If email is alias email
-        await this.pubLookup.attester.replacePubkey(email, mostUsefulPrv.keyInfo.public);
+        await this.pubLookup.attester.submitPubkeyWithConditionalEmailVerification(email, mostUsefulPrv.keyInfo.public);
       }
     } catch (e) {
       ApiErr.reportIfSignificant(e);
@@ -127,7 +127,7 @@ View.run(class KeyserverView extends View {
       return;
     }
     try {
-      const responseText = await this.pubLookup.attester.replacePubkey(String($(target).attr('email')), mostUsefulPrv.keyInfo.public);
+      const responseText = await this.pubLookup.attester.submitPubkeyWithConditionalEmailVerification(String($(target).attr('email')), mostUsefulPrv.keyInfo.public);
       await Ui.modal.info(responseText);
       BrowserMsg.send.closePage(this.parentTabId);
     } catch (e) {
--- a/extension/chrome/settings/modules/keyserver.ts
+++ b/extension/chrome/settings/modules/keyserver.ts
@@ -18,6 +18,8 @@ import { KeyStore } from '../../../js/common/platform/store/key-store.js';
 import { KeyStoreUtil } from "../../../js/common/core/crypto/key-store-util.js";
 import { AcctStore } from '../../../js/common/platform/store/acct-store.js';
 import { KeyUtil } from '../../../js/common/core/crypto/key.js';
+import { InMemoryStore } from '../../../js/common/platform/store/in-memory-store.js';
+import { InMemoryStoreKeys } from '../../../js/common/core/const.js';
 
 type AttesterKeyserverDiagnosis = { hasPubkeyMissing: boolean, hasPubkeyMismatch: boolean, results: Dict<{ pubkeys: string[], match: boolean }> };
 
@@ -96,7 +98,14 @@ View.run(class KeyserverView extends View {
       return;
     }
     try {
-      await this.pubLookup.attester.initialLegacySubmit(String($(target).attr('email')), mostUsefulPrv.keyInfo.public);
+      const email = String($(target).attr('email'));
+      // Use submitPrimaryEmailPubkey if email is primary email
+      if (email === this.acctEmail) {
+        const idToken = await InMemoryStore.get(this.acctEmail, InMemoryStoreKeys.ID_TOKEN);
+        await this.pubLookup.attester.submitPrimaryEmailPubkey(email, mostUsefulPrv.keyInfo.public, idToken!);
+      } else { // If email is alias email
+        await this.pubLookup.attester.replacePubkey(email, mostUsefulPrv.keyInfo.public);
+      }
     } catch (e) {
       ApiErr.reportIfSignificant(e);
       await Ui.modal.error(ApiErr.eli5(e));
--- a/extension/chrome/settings/setup.ts
+++ b/extension/chrome/settings/setup.ts
@@ -3,7 +3,7 @@
 'use strict';
 
 import { Bm, BrowserMsg } from '../../js/common/browser/browser-msg.js';
-import { asyncSome, Url, Value } from '../../js/common/core/common.js';
+import { asyncSome, Url } from '../../js/common/core/common.js';
 import { ApiErr } from '../../js/common/api/shared/api-error.js';
 import { Assert } from '../../js/common/assert.js';
 import { Catch, SubmitPubKeyError } from '../../js/common/platform/catch.js';
@@ -236,7 +236,7 @@ export class SetupView extends View {
     } catch (e) {
       return await Settings.promptToRetry(
         e,
-        Lang.setup.failedToSubmitToAttester,
+        e instanceof SubmitPubKeyError ? Lang.setup.failedToImportUnknownKey : Lang.setup.failedToSubmitToAttester,
         () => this.submitPublicKeys({ submit_main, submit_all }),
         Lang.general.contactIfNeedAssistance(this.isFesUsed())
       );
@@ -337,14 +337,15 @@ export class SetupView extends View {
         const prvs = await KeyStoreUtil.parse(await KeyStore.getRequired(this.acctEmail));
         const parsedPubKeys = await KeyUtil.parseMany(result.pubkeys.join('\n'));
         const hasMatchingKey = await asyncSome(prvs, (async (privateKey) => {
-          return parsedPubKeys.some((parsedPubKey) => Value.arr.hasIntersection(privateKey.key.allIds, parsedPubKey.allIds));
+          return parsedPubKeys.some((parsedPubKey) => privateKey.key.id === parsedPubKey.id);
         }));
         if (!hasMatchingKey) {
           // eslint-disable-next-line max-len
-          throw new SubmitPubKeyError(`Imported private key with ids ${prvs.map(prv => prv.key.id).join(', ')} does not match public keys on company LDAP server with ids ${parsedPubKeys.map(pub => pub.id).join(', ')}. Please ask your help desk.`);
+          throw new SubmitPubKeyError(`Imported private key with ids ${prvs.map(prv => prv.key.id).join(', ')} does not match public keys on company LDAP server with ids ${parsedPubKeys.map(pub => pub.id).join(', ')} for ${this.acctEmail}. Please ask your help desk.`);
         }
       } else {
-        throw new SubmitPubKeyError('Your organization requires public keys to be present on company LDAP server, but no public key was found. Please ask your internal help desk.');
+        // eslint-disable-next-line max-len
+        throw new SubmitPubKeyError(`Your organization requires public keys to be present on company LDAP server, but no public key was found for ${this.acctEmail}. Please ask your internal help desk.`);
       }
     } else {
       // this will actually replace the submitted public key if there was a conflict, better ux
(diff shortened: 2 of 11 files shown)


ioanmo226:
@tomholub Mock tests fail because of below reason.
Attester mock api requires `Authorization` but in `replacePubkey` function, it doesn't pass`Authorization`.
Is mock api wrong or do we have to pass `Authorization` in `replacePubKey`?

https://github.com/FlowCrypt/flowcrypt-browser/blob/343ca0933e5536bdb99c904ffe13a2ede41da4aa/test/source/mock/attester/attester-endpoints.ts#L107

https://github.com/FlowCrypt/flowcrypt-browser/blob/343ca0933e5536bdb99c904ffe13a2ede41da4aa/extension/js/common/api/key-server/attester.ts#L86

tomholub:
I'll have to take a look later today. I believe it needs authentication in some situations, but not others.

tomholub:
The situation is:
 - if authentication is used, key is updated right away (subject to various restrictions)
 - if authentication is missing, and this is a newer version of a previous key (same fingerprint, newer lastModified) then it's also updated
 - if authentication is missing and it's other situation (previously no key, or previously different key) then attester sends a verification email to the user to confirm the change

@martgil please help me confirm that the above is the case for live attester, and if it's not, please file appropriate issues. In particular, I want to ensure that a pubkey is not accepted without authentication and without email confirmation if there was no pubkey before.

Meanwhile @ioanmo226 you can continue work here assuming the above is true.

Now, because authentication is only good for primary account email (the email address listed in the id_token), we should only be sending authentication when we upload public key for the primary account email. In other situation (other aliases) we skip authentication, and the verification email takes care of preventing abuse.

Therefore the answer is, the endpoint accepts both options, and behaves differently.

The mock only needs to check that authentication is present for requests where it's expected (primary account email), and not present for requests where it's not expected (aliases).

martgil:
@tomholub 

> if authentication is used, key is updated right away (subject to various restrictions)

True, the key gets updated right away when authentication is used.

>if authentication is missing, and this is a newer version of a previous key (same fingerprint, newer lastModified) then it's also updated

When I tried submitting a newer version of the previous key same fingerprint with an updated uid. The attester sends a verification email rather than updating the key. Same result when following the same procedures when running flowcrypt-attester-ts.

> if authentication is missing and it's other situation (previously no key, or previously different key) then attester sends a verification email to the user to confirm the change

 True, When I submit a different key or new key while the Authorization header is missing or has no value at all -- the attester sends a verification email to the user. 

I'll create a PR to document this on the appropriate repo and will file an issue.

extension/chrome/settings/modules/keyserver.ts
```diff
@@ -96,7 +98,14 @@ View.run(class KeyserverView extends View {
       return;
     }
     try {
-      await this.pubLookup.attester.initialLegacySubmit(String($(target).attr('email')), mostUsefulPrv.keyInfo.public);
+      const email = String($(target).attr('email'));
+      // Use submitPrimaryEmailPubkey if email is primary email
+      if (email === this.acctEmail) {
+        const idToken = await InMemoryStore.get(this.acctEmail, InMemoryStoreKeys.ID_TOKEN);
```
tomholub, line 104: There is a method in GoogleAuth that always gives you fresh id token if needed. This one may be stale.

extension/chrome/settings/modules/keyserver.ts
```diff
@@ -96,7 +98,14 @@ View.run(class KeyserverView extends View {
...
+        await this.pubLookup.attester.submitPrimaryEmailPubkey(email, mostUsefulPrv.keyInfo.public, idToken!);
+      } else { // If email is alias email
+        await this.pubLookup.attester.replacePubkey(email, mostUsefulPrv.keyInfo.public);
```
tomholub, line 107: Let's call this method `submitPubkeyWithConditionalEmailVerification`

extension/chrome/settings/setup.ts
```diff
@@ -330,16 +330,29 @@ export class SetupView extends View {
...
-    if (this.clientConfiguration.useLegacyAttesterSubmit()) {
+    if (this.clientConfiguration.setupEnsureImportedPrvMatchLdapPub()) {
       // this will generally ignore errors if conflicting key already exists, except for certain orgs
-      await this.pubLookup.attester.initialLegacySubmit(this.acctEmail, pubkey);
+      const result = await this.pubLookup.attester.doLookupLdap(this.acctEmail);
+      if (result.pubkeys.length) {
+        const prvs = await KeyStoreUtil.parse(await KeyStore.getRequired(this.acctEmail));
+        const parsedPubKeys = await KeyUtil.parseMany(result.pubkeys.join('\n'));
+        const hasMatchingKey = await asyncSome(prvs, (async (privateKey) => {
+          return parsedPubKeys.some((parsedPubKey) => Value.arr.hasIntersection(privateKey.key.allIds, parsedPubKey.allIds));
```
tomholub, line 340: can do `=> privateKey.key.id === parsedPubKey.id`. Subkey fingerprints don't matter

extension/js/common/client-configuration.ts
```diff
@@ -192,12 +192,10 @@ export class ClientConfiguration {
...
   /**
-   * Some orgs use flows that are only implemented in POST /initial/legacy_submit and not in POST /pub/email@corp.co:
-   *  -> enforcing that submitted keys match customer key server
-   * Until the newer endpoint is ready, this flag will point users in those orgs to the original endpoint
+   * Some orgs will require user's imported private key to match their LDAP pub key result
    */
-  public useLegacyAttesterSubmit = (): boolean => {
-    return (this.clientConfigurationJson.flags || []).includes('USE_LEGACY_ATTESTER_SUBMIT');
+  public setupEnsureImportedPrvMatchLdapPub = (): boolean => {
+    return (this.clientConfigurationJson.flags || []).includes('SETUP_ENSURE_IMPORTED_PRV_MATCH_LDAP_PUB');
```
tomholub, line 198: I need to ensure I've set this for the customer before merging. And let the customer test before we submit a new release.

test/source/tests/setup.ts
```diff
@@ -499,6 +499,17 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
     }));
 
+    ava.default('invalid.pub@client-configuration-test.flowcrypt.test - no backup, no keygen', testWithBrowser(undefined, async (t, browser) => {
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'invalid.pub@client-configuration-test.flowcrypt.test');
+      await SetupPageRecipe.manualEnter(settingsPage, 'has.pub.client.configuration.test', { noPrvCreateClientConfiguration: true, enforceAttesterSubmitClientConfiguration: true, fillOnly: true },
+        { isSavePassphraseChecked: false, isSavePassphraseHidden: false });
+      await settingsPage.waitAndClick('@input-step2bmanualenter-save');
+      await settingsPage.waitAll(['@container-overlay-prompt-text', '@action-overlay-retry']);
+      const renderedErr = await settingsPage.read('@container-overlay-prompt-text');
+      expect(renderedErr).to.contain('Failed to submit to Attester');
```
tomholub, line 509: In this exact situation where we cross check against LDAP, the error should instead say `Attempting to import unknown key`

test/source/tests/setup.ts
```diff
@@ -507,7 +518,7 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await settingsPage.waitAll(['@container-overlay-prompt-text', '@action-overlay-retry']);
       const renderedErr = await settingsPage.read('@container-overlay-prompt-text');
       expect(renderedErr).to.contain(`Failed to submit to Attester`);
-      expect(renderedErr).to.contain(`Could not find LDAP pubkey on a LDAP-only domain for email no.pub@client-configuration-test.flowcrypt.test on server keys.flowcrypt.test`);
+      expect(renderedErr).to.contain(`Your organization requires public keys to be present on company LDAP server, but no public key was found. Please ask your internal help desk.`);
```
tomholub, line 521: Please update to add ... no public key was found **for email address XXX**.

test/source/tests/setup.ts
```diff
@@ -499,6 +499,17 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
 
+    ava.default('invalid.pub@client-configuration-test.flowcrypt.test - no backup, no keygen', testWithBrowser(undefined, async (t, browser) => {
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, 'invalid.pub@client-configuration-test.flowcrypt.test');
+      await SetupPageRecipe.manualEnter(settingsPage, 'has.pub.client.configuration.test', { noPrvCreateClientConfiguration: true, enforceAttesterSubmitClientConfiguration: true, fillOnly: true },
+        { isSavePassphraseChecked: false, isSavePassphraseHidden: false });
+      await settingsPage.waitAndClick('@input-step2bmanualenter-save');
+      await settingsPage.waitAll(['@container-overlay-prompt-text', '@action-overlay-retry']);
+      const renderedErr = await settingsPage.read('@container-overlay-prompt-text');
+      expect(renderedErr).to.contain('Failed to submit to Attester');
+      expect(renderedErr).to.contain('Imported private key with ids 576C48E8E9E33B772FF07B11BC614F7068DB6E23 does not match public keys on company LDAP server with ids AB8CF86E37157C3F290D72007ED43D79E9617655. Please ask your help desk.');
```
tomholub, line 510: Please update to say ... AB8CF86E37157C3F290D72007ED43D79E9617655 **for address email XXX**.

tomholub (commented):
Thanks! comments below

extension/chrome/settings/modules/keyserver.ts
ioanmo226: Hmm.. Which method do you mean in GoogleAuth?

test/source/tests/setup.ts
ioanmo226: In this situation(no keys found), don't we need to update `Failed to submit to Attester` error message like above?

> In this exact situation where we cross check against LDAP, the error should instead say `Attempting to import unknown key`

tomholub: Yes, also here

extension/chrome/settings/modules/keyserver.ts
tomholub: You're right, there isn't any. And the id token is kept reasonably fresh through other mechanism (when Google API is called and when access token is refreshed). The browser extension is not designed ideally there, but it does work, and you've done it correctly.

extension/js/common/platform/catch.ts
```diff
@@ -5,6 +5,7 @@
 import { FLAVOR, VERSION } from '../core/const.js';
 
 export class UnreportableError extends Error { }
+export class SubmitPubKeyError extends UnreportableError { }
```
tomholub, line 8: Let's call it `CompanyLdapKeyMismatchError`

extension/js/common/client-configuration.ts
tomholub: Ok I've updated it now.

tomholub (approved):

extension/js/common/client-configuration.ts
tomholub: #4707 4707

tomholub:
Very good. I think, this will allow us to completely drop an ageing Python flowcrypt-attester implementation, and fully transition to flowcrypt-attester-ts instead. If so, that would be really good news.

PR #4721, opened 2022-09-28
#4694 Require flags parameter in client configuration

### LINKED ISSUE #4694: require `flags` parameter in client configuration
Today, if the browser extension receives the following client configuration `{}` it will interpret it as `{"flags": []}`, meaning empty.

I would instead like the `flags` parameter to be required in the API response. If there is a 200 response and the `flags` is missing, we should throw an error (which will then block setup procedure). present `flags` that is empty is no problem.

Reason: there was a problem with one deployment that had bad client configuration format, and it was really hard to track down. After this, similar mistakes would stand out.

ioanmo226:
This PR requires flags parameter in client configuration

close #4694 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/client-configuration.ts
+++ b/extension/js/common/client-configuration.ts
@@ -5,6 +5,7 @@
 import { Str } from './core/common.js';
 import { AcctStore } from './platform/store/acct-store.js';
 import { KeyAlgo } from './core/crypto/key.js';
+import { UnreportableError } from './platform/catch.js';
 
 type ClientConfiguration$flag = 'NO_PRV_CREATE' | 'NO_PRV_BACKUP' | 'PRV_AUTOIMPORT_OR_AUTOGEN' | 'PASS_PHRASE_QUIET_AUTOGEN' |
   'ENFORCE_ATTESTER_SUBMIT' | 'NO_ATTESTER_SUBMIT' | 'SETUP_ENSURE_IMPORTED_PRV_MATCH_LDAP_PUB' |
@@ -34,7 +35,7 @@ export class ClientConfiguration {
     }
     const storage = await AcctStore.get(email, ['rules']);
     if (!storage.rules?.flags) {
-      throw new Error('Flags is required for client configuration.');
+      throw new UnreportableError('Missing client configuration flags.');
     }
     return new ClientConfiguration(storage.rules!, Str.getDomainFromEmailAddress(acctEmail));
   };
--- a/extension/js/common/client-configuration.ts
+++ b/extension/js/common/client-configuration.ts
@@ -34,7 +34,7 @@ export class ClientConfiguration {
     }
     const storage = await AcctStore.get(email, ['rules']);
     if (!storage.rules?.flags) {
-      throw new Error(`Flags is required for client configuration`);
+      throw new Error('Flags is required for client configuration.');
     }
     return new ClientConfiguration(storage.rules!, Str.getDomainFromEmailAddress(acctEmail));
   };
--- a/extension/js/common/client-configuration.ts
+++ b/extension/js/common/client-configuration.ts
@@ -27,15 +27,16 @@ export type ClientConfigurationJson = {
  */
 export class ClientConfiguration {
 
-  private static readonly default = { flags: [] };
-
   public static newInstance = async (acctEmail: string): Promise<ClientConfiguration> => {
     const email = Str.parseEmail(acctEmail).email;
     if (!email) {
       throw new Error(`Not a valid email`);
     }
     const storage = await AcctStore.get(email, ['rules']);
-    return new ClientConfiguration(storage.rules || ClientConfiguration.default, Str.getDomainFromEmailAddress(acctEmail));
+    if (!storage.rules?.flags) {
+      throw new Error(`Flags is required for client configuration`);
+    }
+    return new ClientConfiguration(storage.rules!, Str.getDomainFromEmailAddress(acctEmail));
   };
 
   protected constructor(
--- a/test/source/test.ts
+++ b/test/source/test.ts
@@ -128,8 +128,6 @@ ava.default.after.always('evaluate Catch.reportErr errors', async t => {
     ].includes(e.message))
     // below for test "user4@standardsubdomainfes.test:8001 - PWD encrypted message with FES web portal - a send fails with gateway update error"
     .filter(e => !e.message.includes('Test error'))
-    // below for test "user@no-flags-client-configuration.flowcrypt.test - should show error when no flags is present"
-    .filter(e => !e.message.includes('Flags is required for client configuration.'))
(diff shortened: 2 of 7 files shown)


extension/js/common/client-configuration.ts
```diff
@@ -27,15 +27,16 @@ export type ClientConfigurationJson = {
...
-
   public static newInstance = async (acctEmail: string): Promise<ClientConfiguration> => {
     const email = Str.parseEmail(acctEmail).email;
     if (!email) {
       throw new Error(`Not a valid email`);
     }
     const storage = await AcctStore.get(email, ['rules']);
-    return new ClientConfiguration(storage.rules || ClientConfiguration.default, Str.getDomainFromEmailAddress(acctEmail));
+    if (!storage.rules?.flags) {
+      throw new Error('Flags is required for client configuration.');
```
tomholub, line 37: `Missing client configuration flags.`

test/source/test.ts
```diff
@@ -128,6 +128,8 @@ ava.default.after.always('evaluate Catch.reportErr errors', async t => {
     ].includes(e.message))
     // below for test "user4@standardsubdomainfes.test:8001 - PWD encrypted message with FES web portal - a send fails with gateway update error"
     .filter(e => !e.message.includes('Test error'))
+    // below for test "user@no-flags-client-configuration.flowcrypt.test - should show error when no flags is present"
+    .filter(e => !e.message.includes('Flags is required for client configuration.'))
```
tomholub, line 132: The original error should throw `UnreportableError`, then we don't have to ignore it here.

test/source/tests/compose.ts
```diff
@@ -873,7 +873,7 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
     }));
 
     ava.default('compose - enforce message signing when encrypting', testWithBrowser('compatibility', async (t, browser) => {
-      const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
+      const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
```
tomholub, line 876: Unsure why this change

tomholub (commented):
looks good, with some comments. Thanks!

test/source/tests/compose.ts
ioanmo226: well, if we pass `compose` to `ComposePageRecipe.openStandalone` method, compose.htm `account_email` becomes `ci.tests.gmail@flowcrypt.test` which is not correct (logged in account email is `flowcrypt.compatibility@gmail.com`)

https://github.com/FlowCrypt/flowcrypt-browser/blob/4736d924ef7d7454401f38037f6ea00eb0c39173/test/source/tests/page-recipe/compose-page-recipe.ts#L27

tomholub: Generally we use `compose` account for tests that compose emails, and `compatibility` email for tests that read / check emails. I guess that doesn't really matter in particular.

But I'm wondering why now? If the test was broken, it would have to be already failing before?

ioanmo226: It matters with new change and I can see that previous logic was incorrect.
Current logged in user email is `flowcrypt.compatibility@gmail.com` and test is trying to open compose.htm page with `account_email` as `ci.tests.gmail@flowcrypt.test`.
As a result, storage.rules becomes null because provided `account_email` is incorrect

image
image

ioanmo226: Previously tests didn't fail because we didn't throw error for invalid/empty rules from AcctStore.

ioanmo226: Any reason why we need to use `ci.tests.gmail@flowcrypt.test` for `compose` group?

image

tomholub: No strong reason. It's generally preferred to use our own test email on a `.test` domain to make things clear. So we generally always try to use a `.test` domain, except when we cannot.

Many tests use exported data specifically from `flowcrypt.compatibility` account, and for those, we use the same account for simplicity. For tests that don't need those exported messages, like compose tests, we use the `.test` account.

It's not a big deal on its own. I'm just trying to get to the bottom of what happened above. I get it conceptually, I think, but something still smells :-) That's why I'd like to see what will it look like when using the original `compose` test email, and fixing whatever needed fixing. That will surface the needed change more clearly.

ioanmo226: Hmm... Not sure why you want that change..
Current app logic is to receive `acct_email` from url parameters.
`openStandalone` test method opens `compose.htm` page with incorrect `account_email` parameter.

You mean you want to change current app logic not to receive `acct_email` from url parameters? And receive `acct_email` from somewhere else

tomholub: What I meant is this:
 - openStandalone is passed with `compose` as before
 - that uses `ci.tests.gmail@flowcrypt.test` for this test as before

If that breaks the test, fix the breakage by fixing that test, without changing on which account it runs. It should continue to run on `ci.tests.gmail@flowcrypt.test`. If it doesn't now, fix whatever makes it not be able to run as before. Does that make sense?

rrrooommmaaa: > It should continue to run on `ci.tests.gmail@flowcrypt.test`.

It used to run on `compatibility` account, that's the point. So we need to put `compose` (if it is accepted there) or `ci.tests.gmail` to line 875, I guess

tomholub: Ah, now I see the line 875 above, as you mentioned:

```
    ava.default('compose - enforce message signing when encrypting', testWithBrowser('compatibility', async (t, browser) => {
      const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
```

Yes, setting 875 to `ci.tests.gmail` and 876 to `compose` will work.

Sorry for the confusion, I didn't notice the mismatch before.

tomholub: Actually, running on compatibility is fine to. So I'm merging this. Thank you for explaining this to me.

tomholub (approved):

ioanmo226:
@tomholub I tried to remove & re-install extension and i see this error with `master` branch as soon as I click flowcrypt-browser extension icon.
This issue happens because `acctEmail` is set in settings.htm  url params but no datas are stored in AcctStore(as well as storage.rules)
storage.rules is null for this case therefore, below error occurred.
Maybe we need to change condition like this?

`if (storage.rules && !storage.rules.flags) {..`

https://github.com/FlowCrypt/flowcrypt-browser/blob/23946dc828e819f7a5d6244bf2975f05924975c4/extension/js/common/client-configuration.ts#L37
 
image

tomholub:
Yes. Is there no test for that? The mock tests should be updated to test this scenario.

PR #4723, opened 2022-09-28
Remove misleading visible:false option for waitAny()

### LINKED ISSUE #4722: Disallow `visible: false` in waitAny()
Current `waitAny()` parameters are misleading.
`visible: false` should be treated as "element must not be visible".
However, as we don't actually use this in tests, let's remove this option, and only leave `true | undefined`.
Note: `waitAll()` treats all 3 options (true, false and undefined) correctly.

_Originally posted by @rrrooommmaaa in https://github.com/FlowCrypt/flowcrypt-browser/pull/4708#discussion_r979285764_

rrrooommmaaa:
This PR removes unsupported `visible: false` option in `waitAny()` (all the tests use `visible: true` or `visible: undefined`, or `waitAll()` instead

close #4722

- Tests updated

### To be filled by reviewers


test/source/browser/controllable.ts
```diff
@@ -76,8 +76,9 @@ abstract class ControllableBase {
     this.log(`wait_all:7:${selectors.join(',')}`);
   };
 
-  public waitAny = async (selector: string | string[], { timeout = TIMEOUT_ELEMENT_APPEAR, visible = true }: { timeout?: number, visible?: boolean } = {}): Promise<ElementHandle> => {
-    return await this.waitAnyInternal(this.selsAsProcessedArr(selector), { timeout, visible });
+  public waitAny = async (selector: string | string[], properties: { timeout?: number, visible: true | undefined } | { timeout?: number } = {}): Promise<ElementHandle> => {
```
ioanmo226, line 79: Why `visible` type becomes `true | undefined`? Any problem with `visible?: boolean`?

ioanmo226: Never mind, Just figured out because visible `false` option should not be passed.

ioanmo226 (approved):
👍

test/source/browser/controllable.ts
rrrooommmaaa: In `waitAll` the meanings are:
`true` element must exist and be visible
`false` element must exist and be not visible.
`undefined` element must exist

`waitAny` doesn't support the option: element must exist and be not visible.
This PR makes `waitAny` in line with `waitAll` meanings

Issue #1509, opened 2019-01-25
google chrome manifest v3 web extensions proposal [June 2024]


tomholub:
https://docs.google.com/document/d/1nPu6Wy4LWR66EFLeYInl3NzzhHzc-qnk4w4PX-0XMw8/mobilebasic#h.z7s2dbchx38y

tomholub:
First should do some prep work - move what we can from background script to content script

tomholub:
https://developer.chrome.com/extensions/migrating_to_manifest_v3

tomholub:
Available in Chrome Canary

tomholub:
> Dear Developer,
> 
> Last year, we announced the availability of Manifest V3 - a more secure, performant, and privacy-preserving iteration of the extension platform. Since then, we have added new functionality to and improved the robustness of Manifest V3 with feedback from our developer community.
> 
> We recently published a timeline of our plans to move forward with the deprecation of Manifest V2 and shifting focus fully to the new iteration of the platform. This is a notice that beginning January 17, 2022, the Chrome Web Store will stop accepting any new Manifest V2 extensions with visibility set to “Public” or “Unlisted”. Existing Manifest V2 extensions that have previously been published publicly will not be affected by this change and updates to those already-published extensions will continue to be accepted.
> 
> Additionally, beginning in June of 2022, this restriction on new extensions will expand to include items set to “Private” visibility as well. For more details and dates relating to the phase-out of Manifest V2, please refer to the Manifest V2 Support Timeline page, which will be kept up to date as more exact dates and milestone details are available.
> 
> If you have not yet begun to do so, we recommend that you begin migrating your extensions to Manifest V3. If you have feedback around migration, please please post to the chromium-extensions Google Group.
> 
> Thank you for your cooperation and for your participation in the Chrome extension ecosystem!
> 
> - The Google Chrome Web Store team

tomholub:
![image](https://user-images.githubusercontent.com/6306961/214399995-fe235fab-3701-4bdd-90e6-51026d0160dc.png)

rrrooommmaaa:
> First should do some prep work - move what we can from background script to content script

We can get rid of `pgpMsgDecrypt` message handling in the background page, send the request for a decrypted attachment to the content script (handled by MessageRenderer using the cache)

Do you approve this sub-task, @tomholub ?

tomholub:
yup, that sounds good

sosnovsky:
[...]

PR #4725, opened 2022-09-29
Switch to OpenPGP v5

### LINKED ISSUE #3324: switch browser extension to OpenPGP v5
Not ready yet, but hopefully in a few months it will be

### LINKED ISSUE #4904: could key be sent to background page binary / unarmored?
Would it be faster to send the key in binary representation (unarmored openpgp key)? Armoring/de-armoring is oddly expensive, about 10ms per key. Could be separate PR.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4725#discussion_r1085549366_

rrrooommmaaa:
This PR switches to OpenPGP v5

close #3324
close #4904

- Tests added or updated

### To be filled by reviewers


tomholub:
looking great, finally :-)

are we able to drop any functionality related to dropping weak algorithms? There should be settings on v5 to do this natively.

rrrooommmaaa:
> @rrrooommmaaa how is this looking at this moment?

I need to find a way to neatly wire the new library to the browser page, most of incompatibilities were resolved 2 months ago and unit tests are looking good.
I had to suspend the development in favor of review, now I'm able to finish it.

rrrooommmaaa:
While waiting for https://github.com/openpgpjs/openpgpjs/pull/1583 to be merged and published, looking at other issues

rrrooommmaaa:
@tomholub We're getting an error when trying to return `OpenPGP.Key` object from background handlers:
`Could not serialize message`
It happens because of cyclic nesting -- each of `subkeys` and `users` holds a reference to the `mainKey`.
Should we implement a workaround or... ?

rrrooommmaaa:
> Should we implement a workaround or... ?

Perhaps, fix `OpenPGP.js` to allow transportation of the object?
For this we need to find out exactly what this `mainKey` is used for.
Perhaps, we can add a function in `OpenPGP.js` that removes all the `mainKey` references,
and add functionality in functions that search for subkeys and userids to restore this field if it's missing.

tomholub:
I think we should not be expecting to successfully transfer OpenPGP objects across the boundary.

`OpenPGP.Message` or `OpenPGP.Key` should stay within their original boundary, and across the boundary we share the message as bytes, or key as bytes.

Which current action causes deserialized key to be sent across?

rrrooommmaaa:
> `OpenPGP.Message` or `OpenPGP.Key` should stay within their original space. Across the boundary we only share the message as bytes, or key as bytes.
> 
> Which current action causes deserialized key to be sent across?

We have "hidden" fields in our own `Key` type: `rawKey`, `internal` and `rawArmored`.
`rawKey` is `OpenPGP.Key`, `internal` is safe to use `OpenPGP.Key` (without weak subkeys).

So wherever we pass `Key` (inside `PubkeyInfo`), we pass the underlying `OpenPGP.Key` objects.
These are returned by `ContactStore` 
- `getEncryptionKeys`
- `getOneWithAllPubkeys`
- `getPubkeyInfos`

Some `ContactStore` methods accept `Key` arguments, so it seems natural to return `Key` too.
Are you thinking about refactoring `ContactStore`?

rrrooommmaaa:
We can also think about wiping `rawKey` and `internal` fields when sending/receiving messages,
and restoring it on-demand in `extension/js/common/core/crypto/pgp/openpgp-key.ts` functions

tomholub:
> Some ContactStore methods accept Key arguments, so it seems natural to return Key too.
> Are you thinking about refactoring ContactStore?

I never intended to be sending OpenPGP.js objects across the wire, to my knowledge the objects undergo JSON serialization / deserialization in the process and last time I tried, that didn't work for OpenPGP.js objects. I'm surprised it worked that way.

All background methods should be refactored to not not accept or return these, and if that includes the ContactStore, than that should be refactored also to send simple objects across.

rrrooommmaaa:
Yes, all existing tests now pass.
I wanted to thoroughly review the changes and possibly add some additional tests concerning the changed logic.

tomholub:
Great news 👍

tomholub:
Please also do rudimentary manual test on Firefox, to double check nothing is horribly broken, like the content script not loading or not working properly, and setup/send/receive emails with attachments.

extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -145,7 +143,14 @@ export class MsgUtil {
         // This does not 100% mean it's OpenPGP message
         // But it's a good indication that it may be
         const t = opgp.enums.packet;
-        const msgTpes = [t.symEncryptedIntegrityProtected, t.modificationDetectionCode, t.symEncryptedAEADProtected, t.symmetricallyEncrypted, t.compressed];
+        const msgTpes = [
+          t.publicKeyEncryptedSessionKey,
```
rrrooommmaaa, line 147: I added `publicKeyEncryptedSessionKey` type here as a pubkey-encrypted message started with this type of packet. So this method isn't used often and tested properly.
We can review and test it in a separate issue

tomholub: ok

extension/js/common/core/crypto/key.ts
```diff
@@ -221,9 +222,9 @@ export class KeyUtil {
           allErr.push(e as Error);
         }
       }
-      if (err) {
+      /* todo: re-throw? if (err) {
         allErr.push(...err);
-      }
+      } */
```
tomholub, line 227: I wanted to allow reading the results partially. Like if WKD sends a mixture of keys that some throw when processing and some don't, I'd like to see the valid keys and then the errors side by side (code-wise), so that the app may choose to still use the valid ones and ignoring the errors. That's why we are returning errors instead of throwing them, unlike everywhere else in the codebase (I suppose this is what your comment was about).

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -27,21 +34,32 @@ export class OpenPGPKey {
...
 
-  public static validateAllDecryptedPackets = async (key: OpenPGP.key.Key): Promise<void> => {
+  // remove crypto-library objects (useful when sending the object to/from background)
+  public static pack = (key: Key) => {
+    const keyWithPrivateFields = key as unknown as KeyWithPrivateFields;
+    if (typeof keyWithPrivateFields.internal !== 'string') {
+      keyWithPrivateFields.internal = keyWithPrivateFields.internal.armor();
+      keyWithPrivateFields.rawKey = undefined;
+    }
+  };
```
tomholub, line 58: Simplicity looks good.

Would it be faster to send the key in binary representation (unarmored openpgp key)? Armoring/de-armoring is oddly expensive, about 10ms per key. Could be separate PR.

extension/js/common/core/crypto/pgp/msg-util.ts
tomholub: I don't think we necessarily have to unit-test this method, it's not like we normally edit it or plan to do so. I think it's likely broadly covered by other tests.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -27,21 +34,32 @@ export class OpenPGPKey {
   };
 
   public static parseMany = async (text: string): Promise<Key[]> => {
-    const result = await opgp.key.readArmored(text);
+    const resultKeys = await opgp.readKeys({ armoredKeys: text });
+    /* todo: test exception and re-wrap?
     if (result.err) {
       throw new Error('Cannot parse OpenPGP key: ' + result.err + ' for: ' + text);
     }
```
tomholub, line 41: Not checking this follows previous behavior, right? If so, we can file this for another issue, review usages, and decide how to handle these.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -118,74 +113,67 @@ export class OpenPGPKey {
...
-    }
-    if (pwd) {
-      options.passwords = [pwd];
-    }
     if (!pubkeys && !pwd) {
       throw new Error('no-pubkeys-no-challenge');
     }
+    const message = await opgp.createMessage({ binary: data, filename, date });
+    const encryptionKeys = await Promise.all(pubkeys?.map(OpenPGPKey.extractExternalLibraryObjFromKey) ?? []);
+    // TODO: Investigate unwrapping?
```
tomholub, line 137: Highlighting, although unsure about what the todo means

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -194,40 +182,25 @@ export class OpenPGPKey {
    * @param keyToUpdate - an existing Key object to update, optional. Useful in encryptKey and decryptKey, because the operation
    *    is done on the original supplied object.
    */
-  public static convertExternalLibraryObjToKey = async (opgpKey: OpenPGP.key.Key, keyToUpdate?: Key): Promise<Key> => {
+  public static convertExternalLibraryObjToKey = async (opgpKey: OpenPGP.Key, keyToUpdate?: Key): Promise<Key> => {
     const { isPrimaryKeyStrong, keyWithoutWeakPackets } = OpenPGPKey.removeWeakKeyPackets(opgpKey);
```
tomholub, line 186: I suppose there was new functionality for this in OpenPGP.js v5, where we can simply configure which algorithms to use or forbid? Either in global config or when doing particular actions (forgot about their implementation). In which case we could stop using `removeWeakKeyPackets` / remove it. Can be another issue.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -269,114 +244,99 @@ export class OpenPGPKey {
...
+    const message = await opgp.createMessage({ text });
+    const signRes = await opgp.sign({ message, format: 'armored', signingKeys: [signingPrv], detached });
+    return signRes;
   };
 
   public static revoke = async (key: Key): Promise<string | undefined> => {
-    let prv = OpenPGPKey.extractExternalLibraryObjFromKey(key);
+    let prv = await OpenPGPKey.extractExternalLibraryObjFromKey(key);
+    if (!prv.isPrivate()) {
+      return undefined; // todo: or throw?
```
tomholub, line 282: I'd throw

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -269,114 +244,99 @@ export class OpenPGPKey {
...
+    }
     if (!(await prv.isRevoked())) {
-      prv = await prv.revoke({});
+      const keypair = await opgp.revokeKey({ key: prv, format: 'object' });
+      // todo: save this data into `key`?
```
tomholub, line 286: I'm not entirely sure now what is it that we are actually doing. Did we revoke the key, or did we create a revocation certificate that can be saved and later uploaded to public key servers?

If the latter (that's what I thought), people sometimes do that as a form of forward-safety, save a revocation cert somewhere more accessible than the private key itself. not that any FlowCrypt usecases rely on that.

I think the method naming is ambiguous, UI says `revocation certificate` but this method says `revoke` when it should say `newRevocationCertificate` or similar, in which case your note would not apply.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -505,40 +498,28 @@ export class OpenPGPKey {
...
+    const opgpKeys = await Promise.all(validKeys.map(x => OpenPGPKey.extractExternalLibraryObjFromKey(x)));
     // todo: expired?
     try {
-      // this is here to ensure execution order when 1) verify, 2) read data, 3) processing signatures
-      // Else it will hang trying to read a stream: https://github.com/openpgpjs/openpgpjs/issues/916#issuecomment-510620625
-      const verifications = await msg.verify(opgpKeys); // first step
-      const stream = msg instanceof opgp.message.Message ? msg.getLiteralData() : msg.getText();
-      if (stream) {
+      const signerLongids = msg.getSigningKeyIDs().map(kid => OpenPGPKey.bytesToLongid(kid.bytes));
+      const text = msg instanceof opgp.CleartextMessage ? msg.getText() : msg.getLiteralData(); // todo: is this important?
```
tomholub, line 507: On v4 it was, at least. `getText` was for cleartext signed data, `getLiteralData` was for signed encrypted data :shrug:

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -618,127 +601,154 @@ export class OpenPGPKey {
...
+      return keyWithPrivateFields.internal;
     }
-    return opgpKey;
+    return internal;
   };
 
-  private static extractStrengthUncheckedExternalLibraryObjFromKey = (pubkey: Key) => {
-    if (pubkey.family !== 'openpgp') {
-      throw new UnexpectedKeyTypeError(`Key type is ${pubkey.family}, expecting OpenPGP`);
+  private static extractStrengthUncheckedExternalLibraryObjFromKey = async (key: Key) => {
```
tomholub, line 724: Also these "StrengthUnchecked" etc checks could be removed if the above comment is true. Doesn't have to be now.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -16,12 +16,14 @@ import { Gmail } from '../../common/api/email-provider/gmail/gmail.js';
 import { Injector } from '../../common/inject.js';
 import { PubLookup } from '../../common/api/pub-lookup.js';
 import { Notifications } from '../../common/notifications.js';
+// note: only types are supported for OpenPGP.js or web-stream-tools, their function calls will fail
```
tomholub, line 19: This is because last time we tried to use OpenPGP.js in the content script, it didn't work (on Firefox). Actually, it would be nice to load it there. Then we wouldn't have to make so many calls to background_script (which we have to eventually deprecate anyway because of Google Chrome extension manifest v3. https://github.com/FlowCrypt/flowcrypt-browser/issues/1509)

I'll file an issue to try to attempt running openpgp.js in content script again, at which point this comment could be removed, and some things simiplified, in particular:
 - background calls from content script to background page that are only done because OpenPGP.js can be removed, and instead appropriate methods (which rely on OpenPGP.js) can be called directly
 - pgp_block.htm/ts that today calls into background page can instead call the content script when it needs OpenPGP.js, and those background call handlers can be moved from background page to content script

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -16,12 +16,14 @@ import { Gmail } from '../../common/api/email-provider/gmail/gmail.js';
...
 import { PgpArmor } from '../../common/core/crypto/pgp/pgp-armor.js';
 import { Ui } from '../../common/browser/ui.js';
 import { WebmailCommon } from '../../common/webmail.js';
 import { Xss } from '../../common/platform/xss.js';
 import { ClientConfiguration } from '../../common/client-configuration.js';
 import { SendAsAlias } from '../../common/platform/store/acct-store.js';
+// todo: can we somehow define a purely relay class for ContactStore to clearly show that crypto-libraries are not loaded and can't be used?
 import { ContactStore } from '../../common/platform/store/contact-store.js';
```
tomholub, line 27: Good idea, we could separate it, which would be less messy. I'm not filing an issue yet, we'll leave this comment here, and see what to do in the context of https://github.com/FlowCrypt/flowcrypt-browser/issues/1509

tomholub (commented):
Looking great, with some comments, mostly for the future

extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -145,7 +143,14 @@ export class MsgUtil {
         // This does not 100% mean it's OpenPGP message
         // But it's a good indication that it may be
         const t = opgp.enums.packet;
-        const msgTpes = [t.symEncryptedIntegrityProtected, t.modificationDetectionCode, t.symEncryptedAEADProtected, t.symmetricallyEncrypted, t.compressed];
+        const msgTpes = [
```
sosnovsky, line 146: It's probably should be `msgTypes`?

extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -213,10 +218,12 @@ export class MsgUtil {
...
       }
       // cleartext and PKCS#7 are gone by this line
-      const msg = prepared.message as OpenPGP.message.Message;
+      const msg = prepared.message as OpenPGP.Message<OpenPGP.Data>;
       const packets = msg.packets;
-      const isSymEncrypted = packets.filter(p => p.tag === opgp.enums.packet.symEncryptedSessionKey).length > 0;
-      const isPubEncrypted = packets.filter(p => p.tag === opgp.enums.packet.publicKeyEncryptedSessionKey).length > 0;
+      // todo: remove this patch after openpgpjs/openpgpjs#1583 is resolved
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const isSymEncrypted = packets.filter(p => p instanceof (opgp as any).SymEncryptedSessionKeyPacket).length > 0; // tslint:disable-line:no-unsafe-any
```
sosnovsky, line 225: `tslint` comments not needed anymore

package.json
```diff
@@ -34,14 +35,16 @@
...
     "pdfjs-dist": "3.2.146",
     "prettier": "^2.8.3",
     "puppeteer": "19.5.2",
+    "stream-browserify": "^3.0.0",
     "stylelint": "14.16.1",
     "stylelint-config-standard": "29.0.0",
     "typescript": "4.9.4",
-    "web-ext": "7.4.0"
+    "web-ext": "7.4.0",
+    "webpack-cli": "^5.0.1"
```
sosnovsky, line 47: Is `webpack` used somewhere in the project? Haven't found usage in the code

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -269,114 +244,99 @@ export class OpenPGPKey {
...
-            await key.decrypt(passphrase); // throws on password mismatch
-            return 'success';
-          } catch (e) {
-            if (e instanceof Error && e.message.toLowerCase().includes('incorrect key passphrase')) {
-              return 'INCORRECT PASSPHRASE';
-            } else {
-              throw e;
-            }
-          }
+          return (await OpenPGPKey.decryptPrivateKey(key, passphrase)) ? 'success' : 'INCORRECT PASSPHRASE';
```
sosnovsky, line 335: Are other error types ignored intentionally here? As previously errors were rethrown

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -238,21 +211,23 @@ export class OpenPGPKey {
     if (!fingerprint) {
       throw new Error('Key does not have a fingerprint and cannot be parsed.');
     }
-    const algoInfo = keyWithoutWeakPackets.primaryKey.getAlgorithmInfo();
-    const key = keyToUpdate || ({} as Key); // if no key to update, use empty object, will get props assigned below
-
+    const algoInfo = keyWithoutWeakPackets.keyPacket.getAlgorithmInfo();
+    // tslint:disable-next-line:no-unnecessary-initializer
```
sosnovsky, line 215: `tslint:disable` not needed

sosnovsky (commented):
Wow, it's a massive PR, looks great!

extension/js/background_page/background_page.ts
```diff
@@ -20,8 +19,6 @@ import { migrateGlobal, moveContactsToEmailsAndPubkeys, updateOpgpRevocations, u
 
 console.info('background_process.js starting');
 
-opgp.initWorker({ path: '/lib/openpgp.worker.js' });
-
```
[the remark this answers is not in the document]
tomholub: https://github.com/FlowCrypt/flowcrypt-browser/issues/4903

extension/js/common/core/crypto/pgp/openpgp-key.ts
tomholub: https://github.com/FlowCrypt/flowcrypt-browser/issues/4904

extension/js/common/core/crypto/pgp/openpgp-key.ts
tomholub: https://github.com/FlowCrypt/flowcrypt-browser/issues/4905

extension/js/common/core/crypto/pgp/openpgp-key.ts
tomholub: #4905

extension/js/content_scripts/webmail/gmail-element-replacer.ts
tomholub: https://github.com/FlowCrypt/flowcrypt-browser/issues/4906

test/source/tests/unit-node.ts
```diff
@@ -1291,19 +1353,18 @@ ZAvn6PBX7vsaReOVa2zsnuY5g70xCxvzHIwR94POu5cENwRtCkrppFnISALpQ1kA
       expect(result.get('Sign/Verify test')).to.equal('[-] skipped, not fully decrypted');
       expect(result.get('SK 0 > LongId')).to.equal('[-] 0485D618EAA64B05');
       expect(result.get('SK 0 > Created')).to.equal('[-] 1606140328 or 2020-11-23T14:05:28.000Z');
-      expect(result.get('SK 0 > Algo')).to.equal('[-] rsa_encrypt_sign');
+      expect(result.get('SK 0 > Algo')).to.equal('[-] rsaEncryptSign');
       expect(result.get('SK 0 > Verify')).to.equal('[-] OK');
-      expect(result.get('SK 0 > Subkey tag')).to.equal('[-] 7');
+      expect(result.get('SK 0 > Subkey object type')).to.equal('[-] SecretSubkeyPacket');
```
rrrooommmaaa, line 1358: There might be a way to restore the original behavior using `constructor::tag`.
I haven't tried it yet, should I, @tomholub ?

package.json
rrrooommmaaa: I removed the extra packages

extension/js/common/core/crypto/pgp/msg-util.ts
rrrooommmaaa: removed it

extension/js/common/core/crypto/pgp/msg-util.ts
rrrooommmaaa: renamed it

test/source/tests/unit-node.ts
tomholub: this looks good as is

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: Yes, this generally follow the previous behavior, but less descriptive, e.g.
`Misformed armored text` instead of `Cannot parse OpenPGP key: Misformed armored text for: ...`
I added unit tests for this.

build.sh
```diff
@@ -65,14 +66,20 @@ mkdir -p $OUTDIR/lib/bootstrap
 cp node_modules/bootstrap/dist/js/bootstrap.min.js $OUTDIR/lib/bootstrap/bootstrap.min.js
 cp node_modules/bootstrap/dist/css/bootstrap.min.css $OUTDIR/lib/bootstrap/bootstrap.min.css
 
+mkdir -p $OUTDIR/lib/streams
+cp node_modules/@openpgp/web-stream-tools/lib/*.js $OUTDIR/lib/streams
+# patch imports with .js, e.g. replace './streams' with './streams.js'
+# until https://github.com/openpgpjs/web-stream-tools/pull/20 is resolved
+sed -i -E "s/'\.\/(streams|util|writer|reader|node-conversions)'/'\.\/\1\.js'/g" $OUTDIR/lib/streams/*
```
rrrooommmaaa, line 73: I can add a more complex regex to search for
`import { ... } from './file'`
and `import * as ... from './file'`

What do you think?

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: Right, I renamed this method to `getOrCreateRevocationCertificate()` as it either extracts the revocation certificate from the key (both public and private) if it's already revoked, or creates a new revocation certificate (if the key is private, otherwise throws an exception).
I also added `applyRevocationCertificate()` method and a unit test.

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: I added this into this PR

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: Done

extension/js/common/core/crypto/key.ts
```diff
@@ -203,7 +204,7 @@ export class KeyUtil {
       allErr: Error[] = [];
     let uncheckedOpgpKeyCount = 0;
     try {
-      const { keys, err } = await opgp.key.read(key);
+      const keys = await opgp.readKeys({ binaryKeys: key }); // todo: opgp.readKey ?
```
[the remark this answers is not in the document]
rrrooommmaaa: The new `opgp.readKeys` function differs in that it doesn't wrap each `new PrivateKey/new PublicKey` constructor in a `try/catch` block unlike the 4.10.10 implementation. So if we stumble on the some non-last key, the remaining keys won't be processed.
That got me confused

extension/js/common/core/crypto/key.ts
rrrooommmaaa: The results may differ compared to 4.10.10 implementation so we need to decide how to deal with this (see comment above).

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: And also a concern about premature processing stop (same as with `KeyUtil.readBinary`)

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: If an exception occurs, `formatResultAsync` will display the exception data in square brackets
```
    } catch (e) {
      return `[${String(e)}]`;
    }
```

build.sh
tomholub: It's a temporary fix, right? I think ok as is.

tomholub (dismissed):
reviewing delta

extension/js/common/core/crypto/key.ts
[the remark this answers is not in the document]
tomholub: Ah. Not ideal but ok to proceed as is and file an issue.

extension/js/common/core/crypto/key.ts
tomholub: let's file the comment above for another issue.

extension/js/common/core/crypto/key.ts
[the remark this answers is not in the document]
tomholub: also https://github.com/FlowCrypt/flowcrypt-browser/pull/4725#discussion_r1089731626

extension/js/common/core/crypto/pgp/openpgp-key.ts
tomholub: mentioned here https://github.com/FlowCrypt/flowcrypt-browser/pull/4725#discussion_r1090981268

extension/js/common/core/crypto/pgp/openpgp-key.ts
tomholub: sounds good

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -618,127 +610,150 @@ export class OpenPGPKey {
...
-      opgp.enums.signature.subkey_binding,
-      dataToVerify,
-      date
-    );
+    const bindingSignature = await OpenPGPKey.getLatestValidSignature(subKey.bindingSignatures, primaryKey, opgp.enums.signature.subkeyBinding, dataToVerify);
     if (
       bindingSignature &&
       bindingSignature.embeddedSignature &&
-      (await OpenPGPKey.getLatestValidSignature([bindingSignature.embeddedSignature], subKey.keyPacket, opgp.enums.signature.key_binding, dataToVerify, date))
+      (await OpenPGPKey.getLatestValidSignature([bindingSignature.embeddedSignature], subKey.keyPacket, opgp.enums.signature.keyBinding, dataToVerify))
```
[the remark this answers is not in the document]
tomholub: I'm not sure and I'm ok to not worry about it if it passes existing tests. However, it would be good to file it for another issue to consider further.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -618,178 +607,226 @@ export class OpenPGPKey {
...
-    if (firstTry) {
-      return firstTry;
+  private static getSigningAndEncryptionKeys = async (key: OpenPGP.Key) => {
+    // encryptionKey and signingKey are obtained with undefined value of the date parameter, which is considered current date
+    const encryptionKey = await Catch.undefinedOnException(key.getEncryptionKey());
+    const signingKey = await Catch.undefinedOnException(key.getSigningKey());
+    const possibleExpirations: number[] = [];
+    const primaryKeyExpiration = OpenPGPKey.getExpirationAsDateOrUndefined(await key.getExpirationTime())?.getTime();
+    if (!encryptionKey || !signingKey) {
+      possibleExpirations.push(
```
rrrooommmaaa, line 744: I reworked this method to make use of `getSigningKey` and `getEncryptionKey` calls from OpenPGP.js instead of trying to replicate OpenPGP.js behaviour when selecting keys, but it turned out to be not so simple. Supplying `date: null` (meaning: ignore date) parameter instead of `date: undefined` (meaning: take current date) it is possible to find a (sub)key for singing/encryption, but it may not be the subkey we're looking for ( 1) the subkey may have `created` property after primary key's expiration is in the `key was never usable` unit test, 2) the subkey may tell us an incorrect expiration date, as there may be another subkey with later expiration that didn't show up because OpenPGP.js sorts subkeys by `created` property)
So, for already expired keys the following algo is implemented:
1. create a list of all the subkey's expiration dates (prior to primary key's expiration) 
(todo: we can make it faster by manually collecting expirations from signatures?)
2. call  `getEncryptionKey`/`getSigningKey`  with dates from the list in descending order until we get a usable key.
If a not-expired encryption key is returned for the current date, then we'll call `getEncryptionKey` with the expiration date of the found key to find a next valid key to figure out the final `expiration` date over all the subkeys.
What do you think, @tomholub ? Is this approach good?

tomholub (commented):
delta review - looks good

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: Perhaps, it's outdated.
The keys coming from the background are now `wrapped` (packed) as of this PR

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: Done

rrrooommmaaa:
> Please also do rudimentary manual test on Firefox, to double check nothing is horribly broken, like the content script not loading or not working properly, and setup/send/receive emails with attachments.

I tested the extension in Firefox and it seems to setup and send/receive messages with attachments ok

tomholub (approved):

PR #4732, opened 2022-10-06
#4698 move live test to mock test

### LINKED ISSUE #4698: Move live tests to mock test
Several live test can be moved to mock test.
Ex: `mail.google.com - setup prompt notif + hides when close clicked + reappears + setup link opens settings`, `mail.google.com/chat`

ioanmo226:
This PR moved several live test to mock test

close #4698 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/mock/lib/oauth.ts
+++ b/test/source/mock/lib/oauth.ts
@@ -60,17 +60,16 @@ export class OauthMock {
     }
   };
 
-  public checkAuthorizationHeaderWithAccessToken = (authorization: string | undefined, checkAutorization = true) => {
+  public checkAuthorizationHeaderWithAccessToken = (authorization: string | undefined, fallbackAcct?: string) => {
     if (!authorization) {
       throw new HttpClientErr('Missing mock bearer authorization header', Status.UNAUTHORIZED);
     }
     const accessToken = authorization.replace(/^Bearer /, '');
-    const acct = this.acctByAccessToken[accessToken];
-    if (!acct && checkAutorization) {
+    const acct = this.acctByAccessToken[accessToken] ?? fallbackAcct;
+    if (!acct) {
       throw new HttpClientErr('Invalid mock auth token', Status.UNAUTHORIZED);
     }
-    // Return flowcrypt.compatibility@gmail.com when authorization is not needed
-    return acct ?? 'flowcrypt.compatibility@gmail.com';
+    return acct;
   };
 
   /**
--- a/test/source/mock/lib/oauth.ts
+++ b/test/source/mock/lib/oauth.ts
@@ -60,16 +60,17 @@ export class OauthMock {
     }
   };
 
-  public checkAuthorizationHeaderWithAccessToken = (authorization: string | undefined) => {
+  public checkAuthorizationHeaderWithAccessToken = (authorization: string | undefined, checkAutorization = true) => {
     if (!authorization) {
       throw new HttpClientErr('Missing mock bearer authorization header', Status.UNAUTHORIZED);
     }
     const accessToken = authorization.replace(/^Bearer /, '');
     const acct = this.acctByAccessToken[accessToken];
-    if (!acct) {
+    if (!acct && checkAutorization) {
       throw new HttpClientErr('Invalid mock auth token', Status.UNAUTHORIZED);
     }
-    return acct;
+    // Return flowcrypt.compatibility@gmail.com when authorization is not needed
+    return acct ?? 'flowcrypt.compatibility@gmail.com';
   };
 
   /**
--- a/test/source/tests/setup.ts
+++ b/test/source/tests/setup.ts
@@ -483,7 +483,7 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, await getAuthorizationHeader(t, browser, acct));
       await gmailPage.waitAll(['@webmail-notification', '@notification-setup-action-open-settings', '@notification-setup-action-dismiss', '@notification-setup-action-close']);
       await gmailPage.waitAndClick('@notification-setup-action-close', { confirmGone: true });
-      gmailPage.page.reload();
+      await gmailPage.page.reload();
       await gmailPage.waitAll(['@webmail-notification', '@notification-setup-action-open-settings', '@notification-setup-action-dismiss', '@notification-setup-action-close']);
       const newSettingsPage = await browser.newPageTriggeredBy(t, () => gmailPage.waitAndClick('@notification-setup-action-open-settings'));
       await newSettingsPage.waitAll('@action-connect-to-gmail');
--- a/test/source/tests/setup.ts
+++ b/test/source/tests/setup.ts
@@ -423,10 +423,9 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
 
     ava.default('test re-auth after updating chrome extension', testWithBrowser('compatibility', async (t, browser) => {
       const acctEmail = 'flowcrypt.compatibility@gmail.com';
-      const extraAuthHeaders = await getAuthorizationHeader(t, browser, acctEmail);
       // Wipe google tokens to test re-auth popup
       await Util.wipeGoogleTokensUsingExperimentalSettingsPage(t, browser, acctEmail);
-      const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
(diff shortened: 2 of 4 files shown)


tomholub:
excellent, this is badly needed

ioanmo226:
@tomholub To mock other tests like below one, we need to mock gmail detail page and list page simliar to real gmail page.
(Same classes, same DOM architecture, etc).
I'm gonna work on this.
But before doing, this I wanna ask you if it's worth our time to mock this.

```ts
ava.default('mail.google.com - pubkey gets rendered with new signed and encrypted Thunderbird signature', testWithBrowser('ci.tests.gmail', async (t, browser) => {
      const gmailPage = await openGmailPage(t, browser);
      await gotoGmailPage(gmailPage, '/FMfcgzGkbDZKPLBqWFzbgWqCrplTQdNz');
      const pgpBlockUrls = await gmailPage.getFramesUrls(['/chrome/elements/pgp_block.htm'], { sleep: 10, appearIn: 25 });
      const url = pgpBlockUrls[0].split('/chrome/elements/pgp_block.htm')[1];
      await BrowserRecipe.pgpBlockVerifyDecryptedContent(t, browser, {
        params: url,
        content: ['Encrypted Subject: [ci.test] Thunderbird html signed + encrypted', '1234'],
        encryption: 'encrypted',
        signature: 'not signed'
      });
      await pageHasSecureReplyContainer(t, browser, gmailPage);
      const pubkeyPage = await gmailPage.getFrame(['/chrome/elements/pgp_pubkey.htm']);
    }));
```

tomholub:
I would say not yet. I would, at this stage, only mock the kind of tests that only use minimal Gmail DOM. Especially the ones about setup, notifications, and using the compose button.

There is some value in having the more DOM-reliant tests run live. Because from time to time, Gmail actually changes its DOM, and it would be good to catch these situations, which can only be done with live tests, unfortunately.

tomholub:
Although I hoped we have more such live tests that could be transferred over - do all of the other ones use Gmail DOM heavily?

tomholub:
sounds good

test/source/mock/lib/oauth.ts
```diff
@@ -60,16 +60,17 @@ export class OauthMock {
     }
   };
 
-  public checkAuthorizationHeaderWithAccessToken = (authorization: string | undefined) => {
+  public checkAuthorizationHeaderWithAccessToken = (authorization: string | undefined, checkAutorization = true) => {
```
rrrooommmaaa, line 63: It looks more like `fallbackAcct` parameter, because even if `checkAuthorization = false`, the lookup happens anyway.
I suggest to refactor the parameter to `fallbackAcct?: string` and pass it from the `google-endpoints.ts` method

rrrooommmaaa (changes requested):
I suggest to refactor `checkAuthorizationHeaderWithAccessToken` as described

test/source/mock/lib/oauth.ts
ioanmo226: Sounds good 👍

test/source/tests/setup.ts
```diff
@@ -436,6 +446,60 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+      // reload and test that it has no more notifications
+      await gmailPage.page.reload();
+      await gmailPage.waitAndClick('@action-secure-compose');
+      await Util.sleep(1);
+      await gmailPage.notPresent(['@webmail-notification']);
+    }));
+
+    ava.default('mail.google.com - setup prompt notif + hides when close clicked + reappears + setup link opens settings', testWithBrowser(undefined, async (t, browser) => {
+      const acct = 'flowcrypt.compatibility@gmail.com';
+      const settingsPage = await openMockGmailPage(t, browser, acct);
```
rrrooommmaaa, line 481: What are lines 481 and 482 actually needed?

test/source/tests/setup.ts
```diff
@@ -436,6 +446,60 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+      await settingsPage.close();
+      const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, await getAuthorizationHeader(t, browser, acct));
```
rrrooommmaaa, line 483: how is this different from `const gmailPage = await openMockGmailPage(t, browser, acct);`?

test/source/tests/setup.ts
```diff
@@ -436,6 +446,60 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+      await gmailPage.waitAndClick('@notification-setup-action-close', { confirmGone: true });
+      await gmailPage.page.reload();
+      await gmailPage.waitAll(['@webmail-notification', '@notification-setup-action-open-settings', '@notification-setup-action-dismiss', '@notification-setup-action-close']);
+      const newSettingsPage = await browser.newPageTriggeredBy(t, () => gmailPage.waitAndClick('@notification-setup-action-open-settings'));
+      await newSettingsPage.waitAll('@action-connect-to-gmail');
+    }));
+
+    ava.default('mail.google.com - setup prompt notification shows up + dismiss hides it + does not reappear if dismissed', testWithBrowser(undefined, async (t, browser) => {
+      const acct = 'flowcrypt.compatibility@gmail.com';
+      const settingsPage = await BrowserRecipe.openSettingsLoginButCloseOauthWindowBeforeGrantingPermission(t, browser, acct);
```
rrrooommmaaa, line 494: Looks like lines 494 and 495 don't affect anything?

test/source/tests/setup.ts
```diff
@@ -436,6 +446,60 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
       await gmailPage.notPresent(['@webmail-notification']);
     }));
 
+    ava.default('mail.google.com - success notif after setup, click hides it, does not re-appear + offers to reauth', testWithBrowser('compatibility', async (t, browser) => {
+      const acct = 'flowcrypt.compatibility@gmail.com';
+      const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, await getAuthorizationHeader(t, browser, acct));
```
rrrooommmaaa, line 451: can substitute with the shortcut `const gmailPage = await openMockGmailPage(t, browser, acct);`

test/source/tests/setup.ts
```diff
@@ -17,14 +17,26 @@ import { testConstants } from './tooling/consts';
...
+import { BrowserHandle, ControllablePage } from '../browser';
 import { OauthPageRecipe } from './page-recipe/oauth-page-recipe';
+import { AvaContext } from './tooling';
 
 // tslint:disable:no-blank-lines-func
 // tslint:disable:no-unused-expression
 /* eslint-disable no-unused-expressions */
 /* eslint-disable max-len */
 
+const getAuthorizationHeader = async (t: AvaContext, browser: BrowserHandle, acctEmail: string) => {
```
rrrooommmaaa, line 29: Does this always return `Bearer undefined` for the new 3 tests?
If so, why bother? We could just as well send `Bearer undefined` or something more descriptive like `Bearer emulating-not-properly-registered-extension` right away.

test/source/tests/setup.ts
rrrooommmaaa: or something like 
`await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, { Authorization: 'Bearer no-access-for-extension'});` or some other dummy descriptive token might work

test/source/tests/setup.ts
rrrooommmaaa: or something like 
`await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, { Authorization: 'Bearer no-access-for-extension'});` or some other dummy descriptive token might work

test/source/tests/setup.ts
rrrooommmaaa: If we want to double-check the test setup itself (the fact that the extension isn't actually set up), we can try to extract google access token and `expect` it to be `undefined` right here (or in the beginning of each test). This would be more obvious, then sending incorrectly extracted value.

test/source/tests/setup.ts
rrrooommmaaa: If they are a separate mini-test, it'd be better to put them in a separate test.

test/source/tests/setup.ts
ioanmo226: It's indeed not needed

test/source/tests/setup.ts
ioanmo226: You are right, Refactoring is needed

rrrooommmaaa (approved):

PR #4736, opened 2022-10-13
#3824 filter sensitive fields from logs

### LINKED ISSUE #3824: sensitive fields should be filtered from logs
```
### Catch.reportErr calling stack ###
# at Function.Catch.formatExceptionForReport (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/platform/catch.js:260:90)
# at Function.Catch.onErrorInternalHandler (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/platform/catch.js:74:29)
# at Function.Catch.reportErr (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/platform/catch.js:90:18)
# at PgpBlockViewErrorModule.handleInitializeErr (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/pgp_block_modules/pgp-block-error-module.js:93:27)
# at PgpBlockViewDecryptModule.initialize (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/pgp_block_modules/pgp-block-decrypt-module.js:56:49)
# at async PgpBlockView.render (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/pgp_block.js:30:21)
# at async chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/view.js:27:21
#
# url: chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/pgp_block.htm?frameId=frame_djELzzcyYx&message=blahblah&etc
```

On all pages, the following parameters should be filtered in the report, in the url at the bottom:
 - senderEmail
 - acctEmail
 - message

Instead of the original content, please show `[SCRUBBED]`. I think we already do something similar in `onErrorInternalHandler` so if it can be implemented there, that would be good.

ioanmo226:
This PR filters sensitive fields from logs

close #3824 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -316,7 +316,7 @@ export class Url {
   };
 
   public static replaceUrlParam = (url: string, key: string, value: string) => {
-    const regex = new RegExp(`(${key}=).*?(&|$)`);
+    const regex = new RegExp(`([?:\?|&]${key}=).*?(&|$)`, 'i');
     return url.replace(regex, '$1' + value + '$2');
   };
 
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -315,6 +315,11 @@ export class Url {
     return `${urlParts[0]}?${params.toString()}`;
   };
 
+  public static replaceUrlParam = (url: string, key: string, value: string) => {
+    const regex = new RegExp(`(${key}=).*?(&)`);
+    return url.replace(regex, '$1' + value + '$2');
+  };
+
   private static snakeCaseToCamelCase = (s: string) => {
     return s.replace(/_[a-z]/g, boundary => boundary[1].toUpperCase());
   };
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -316,7 +316,7 @@ export class Url {
   };
 
   public static replaceUrlParam = (url: string, key: string, value: string) => {
-    const regex = new RegExp(`([?:\?|&]${key}=).*?(&|$)`, 'i');
+    const regex = new RegExp(`([?|&]${key}=).*?(&|$)`, 'i');
     return url.replace(regex, '$1' + value + '$2');
   };
 
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -316,7 +316,7 @@ export class Url {
   };
 
   public static replaceUrlParam = (url: string, key: string, value: string) => {
-    const regex = new RegExp(`(${key}=).*?(&)`);
+    const regex = new RegExp(`(${key}=).*?(&|$)`);
     return url.replace(regex, '$1' + value + '$2');
   };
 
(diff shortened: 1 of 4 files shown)


test/source/tests/browser-unit-tests/unit-Catch.js
```diff
@@ -48,3 +48,16 @@ BROWSER_UNIT_TEST_NAME(`Catcher does report on consumer`).consumer;
   }
   return "pass";
 })();
+
+BROWSER_UNIT_TEST_NAME(`Catcher does report sensitive infos`);
+(async () => {
+  const url = 'chrome://extension-id/pgp_block.htm?frameId=id&message=blahblah&senderEmail=blahblah&acctEmail=blahblah';
+  const censoredUrl = Catch.censoredUrl(url);
+  const urlParams = Url.parse(['message', 'senderEmail', 'acctEmail'], censoredUrl);
```
rrrooommmaaa, line 56: Why don't we just compare `censoredUrl` against the expected string?
`chrome://extension-id/pgp_block.htm?frameId=id&message=[SCRUBBED]&senderEmail=[SCRUBBED]&acctEmail=[SCRUBBED]`
This test doesn't show that `frameId` remains not scrubbed.

extension/js/common/core/common.ts
```diff
@@ -315,6 +315,11 @@ export class Url {
     return `${urlParts[0]}?${params.toString()}`;
   };
 
+  public static replaceUrlParam = (url: string, key: string, value: string) => {
+    const regex = new RegExp(`(${key}=).*?(&|$)`);
```
rrrooommmaaa, line 319: This also reacts to prefixed keys, like `xyzacctEmail=`.
We can specify a non-capturing group requiring the match to start from either ? or & (or beginning of string?)
`((?:\?|&)acctEmail=).*?(&|$)`
I'd also suggest to add case-insensitive option 'i'.

test/source/tests/browser-unit-tests/unit-Catch.js
rrrooommmaaa: It'd be also nice to add more edge test cases with scrubbed parameters interspersed with non-scrubbed. e.g.
`chrome://extension-id/pgp_block.htm?frameId=id&message=blahblah&some=1&senderEmail=blahblah&bbb=123`
that make sure our regex is acting in greedy mode and leaves `frameId`, `some` and `bbb` unscrubbed

extension/js/common/core/common.ts
rrrooommmaaa: Perhaps, for better security we can leave the regex as it is to scrub even prefixed parameters, but adding case-insensitivity option 'i' for better security would also be good.

test/source/tests/browser-unit-tests/unit-Catch.js
ioanmo226: Trying to display which params didn't exactly scrubbed. But comparing against expected string also looks okay to me.
> Why don't we just compare censoredUrl against the expected string?

rrrooommmaaa (approved):

PR #4742, opened 2022-10-19
#4737 Oddity when replying to a multiple-recipient encrypted message

### LINKED ISSUE #4737: Oddity when replying to a multiple-recipient encrypted message
When pressing "Secure Reply" button on this message https://mail.google.com/mail/u/0/#inbox/FMfcgzGqQvvQQTzfBdzFCTbrWRWjFJnL,
I can only see one of the recipients in `To:` field (the last one in the original message's `To:` list) -- glitch 1.
When I clicking on the recipient, the `To:` field control is loaded with the original message's sender email address in its place -- glitch 2.
When I trying to delete the recipient by clicking on the `x`, I'm getting this (glitch 3):
image

And on clicking "Encrypt, Sign and Send" button, the reply is sent to the last recipient of the original message's `To:` list.

Expected behaviour (if "Secure Reply" means replying to the original sender only): The original sender's email address should appear in `To:` list.
Question: Should I be able to add more "To:", "Cc:" or "Bcc:" recipients or remove the original sender from the list?
Anyhow, there shouldn't be the "unexpected error" popup on trying to delete the original sender.

ioanmo226:
This PR fixed oddity when replying to a multiple-recipient encrypted message

close #4737 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


rrrooommmaaa (approved):
Good catch!

PR #4751, opened 2022-10-20
#4740 Enable users to change reply option while compose

### LINKED ISSUE #4740: Show reply/reply all/forward options through composing message
One thing is though, Gmail also has a way for me to switch to "reply all" or "forward" after the reply box is already open. Even mid way through composing my message. That's why they don't need the extra step, and that's why they don't need so many buttons. So I suppose we may need to do that first?

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4496#issuecomment-1227609174_

ioanmo226:
This PR enables users to change reply option while compose

close #4740 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


test/source/mock/google/exported-messages/message-export-183ec175f060b2c2.json
```diff
@@ -0,0 +1,76 @@
+{
```
ioanmo226, line 1: @rrrooommmaaa FYI Need to review https://github.com/FlowCrypt/flowcrypt-browser/pull/4742 first

ioanmo226:
chrome-capture-2022-9-21

extension/chrome/elements/compose-modules/compose-render-module.ts
```diff
@@ -191,6 +195,25 @@ export class ComposeRenderModule extends ViewModule<ComposeView> {
...
+  public changeReplyOption = async (option: ReplyOptions) => {
+    if (!this.view.replyParams) {
+      return;
+    }
+    this.view.recipientsModule.clearRecipients();
+    if (option === 'a_forward') {
+      await this.view.quoteModule.addTripleDotQuoteExpandFooterAndQuoteBtn(this.view.replyMsgId, 'forward');
+    } else {
+      const recipients: Recipients = { to: this.view.replyParams.to, cc: this.view.replyParams.cc, bcc: this.view.replyParams.bcc };
+      this.view.recipientsModule.addRecipients(recipients, false).catch(Catch.reportErr);
```
rrrooommmaaa, line 207: We should use `this.view.replyParams` directly, as the new object still has references to the old object's arrays, and we would need to care about all the places they could be possibly modified.
`this.view.recipientsModule.addRecipients(this.view.replyParams, ...`

extension/chrome/elements/compose-modules/compose-reply-btn-popover-module.ts
```diff
@@ -0,0 +1,71 @@
...
+import { ViewModule } from '../../../js/common/view-module.js';
+import { ComposeView } from '../compose.js';
+
+export type ReplyOptions = 'a_reply' | 'a_reply_all' | 'a_forward';
+
+export class ComposeReplyBtnPopoverModule extends ViewModule<ComposeView> {
+
+  private popoverItems: Record<ReplyOptions, { text: string, iconPath: string }> = {
+    a_reply: { text: 'Reply', iconPath: '/img/reply-icon.png' },
+    a_reply_all: { text: 'Replay All', iconPath: '/img/reply-all-icon.png' },
```
rrrooommmaaa, line 16: There is a typo: Repl`a`y All

test/source/tests/compose.ts
```diff
@@ -816,6 +816,20 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      await composePage.waitAndClick('@action-accept-reply-all-prompt');
+      await composePage.waitForContent('@recipients-preview', 'sender@domain.comtest@gmail.comtest2@gmail.comtest3@gmail.comtest4@gmail.comtest5@gmail.com');
+      await composePage.waitAndClick('@action-show-reply-options-popover');
+      await composePage.waitAndClick('@action-toggle-a_reply');
+      await composePage.waitForContent('@recipients-preview', 'sender@domain.com');
+      await composePage.waitAndClick('@action-show-reply-options-popover');
+      await composePage.waitAndClick('@action-toggle-a_forward');
+      await composePage.waitUntilFocused('@input-to');
+      await expectRecipientElements(composePage, { to: [], cc: [], bcc: [] });
+    }));
```
rrrooommmaaa, line 831: Why don't we test `@action-toggle-a_reply_all` while we're at it?

rrrooommmaaa (changes requested):
It looked good when I tested it.
There is one typo and a couple of tests that I think are missing.

rrrooommmaaa (approved):

PR #4761, opened 2022-10-27
#4496 fix: switch to encrypted reply

### LINKED ISSUE #4496: "switch to encrypted reply" no longer works
image

To reproduce, in my case:
 - receive several encrypted messages in a thread
 - pick the one in the middle
 - click "reply all" using standard gmail interface
 - then see above banner
 - clicking `Switch to encrypted reply` doesn't work

Actually, the reason this happened is because I didn't see any "encrypted reply" option on the message itself. It seems that it's missing on encrypted messages that have many recipients:

image

but it's present and functioning on encrypted messages with just me as a recipient:

image

So that needs to be addressed as well.

ioanmo226:
This PR fixes switch to encrypted reply issue for messages (Thread messages which is not the last message)

close #4496 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/gmail.ts
+++ b/test/source/tests/gmail.ts
@@ -302,7 +302,6 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       await gmailPage.waitAndClick('[role="listitem"] .adf.ads', { delay: 1 }); // click first message of thread
       await Util.sleep(3);
       await gmailPage.waitAndClick('[data-tooltip="Reply"]', { delay: 1 });
-      await gmailPage.waitTillGone('.reply_message');
       await gmailPage.waitAll('[data-tooltip^="Send"]'); // The Send button from the Standard reply box
       await gmailPage.waitForContent('.reply_message_evaluated .error_notification', 'The last message was encrypted, but you are composing a reply without encryption.');
       await gmailPage.waitAndClick('[data-tooltip="Secure Reply"]'); // Switch to encrypted reply
--- a/test/source/tests/gmail.ts
+++ b/test/source/tests/gmail.ts
@@ -24,10 +24,10 @@ export type GmailCategory = 'inbox' | 'sent' | 'drafts' | 'spam' | 'trash';
 export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: TestWithBrowser) => {
 
   if (testVariant === 'CONSUMER-LIVE-GMAIL') {
+    const pageHasSecureReplyContainer = async (t: AvaContext, browser: BrowserHandle, gmailPage: ControllablePage, { isReplyPromptAccepted, composeCount = 1}: { isReplyPromptAccepted?: boolean, composeCount?: number } = {}) => {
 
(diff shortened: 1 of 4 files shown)


test/source/tests/gmail.ts
```diff
@@ -24,10 +24,21 @@ export type GmailCategory = 'inbox' | 'sent' | 'drafts' | 'spam' | 'trash';
...
+      t: AvaContext,
+      browser: BrowserHandle,
+      gmailPage: ControllablePage,
+      {
+        isReplyPromptAccepted,
+        composeCount
+      }:
+      {
+        isReplyPromptAccepted?: boolean,
+        composeCount?: number
```
rrrooommmaaa, line 37: I suggest a more precise name: `composeFrameCount`

test/source/tests/gmail.ts
```diff
@@ -296,6 +307,19 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       await gmailPage.notPresent('.reply_message_evaluated .error_notification'); // should not show the warning about switching to encrypted reply
     }));
 
+    ava.default('mail.google.com - switch to encrypted reply for middle message', testWithBrowser('ci.tests.gmail', async (t, browser) => {
+      const gmailPage = await openGmailPage(t, browser);
+      await gotoGmailPage(gmailPage, '/FMfcgzGqRGfPBbNLWvfPvDbxnHBwkdGf'); // plain convo
```
rrrooommmaaa, line 312: About a year ago we started to put test messages from the live account to the `test/source/mock/google/live-gmail-test-emails` folder in plain EML format to be able to restore the message in case it's accidentally deleted from the server. It's appropriate to have the id from the test `FMfcgzGqRGfPBbNLWvfPvDbxnHBwkdGf` in the file name.
Of course, need to take care that the message doesn't contain any sensitive data (such as email addresses of people who wouldn't want to disclose their email addresses).

rrrooommmaaa (changes requested):
The implementation looking good.
There are a few things we can update in tests, please see comments.

test/source/tests/gmail.ts
```diff
@@ -24,10 +24,21 @@ export type GmailCategory = 'inbox' | 'sent' | 'drafts' | 'spam' | 'trash';
...
+      {
+        isReplyPromptAccepted?: boolean,
+        composeCount?: number
+      } = {}
+    ) => {
       const urls = await gmailPage.getFramesUrls(['/chrome/elements/compose.htm'], { sleep: 0 });
-      expect(urls.length).to.equal(1);
+      expect(urls.length).to.equal(composeCount ?? 1);
       if (typeof isReplyPromptAccepted !== 'undefined') {
         const replyBox = await browser.newPage(t, urls[0]);
```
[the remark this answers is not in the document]
ioanmo226: To be frank, I also agree with you.
I think original author's thought was like this.
Will follow your way.

> Is this because in our particular test the "secure reply compose container" is located higher than the other "reply or forward buttons" container

rrrooommmaaa (approved):

PR #4768, opened 2022-11-02
#4703 Notify users when their local keys expire soon

### LINKED ISSUE #4703: notify users when their keys expire soon
If user keys expire within the next 30 days, add the following notification:

- if EKM is used: `Your local keys expire in X days. To receive the latest keys, please ensure that you can connect to your corporate network either through VPN or in person and reload Gmail. If this notification still shows after that, please contact your Help Desk.`
- else, if `Your keys are expiring in X days. Please import a newer set of keys to use.`

If EKM keys are being repulled automatically, then this check should run after EKM keys were pulled when browser extension is being loaded (successfully or not). So that the user is able to actually follow the instructions: connect to VPN and reload the extension, which will make the problem go away (instead they will see a successful message from the EKM sync).

For a test, when using EKM:
 - set up with EKM and a key that expires in 29 days
 - tell EKM to start returning error 500 for this user
 - load Gmail mock
 - observe that there is the warning notification
 - re-enable mock EKM
 - add **another key** that doesn't expire for that user on EKM (now, both keys will be returned. This mimics prod behavior during key rotation)
 - reload Gmail mock
 - observe "keys successfully updated" or similar message (already implemented)

Second test, not using EKM:
 - set up with a key that expires in 29 days
 - load Gmail mock
 - observe that there is the warning notification
 - import **another key** in the settings
 - observe no warning
 - remove that added **another key** in settings
 - observe warning again
 - add **an updated version of the original key**, that was updated to not expire
 - observe no warning

When generating keys for this test, you'll have to do it during the running of each test using OpenPGP.js instead of checking the keys in, because of the expiration that is relative to current time. Make sure these are ECC keys (eg curve25519, should be the default) so that the keygen is fast.

ioanmo226:
This PR notifys users when their local keys expire soon

close #4703 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/helpers.ts
+++ b/extension/js/common/helpers.ts
@@ -164,3 +164,12 @@ export const processAndStoreKeysFromEkmLocally = async (
   }
   return { updateCount: encryptedKeys?.keys.length ?? 0 + (existingKeys.length - keysToRetain.length), noKeysSetup: !(encryptedKeys?.keys.length || keysToRetain.length) };
 };
+
+export const getLocalKeyExpiration = async ({ acctEmail }: Bm.GetLocalKeyExpiration): Promise<Bm.Res.GetLocalKeyExpiration> => {
+  const keys = await KeyStore.get(acctEmail);
+  const keyInfos = await Promise.all(keys.map(async (key) => await KeyUtil.parse(key.public)));
+  console.log(keyInfos);
+  const keyExpiration = keyInfos
+    .sort((key1, key2) => (key1.expiration ?? 0) < (key2.expiration ?? 0) ? 1 : -1)[0]?.expiration;
+  return keyExpiration;
+}
--- a/extension/js/common/helpers.ts
+++ b/extension/js/common/helpers.ts
@@ -172,4 +172,4 @@ export const getLocalKeyExpiration = async ({ acctEmail }: Bm.GetLocalKeyExpirat
   const keyExpiration = keyInfos
     .sort((key1, key2) => (key1.expiration ?? 0) < (key2.expiration ?? 0) ? 1 : -1)[0]?.expiration;
   return keyExpiration;
-}
+};
--- a/extension/js/content_scripts/webmail/setup-webmail-content-script.ts
+++ b/extension/js/content_scripts/webmail/setup-webmail-content-script.ts
@@ -330,13 +330,13 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
     }
     let warningMsg;
     if (clientConfiguration.usesKeyManager()) {
-      warningMsg = `Your local keys expire in ${expireInDays} days.<br/>To receive the latest keys, please ensure that you can connect to your corporate network either through VPN or in person and reload Gmail.<br/>` + 
+      warningMsg = `Your local keys expire in ${expireInDays} days.<br/>` +
+      `To receive the latest keys, please ensure that you can connect to your corporate network either through VPN or in person and reload Gmail.<br/>` +
       `If this notification still shows after that, please contact your Help Desk.`;
     } else {
       warningMsg = `Your keys are expiring in ${expireInDays} days. Please import a newer set of keys to use.`;
     }
     warningMsg += `<a href="#" class="close" data-test="notification-close-expiration-popup">close</a>`;
-    console.log(warningMsg);
     notifications.clear();
     notifications.show(warningMsg);
   };
--- a/extension/js/content_scripts/webmail/setup-webmail-content-script.ts
+++ b/extension/js/content_scripts/webmail/setup-webmail-content-script.ts
@@ -2,26 +2,26 @@
 
 'use strict';
 
-import { Bm, BrowserMsg, TabIdRequiredError } from '../../common/browser/browser-msg.js';
-import { Env, WebMailName } from '../../common/browser/env.js';
-import { WebmailVariantString, XssSafeFactory } from '../../common/xss-safe-factory.js';
-import { BrowserMsgCommonHandlers } from '../../common/browser/browser-msg-common-handlers.js';
-import { Catch } from '../../common/platform/catch.js';
-import { ContentScriptWindow } from '../../common/browser/browser-window.js';
-import { Injector } from '../../common/inject.js';
-import { Notifications } from '../../common/notifications.js';
 import Swal from 'sweetalert2';
+import { AccountServer } from '../../common/api/account-server.js';
+import { KeyManager } from '../../common/api/key-server/key-manager.js';
+import { ApiErr, BackendAuthErr } from '../../common/api/shared/api-error.js';
+import { BrowserMsgCommonHandlers } from '../../common/browser/browser-msg-common-handlers.js';
+import { Bm, BrowserMsg, TabIdRequiredError } from '../../common/browser/browser-msg.js';
+import { ContentScriptWindow } from '../../common/browser/browser-window.js';
+import { Env, WebMailName } from '../../common/browser/env.js';
 import { Ui } from '../../common/browser/ui.js';
+import { ClientConfiguration } from '../../common/client-configuration.js';
+import { Url } from '../../common/core/common.js';
 import { InMemoryStoreKeys, VERSION } from '../../common/core/const.js';
+import { Injector } from '../../common/inject.js';
+import { Lang } from '../../common/lang.js';
+import { Notifications } from '../../common/notifications.js';
+import { Catch } from '../../common/platform/catch.js';
 import { AcctStore } from '../../common/platform/store/acct-store.js';
 import { GlobalStore } from '../../common/platform/store/global-store.js';
-import { ClientConfiguration } from '../../common/client-configuration.js';
-import { KeyManager } from '../../common/api/key-server/key-manager.js';
 import { InMemoryStore } from '../../common/platform/store/in-memory-store.js';
-import { AccountServer } from '../../common/api/account-server.js';
-import { ApiErr, BackendAuthErr } from '../../common/api/shared/api-error.js';
-import { Url } from '../../common/core/common.js';
-import { Lang } from '../../common/lang.js';
+import { WebmailVariantString, XssSafeFactory } from '../../common/xss-safe-factory.js';
 
 export type WebmailVariantObject = { newDataLayer: undefined | boolean, newUi: undefined | boolean, email: undefined | string, gmailVariant: WebmailVariantString };
 export type IntervalFunction = { interval: number, handler: () => void };
@@ -318,6 +318,29 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
 
   };
 
+  const notifyExpiringKeys = async (acctEmail: string, clientConfiguration: ClientConfiguration, notifications: Notifications) => {
+    const expiration = await BrowserMsg.send.bg.await.getLocalKeyExpiration({acctEmail});
+    if (expiration === undefined) {
+      return;
+    }
+    const expireInDays = Math.ceil((expiration - Date.now()) / 1000 / 60 / 60 / 24);
+    console.log(`expireInDays: ${expireInDays}`);
+    if (expireInDays > 30) {
+      return;
+    }
+    let warningMsg;
+    if (clientConfiguration.usesKeyManager()) {
+      warningMsg = `Your local keys expire in ${expireInDays} days.<br/>To receive the latest keys, please ensure that you can connect to your corporate network either through VPN or in person and reload Gmail.<br/>` + 
+      `If this notification still shows after that, please contact your Help Desk.`;
+    } else {
+      warningMsg = `Your keys are expiring in ${expireInDays} days. Please import a newer set of keys to use.`;
+    }
+    warningMsg += `<a href="#" class="close" data-test="notification-close-expiration-popup">close</a>`;
+    console.log(warningMsg);
+    notifications.clear();
+    notifications.show(warningMsg);
+  };
+
   const entrypoint = async () => {
     try {
       const acctEmail = await waitForAcctEmail();
@@ -329,6 +352,7 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
       const clientConfiguration = await ClientConfiguration.newInstance(acctEmail);
       await startPullingKeysFromEkm(acctEmail, clientConfiguration, factory, ppEvent);
       await webmailSpecific.start(acctEmail, clientConfiguration, inject, notifications, factory, notifyMurdered);
+      await notifyExpiringKeys(acctEmail, clientConfiguration, notifications);
     } catch (e) {
       if (e instanceof TabIdRequiredError) {
         console.error(`FlowCrypt cannot start: ${String(e)}`);
--- a/test/source/mock/key-manager/key-manager-endpoints.ts
+++ b/test/source/mock/key-manager/key-manager-endpoints.ts
@@ -8,6 +8,7 @@ import { Dict } from '../../core/common';
 import { expect } from 'chai';
 import { KeyUtil } from '../../core/crypto/key';
 import { testConstants } from '../../tests/tooling/consts';
+import { opgp } from '../../core/crypto/pgp/openpgpjs-custom';
 
 // tslint:disable:max-line-length
 /* eslint-disable max-len */
@@ -227,6 +228,10 @@ export const mockKeyManagerEndpoints: HandlersDefinition = {
       }
       if (acctEmail === 'get.key@no-submit-client-configuration.key-manager-autogen.flowcrypt.test') {
         return { privateKeys: [{ decryptedPrivateKey: prvNoSubmit }] };
+      }
+      if (acctEmail === 'flowcrypt.notify.expiring.keys@gmail.com') {
+        const key = await opgp.generateKey({ curve: 'curve25519', userIds: [{ email: acctEmail }], keyExpirationTime: 20 * 24 * 60 * 60});
+        return { privateKeys: [{ decryptedPrivateKey:  key.privateKeyArmored }] };
       }
       if (acctEmail === 'two.keys@key-manager-autogen.flowcrypt.test') {
         return { privateKeys: [{ decryptedPrivateKey: twoKeys1 }, { decryptedPrivateKey: twoKeys2 }] };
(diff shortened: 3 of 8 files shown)


ioanmo226:
Without using EKM
image

Using EKM
image

extension/js/common/helpers.ts
```diff
@@ -164,3 +164,11 @@ export const processAndStoreKeysFromEkmLocally = async (
   }
   return { updateCount: encryptedKeys?.keys.length ?? 0 + (existingKeys.length - keysToRetain.length), noKeysSetup: !(encryptedKeys?.keys.length || keysToRetain.length) };
 };
+
+export const getLocalKeyExpiration = async ({ acctEmail }: Bm.GetLocalKeyExpiration): Promise<Bm.Res.GetLocalKeyExpiration> => {
+  const keys = await KeyStore.get(acctEmail);
+  const keyInfos = await Promise.all(keys.map(async (key) => await KeyUtil.parse(key.public)));
+  const keyExpiration = keyInfos
+    .sort((key1, key2) => (key1.expiration ?? 0) < (key2.expiration ?? 0) ? 1 : -1)[0]?.expiration;
```
rrrooommmaaa, line 172: Shouldn't `undefined` expiration be treated like `Infinity`?

extension/js/common/helpers.ts
```diff
@@ -164,3 +164,11 @@ export const processAndStoreKeysFromEkmLocally = async (
   }
   return { updateCount: encryptedKeys?.keys.length ?? 0 + (existingKeys.length - keysToRetain.length), noKeysSetup: !(encryptedKeys?.keys.length || keysToRetain.length) };
 };
+
+export const getLocalKeyExpiration = async ({ acctEmail }: Bm.GetLocalKeyExpiration): Promise<Bm.Res.GetLocalKeyExpiration> => {
+  const keys = await KeyStore.get(acctEmail);
+  const keyInfos = await Promise.all(keys.map(async (key) => await KeyUtil.parse(key.public)));
```
rrrooommmaaa, line 170: We often name the other way around. `KeyStore.get` returns KeyInfo's (kis) -- armored keys with some additional info.
And parsed keys can be named `parsedKeys` or simply `keys`

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -329,6 +351,7 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
       const clientConfiguration = await ClientConfiguration.newInstance(acctEmail);
       await startPullingKeysFromEkm(acctEmail, clientConfiguration, factory, ppEvent);
       await webmailSpecific.start(acctEmail, clientConfiguration, inject, notifications, factory, notifyMurdered);
+      await notifyExpiringKeys(acctEmail, clientConfiguration, notifications);
```
rrrooommmaaa, line 354: Line 352 only starts pulling keys from EKM, the actual pull and processing are lines 295 and 296, so notification should be performed after line 296 (possibly in `catch` or `finally` block to properly handle a fetch error)

test/source/tests/setup.ts
```diff
@@ -496,6 +497,84 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+      await gmailPage.waitForContent('@webmail-notification', warningMsg);
+      const updatedKey = await opgp.generateKey({
+        curve: 'curve25519',
+        userIds: [{ email: acctEmail }, { email: 'demo@gmail.com', name: 'Demo user'}],
+        keyExpirationTime: 100 * 24 * 60 * 60
+      });
+      MOCK_KM_KEYS[acctEmail].response = { privateKeys: [{ decryptedPrivateKey: key.privateKeyArmored }, { decryptedPrivateKey: updatedKey.privateKeyArmored }] };
+      console.log('MOCK_KM_KEYS');
+      await gmailPage.page.reload();
+      await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, 'Account keys updated');
```
rrrooommmaaa, line 573: What happens (or should happen) if one of the keys has no expiration?

test/source/mock/key-manager/key-manager-endpoints.ts
```diff
@@ -218,8 +228,8 @@ export const mockKeyManagerEndpoints: HandlersDefinition = {
       if (acctEmail === 'get.key@key-manager-autogen.flowcrypt.test') {
         return { privateKeys: [{ decryptedPrivateKey: testConstants.existingPrv }] };
       }
-      if (acctEmail.includes('updating.key')) {
-        const { response, badRequestError } = MOCK_KM_UPDATING_KEY[acctEmail];
+      if (acctEmail.includes('updating.key') || acctEmail === 'flowcrypt.notify.expiring.keys@key-manager-autogen.flowcrypt.test') {
```
rrrooommmaaa, line 231: Couldn't we just use the marker `updating.key` (or any other) for the tests that use `MOCK_KM_KEYS`?

test/source/tests/setup.ts
```diff
@@ -496,6 +497,84 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+      });
+      // Setup with above key
+      await SetupPageRecipe.manualEnter(settingsPage, 'unused', {
+        submitPubkey: false,
+        usedPgpBefore: false,
+        key: {
+          title: '?',
+          armored: key.privateKeyArmored,
+          passphrase,
+          longid: '1b383d0334e38b28',
```
rrrooommmaaa, line 521: You just generated a new key and supplying a longid from a totally different key?
If `longid` is not needed, we'd better make it obvious, like
```
longid: '0000000000000000', // dummy -- not needed
```

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
rrrooommmaaa: In case `clientConfiguration.usesKeyManager()` is `false` then we can call it here

extension/js/common/helpers.ts
ioanmo226: Nice catch. Yeah, it should be treated as Infinity.

test/source/tests/setup.ts
ioanmo226: Updated at above.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -286,17 +286,24 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
...
-  const startPullingKeysFromEkm = async (acctEmail: string, clientConfiguration: ClientConfiguration, factory: XssSafeFactory, ppEvent: { entered?: boolean }) => {
-    if (clientConfiguration.usesKeyManager()) {
-      const idToken = await InMemoryStore.get(acctEmail, InMemoryStoreKeys.ID_TOKEN);
-      if (idToken) {
-        const keyManager = new KeyManager(clientConfiguration.getKeyManagerUrlForPrivateKeys()!);
-        Catch.setHandledTimeout(async () => {
-          const { privateKeys } = await keyManager.getPrivateKeys(idToken);
-          await processKeysFromEkm(acctEmail, privateKeys.map(entry => entry.decryptedPrivateKey), clientConfiguration, factory, idToken, ppEvent);
-        }, 0);
+  const startPullingKeysFromEkm = async (acctEmail: string, clientConfiguration: ClientConfiguration, factory: XssSafeFactory, ppEvent: { entered?: boolean }): Promise<void> => {
```
rrrooommmaaa, line 289: So this function is no longer `startPulling...` due to the latest modifications, but actually `pull...`?
I can't see a good reason why we should await for the completion of  `getPrivateKeys` and `processKeysFromEkm` before starting the web client. We can simply pass an additional function to call after line 297 in a `try {} catch {} finally {}` block

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -329,6 +358,7 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
       const clientConfiguration = await ClientConfiguration.newInstance(acctEmail);
       await startPullingKeysFromEkm(acctEmail, clientConfiguration, factory, ppEvent);
       await webmailSpecific.start(acctEmail, clientConfiguration, inject, notifications, factory, notifyMurdered);
```
rrrooommmaaa, line 360: why should we wait for completion of pulling the keys before `start`-ing the `webmailSpecific`?
This, of course, leads to less code, but worse UX.
They can be nicely run in parallel, I guess (by passing `() => notifyExpiringKeys(...)` as an argument to `startPulling..` if you like)

extension/js/common/helpers.ts
```diff
@@ -164,3 +164,11 @@ export const processAndStoreKeysFromEkmLocally = async (
   }
   return { updateCount: encryptedKeys?.keys.length ?? 0 + (existingKeys.length - keysToRetain.length), noKeysSetup: !(encryptedKeys?.keys.length || keysToRetain.length) };
 };
+
+export const getLocalKeyExpiration = async ({ acctEmail }: Bm.GetLocalKeyExpiration): Promise<Bm.Res.GetLocalKeyExpiration> => {
+  const kis = await KeyStore.get(acctEmail);
+  const parsedKeys = await Promise.all(kis.map(async (ki) => await KeyUtil.parse(ki.public)));
+  const keyExpiration = parsedKeys
+    .sort((key1, key2) => (key1.expiration ?? Number.MAX_SAFE_INTEGER) < (key2.expiration ?? Number.MAX_SAFE_INTEGER) ? 1 : -1)[0]?.expiration;
```
rrrooommmaaa, line 172: I suggest to simplify this: we don't need the sorted array, only the `min` value?
We can simply `.map(key => key.expiration ?? Number.MAX_SAFE_INTEGER)` (or during parsing `(await KeyUtil.parse(ki.public)).expiration ?? Number.MAX_SAFE_INTEGER`
and then find the min value using Math.min or Array.prototype.reduce
as shown in these examples https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/max

rrrooommmaaa: expirations.reduce((a, b) => Math.min(a, b), 0);

rrrooommmaaa: or even without mapping (though in our case mapping would simplify the code) as the reduced valued type need not match the array element types:
`const keyExpiration = parsedKeys.reduce((expiration, key2) => Math.min(expiration, key2.expiration ?? Number.MAX_SAFE_INTEGER), parsedKeys[0].expiration ?? Number.MAX_SAFE_INTEGER);`

test/source/tests/setup.ts
```diff
@@ -496,6 +497,82 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        date: new Date(Date.now() - 2 * 24 * 60 * 60 * 1000)
+      });
+      MOCK_KM_KEYS[acctEmail] = {response: { privateKeys: [{ decryptedPrivateKey: key.privateKeyArmored }] }};
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);
+      await SetupPageRecipe.autoSetupWithEKM(settingsPage);
+      const gmailPage = await openMockGmailPage(t, browser, acctEmail);
+      // Check if notification presents
+      const warningMsg = 'Your local keys expire in 18 days.\nTo receive the latest keys, please ensure that you can connect to your corporate network either through VPN or in person and reload Gmail.\nIf this notification still shows after that, please contact your Help Desk.';
+      await gmailPage.waitForContent('@webmail-notification', warningMsg);
+      const updatedKey = await opgp.generateKey({
```
rrrooommmaaa, line 565: We can save some time by taking a non-expiring key from a constant instead of generating it,

rrrooommmaaa:
Btw, can you turn on auto-formatting in VS Code? VS Code tries to put extra spaces to the code you submitted.
image

test/source/tests/setup.ts
```diff
@@ -496,6 +497,82 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        curve: 'curve25519',
+        userIds: [{ email: acctEmail }, { email: 'demo@gmail.com', name: 'Demo user'}]
+      });
+      MOCK_KM_KEYS[acctEmail].response = { privateKeys: [{ decryptedPrivateKey: key.privateKeyArmored }, { decryptedPrivateKey: updatedKey.privateKeyArmored }] };
+      await gmailPage.page.reload();
+      await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, 'Account keys updated');
+      await gmailPage.page.reload();
+      await gmailPage.notPresent('@webmail-notification');
+    }));
+
```
rrrooommmaaa, line 575: Are we missing a test case that shows the "VPN warning" on network failure? We have `badRequestError` property in `MOCK_KM_KEYS` to emulate this.

test/source/tests/setup.ts
ioanmo226: Well, I use generated key (when running tests) because we have to test expiration time which is relevant to current time.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
ioanmo226: Updated

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
ioanmo226: Updated

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -286,16 +286,26 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
...
       const idToken = await InMemoryStore.get(acctEmail, InMemoryStoreKeys.ID_TOKEN);
       if (idToken) {
         const keyManager = new KeyManager(clientConfiguration.getKeyManagerUrlForPrivateKeys()!);
         Catch.setHandledTimeout(async () => {
           const { privateKeys } = await keyManager.getPrivateKeys(idToken);
           await processKeysFromEkm(acctEmail, privateKeys.map(entry => entry.decryptedPrivateKey), clientConfiguration, factory, idToken, ppEvent);
-        }, 0);
+        }, 0, completion);
+      } else {
+        completion();
```
rrrooommmaaa, line 305: This would be a hanging promise?  I wonder why linter isn't giving warning about this.
Wouldn't it be more flexible to return the function in case it can't be consumed?
So `startPullingKeysFromEkm` accepts parameter `completion: () => Promise<void>`
and returns `Promise<Promise<void> | undefined>`?
So we can await the completion properly (if needed) or attach `catch` handler for it properly.

extension/js/common/platform/catch.ts
```diff
@@ -123,15 +123,23 @@ export class Catch {
...
+          isPromise = true;
           r.catch(Catch.reportErr);
+          void r.finally(callback);
         }
       } catch (codeErr) {
         Catch.reportErr(codeErr);
+      } finally {
+        // Do not call callback for promise function because it's already called in line 134.
+        if (!isPromise && callback) {
+          callback();
```
rrrooommmaaa, line 141: Don't we need to `catch` this too? This looks excessively complicated.
Why don't we add a plain `try .. catch .. finally` around lines 301 and 302?
```
try {
         const { privateKeys } = await keyManager.getPrivateKeys(idToken);
         await processKeysFromEkm(acctEmail, privateKeys.map(entry => entry.decryptedPrivateKey), clientConfiguration, factory, idToken, ppEvent);
} finally {
         await comlpetion();
}
```

test/source/tests/setup.ts
rrrooommmaaa: The `updatedKey` doesn't have expiration, does it? So it may be retrieved from a constant

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -327,7 +362,9 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
       const ppEvent: { entered?: boolean } = {};
       browserMsgListen(acctEmail, tabId, inject, factory, notifications, ppEvent);
       const clientConfiguration = await ClientConfiguration.newInstance(acctEmail);
-      await startPullingKeysFromEkm(acctEmail, clientConfiguration, factory, ppEvent);
+      await startPullingKeysFromEkm(acctEmail, clientConfiguration, factory, ppEvent, Catch.try(async () => {
+        await notifyExpiringKeys(acctEmail, clientConfiguration, notifications);
```
rrrooommmaaa, line 366: We don't need `await` here, it's enough to simply return a promise
```
Catch.try(() => notifyExpiringKeys(acctEmail, clientConfiguration, notifications)));
```

test/source/tests/setup.ts
```diff
@@ -461,39 +462,120 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        keyExpirationTime: 20 * 24 * 60 * 60,
+        date: new Date(Date.now() - 2 * 24 * 60 * 60 * 1000)
+      });
+      MOCK_KM_KEYS[acctEmail] = { response: { privateKeys: [{ decryptedPrivateKey: key.privateKeyArmored }] } };
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);
+      await SetupPageRecipe.autoSetupWithEKM(settingsPage);
+      const gmailPage = await openMockGmailPage(t, browser, acctEmail);
+      // Check if notification presents
+      const warningMsg = 'Your local keys expire in 18 days.\nTo receive the latest keys, please ensure that you can connect to your corporate network either through VPN or in person and reload Gmail.\nIf this notification still shows after that, please contact your Help Desk.';
+      await gmailPage.waitForContent('@webmail-notification-notify_expiring_keys', warningMsg);
```
rrrooommmaaa, line 564: It'd be great to add a test for multiple notifications (#4779) but we can file a separate low-priority issue for this as it seems to work ok

test/source/tests/setup.ts
ioanmo226: You are right.
This key doesn't have expiration and we can have it as a constant.

test/source/tests/setup.ts
ioanmo226: I didn't add specific test because we have several tests which cover multiple notification feature test.
But it's good to have specific test.

ioanmo226:
@DenBond As you can see, `ci/semaphoreci/push: Tests` is not reported.
Maybe you have permission to update/fix this issue?

rrrooommmaaa (approved):

Issue #2538, opened 2020-02-07
forbid sending rich text messages with more than 19 mb atts


tomholub:
- it's hard on the recipient, especially on mobile
- gmail may not accept it due to size problems (over 20mb or so) due to several layers of encoding (size bloat)

tomholub:
Interestingly didn't run into this again yet, but is likely still a problem. May need to have a more hollistic look into encoding, attachments, rich text.

tomholub:
would become irrelevant if #3032 is resolved by Gmail

tomholub:
@martgil please try, on firefox, when switched to "Rich text" mode, what is the largest size of attachment that gets accepted and also successfully sent. Does it work with a 25mb attachment? Or is it less?

martgil:
These are what I noticed when sending an email with attachment in "Rich text" mode.

![image](https://user-images.githubusercontent.com/46025304/229971074-f336768f-73e1-4f02-b350-b72788fd1724.png)
For Figure 1: When attaching a 21mb image file in rich text mode, the FlowCrypt secure compose window have no problem when including the image. However, an error will occur when sending the secure email with attachment saying "couldn't send the email because the attachment is too large. For comparison, I tried to send the same 21mb image file in non-rich text mode and the email went through.
![image](https://user-images.githubusercontent.com/46025304/229971680-bdf63442-bb1e-45e1-87ac-04af339eb235.png)
(figure no. 1)

For Figure 2: In rich text mode, when attaching 2 or more images with 21mb file size each, a warning prompt will appear. While it is a correct, if we watched video no. 1, the attachment prompt continues to show a warning prompt but the images get appended in the Rich text editor. There is also an additional error that says "too much recursion" and it occurs even when the entire attachments/images in rich text are removed.

![image](https://user-images.githubusercontent.com/46025304/229973388-3d8e3fcc-3e48-4026-9833-c3b4f5f6760c.png)

(figure no. 2)

[Screencast from 2023-04-05 11-30-01.webm](https://user-images.githubusercontent.com/46025304/229978432-0299051c-35a5-4f38-9b9d-ed2f2875c5ac.webm)

(video no. 1)

Link to the 21mb image file: https://sample-videos.com/img/Sample-jpg-image-20mb.jpg
other source of test images: https://7132.com/en/corporate/press/high-resolution-images

[...]

PR #4773, opened 2022-11-03
#4702 fix: send attachment when sending through FES

### LINKED ISSUE #4702: attachments not sent through email provider for fes message
1) send message on enterprise account from browser extension through the web portal, with attachments
2) navigate to sent message in sent folder

actual: attachments not rendering

expected: sent attachments to show

Before implementing the fix, I'd like the issue just investigated. How are the attachments encoded? Is it pgp/mime? Or some other method?

ioanmo226:
This PR fixed send attachment issue when sending through FES

close #4702 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
+++ b/extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
@@ -79,7 +79,7 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
         { isDraft: this.isDraft });
     }
     // rich text: PGP/MIME - https://tools.ietf.org/html/rfc3156#section-4
-    const attachments = this.formatEncryptedMimeDataAsPgpMimeMetaAttachments(encrypted);
+    const attachments = await this.formatEncryptedMimeDataAsPgpMimeMetaAttachments(encrypted, pubkeys);
     return await SendableMsg.createPgpMime(this.acctEmail, this.headers(newMsg), attachments, { isDraft: this.isDraft });
   };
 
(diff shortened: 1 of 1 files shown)


ioanmo226:
Please review implementation/fix.
I'll add test after review.

rrrooommmaaa:
@ioanmo226 Did you want to review this or Tom?

ioanmo226:
I prefer both of you review.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -206,8 +206,8 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
     return await SendableMsg.createSMimeEncrypted(this.acctEmail, this.headers(newMsg), encryptedMessage.data, { isDraft: this.isDraft });
   };
 
-  private formatEncryptedMimeDataAsPgpMimeMetaAttachments = (data: Uint8Array) => {
-    const attachments: Attachment[] = [];
+  private formatEncryptedMimeDataAsPgpMimeMetaAttachments = async (data: Uint8Array, pubkeys: PubkeyResult[]) => {
+    const attachments: Attachment[] = await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubkeys);
```
rrrooommmaaa, line 210: As far as I remember, on the top level of a PGP/Mime message there should be only 2 parts: version and the encrypted message. If we need to include the attachment, we should probably include it in the second part (which is a MIME message itself).
As they say, 
```
An advantage to this scheme is that the message structure, like attachment metadata, is not leaked to someone who intercepts the encrypted message... However, to be able to read just the message body, it is necessary to download the whole message, including all attachments, because everything is encrypted together.
```
https://proton.me/support/pgp-mime-pgp-inline

tomholub:
I will close it for now, I don't have the capacity at the moment to evaluate whether the fix is pgp compliant, plus there are downsides to the fix (email size bloat). We'll have to look into this more properly later, after evaluating and possibly fixing size bloat mentioned here https://github.com/FlowCrypt/flowcrypt-browser/issues/2538

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
ioanmo226: Yeah, You are right. Attachment metadata could be leaked if we use PGP/Inline.

@sosnovsky Which way do we have to go?
Do we have to encrypt everything including attachments (PGP/MIME) or current way?

sosnovsky: Let's use `pgp/inline` for now, like we do for common encrypted messages - attachments metadata is visible there too.
Later, after we'll finalize `pgp/mime` support and enable it in Chrome extension, we can use `pgp/mime` structure for password-protected messages too.

PR #4774, opened 2022-11-04
#4681 fix: visual glitch opening passphrase dialog

### LINKED ISSUE #4681: Visual glitch opening passphrase dialog
I noticed that the default `<h1>` text "Enter FlowCrypt pass phrase to read encrypted email" is visible for a second,
and then gets replaced with a correct text based on query arguments.
https://github.com/FlowCrypt/flowcrypt-browser/blob/754e0b6d7a453b7ca0fd12672fa806b7d32298e7/extension/chrome/elements/passphrase.htm#L21

(I saw this effect with `update_key` tests, perhaps, the same effect happens for all the other cases)
https://github.com/FlowCrypt/flowcrypt-browser/blob/754e0b6d7a453b7ca0fd12672fa806b7d32298e7/extension/chrome/elements/passphrase.ts#L79

I think it'd be good to display the correct text at once.

ioanmo226:
This PR fixed visual glitch opening passphrase dialog

close #4681 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -364,6 +364,6 @@ export const asyncSome = async<T>(arr: Array<T>, predicate: (e: T) => Promise<bo
   return false;
 };
 
-export const stringTuple = <T extends [string] | string[]>(...data: T): T => {
+export const stringTuple = <T extends | string[]>(...data: T): T => {
   return data;
 };
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -363,3 +363,7 @@ export const asyncSome = async<T>(arr: Array<T>, predicate: (e: T) => Promise<bo
   }
   return false;
 };
+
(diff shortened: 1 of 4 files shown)


extension/js/common/core/common.ts
```diff
@@ -362,4 +362,8 @@ export const asyncSome = async<T>(arr: Array<T>, predicate: (e: T) => Promise<bo
     if (await predicate(e)) return true;
   }
   return false;
+};
+
+export const stringTuple = <T extends [string] | string[]>(...data: T): T => {
```
rrrooommmaaa, line 367: Wouldn't `<T extends string[]>` do?
I see no benefit of `| [string]`

ioanmo226: You are right. 👍 
Just updated

rrrooommmaaa (approved):

PR #4799, opened 2022-11-17
#4792 fix: Password protected message issue when FES is not used

### LINKED ISSUE #4792: Account reconnection pops-up when sending a password-protected messages
Users cannot send a password-protected message as the account reconnection pops up every time the send button was clicked.

Steps to reproduce:
1. Compose a pw-protected message
2. Send the pw-protected message
3. Account reconnection message will pop up -- try to reconnect the account the refresh the webpage
4. repeat step 1 & 2 and the account reconnection message will appear again preventing the user from sending the pw-message.

Screenshot:
image

Reference: 
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZfCdPtdKFTbMWkkrtqctmJB
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZgLpSTPjSWsdNBfkVNbFBkC
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZjcnFHfBwGNKvMgznTbDbzh
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThfFwlRtBWnPxjgwGtNnD
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZjcdjsFWLMXvsktCbbWfkTq
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTqwFLQmngvRvzmQSJDtTr
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTqrhnNlNPxsrQLgzPrpss
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTqpQgpGPTHLwvKkBRQkDw
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThfGjjvFmZQKFMLdSLcQc
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThXXwcSkGkbHSfKDWPTlc
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTFVbFHqVgFZrgfkDxQHQZ
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTNqpLZqSpNWkJtkfrsXWQ
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTXDKgtnmwBxFxmVRVJsGZ
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTXHjJGvfgJVrcrNMSkXNs
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhTXJpSpspRWgKhPnxblgkC
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZgLVZXWzjSGxZVWCXmRTMxJ
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThQmNLnBBfjGBMxcRdVzq
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThRsPgHFpLpMbjWXgHrpm
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThVKgMksXVPkBJbPFStMV
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThcwGLWpVbMhgrZBJDvSJ
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRZhThWQhfthVjwgglQfWVCJG

ioanmo226:
This PR password protected message send issue when FES is not used

close #4792 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/mock/backend/backend-endpoints.ts
+++ b/test/source/mock/backend/backend-endpoints.ts
@@ -14,11 +14,7 @@ export const mockBackendData = new BackendData();
 export const mockBackendEndpoints: HandlersDefinition = {
   '/api/account/get': async ({ }, req) => {
     throwIfNotPost(req);
-    const idToken = req.headers.authorization?.replace(/^Bearer /, '');
-    if (!idToken) {
-      throw new HttpClientErr('backend mock: Missing id_token');
-    }
-    const email = oauth.extractEmailFromIdToken(idToken);
+    const email = throwIfInvalidOrMissingIdToken(req);
     return JSON.stringify({
       account: mockBackendData.getAcctRow(email!),
       domain_org_rules: mockBackendData.getClientConfiguration(email!),
@@ -41,7 +37,8 @@ export const mockBackendEndpoints: HandlersDefinition = {
     expect((body as { email: string }).email).to.equal('flowcrypt.compatibility@gmail.com');
     return { sent: true, text: 'Feedback sent' };
   },
-  '/api/message/upload': async ({ }) => {
+  '/api/message/upload': async ({ }, req) => {
+    throwIfInvalidOrMissingIdToken(req);
     return { short: 'mockmsg000' };
   },
   '/api/link/me': async ({ }, req) => {
@@ -53,4 +50,16 @@ const throwIfNotPost = (req: IncomingMessage) => {
   if (!isPost(req)) {
     throw new HttpClientErr('Backend mock calls must use POST method');
   }
+};
+
+const throwIfInvalidOrMissingIdToken = (req: IncomingMessage) => {
+  const idToken = req.headers.authorization?.replace(/^Bearer /, '');
+  if (!idToken) {
+    throw new HttpClientErr('backend mock: Missing id_token');
+  }
+  const email = oauth.extractEmailFromIdToken(idToken);
+  if (!email) {
+    throw new HttpClientErr('Invalid Id token. Missing email.');
+  }
+  return email;
 };
--- a/test/source/mock/backend/backend-endpoints.ts
+++ b/test/source/mock/backend/backend-endpoints.ts
@@ -14,7 +14,7 @@ export const mockBackendData = new BackendData();
 export const mockBackendEndpoints: HandlersDefinition = {
   '/api/account/get': async ({ }, req) => {
     throwIfNotPost(req);
-    const email = throwIfInvalidOrMissingIdToken(req);
+    const email = getEmailFromIdTokenOrThrow(req);
     return JSON.stringify({
       account: mockBackendData.getAcctRow(email!),
       domain_org_rules: mockBackendData.getClientConfiguration(email!),
@@ -38,7 +38,7 @@ export const mockBackendEndpoints: HandlersDefinition = {
     return { sent: true, text: 'Feedback sent' };
   },
   '/api/message/upload': async ({ }, req) => {
-    throwIfInvalidOrMissingIdToken(req);
+    getEmailFromIdTokenOrThrow(req);
     return { short: 'mockmsg000' };
   },
   '/api/link/me': async ({ }, req) => {
@@ -52,7 +52,7 @@ const throwIfNotPost = (req: IncomingMessage) => {
   }
 };
 
-const throwIfInvalidOrMissingIdToken = (req: IncomingMessage) => {
+const getEmailFromIdTokenOrThrow = (req: IncomingMessage) => {
   const idToken = req.headers.authorization?.replace(/^Bearer /, '');
   if (!idToken) {
     throw new HttpClientErr('backend mock: Missing id_token');
--- a/test/source/tests/gmail.ts
+++ b/test/source/tests/gmail.ts
@@ -126,14 +126,6 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       expect(urls.length).to.eq(1);
     }));
 
-    ava.default('mail.google.com - send password protected message', testWithBrowser('ci.tests.gmail', async (t, browser) => {
-      const gmailPage = await BrowserRecipe.openGmailPageAndVerifyComposeBtnPresent(t, browser);
-      const composePage = await GmailPageRecipe.openSecureCompose(t, gmailPage, browser);
-      const subject = `New Password Protected Message ${Util.lousyRandom()}`;
-      await ComposePageRecipe.fillMsg(composePage, { to: 'test@email.com' }, subject, 'Test');
-      await ComposePageRecipe.sendAndClose(composePage, { password: 'gO0d-pwd' });
-    }));
-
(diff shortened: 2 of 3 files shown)


ioanmo226:
@martgil Please try if this branch fixed current password protected message issue or try after this branch is merged to master

martgil:
I can try it before it gets merged. I'll let you know my feedback here in this PR shortly.

martgil:
I tested it and it seemed the issue was fixed 👍

test/source/tests/gmail.ts
```diff
@@ -126,6 +126,14 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       expect(urls.length).to.eq(1);
     }));
 
+    ava.default('mail.google.com - send password protected message', testWithBrowser('ci.tests.gmail', async (t, browser) => {
```
rrrooommmaaa, line 129: The fix looks good, but why do we need a live test?
Is this because we don't check authorization in our '/api/message/upload'  handler in `backend-endpoints.ts`  (Flowcrypt API emulation)? We can try to add the extra checks there too, have you tried that? This way, I suppose, the previous compose test would suffice to cover the issue.

ioanmo226: You are right. 👍 
We indeed don't need live test for this.

test/source/mock/backend/backend-endpoints.ts
```diff
@@ -53,4 +50,16 @@ const throwIfNotPost = (req: IncomingMessage) => {
   if (!isPost(req)) {
     throw new HttpClientErr('Backend mock calls must use POST method');
   }
+};
+
+const throwIfInvalidOrMissingIdToken = (req: IncomingMessage) => {
```
rrrooommmaaa, line 55: Let's rename it to `getEmailFromIdTokenOrThrow` as `throwIf...` doesn't suppose a return value.

We can merge this, however, I noticed that the API endpoint of `/api/message/upload` is called twice successfully. We can discuss and analyze in a separate issue whether we can retain `UploadedMessageData` from the first successful call to re-use it in the subsequent sending.

ioanmo226: What do you mean called twice? was this api called twice somewhere?
> I noticed that the API endpoint of /api/message/upload is called twice successfully.

tomholub (approved):

test/source/mock/backend/backend-endpoints.ts
rrrooommmaaa: > What do you mean called twice? was this api called twice somewhere?

Yes, in `compose - test compose after reconnect account` test, the endpoint `/api/message/upload` was accessed twice -- before account reconnection and after account reconnection, both times successfully (the authentication header was obviously valid) -- so as I understand the message was uploaded to the backend twice (I'm planning to re-check the FES upload too)

ioanmo226: I checked but it's called twice because of below reason.
When it tries to send pwd-protected message first time(it calls upload api 1st time), it returns error because id_token is missing.
After that app asks user to re-authorize google extension and then re-tries to send message again. (this time it calls upload api 2nd time)

rrrooommmaaa: > When it tries to send pwd-protected message first time(it calls upload api 1st time), it returns error because id_token is missing.

Sending a PWD-protected message message consists of 2 phases:
1) uploading the PWD-encrypted message to a flowcrypt/FES server, getting message link(s)
2) sending a message containing links/pubkey-encrypted message to `gmail`.
Which phase gives an error? In the mock environment, only the 2. Perhaps, it's worthwhile to update the mock environment/test to mimic the real situation more precisely.

Anyway, this isn't critical, just a consideration.
A possible scenario:
1) PWD-encrypted message gets successfully uploaded, we have `UploadedMessageData`
2) sending message(s) to `gmail` fails (for whatever reason), but auto-retried (e.g. after reconnecting the account)
3) If possible, PWD-encrypted message shouldn't be re-uploaded, provided the inputs are the same (the sender, the recipients, the text etc.) -- which is probably true for auto-resends.
4) sending message(s) to `gmail` using either previous `UploadedMessageData` or the previously built messages -- but the point of concern here is the situation when the user "reconnected" to another gmail account if that's possible.

ioanmo226: I think 1st phase gives error in mock environment.
1st phase tries to upload PWD-encrypted message to FES using /message/upload url. (Mock environment returns error when id_token is missing or invalid)
Therefore, I think current logic is correct. 
> Which phase gives an error? In the mock environment, only the 2.

rrrooommmaaa: What I saw during the test was:
1) The mock FES endpoint accepted the token as valid.
2) The mock Google endpoint rejected the token as invalid.
Is this a peculiarity of the test setup or a possible live situation? Please describe your thoughts on this.

Anyway, Google endpoint may possibly reject a message for reasons other than invalid token. The point of #4801 is when we have successfully uploaded the pwd-encrypted message in phase 1, we should retain the message tokens (returned by FES) by reasonably simple means, to re-use them in automatic google-message re-sends. Of course, these tokens should be invalidated if a recipient of the message is added (removed?) or the message text is changed.

So #4801 is worth considering without direct relation to the invalid token issue.

ioanmo226: Hmm... I didn't experience this. So mock FES treats id_token as valid but mock google endpoint treats id_token as invalid?
Could you tell me reproduce steps?
> What I saw during the test was:

rrrooommmaaa: > Could you tell me reproduce steps?

The test `compose - test compose after reconnect account`

ioanmo226: 👍  let me check

rrrooommmaaa: > So mock FES treats id_token as valid

Not FES, `backend` -- flowcrypt API. but this doesn't really matter

ioanmo226: Just finished checking and I now fully understand your concern.
But if we want to change like above way, we need to change compose send infra and need to change several tens of lines of codes and might make code messy.
Moreover, I think it's fine to call /message/upload twice because id_token will change after google re-auth(or something like that) and backend might need to return different data than previous one.

Let me know your thought @rrrooommmaaa

PR #4807, opened 2022-11-23
#4795 fix browser extension tries to decrypt plain message attachment message issue

### LINKED ISSUE #4795: The browser extension tries to decrypt an attachment named 'message' even though it contains plain/html content
Description:
The browser extension tries to decrypt an attachment named "message" despite it contains plain or HTML content.

It will produce the following error: 
image

For now, the affected users would have to click the 'See original' button to view the original email and 'message' attachment nicely.
image

Sample email to replicate the issue (the one I crafted and use to replicate the issue):
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGqRZhStqHVPKPCWQGqJbbNTxfg

Sample email shared by the customers:
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGqRZhSlXHNCTqcnfjhKcTssVCV
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGqRZhSlXHNCTGSpHJRSBclGDGW

ioanmo226:
This PR fixed browser extension tries to decrypt plain message attachment message issue

close #4795 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


ioanmo226:
Forgot to add live test message to /live-gmail-test-emails folder.
@rrrooommmaaa  Thank you for adding that.

rrrooommmaaa:
I also added this message for the mock test.

ioanmo226:
I think live test would be sufficient?

test/source/tests/decrypt.ts
```diff
@@ -47,6 +48,17 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
 
+    ava.default('decrypt - encrypted text inside "message" attachment is correctly decrypted', testWithBrowser('ci.tests.gmail', async (t, browser) => {
+      const acctEmail = 'ci.tests.gmail@flowcrypt.test';
+      const key = Config.key('flowcrypt.compatibility.1pp1')!;
+      await SettingsPageRecipe.addKeyTest(t, browser, acctEmail, key.armored!, key.passphrase, {}, false);
+      await InboxPageRecipe.checkDecryptMsg(t, browser, {
+        acctEmail,
+        threadId: '184a474fc1bd59b8',
+        expectedContent: 'This message contained the actual encrypted text inside a "message" attachment.'
+      });
```
rrrooommmaaa, line 59: This test would succeed on `master`, this is how this message is rendered on `master` (and the live gmail test).
image
But on this branch the `message` is considered to be non-encrypted
image
So the gmail rendering and inbox-rendering have diverged, that's not too good.

rrrooommmaaa: On the opposite, we want the non-encrypted rendering of the `message` in case there is any text in the message body. So we should add such a message to the tests too and make sure it's rendered like this:
image

ioanmo226: You are right. Let me fix it

test/source/tests/decrypt.ts
```diff
@@ -47,6 +47,31 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+      const threadId = '184a87a7b32dd009';
+      const acctEmail = 'ci.tests.gmail@flowcrypt.test';
+      const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId}`));
+      await inboxPage.waitForSelTestState('ready');
+      await inboxPage.waitAll('iframe');
+      expect(await inboxPage.isElementPresent('@container-attachments')).to.equal(true);
+      await inboxPage.waitForContent('.message.line', 'Plain message');
+      // expect no pgp blocks
+      const urls = await inboxPage.getFramesUrls(['/chrome/elements/pgp_block.htm']);
+      expect(urls.length).to.equal(0);
```
rrrooommmaaa, line 71: I added this check (it was the `pgp_block.htm` that previously showed error to the user).
Please add a `gmail` test for this message too, as the end user mainly sees the result of `GmailElementReplacer`.
I think we can later test `GmailElementReplacer` in our mock environment, we would need to develop our mock gmail page in a separate issue.

test/source/tests/gmail.ts
```diff
@@ -392,6 +392,17 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       await gmailPage.close();
     }));
 
+    ava.default(`mail.google.com - render plain text for "message" attachment (which has plain text)`, testWithBrowser('ci.tests.gmail', async (t, browser) => {
+      const gmailPage = await openGmailPage(t, browser);
+      await gotoGmailPage(gmailPage, '/184a87a7b32dd009');
```
rrrooommmaaa, line 397: FMfcgzGrbHrBdFGBXqpFZvSkcQpKkvrM
?

ioanmo226: sorry? why do we need that id?
`184a87a7b32dd009` is gmail thread which contains plain text + plain `message` attachment?

ioanmo226: sorry? why do we need that id?
`184a87a7b32dd009` is gmail thread which contains plain text + plain `message` attachment?

rrrooommmaaa: oh interesting. Then why do we use this type of link for `mail.google.com - encrypted text inside "message" attachment` test?

rrrooommmaaa: Also, please add `eml` from the live account to the repo

ioanmo226: I'm not sure what you mean.
We test `FMfcgzGrbHprlHvtTJscCJQpZcqrKQbg` thread for `mail.google.com - encrypted text inside "message" attachment`  test.
And we test `184a87a7b32dd009` thread for `mail.google.com - render plain text for "message" attachment (which has plain text)` test

ioanmo226: Ohhh.. you are right.
we need to use `FMfcgzGrbHrBdFGBXqpFZvSkcQpKkvrM` for `mail.google.com - render plain text for "message" attachment (which has plain text) `.
I was confused with mock thread id.

ioanmo226: If I enter https://mail.google.com/mail/u/0/#inbox/184a87a7b32dd009, it redirects to https://mail.google.com/mail/u/0/#inbox/FMfcgzGrbHrBdFGBXqpFZvSkcQpKkvrM though

ioanmo226: Therefore both ids are fine

rrrooommmaaa (approved):

test/source/tests/gmail.ts
rrrooommmaaa: This is an interesting fact!

rrrooommmaaa: Let's merge it as it is now

PR #4809, opened 2022-11-24
#4675 Use google api to get allowed google scopes

### LINKED ISSUE #4675: Use google api to get `allowed google scopes` instead of from AcctStore
I totally agree. User might remove it manually from google account page and it can get out of sync easily.
I think there must be functionality to check authorized scopes in web too as we did successfully in iOS app.
By the way, I think better to do this in another task as this one is big enough task.

> Maybe it wasn't such a good idea to keep this in AcctStore. It can get out of sync. Is there a Google endpoint where we can figure out current authorizations? Otherwise, any ideas?

_Originally posted by @ioanmo226 in https://github.com/FlowCrypt/flowcrypt-browser/issues/4657#issuecomment-1233147850_

ioanmo226:
This PR uses google api to get allowed google scopes

close #4675 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -387,7 +387,8 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     try {
       const scopes = await AcctStore.getScopes(this.view.acctEmail);
       return scopes.readContacts && scopes.readOtherContacts;
-    } catch {
+    } catch (e) {
+      this.view.errModule.debug(`googleContactsSearchEnabled: Error occurred while fetching result: ${e}`);
       return undefined;
     }
   };
@@ -570,7 +571,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
         return;
       }
       let foundOnGoogle: EmailProviderContact[] = [];
-      if (await this.googleContactsSearchEnabled) {
+      if (await this.googleContactsSearchEnabled || await this.googleContactsSearchEnabled === undefined) {
         this.view.errModule.debug(`searchContacts 3`);
         foundOnGoogle = await this.searchContactsOnGoogle(substring, contacts);
         await this.addApiLoadedContactsToDb(foundOnGoogle);
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -550,6 +550,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
   private searchContacts = async (input: JQuery<HTMLElement>): Promise<void> => {
     try {
       this.view.errModule.debug(`searchContacts`);
+      this.view.errModule.debug(`googleContactsSearchEnabled: ${await this.googleContactsSearchEnabled}`);
       const substring = Str.parseEmail(String(input.val()), 'DO-NOT-VALIDATE').email;
       this.view.errModule.debug(`searchContacts.query.substring(${JSON.stringify(substring)})`);
       if (!substring) {
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -49,11 +49,6 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     void this.initScopes();
   }
 
-  private initScopes = async () => {
-    const scopes = await AcctStore.getScopes(this.view.acctEmail);
-    this.googleContactsSearchEnabled = scopes.readContacts && scopes.readOtherContacts;
-  };
-
   public setHandlers = (): void => {
     let preventSearchContacts = false;
     const inputs = this.view.S.cached('recipients_inputs');
@@ -386,6 +381,11 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     }
     this.showHideCcAndBccInputsIfNeeded();
     this.setEmailsPreview();
+  };
+
+  private initScopes = async () => {
+    const scopes = await AcctStore.getScopes(this.view.acctEmail);
+    this.googleContactsSearchEnabled = scopes.readContacts && scopes.readOtherContacts;
   };
 
   private inputsBlurHandler = async (target: HTMLElement, e: JQuery.Event<HTMLElement, null>) => {
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -42,12 +42,17 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
 
   private dragged: Element | undefined = undefined;
 
-  private googleContactsSearchEnabled: boolean;
+  private googleContactsSearchEnabled = false;
 
   constructor(view: ComposeView) {
     super(view);
-    this.googleContactsSearchEnabled = this.view.scopes.readContacts && this.view.scopes.readOtherContacts;
+    void this.initScopes();
   }
+
+  private initScopes = async () => {
+    const scopes = await AcctStore.getScopes(this.view.acctEmail);
+    this.googleContactsSearchEnabled = scopes.readContacts && scopes.readOtherContacts;
+  };
 
   public setHandlers = (): void => {
     let preventSearchContacts = false;
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -550,7 +550,6 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
   private searchContacts = async (input: JQuery<HTMLElement>): Promise<void> => {
     try {
       this.view.errModule.debug(`searchContacts`);
-      this.view.errModule.debug(`googleContactsSearchEnabled: ${await this.googleContactsSearchEnabled}`);
       const substring = Str.parseEmail(String(input.val()), 'DO-NOT-VALIDATE').email;
       this.view.errModule.debug(`searchContacts.query.substring(${JSON.stringify(substring)})`);
       if (!substring) {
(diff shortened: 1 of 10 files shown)


extension/js/common/platform/store/acct-store.ts
```diff
@@ -125,19 +129,30 @@ export class AcctStore extends AbstractStore {
...
-          result[scopeName] = google_token_scopes.includes(GoogleAuth.OAUTH.scopes[scopeName as keyof typeof GoogleAuth.OAUTH.scopes]);
-        } else if (scopeName in GoogleAuth.OAUTH.legacy_scopes) {
-          result[scopeName] = google_token_scopes.includes(GoogleAuth.OAUTH.legacy_scopes[scopeName as keyof typeof GoogleAuth.OAUTH.legacy_scopes]);
-        }
+    if (!accessToken) {
+      return result;
+    }
+    let allowedScopes: string[] = [];
+    try {
+      const { scope } = await Api.ajax({ url: `https://www.googleapis.com/oauth2/v1/tokeninfo?access_token=${accessToken}` }, Catch.stackTrace()) as { scope: string };
```
rrrooommmaaa, line 143: Should use `OAUTH_GOOGLE_API_HOST` (preferred)
or `GMAIL_GOOGLE_API_HOST` so this doesn't make its way to mock builds

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -42,11 +42,11 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
 
   private dragged: Element | undefined = undefined;
 
-  private googleContactsSearchEnabled: boolean;
+  private googleContactsSearchEnabled = false;
 
   constructor(view: ComposeView) {
     super(view);
-    this.googleContactsSearchEnabled = this.view.scopes.readContacts && this.view.scopes.readOtherContacts;
+    void this.initScopes();
```
rrrooommmaaa, line 49: In this case, `googleContactsSearchEnabled` may remain wrongly `false` until the execution completes.
Do we have a unit test that relies on `googleContactsSearchEnabled=true`? It would fail if we introduce a delay in our '/oauth2/v1/tokeninfo' mock handler.
Therefore, I'd recommend to make `googleContactsSearchEnabled: Promise<boolean>` or `googleContactsSearchEnabled: boolean | Promise<boolean>`, so we could do something like `if (await this.googleContactsSearchEnabled)`,
but we'll have to make `renderSearchRes` function `async`
It'd be better

rrrooommmaaa: FYU: it's ok to `await` a promise more than once, it will return the same value once resolved

ioanmo226: Nice suggestion 👍 
Thanks

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -545,6 +550,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
   private searchContacts = async (input: JQuery<HTMLElement>): Promise<void> => {
     try {
       this.view.errModule.debug(`searchContacts`);
+      this.view.errModule.debug(`googleContactsSearchEnabled: ${await this.googleContactsSearchEnabled}`);
```
rrrooommmaaa, line 553: why would we pause searchContacts here?
we can exit early at line 561/567 etc.?

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -555,19 +561,19 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
       this.view.errModule.debug(`searchContacts db count: ${contacts.length}`);
-      this.renderSearchRes(input, contacts, { substring });
+      await this.renderSearchRes(input, contacts, { substring });
       if (contacts.length >= this.MAX_CONTACTS_LENGTH) {
         this.view.errModule.debug(`searchContacts 2, count: ${contacts.length}`);
         return;
       }
       let foundOnGoogle: EmailProviderContact[] = [];
-      if (this.googleContactsSearchEnabled) {
+      if (await this.googleContactsSearchEnabled) {
```
rrrooommmaaa, line 570: I also wonder how the UI would look like if this request hangs for a minute, for example. Should we introduce some sort of timeout? Or run this block concurrently with `searchContacts 4`?

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -610,7 +616,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
   };
 
   private searchContactsOnGoogle = async (query: string, knownContacts: ContactPreview[]): Promise<EmailProviderContact[]> => {
-    if (this.googleContactsSearchEnabled) {
+    if (await this.googleContactsSearchEnabled) {
```
rrrooommmaaa, line 619: I suggest to remove this condition, as we already check this before calling `searchContactsOnGoogle`

extension/chrome/elements/compose-modules/compose-recipients-module.ts
ioanmo226: This condition was added because I need some sort of debugging.
Can remove this line

extension/chrome/elements/compose-modules/compose-recipients-module.ts
ioanmo226: Timeout makes sense to me
Maybe need to update `api` infra to accept optional timeout param and pass it to $.ajax if passed?
https://github.com/FlowCrypt/flowcrypt-browser/blob/01c1e0087013661c1df82a0435f070d3f0cb7354/extension/js/common/api/shared/api.ts#L73

rrrooommmaaa: If the request times out, should we try to search contacts? E.g. introduce a third state (disabled, enabled, maybe-enabled)?

ioanmo226: I think so

extension/chrome/settings/setup/setup-render.ts
```diff
@@ -107,11 +107,12 @@ export class SetupRenderModule {
       return await Settings.promptToRetry(e, Lang.setup.failedToCheckIfAcctUsesEncryption, () => this.renderSetupDialog(),
         Lang.general.contactIfNeedAssistance(this.view.isFesUsed()));
     }
+    const scopes = await AcctStore.getScopes(this.view.acctEmail);
     if (keyserverRes.pubkeys.length) {
       if (!this.view.clientConfiguration.canBackupKeys()) {
         // they already have a key recorded on attester, but no backups allowed on the domain. They should enter their prv manually
         this.displayBlock('step_2b_manual_enter');
-      } else if (this.view.storage!.email_provider === 'gmail' && this.view.scopes!.modify) {
+      } else if (this.view.storage!.email_provider === 'gmail' && scopes.modify) {
```
rrrooommmaaa, line 115: If I understand correctly, we decided to not check `scopes.modify` and this was implemented by your fix #4611

I looked at the code and see that `GoogleAuth.newOpenidAuthPopup` is never used and can be removed, hence `GoogleAuth.defaultScopes('openid')` is never used and this option can be removed, therefore `modify` should be present for all (at least new) setups, right?

ioanmo226: Yeah, I agree

extension/chrome/elements/compose-modules/compose-recipients-module.ts
rrrooommmaaa: Please remove

extension/chrome/elements/compose-modules/compose-recipients-module.ts
rrrooommmaaa: Please remove

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -383,6 +383,11 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     this.setEmailsPreview();
   };
 
+  private queryIfGoogleSearchEnabled = async () => {
+    const scopes = await AcctStore.getScopes(this.view.acctEmail);
+    return scopes.readContacts && scopes.readOtherContacts;
```
rrrooommmaaa, line 388: I'd me more comfortable, if we wrap this call in try/catch and return `undefined` in case of exception. (So we can try and search Google contacts in case of `!== false` and leave
```
      if (!(await this.googleContactsSearchEnabled)) {
        this.addBtnToAllowSearchContactsFromGoogle(input);
      }
```
as it is now.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
rrrooommmaaa: We agreed to search google contacts on `undefined` too.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -383,6 +383,15 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     this.setEmailsPreview();
   };
 
+  private queryIfGoogleSearchEnabled = async () => {
+    try {
+      const scopes = await AcctStore.getScopes(this.view.acctEmail);
+      return scopes.readContacts && scopes.readOtherContacts;
+    } catch {
+      return undefined;
```
rrrooommmaaa, line 391: It'd be nice to log something to console or `debug`

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -555,19 +565,19 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
       this.view.errModule.debug(`searchContacts db count: ${contacts.length}`);
-      this.renderSearchRes(input, contacts, { substring });
+      await this.renderSearchRes(input, contacts, { substring });
       if (contacts.length >= this.MAX_CONTACTS_LENGTH) {
         this.view.errModule.debug(`searchContacts 2, count: ${contacts.length}`);
         return;
       }
       let foundOnGoogle: EmailProviderContact[] = [];
-      if (this.googleContactsSearchEnabled) {
+      if (await this.googleContactsSearchEnabled || await this.googleContactsSearchEnabled === undefined) {
```
rrrooommmaaa, line 574: please....
`(await this.googleContactsSearchEnable) !== false`

ioanmo226: Yeah, That looks more clear.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -709,7 +717,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     } else {
       this.setContactPopupStyle(input);
       contactEl.find('ul').html('<li data-test="no-contact-found">No Contacts Found</li>'); // xss-direct
-      if (!this.googleContactsSearchEnabled) {
+      if (!(await this.googleContactsSearchEnabled)) {
```
rrrooommmaaa, line 720: Do you think we should show "Enable Google Contact Search" button when a request to get token info fails? Wouldn't that be misleading?
I suggest to use strictly `(await this.googleContactsSearchEnabled) === false` here

ioanmo226: I agree

rrrooommmaaa (approved):

PR #4816, opened 2022-11-28
Prevent http leaks when viewing/downloading attachments

martgil:
This PR adds a domain check before allowing users to download/preview an attachment. 

close https://github.com/FlowCrypt/flowcrypt-security/issues/225

Screenshot:

image

- Test was added

### To be filled by reviewers

- [x] is accompanied with tests


ioanmo226:
@martgil Please fix test

extension/chrome/elements/pgp_block_modules/pgp-block-attachmens-module.ts
```diff
@@ -100,7 +100,7 @@ export class PgpBlockViewAttachmentsModule {
   };
 
   private isAttachmentURLInvalid = (attachment: Attachment): boolean => {
-    return new URL(String(attachment.url)).host !== 'flowcrypt.s3.amazonaws.com';
+    return attachment.url !== undefined && new URL(attachment.url).host !== 'flowcrypt.s3.amazonaws.com';
```
ioanmo226, line 103: `attachment.url !== undefined` can be shortened to `!attachment.url` (which also checks empty string)

martgil: Thank you - I learned something new here.

martgil:
@rrrooommmaaa ready for review, sir. Thank you

extension/chrome/elements/pgp_block_modules/pgp-block-attachmens-module.ts
```diff
@@ -91,4 +99,8 @@ export class PgpBlockViewAttachmentsModule {
     }
   };
 
+  private isAttachmentURLInvalid = (attachment: Attachment): boolean => {
+    return !!attachment.url && new URL(attachment.url).host !== 'flowcrypt.s3.amazonaws.com';
```
rrrooommmaaa, line 103: There are several (two?) paths where an attachment can be initialized with `url` parameter.
One of them is `processGoogleDriveAttachments` and obviously the url `flowcrypt.s3.amazonaws.com` doesn't fit there.
The other one is `MsgBlockParser.extractFcAttachments`, so It'd be correct to exclude malicious urls in that method, by including the URL in the regex.
Please add a test with decrypting a message containing a malicious URL so we can see how it looks and  that it's not converted to an attachment.

rrrooommmaaa (changes requested):

martgil:
@rrrooommmaaa @ioanmo226 ready for a review, again. Thank you!

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -228,7 +228,11 @@ export class PgpBlockViewRenderModule {
     // todo - replace with MsgBlockParser.fmtDecryptedAsSanitizedHtmlBlocks, then the extract/strip methods could be private?
     if (!Mime.resemblesMsg(decryptedBytes)) {
       const fcAttachmentBlocks: MsgBlock[] = [];
-      decryptedContent = MsgBlockParser.extractFcAttachments(decryptedContent, fcAttachmentBlocks);
+      const extractedFcAttachmentsResult = await MsgBlockParser.extractFcAttachments(decryptedContent, fcAttachmentBlocks);
+      if (extractedFcAttachmentsResult.error) {
+        await Ui.modal.warning(extractedFcAttachmentsResult.error);
+      }
+      decryptedContent = extractedFcAttachmentsResult.decryptedContent;
```
ioanmo226, line 235: Cab be shortened like below

```ts
const { decryptedContent: decrypted, error } = await MsgBlockParser.extractFcAttachments(decryptedContent, fcAttachmentBlocks);
if (error) {
  await Ui.modal.warning(error);
}
decryptedContent = MsgBlockParser.stripFcTeplyToken(decrypted);
```

ioanmo226: Also I think we need to add `return` statement in `if(error)` block

extension/js/common/core/msg-block-parser.ts
```diff
@@ -84,20 +85,29 @@ export class MsgBlockParser {
     return { blocks, subject: decoded.subject, isRichText, webReplyToken };
   };
 
-  public static extractFcAttachments = (decryptedContent: string, blocks: MsgBlock[]) => {
+  public static extractFcAttachments = async (decryptedContent: string, blocks: MsgBlock[]) => {
     // these tags were created by FlowCrypt exclusively, so the structure is rigid (not arbitrary html)
     // `<a href="${attachment.url}" class="cryptup_file" cryptup-data="${fcData}">${linkText}</a>\n`
     // thus we use RegEx so that it works on both browser and node
+    let error = '';
+    const fcAttachmentsPattern = /<a\s+href="([^"]+)"\s+class="cryptup_file"\s+cryptup-data="([^"]+)"\s*>[^<]+<\/a>\n?/gm;
```
rrrooommmaaa, line 93: Thank you for the demo, @martgil.
My initial suggestion was to include the URL in the regex, like
`/<a\s+href="(https?:\/\/flowcrypt\.s3\.amazonaws\.com\/[^"]+)"\s+class="cryptup_file"\s+cryptup-data="([^"]+)"\s*>[^<]+<\/a>\n?/gm`
so any external URL will be naturally skipped and result in simple text rendering like
image
Your implementation gives user some insight into a possible attack, but vaguely.
image
We need to decide first whether we actually need this extra UI (and making the user press 'OK' may also be redundant), especially by complicating the code.
If we do need to display something special for this case, we can think about simplifying the parsing/rendering code next.
@ioanmo226, @tomholub What would you vote for?

ioanmo226: Well, I think we need to show original content (like @rrrooommmaaa suggested) + notify user via `toast` (something like `Skipping attachment rendering and show original content because attachment url is modified/invalid`

tomholub: Whatever is simpler to implement. Could be:
 - toast
 - modal
 - changing the content of the message to include `[skipped attachment due to invalid url]`

test/source/tests/decrypt.ts
```diff
@@ -800,6 +800,17 @@ XZ8r4OC6sguP/yozWlkG+7dDxsgKQVBENeG6Lw==
...
 
+    ava.default('decrypt - prevent rendering of attachments from domain sources other than flowcrypt.s3.amazonaws.com1', testWithBrowser('compatibility', async (t, browser) => {
+      const threadId1 = '184cc6aa8e884397';
+      const acctEmail = 'flowcrypt.compatibility@gmail.com';
+      const inboxPage = await browser.newPage(t, TestUrls.extension(`chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId1}`));
+      await inboxPage.waitAll('iframe');
+      const pgpBlock = await inboxPage.getFrame(['pgp_block.htm']);
+      await pgpBlock.waitForSelTestState('ready');
+      await pgpBlock.waitForContent('@pgp-block-content', '[skipped attachment due to invalid url]');
+      await pgpBlock.notPresent(['.preview-attachment', '@download-attachment-0']);
```
rrrooommmaaa, line 811: Good, you added a positive test -- the warning should appear.
I added a negative test -- the attachment should not appear.

extension/js/common/core/msg-block-parser.ts
```diff
@@ -90,6 +90,9 @@ export class MsgBlockParser {
     // thus we use RegEx so that it works on both browser and node
     if (decryptedContent.includes('class="cryptup_file"')) {
       decryptedContent = decryptedContent.replace(/<a\s+href="([^"]+)"\s+class="cryptup_file"\s+cryptup-data="([^"]+)"\s*>[^<]+<\/a>\n?/gm, (_, url, fcData) => {
+        if (!decryptedContent.match(/<a\s+href="(https?:\/\/flowcrypt\.s3\.amazonaws\.com\/[^"]+)"\s+class="cryptup_file"\s+cryptup-data="([^"]+)"\s*>[^<]+<\/a>\n?/gm)) {
```
rrrooommmaaa, line 93: Don't need to call the regex.match() twice.
Please check `url` parameter here. It has the value of what is captured inside of `href="..."`
It's ok to extract the host from there like you previously did.

rrrooommmaaa (approved):

Issue #4734, opened 2022-10-12
one mock instance per test


tomholub:
The contents of "First PR" as defined in #4494

PR #4821, opened 2022-11-30
#4815 Ensure gracious behavior & ui should remain functional when updating client configuration

### LINKED ISSUE #4815: when updating client configuration on already set up browser extension, ensure gracious behavior & ui should remain functional
ensure that the browser extension handles badly formatted client configuration during client configuration updates more graciously:
 - it should not hide the interface from the user if it was already set up with previous client configuration (probably happens due to unhandled exception)
 - Instead, it should render an error when it receives invalid configuration update, and retain original functioning configuration

ioanmo226:
This PR ensures gracious behavior & ui should remain functional when updating client configuration

close #4815 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/settings.ts
+++ b/test/source/tests/settings.ts
@@ -999,9 +999,11 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
       };
       const extraAuthHeaders = { Authorization: `Bearer ${accessToken}` };
       const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
-      await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, 'Failed to update FlowCrypt Client Configuration: Missing client configuration flags.');
+      const errorMsg = 'Failed to update FlowCrypt Client Configuration: Missing client configuration flags.';
+      await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, errorMsg);
       // Ensure previous client configuration remains same
       const settingsPage = await browser.newPage(t, TestUrls.extensionSettings(acct));
+      await PageRecipe.waitForToastToAppearAndDisappear(settingsPage, errorMsg);
       const {
         cryptup_testupdatesettingsflowcrypttest_rules: rules2
       } = await settingsPage.getFromLocalStorage([
--- a/test/source/tests/settings.ts
+++ b/test/source/tests/settings.ts
@@ -971,14 +971,14 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
     }));
 
     ava.default('settings - ensure gracious behavior & ui should remain functional when updating client configuration', testWithBrowser(undefined, async (t, browser) => {
-      const acct = 'settings@settings.flowcrypt.test';
+      const acct = 'test-update@settings.flowcrypt.test';
       mockBackendData.clientConfigurationByAcctEmail[acct] = keyManagerAutogenRules;
       const setupPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acct);
       await SetupPageRecipe.autoSetupWithEKM(setupPage);
       const {
-        cryptup_settingssettingsflowcrypttest_rules: rules1
+        cryptup_testupdatesettingsflowcrypttest_rules: rules1
       } = await setupPage.getFromLocalStorage([
-        'cryptup_settingssettingsflowcrypttest_rules'
+        'cryptup_testupdatesettingsflowcrypttest_rules'
       ]);
       const clientConfiguration1 = rules1 as ClientConfiguration;
       expect(clientConfiguration1.flags).to.eql([
@@ -990,7 +990,6 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
       expect(clientConfiguration1.disallow_attester_search_for_domains).to.eql([]);
       expect(clientConfiguration1.enforce_keygen_algo).to.equal('rsa2048');
       expect(clientConfiguration1.key_manager_url).to.equal('https://localhost:8001/flowcrypt-email-key-manager');
-      // open the settings page
       const accessToken = await BrowserRecipe.getGoogleAccessToken(setupPage, acct);
       await setupPage.close();
       // Set invalid client configuration and check if it ensures gracious behavior & ui remain functional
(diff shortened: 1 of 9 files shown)


test/source/tests/settings.ts
```diff
@@ -970,6 +970,26 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+      await SetupPageRecipe.autoSetupWithEKM(setupPage);
+      await setupPage.close();
+      // Set invalid client configuration and check if it ensures gracious behavior & ui remain functional
+      mockBackendData.clientConfigurationByAcctEmail[acct] = {
+        // flags is required but don't return it (to mock invalid client configuration)
+        key_manager_url: 'https://localhost:8001/flowcrypt-email-key-manager'
+      };
+      // open the settings page
+      const settingsPage = await browser.newPage(t, TestUrls.extensionSettings(acct));
+      const accessToken = await BrowserRecipe.getGoogleAccessToken(settingsPage, acct);
```
rrrooommmaaa, line 986: You can extract the token via `setupPage` before closing it (line 978). No need to open a new page, any extension page would do.

test/source/tests/settings.ts
```diff
@@ -970,6 +970,25 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+      // Set invalid client configuration and check if it ensures gracious behavior & ui remain functional
+      mockBackendData.clientConfigurationByAcctEmail[acct] = {
+        // flags is required but don't return it (to mock invalid client configuration)
+        key_manager_url: 'https://localhost:8001/flowcrypt-email-key-manager'
+      };
+      const extraAuthHeaders = { Authorization: `Bearer ${accessToken}` };
+      // const gmailPage =
+      const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
+      await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, 'Failed to update FlowCrypt Client Configuration: Missing client configuration flags.');
+    }));
```
rrrooommmaaa, line 990: It'd be good to load the client configuration `cryptup_<acct>_rules` from local storage to check that it's unchanged the way we do in `settings - client configuration gets updated on settings and content script reloads` test.
You can do using the same `setupPage` (setupPage.getFromLocalStorage), I guess (no need to close it).
And the client configuration is also updated when opening `settings` page, need to check its behaviour too.

ioanmo226: Yeah, nice suggestion

test/source/tests/settings.ts
```diff
@@ -970,6 +970,56 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
       await settingsPage.close();
     }));
 
+    ava.default('settings - ensure gracious behavior & ui should remain functional when updating client configuration', testWithBrowser(undefined, async (t, browser) => {
+      const acct = 'settings@settings.flowcrypt.test';
+      mockBackendData.clientConfigurationByAcctEmail[acct] = keyManagerAutogenRules;
```
rrrooommmaaa, line 975: This would interfere with `settings - client configuration gets updated on settings and content script reloads` test if run in parallel, as there is the same `acct` used. So we either need a different `acct` for this test or finish #4734 first.

test/source/tests/settings.ts
```diff
@@ -970,6 +970,56 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+      expect(clientConfiguration1.flags).to.eql([
+        'NO_PRV_BACKUP',
+        'ENFORCE_ATTESTER_SUBMIT',
+        'PRV_AUTOIMPORT_OR_AUTOGEN',
+        'PASS_PHRASE_QUIET_AUTOGEN',
+        'DEFAULT_REMEMBER_PASS_PHRASE']);
+      expect(clientConfiguration1.disallow_attester_search_for_domains).to.eql([]);
+      expect(clientConfiguration1.enforce_keygen_algo).to.equal('rsa2048');
+      expect(clientConfiguration1.key_manager_url).to.equal('https://localhost:8001/flowcrypt-email-key-manager');
+      // open the settings page
```
rrrooommmaaa, line 993: This comment is irrelevant, I see no opening of settings page here

test/source/tests/settings.ts
ioanmo226: You are right. Just changed `acct`.

test/source/tests/settings.ts
```diff
@@ -970,6 +970,55 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+      // Set invalid client configuration and check if it ensures gracious behavior & ui remain functional
+      mockBackendData.clientConfigurationByAcctEmail[acct] = {
+        // flags is required but don't return it (to mock invalid client configuration)
+        key_manager_url: 'https://localhost:8001/flowcrypt-email-key-manager'
+      };
+      const extraAuthHeaders = { Authorization: `Bearer ${accessToken}` };
+      const gmailPage = await browser.newPage(t, TestUrls.mockGmailUrl(), undefined, extraAuthHeaders);
+      await PageRecipe.waitForToastToAppearAndDisappear(gmailPage, 'Failed to update FlowCrypt Client Configuration: Missing client configuration flags.');
+      // Ensure previous client configuration remains same
+      const settingsPage = await browser.newPage(t, TestUrls.extensionSettings(acct));
```
rrrooommmaaa, line 1004: I expected to see the same toast here as `accountGetAndUpdateLocalStore` is also called when loading `settings/index.htm` (see `settings/index.ts`) but only saw this 
image


Also, I see that the client configuration is loaded often when composing a pwd-protected message to get `default_message_expire`. So in case of misconfiguration, the compose module won't get its `{ default_message_expire: 180 }` in case of FES, or a value from FlowCryptAPI even if it is there.
Can you please add a test of composing to a pwd recipient here (when client configuration is corrupt but `default_message_expire` is ok, so we can see what the actual result will be?

Still it's a rather strange excuse to update client configuration -- to get the freshest value of `default_message_expire` -- like this is the most important setting. In case of FES -- FES doesn't even return `default_message_expire` -- so this is simply an excuse. Why not just refresh client configuration on every load of compose frame? Or, on the opposite, never refresh the configuration here. And how to notify the user of corrupt configuration. And how often? These are considerations for discussion with @tomholub

ioanmo226: I agree. Just updated.
> I expected to see the same toast here as accountGetAndUpdateLocalStore is also called when loading settings/index.htm (see settings/index.ts) but only saw this

ioanmo226: @tomholub Please confirm this so that I can continue

rrrooommmaaa: It seems to work ok in our main test cases: `gmail` page and `settings` page.
We can create a separate issue for discussion on updating client configuration in compose boxes and respective tests

ioanmo226: You mean we can merge this PR now and take care of above in separate issue?
You have other feedbacks in this PR?

rrrooommmaaa: > You mean we can merge this PR now and take care of above in separate issue?

I think so, yes.

rrrooommmaaa (approved):

PR #4829, opened 2022-12-06
#4771 Better ux for disabled password messages

### LINKED ISSUE #4771: better ux for disabled password messages: December/January
I'd like to add a client configuration flag `DISABLE_FLOWCRYPT_HOSTED_PASSWORD_MESSAGES`.

When present, and recipients are missing a public key, and the user is using flowcrypt.com/api (not FES), it will not give the user option to enter a message password, as if that functionality didn't exist.

need to then check:
 - that the wording / user instructions in that situation still make sense
 - that what happens after clicking the send button (even though missing pubkey makes sense)

in neither case should it be prompting the user to add any password, just focus on public keys

(standard behavior is to prompt for public keys OR a password)

ioanmo226:
This PR introduces DISABLE_FLOWCRYPT_HOSTED_PASSWORD_MESSAGES client configuration flag.
When this flag is present & recipient pub keys are missing & don't use FES, treat like password protected message feature is not there.

close #4771 // if this PR closes an issue

image

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-err-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-err-module.ts
@@ -151,7 +151,7 @@ export class ComposeErrModule extends ViewModule<ComposeView> {
     }
   };
 
-  public throwIfEncryptionPasswordInvalid = async ({ subject, pwd }: { subject: string, pwd?: string }) => {
+  public throwIfEncryptionPasswordInvalidOrDisabled = async ({ subject, pwd }: { subject: string, pwd?: string }) => {
     // When DISABLE_FLOWCRYPT_HOSTED_PASSWORD_MESSAGES present, and recipients are missing a public key, and the user is using flowcrypt.com/api (not FES)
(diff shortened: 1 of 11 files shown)


extension/chrome/elements/compose-modules/compose-err-module.ts
```diff
@@ -152,6 +152,10 @@ export class ComposeErrModule extends ViewModule<ComposeView> {
   };
 
   public throwIfEncryptionPasswordInvalid = async ({ subject, pwd }: { subject: string, pwd?: string }) => {
```
rrrooommmaaa, line 154: We should rename this method to fully reflect its function, e.g. `throwIfEncryptionPasswordInvalidOrDisabled`
Otherwise, all looks goof to me 👍

ioanmo226: Sounds good

rrrooommmaaa (approved):

Issue #4782, opened 2022-11-08
Key import issue


ioanmo226:
![image](https://user-images.githubusercontent.com/77344191/200562577-b2b3c9f3-4d04-40ad-93a3-a60ae8856b3f.png)

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGqRQJkGxhDhWksnJrlwpzgKTsx

ioanmo226:
@martgil I guess you received this message. Could you share human email link here for further reference?

martgil:
Hello @ioanmo226 yes - no problem. By the way, have you already received the private key from the customer?

ioanmo226:
Yes.

ioanmo226:
Not sure this task priority though. Can you confirm @martgil?

martgil:
Usually, @tomholub sets a dedicated prioritization for these. If you have the private key and passphrase, I can take over this so you can focus on your tasks in the meantime. You may share the credentials via email and I'll try to analyze the key first.

ioanmo226:
Thank you.
Just sent email. Let me know if it works.

tomholub:
priority high. it's an important customer

i suspect Roman will have to take a look. i think he is the only one who may have a chance to actually fix it, or be able to clearly see what is wrong compared to similar keys the extension can normally fix

martgil:
@tomholub I take the first step to import the key with `gpg` - that fixes the format of the key and allows me to import/use the key with the FlowCrypt extension.

However, I'm now getting a "missing private key parameters" -- which is caused by keys exported via gpg's "--export-secret-subkeys" flag or the secret key is missing.

```
(1) Primary keys found: 1  
(2) ----- Testing key 0 -----  
(3) PK 0 > Key type: openpgp  
(4) PK 0 > Is Private?: [-] true  
(5) PK 0 > User id 0: User <redacted@domain.tld>  
(6) PK 0 > User id 1: redacted@domain.tld <redacted@domain.tld>  
(7) PK 0 > Primary User: User <redacted@domain.tld>  
(8) PK 0 > Fingerprint: redacted  
(9) PK 0 > Subkeys: [-] 1  
(10) PK 0 > Primary key algo: [-] rsa_encrypt_sign  
(11) PK 0 > Usage flags: [-] [sign_data, certify_keys]  
(12) PK 0 > key decrypt: [-] success  
[...]

PR #4832, opened 2022-12-08
#3346 custom SKS server client configuration test 

### LINKED ISSUE #3346: custom SKS server client configuration test

ioanmo226:
This PR implemented custom SKS server client configuration test 

close #3346 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/setup.ts
+++ b/test/source/tests/setup.ts
@@ -992,6 +992,19 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
         await dbPage.close();
       }));
 
+    ava.default('user@custom-sks.flowcrypt.test - Respect custom key server url', testWithBrowser(undefined, async (t, browser) => {
+      const acct = 'user@custom-sks.flowcrypt.test';
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acct);
+      await SetupPageRecipe.autoSetupWithEKM(settingsPage);
+      const composePage = await ComposePageRecipe.openStandalone(t, browser, acct);
+      await ComposePageRecipe.fillMsg(composePage, { to: 'test@custom-sks.flowcrypt.test' }, 'Respect custom key server url');
+      await composePage.waitForContent('.email_address.has_pgp', 'test@custom-sks.flowcrypt.test');
+      await composePage.close();
+      await SettingsPageRecipe.toggleScreen(settingsPage, 'additional');
+      const contactsFrame = await SettingsPageRecipe.awaitNewPageFrame(settingsPage, '@action-open-contacts-page', ['contacts.htm', 'placement=settings']);
+      await contactsFrame.waitForContent('@custom-key-server-description', 'using custom SKS pubkeyserver: https://localhost:8001');
+    }));
+
     ava.default.todo('DEFAULT_REMEMBER_PASS_PHRASE with auto-generation when all keys are removed by EKM');
     // should we re-use the known passphrase or delete it from the storage in this scenario?
 
--- a/extension/chrome/settings/modules/contacts.ts
+++ b/extension/chrome/settings/modules/contacts.ts
@@ -68,7 +68,8 @@ View.run(class ContactsView extends View {
     let lineActionsHtml = '&nbsp;&nbsp;<a href="#" class="action_export_all">export all</a>&nbsp;&nbsp;' +
       '&nbsp;&nbsp;<a href="#" class="action_view_bulk_import" data-test="action-show-import-public-keys-form">import public keys</a>&nbsp;&nbsp;';
     if (this.clientConfiguration.getCustomSksPubkeyServer()) {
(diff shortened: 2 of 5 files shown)


test/source/tests/setup.ts
```diff
@@ -992,6 +992,19 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
         await dbPage.close();
       }));
 
+    ava.default('user@custom-sks.flowcrypt.test - Respect custom key server url', testWithBrowser(undefined, async (t, browser) => {
+      const acct = 'user@custom-sks.flowcrypt.test';
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acct);
+      await SetupPageRecipe.autoSetupWithEKM(settingsPage);
+      const composePage = await ComposePageRecipe.openStandalone(t, browser, acct);
+      await ComposePageRecipe.fillMsg(composePage, { to: 'test@custom-sks.flowcrypt.test' }, 'Respect custom key server url');
+      await composePage.waitForContent('.email_address.has_pgp', 'test@custom-sks.flowcrypt.test');
```
rrrooommmaaa, line 1001: If you try to actually send a message with this key, it turns out to be invalid for encryption. I created #4833. So this key isn't really `has_pgp`, how did you created it? Save this key to test the implementation of #4833

rrrooommmaaa (commented):
Looks good except that the key isn't usable

test/source/tests/setup.ts
ioanmo226: Nothing special. Created key with GPG Keychain (with no expiration)

ioanmo226: Nice catch though

ioanmo226: Hmmm.. @rrrooommmaaa I tried to reproduce above issue you mentioned but I couldn't.
As I debugged, it says usableForEncryption is true.
How did you reproduce it?

image

rrrooommmaaa: when pressing "Send" button, the key is loaded from the storage, validation happens there

rrrooommmaaa: I noticed too that sometimes validation is obviously skipped (the key is usable), so that's another point of concern

ioanmo226: When pressing send button, i get another error.
image

rrrooommmaaa: Right, I'm getting the same now but I got different thing in the morning!
Looks like this is issue #4782 again!
The key creation time is 1670482106000 which is 	Thu Dec 08 2022 06:48:26 GMT (yesterday)
I suppose this is the time when you created the key? Why wasn't the key valid this morning for me? Hmmm...
Maybe it's a bug specific to OpenPGP v4
Looks like the key is invalid for about 24 hours after creation??

ioanmo226: Yes
> I suppose this is the time when you created the key?
Not sure
> Looks like the key is invalid for about 24 hours after creation??

ioanmo226: But I don't think it's invalid for 24 hours after creation.
Test worked for me & semaphoreci side

ioanmo226: I tried to create new key and test with that key but it works fine with new key.

ioanmo226: Could you try to replicate above issue you mentioned?

rrrooommmaaa: I guess we can merge this PR, and I'm not sure what the system clock was on my virtual machine in the morning -- it may have remained back in the past before I rebooted it, since testing of #4782
Anyway, issue #4833 is still relevant -- the key wasn't usable and was green.
And we have no means to inform the user that an imported pubkey has a date in the future, which isn't convenient

rrrooommmaaa: Looks like my system clock was indeed in the past

rrrooommmaaa (approved):

PR #4840, opened 2022-12-15
Prevent pasting large text on secure compose window

### LINKED ISSUE #4818: Limit size of message text when pasting to the secure compose message box
Need to modify `.on("paste"`  handler as suggested in this discussion

_Originally posted by @rrrooommmaaa in https://github.com/FlowCrypt/flowcrypt-browser/pull/4536#discussion_r944252947_

martgil:
This PR prevents pasting large text on secure compose window

close #4818  

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -205,14 +205,26 @@ export class ComposeView extends View {
     this.S.cached('body').on('focusin', setActiveWindow);
     this.S.cached('body').on('click', setActiveWindow);
     this.S.cached('icon_help').on('click', this.setHandler(async () => await this.renderModule.openSettingsWithDialog('help'), this.errModule.handle(`help dialog`)));
-    this.S.cached('input_text').on('paste', this.setHandler(async (el, ev) => {
+    this.S.cached('input_intro').on('paste', this.setHandler(async (el, ev: JQuery.Event) => {
       const clipboardEvent = ev.originalEvent as ClipboardEvent;
       const maxPlainTextLength = 10;
       if (clipboardEvent.clipboardData) {
         const clipboardLength = clipboardEvent.clipboardData.getData('text').length;
         if (clipboardLength > maxPlainTextLength) {
-          console.log(clipboardLength);
-          ev.preventDefault();
+          ev.preventDefault(); // this works
+        }
+      }
+    }));
+    this.S.cached('input_text').on('paste', this.setHandler(async (el, ev: JQuery.Event) => {
+      const clipboardEvent = ev.originalEvent as ClipboardEvent;
+      const maxPlainTextLength = 10;
+      if (clipboardEvent.clipboardData) {
+        const clipboardLength = clipboardEvent.clipboardData.getData('text').length;
+        console.log(clipboardLength);
+        console.log(clipboardLength, maxPlainTextLength);
+        console.log(clipboardLength > maxPlainTextLength);
+        if (clipboardLength > maxPlainTextLength) {
+          ev.preventDefault(); // this doesn't work
         }
       }
     }));
--- a/extension/chrome/elements/compose.ts
+++ b/extension/chrome/elements/compose.ts
@@ -154,7 +154,7 @@ export class ComposeView extends View {
     const clipboardEvent = ev.originalEvent as ClipboardEvent;
     if (clipboardEvent.clipboardData) {
       const clipboardLength = clipboardEvent.clipboardData.getData('text').length;
-      if (clipboardLength > limit) {
+      if (clipboardLength + el.innerText.length > limit) {
         ev.preventDefault();
       }
     }
@@ -215,11 +215,11 @@ export class ComposeView extends View {
     this.S.cached('body').on('focusin', setActiveWindow);
     this.S.cached('body').on('click', setActiveWindow);
     this.S.cached('icon_help').on('click', this.setHandler(async () => await this.renderModule.openSettingsWithDialog('help'), this.errModule.handle(`help dialog`)));
-    this.S.cached('input_text').on('paste', this.setHandler(async (el, ev: JQuery.Event) => {
-      this.setInputLimit(el, ev, 10); // this works
+    this.S.cached('input_text').on('paste', this.setHandler((el, ev: JQuery.Event) => {
+      this.setInputLimit(el, ev, 10); // doesn't prevent pasting
     }));
-    this.S.cached('input_intro').on('paste', this.setHandler(async (el, ev: JQuery.Event) => {
-      this.setInputLimit(el, ev, 10); // this doesn't work
+    this.S.cached('input_intro').on('paste', this.setHandler((el, ev: JQuery.Event) => {
+      this.setInputLimit(el, ev, 10); // does work
     }));
     this.attachmentsModule.setHandlers();
     this.inputModule.setHandlers();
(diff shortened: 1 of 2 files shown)


martgil:
@DenBond7 looks like `clipboardEvent.clipboardData.getData('text')` is smart enough not to include images when reading clipboard. I only got one problem left, the `ev.preventDefault();` is surprisly didn't work.

I already tried to put only the`ev.preventDefault();` within the on.paste block but it still continues to paste whatever text is on my clipboard.

extension/chrome/elements/compose.ts
```diff
@@ -205,6 +215,12 @@ export class ComposeView extends View {
     this.S.cached('body').on('focusin', setActiveWindow);
     this.S.cached('body').on('click', setActiveWindow);
     this.S.cached('icon_help').on('click', this.setHandler(async () => await this.renderModule.openSettingsWithDialog('help'), this.errModule.handle(`help dialog`)));
+    this.S.cached('input_text').on('paste', this.setHandler(async (el, ev: JQuery.Event) => {
```
rrrooommmaaa, line 218: FYI `S.cached('input_text')` and `S.cached('input_intro')` may work differently because they are defined differently in `extension/chrome/elements/compose.ts`
```
public S = Ui.buildJquerySels({
...
input_text: 'div#input_text',
...
input_intro: '.input_intro',
...
```

Also, please remove `async` from handler function definitions: they are unnecessary and confusing.

martgil: Thanks for remarks. Both contenteditable div are being selected successfully and the `onpaste` event gets triggered but I have no clue why it won't work for `input_text`. I also tried binding the `onpaste` event on its child node but still doesn't give the expected result while `input_intro` works very well.

martgil: I started to get more info with the issue. The `ev.currentTarget` and `ev.target` is different when the event is triggered . The currentTarget is `div#input_text` which is correct, however, the `ev.target` is either the child node within the currentTarget and that is `<div>` and `<br>` so the paste being prevented is only for that two element making the `ev.preventDefault()` ineffective.

extension/chrome/elements/compose.ts
```diff
@@ -217,6 +223,16 @@ export class ComposeView extends View {
   };
 
   public isFesUsed = () => Boolean(this.fesUrl);
+
+  private setInputLimit(el: HTMLElement, ev: JQuery.Event, limit: number) {
+    const clipboardEvent = ev.originalEvent as ClipboardEvent;
```
rrrooommmaaa, line 228: can you try to address the element with `ev.currentTarget`, not `el`?
I have a suspicion that `el` is actually `ComposeView`, not the element pasted to.

extension/chrome/elements/compose.ts
rrrooommmaaa: Oh you mean after these appends:
```
Xss.sanitizeAppend(this.view.S.cached('input_text'), `<br/>\n<br/>\n<br/>\n${Xss.escape(e.data)}`);
```
and
```
Xss.sanitizeAppend(this.view.S.cached('input_text'), this.getTripleDotSanitizedFormattedHtmlContent());
```

Have you tried to use `#input_text` instead of `div#input_text`? Would that help?
That is, add one more mapping like
```
public S = Ui.buildJquerySels({
...
input_text_full: '#input_text',
```
and then attach the handler
` this.S.cached('input_text_full').on('paste' ... `

martgil: > Have you tried to use #input_text instead of div#input_text? Would that help?

Yes, I do but unfortunately, it doesn't work.

martgil: Here is the comparison for an event on `input_text` vs `input_intro`:

`input_text`:
image

`input_intro`:
image

We can notice that input_intro has `currentTarget` and `target` has the same value. Where in input_text, `currentTarget` has the right element and `target` was set to `br` element.

When it comes to element hierarchy, `div#input_text` is the parent and the one should be prioritize to received the paste event over the `br` element?

image

rrrooommmaaa: Technically the event was generated at `br` level.
Can you tried something like https://stackoverflow.com/a/10558535 ?

extension/chrome/elements/compose.ts
martgil: > can you try to address the element with ev.currentTarget, not el?

Sorry I didn't get it.

> I have a suspicion that `el` is actually ComposeView, not the element pasted to

When I `console.log()` the `el` variable, the console returns the DOM tree of the element, for example, the intro_input or input_text.

image

extension/chrome/elements/compose.ts
martgil: During my testing case with somehow blocking child nodes from receiving the event, it doesn't work. `currentTarget: div#input_text` while `target: br`.

martgil: I'd also like to input limits on public key import, do you have any recommendation of what would be a good number for the input? Do you think it would reach 5Kb?

rrrooommmaaa: > During my testing case with somehow blocking child nodes from receiving the event, it doesn't work. `currentTarget: div#input_text` while `target: br`.

Have you played with this fiddle http://jsfiddle.net/xvKNC/2/   
?
Can you create a fiddle setup for the problem?
And test how it works with `.on("paste", "br" ...` (div2 example)

rrrooommmaaa: > I'd also like to input limits on public key import, do you have any recommendation of what would be a good number for the input? Do you think it would reach 5Kb?

If there is a single key, probably not. Need to check whether there is `parse` or `parseMany` etc.

martgil: @rrrooommmaaa I tried to do simple experiment on jsfiddle to conclude if try to prevent paste with `onpaste` here :https://jsfiddle.net/zqfx2h14/ and all contenteditable obeys the ev.preventdefault. 

on the extension, it only prevents pasting on input_intro but not on input_text. We can remember that input_intro doesn't have any change in terms of functionality unlike the input_text. I think the data has been processed/injected on <div> in div#input_text that's why binding paste don't really achieve anything there.

martgil: It seems i found the right place: https://github.com/FlowCrypt/flowcrypt-browser/blob/43255ea7bf508f02fff685693a2e1460f1fb690d/extension/chrome/elements/compose-modules/compose-input-module.ts#L81

martgil:
closing this pr. too messy.

extension/chrome/elements/compose.ts
rrrooommmaaa: > I tried to do simple experiment on jsfiddle to conclude if try to prevent paste with `onpaste` here :https://jsfiddle.net/zqfx2h14/ and all contenteditable obeys the ev.preventdefault.
> 
> on the extension, it only prevents pasting on input_intro but not on input_text. We can remember that input_intro doesn't have any change in terms of functionality unlike the input_text. I think the data has been processed/injected on
> 
> in div#input_text that's why binding paste don't really achieve anything there.

There can be `<br/>`s added programmatically, so for fiddling, I think, we'll need something like this:
```
<div contenteditable style="border: 1px solid orange">something before<br/>something after</div>
```

martgil:
Thank you, sir!

Actually, I have it fixed on my end and now preparing to add a test and
publish the pr.

thank you again sir for your prompt help, and have a nice day!

PR #4841, opened 2022-12-15
Clipped decrypted large messages

martgil:
This PR clipped the decrypted large message to prevent the app from crashing

close https://github.com/FlowCrypt/flowcrypt-security/issues/215

- Tests added or updated

### To be filled by reviewers


martgil:
@rrrooommmaaa ready for review. thank you!

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -222,10 +222,14 @@ export class PgpBlockViewRenderModule {
       this.setFrameColor('gray');
     }
     const publicKeys: string[] = [];
+    const maxDecryptedContentLength = 50000;
     let renderableAttachments: Attachment[] = [];
     let decryptedContent = decryptedBytes.toUtfStr();
     let isHtml: boolean = false;
     // todo - replace with MsgBlockParser.fmtDecryptedAsSanitizedHtmlBlocks, then the extract/strip methods could be private?
+    if (decryptedContent.length > maxDecryptedContentLength) {
+      decryptedContent = decryptedContent.substring(0, maxDecryptedContentLength) + ' [clipped - message too large]';
```
rrrooommmaaa, line 231: ok, if the decrypted content doesn't resemble Mime message, then this `decryptedContent` goes into line 234 etc.
But if it does, then it goes to line 243 and then it's not clipped (restored from the non-clipped `decryptedBytes`).
Won't the browser extension crash if it is too long there?

And also a question: should we clip the message before or after extracting FcAttachments, replyToken and public keys. If after (which is preferred unless it consumes a lot of processing time), then we should also think about limiting the number of attachments and public keys, as too many may also crash the frame or make it otherwise unusable. If before, the output may look a bit ugly, if a pubkey or attachment link is clipped midway.

rrrooommmaaa (commented):
I'd also recommend to implement creation of a dummy message (for both MIME and non-MIME encrypted data) of an arbitrary long size (can be later adjusted) on the fly, so that we can test better how the extension reacts to extremely big messages and where exactly it hangs or crashes. We'll have to customize our mock  `GoogleData.getMessage()` to provide a message of the required size based on id.

martgil:
> I'd also recommend to implement creation of a dummy message (for both MIME and non-MIME encrypted data) of an arbitrary long size (can be later adjusted) on the fly, so that we can test better how the extension reacts to extremely big messages and where exactly it hangs or crashes. We'll have to customize our mock `GoogleData.getMessage()` to provide a message of the required size based on id.

You're right, we should be testing mime and non-mime messages both will crash for extremely long messages. I'm still unsure how we could test for the right length before it crashes. I mean, let's assume that rendering 10 million chars (encrypted) will work on the better machines but that 10 million chars may be too much to handle for a dual-core PC or less (though it's acceptable as it's a comprised limitation of it).

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
martgil: @rrrooommmaaa thank you sir for your helpful comments.

> ok, if the decrypted content doesn't resemble Mime message, then this decryptedContent goes into line 234 etc.
But if it does, then it goes to line 243 and then it's not clipped (restored from the non-clipped decryptedBytes).
Won't the browser extension crash if it is too long there?

thank you. right it's a mistake. things should be clipped in both situation (mime/non-mime).

> And also a question: should we clip the message before or after extracting FcAttachments, replyToken and public keys. If after (which is preferred unless it consumes a lot of processing time), then we should also think about limiting the number of attachments and public keys, as too many may also crash the frame or make it otherwise unusable. If before, the output may look a bit ugly, if a pubkey or attachment link is clipped midway.

Thank you for the question - this gives me more realization about the situation. 

1. We're dealing with inline attachment from web portal so that is one problem when it comes to limiting attachments.
2. If we disregard the item no.1 (dealing with inline attachment) - I think we don't need to worry about attachments as it has 25mb limit?
3. I think it's reasonable to add a limit for number of public keys as we displayed their details on the page (we had check before rendering that is has to be a valid public key).

martgil: > @martgil were there any issues with this PR or something else should be implemented before we can finish this PR?

@sosnovsky No blocking issues. May I hear your suggestion for the following question above? I might just re-create this PR afterward. Thanks.

sosnovsky: I think here we should clip text in decrypted data after extracting all attachments, so it'll affect only rendered message text:
- decrypt full data
- extract all attachments
- if message text is longer than max length - clip it
- for attachments and public keys need to check how they affect page performance and then set MAX_ATTACHMENTS_NUMBER - maybe it'll work well even for 50 attachments

Issue #4837, opened 2022-12-14
LGTM migration to Github code scanning


ioanmo226:
lgtm.com seems deprecated and need to be migrated to github code scanning.

https://github.blog/2022-08-15-the-next-step-for-lgtm-com-github-code-scanning/

cc: @tomholub

tomholub:
https://github.com/FlowCrypt/flowcrypt-ios/issues/1908 here is a related fix

PR #4857, opened 2023-01-02
#4526 Replace TSLint with ESLint

### LINKED ISSUE #4526: Migrate tslint to eslint
Tslint is deprecated and needs to be migrated to eslint.

sosnovsky:
This PR migrates TypeScript linting from deprecated TSLint to ESLint. It also integrates `prettier` for formatting (as recommended by `typescript-eslint` - https://typescript-eslint.io/linting/troubleshooting/formatting)

For found eslint issues I haven't changed code, just added `eslint-disable` comments for these lines (as this PR has too many changed files and it would be difficult to find these code changes among other formatting updates)

I also updated VS Code recommended plugins list to remove deprecated plugins:
`mkaufman.htmlhint -> htmlhint.vscode-htmlhint`,
`philhindle.errorlens -> usernamehw.errorlens`

close #4526

### To be filled by reviewers


sosnovsky:
> #### ⚠️ GitGuardian has uncovered 1 secret following the scan of your pull request.

Should we ignore this alert?

test/source/tests/browser-unit-tests/unit-ContactStore.js
```diff
@@ -118,26 +116,38 @@ BROWSER_UNIT_TEST_NAME(`ContactStore.search ignores top-level domains and common
   return 'pass';
 })();
 
-BROWSER_UNIT_TEST_NAME(`ContactStore.search allows searching for name parts that look like top-level domains and common domains`);
+BROWSER_UNIT_TEST_NAME(`ContactStore.search allows searching for name parts that look like top-level domains and common domains`); // prettier-ignore
```
rrrooommmaaa, line 119: Is this because of backticks?
I suggest to add support for ordinary single quotes instead.
We can replace backtick with single quotes

tomholub: We should probably ignore whole test/source/tests/browser-unit-tests folder. It's not really javascript nor typescript.

extension/chrome/elements/attachment.ts
```diff
@@ -83,15 +103,19 @@ export class AttachmentDownloadView extends View {
...
+        .then(fileSize => {
+          if (typeof fileSize !== 'undefined') {
+            this.size = fileSize;
+          }
+        })
+        .catch(ApiErr.reportIfSignificant);
     }
     try {
-      this.canClickOnAttachment = ! await this.processAsPublicKeyAndHideAttachmentIfAppropriate();
+      this.canClickOnAttachment = !(await this.processAsPublicKeyAndHideAttachmentIfAppropriate());
```
tomholub, line 118: I suppose this is driven by some new eslint rule? `! await something;` is intuitive and nice to me, but maybe because I'm used to it. How do you see this?

extension/chrome/elements/attachment.ts
```diff
@@ -273,16 +329,30 @@ export class AttachmentDownloadView extends View {
...
+      );
     } else if (result.error.type === DecryptErrTypes.needPassphrase) {
-      BrowserMsg.send.passphraseDialog(this.parentTabId, { type: 'attachment', longids: result.longids.needPassphrase });
-      if (! await PassphraseStore.waitUntilPassphraseChanged(this.acctEmail, result.longids.needPassphrase, 1000, this.ppChangedPromiseCancellation)) {
+      BrowserMsg.send.passphraseDialog(this.parentTabId, {
+        type: 'attachment',
+        longids: result.longids.needPassphrase
+      });
+      if (
+        !(await PassphraseStore.waitUntilPassphraseChanged(
```
tomholub, line 341: and here. There will be a TON of places like this. If you actually do prefer `!(await something)` than that is ok with me.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -903,37 +1100,47 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
...
     const changedIndex = this.addedRecipients.findIndex(
-      (addedRecipient) => (!recipient.email || addedRecipient.email === recipient.email) && addedRecipient.id === recipient.id
+      addedRecipient =>
+        (!recipient.email || addedRecipient.email === recipient.email) && addedRecipient.id === recipient.id
     );
     if (changedIndex > -1) {
       this.addedRecipients.splice(changedIndex, 1, recipient);
     }
-    this.view.pwdOrPubkeyContainerModule.showHideContainerAndColorSendBtn(); // tslint:disable-line:no-floating-promises
+    this.view.pwdOrPubkeyContainerModule.showHideContainerAndColorSendBtn(); // eslint-disable-line @typescript-eslint/no-floating-promises
```
tomholub, line 1143: Maybe should still have `.catch(Catch.reportErr);`. That may resolve the floating promise warning anyway. Can be different PR

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -968,9 +1178,11 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     if (container) {
       this.view.sizeModule.resizeInput($(container).find('input'));
     }
-    this.view.S.cached('input_addresses_container_outer').find(`#input-container-${this.addedRecipients[index].sendingType} input`).focus();
+    this.view.S.cached('input_addresses_container_outer')
+      .find(`#input-container-${this.addedRecipients[index].sendingType} input`)
+      .focus();
     this.addedRecipients.splice(index, 1);
-    this.view.pwdOrPubkeyContainerModule.showHideContainerAndColorSendBtn(); // tslint:disable-line:no-floating-promises
+    this.view.pwdOrPubkeyContainerModule.showHideContainerAndColorSendBtn(); // eslint-disable-line @typescript-eslint/no-floating-promises
```
tomholub, line 1185: Same here, could just catch the errors

extension/chrome/elements/compose-modules/compose-render-module.ts
```diff
@@ -32,7 +31,7 @@ export class ComposeRenderModule extends ViewModule<ComposeView> {
     }
     await this.view.replyPopoverModule.render(this.view.isReplyBox);
     this.initComposeBoxStyles();
-    if (!this.view.draftId && await this.view.draftModule.localDraftGet()) {
+    if (!this.view.draftId && (await this.view.draftModule.localDraftGet())) {
```
tomholub, line 34: I'd say this one is even more questionable, `&& await something` looks legit to me and better than `&& (await something)` but since I don't contribute that much here anymore, I'll follow what you think

extension/chrome/settings/setup.ts
```diff
@@ -341,27 +450,35 @@ export class SetupView extends View {
...
         if (!hasMatchingKey) {
-          // eslint-disable-next-line max-len
-          throw new CompanyLdapKeyMismatchError(`Imported private key with ids ${prvs.map(prv => prv.key.id).join(', ')} does not match public keys on company LDAP server with ids ${parsedPubKeys.map(pub => pub.id).join(', ')} for ${this.acctEmail}. Please ask your help desk.`);
+          throw new CompanyLdapKeyMismatchError(
+            `Imported private key with ids ${prvs
+              .map(prv => prv.key.id)
+              .join(', ')} does not match public keys on company LDAP server with ids ${parsedPubKeys
+              .map(pub => pub.id)
+              .join(', ')} for ${this.acctEmail}. Please ask your help desk.`
+          );
```
tomholub, line 463: This formatting looks odd, the `.map` on next line doesn't look correct in the context of the `${}`

.prettierrc
```diff
@@ -0,0 +1,6 @@
+printWidth: 120
```
rrrooommmaaa, line 1: we can leave line width as it is now (190?)

extension/js/common/core/const.ts
```diff
@@ -25,7 +25,7 @@ export const GMAIL_RECOVERY_EMAIL_SUBJECTS = [
   'Your FlowCrypt Backup',
   'Your CryptUp Backup',
   'All you need to know about CryptUP (contains a backup)',
-  'CryptUP Account Backup',
+  'CryptUP Account Backup'
```
tomholub, line 28: I think the separator on every line is better, for languages that support it. It would be best to enforce that through linting, if possible.

Then you can add or remove lines without having to modify other lines that you didn't mean to add/remove (by adding/removing comma there), which keeps git history cleaner.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -41,7 +40,10 @@ export class OpenPGPKey {
   };
 
   public static validateAllDecryptedPackets = async (key: OpenPGP.key.Key): Promise<void> => {
-    for (const prvPacket of key.toPacketlist().filter(OpenPGPKey.isPacketPrivate).filter(packet => packet.isDecrypted())) {
+    for (const prvPacket of key
+      .toPacketlist()
+      .filter(OpenPGPKey.isPacketPrivate)
+      .filter(packet => packet.isDecrypted())) {
```
tomholub, line 46: I'm thinking maybe the `) {` could be on its own line, it would be a bit more readable

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -51,26 +53,43 @@ export class OpenPGPKey {
...
+      .filter(OpenPGPKey.isPacketPrivate) as PrvPacket[];
     if (!chosenPrvPackets.length) {
-      throw new Error(`No private key packets selected of ${prv.getKeys().map(k => k.keyPacket).filter(OpenPGPKey.isPacketPrivate).length} prv packets available`);
+      throw new Error(
+        `No private key packets selected of ${
+          prv
+            .getKeys()
+            .map(k => k.keyPacket)
+            .filter(OpenPGPKey.isPacketPrivate).length
+        } prv packets available`
```
tomholub, line 84: This looks awkward

extension/types/node-forge.d.ts
```diff
@@ -18,1081 +18,1085 @@
```
tomholub, line 18: also. Maybe we should ignore the whole `/types/` folder

test/source/tests/browser-unit-tests/unit-ContactStore.js
tomholub: Could be length (149 chars). Do you think we should just ignore the length here, if it eslint works ok for the rest?

tooling/tsc-compiler.ts
```diff
@@ -43,27 +43,41 @@ const getNameAndPos = (f: ts.FunctionLike) => {
  * This transformer will wrap content of all async functions with a try/catch that helps preserve proper async stack traces
  */
 const preserveAsyncStackTracesTransformerFactory = () => {
-  const createStackTracePreservingCatchBlockStatements = (f: ts.FunctionLike): ts.Statement[] => {
+  const createStackTracePreservingCatchBlockStatements = (f: ts.SignatureDeclaration): ts.Statement[] => {
     const statements: ts.Statement[] = [];
     const addStackLine = `\\n    at <async> ${getNameAndPos(f)}`;
     const code = `if(t instanceof Error){t.stack+="${addStackLine}";throw t}const e=new Error("Thrown["+typeof t+"]"+t);e.thrown=t;throw e`;
-    statements.push(ts.createStatement(ts.factory.createIdentifier(code)));
+    statements.push(ts.factory.createExpressionStatement(ts.factory.createIdentifier(code)));
```
tomholub, line 50: Are these changes related to updated TS version, or something else?

tooling/tslint-rules/noReturnAnyRule.ts
```diff
@@ -1,111 +0,0 @@
-/**
```
tomholub, line 1: This was a pretty good rule, was there a ready-made one from eslint?

tomholub:
@sosnovsky ok, this concludes my review. Quite a change - didn't expect that, and I suppose you didn't expect it either. While I'm not a fan of every formatting rule applied, consistency is key, so it's a step in the right direction. My comments above. Thanks!

tomholub:
> > #### warning GitGuardian has uncovered 1 secret following the scan of your pull request.
> 
> Should we ignore this alert?

Yes, while it looks like a secret (it's called client_secret), per available information for web apps that are client-side, it's not actually a secret. So we can ignore this.

test/source/tests/browser-unit-tests/unit-ContactStore.js
sosnovsky: I added `prettier-ignore` here as it was automatically wrapping text, so test case name was on the next line and it broke test file parsing at https://github.com/FlowCrypt/flowcrypt-browser/blob/feature/issue-4526-eslint/test/source/tests/unit-browser.ts#L68

sosnovsky: > We should probably ignore whole test/source/tests/browser-unit-tests folder. It's not really javascript nor typescript.

Yes, I disabled `eslint` in this folder, text wrapping was done by `prettier`

rrrooommmaaa: > I added `prettier-ignore` here as it was automatically wrapping text, so test case name was on the next line and it broke test file parsing at https://github.com/FlowCrypt/flowcrypt-browser/blob/feature/issue-4526-eslint/test/source/tests/unit-browser.ts#L68

If we allow length of 190 (as it was before), there would be much less changes which is good as it'd be easier to `git blame` etc.

tooling/tslint-rules/noReturnAnyRule.ts
sosnovsky: In iOS `Core` code we replaced it with [no-unsafe-return](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-unsafe-return.md) rule, it should work the same

tooling/tsc-compiler.ts
sosnovsky: Yes, these functions were marked as deprecated:

Screenshot 2023-01-02 at 20 18 21

No test failed after this change, so probably it should work the same as before.

extension/types/jquery.d.ts
```diff
@@ -24,8 +24,6 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
```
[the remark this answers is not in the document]
sosnovsky: By the way, do we really need this file for jquery definitions when we already have `@types/jquery` in `package.json`?

extension/js/common/core/const.ts
sosnovsky: Agree, `trailingComma` rule was enabled by default, but I disabled it as such separators were rarely used in FlowCrypt codebase. Will enable it back

test/source/tests/browser-unit-tests/unit-ContactStore.js
sosnovsky: 190 seems to be too much, as it makes some expressions quite difficult to read.

I think setting 120 as max line length will lead to better readability and encourage us to split complex expressions into simpler ones. For sure, there are some cases when `prettier` makes strange decisions about code wrap, but we can fix them manually or just disable `prettier` for some lines to produce code without any formatting.

There are too many format changes in this PR because I ran `prettier` across all files to make code look consistent, but all future code will be automatically formatted for any developer before commit.

By the way, it's not really 'max-length' rule, but our desired line length, as `prettier` documentation says:

> For readability we recommend against using more than 80 characters:
> 
> In code styleguides, maximum line length rules are often set to 100 or 120. However, when humans write code, they don’t strive to reach the maximum number of columns on every line. Developers often use whitespace to break up long lines for readability. In practice, the average line length often ends up well below the maximum.
> 
> Prettier’s printWidth option does not work the same way. It is not the hard upper allowed line length limit. It is a way to say to Prettier roughly how long you’d like lines to be. Prettier will make both shorter and longer lines, but generally strive to meet the specified printWidth.
> 
> Remember, computers are dumb. You need to explicitly tell them what to do, while humans can make their own (implicit) judgements, for example on when to break a line.
> 
> In other words, don’t try to use printWidth as if it was ESLint’s [max-len](https://eslint.org/docs/rules/max-len) – they’re not the same. max-len just says what the maximum allowed line length is, but not what the generally preferred length is – which is what printWidth specifies.

@rrrooommmaaa are you okay with trying to use `120` as our preferred line length?

extension/chrome/elements/attachment.ts
sosnovsky: I used to `!(await ...)` and it's default behavior of `prettier` formatter, probably because of consistency for different cases like:
```js
if (!(await wkd.lookupEmail(email)).pubkeys.length) { ...

// because it won't work if you'll forget to add brackets and write it this way:
if (! await wkd.lookupEmail(email).pubkeys.length) {...
```

extension/js/common/core/crypto/pgp/openpgp-key.ts
sosnovsky: Updated it to
```js
const availablePackets = prv
  .getKeys()
  .map(k => k.keyPacket)
  .filter(OpenPGPKey.isPacketPrivate);
throw new Error(`No private key packets selected of ${availablePackets.length} prv packets available`);
```

extension/js/common/core/crypto/pgp/openpgp-key.ts
sosnovsky: Updated it to
```js
const decryptedPackets = key
  .toPacketlist()
  .filter(OpenPGPKey.isPacketPrivate)
  .filter(packet => packet.isDecrypted());
for (const prvPacket of decryptedPackets) {
  await prvPacket.validate(); // gnu-dummy never raises an exception, invalid keys raise exceptions
}
```

extension/chrome/settings/setup.ts
sosnovsky: Updated to
```js
const keyIds = prvs.map(prv => prv.key.id).join(', ');
const pubKeyIds = parsedPubKeys.map(pub => pub.id).join(', ');
throw new CompanyLdapKeyMismatchError(
  `Imported private key with ids ${keyIds} does not match public keys on company LDAP server with ids ${pubKeyIds} for ${this.acctEmail}. Please ask your help desk.`
);
```

extension/chrome/elements/compose-modules/compose-render-module.ts
sosnovsky: Yeah, here `await something` looks cleaner, but I think formatter automatically adds brackets to be consistent for all `await` expressions and to avoid unexpected behavior when, for example, in the future we'll change code to something like
```js
await this.view.draftModule.localDraftGet().id

// which will work only with brackets 
(await this.view.draftModule.localDraftGet()).id
```

.prettierrc
sosnovsky: https://github.com/FlowCrypt/flowcrypt-browser/pull/4857#discussion_r1060167417

sosnovsky:
> @sosnovsky ok, this concludes my review. Quite a change - didn't expect that, and I suppose you didn't expect it either. While I'm not a fan of every formatting rule applied, consistency is key, so it's a step in the right direction. My comments above. Thanks!

Yeah, iOS migration wasn't so huge as we have less TypeScript code there, while here almost all code is TypeScript :)

I agree that some code parts looked weird after updated formatting, but it was just initial setup of `prettier` formatter and there were too many changes, quite difficult to check all of them. But for all future code we'll have consistent formatting and as `prettier` is the most popular JS formatter, it'll be easier for any developer to start working on the FlowCrypt codebase.

extension/types/jquery.d.ts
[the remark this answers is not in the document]
tomholub: possibly not, filed in https://github.com/FlowCrypt/flowcrypt-browser/issues/4860

tomholub (commented):
reviewing the delta

extension/chrome/elements/compose-modules/compose-render-module.ts
tomholub: This behavior is not configurable whatsoever?

extension/chrome/elements/attachment.ts
tomholub: Yup, in the second case the brackets need to be added.

Doesn't bother me that much. I just miss the clean look, since there are very few examples of the second case, I think.

extension/chrome/elements/compose-modules/compose-render-module.ts
sosnovsky: Yes, I didn't find exact option for removing these brackets

test/source/tests/browser-unit-tests/unit-ContactStore.js
tomholub: I suspect before this refactor, there were still many lines far beyond 190. The limit used to be 240 on this codebase when it only had one contributor, because that contributor had screen more wide than tall :-)

I suppose a print-width of 160 would be a reasonable compromise. A lot of developers chose one or two-word method/variable names, but for sake of clarity (thus security), some of our methods and variables are quite wordy, to make it _very_ clear what is being done and how. An 80 limit would be unusable on our codebase. I agree with Roman that 120 is still quite small target for this particular codebase.

If this is an automated process reformatting it, I propose to try 160. Then we'll evaluate it. If Roma finds it still too dense, we can try 150, or 140 as an alternative.

extension/chrome/elements/compose-modules/compose-render-module.ts
tomholub: Alright

test/source/tests/browser-unit-tests/unit-ContactStore.js
sosnovsky: 160 is okay for me, I pushed commit with increased `printWidth` value

tomholub (approved):
This looks great to me - the ridiculously long lines were addressed, without this being too aggressive a change.

sosnovsky:
@tomholub it seems `LGTM analysis` check doesn't run anymore, after LGTM was deprecated and closed by Github.

tomholub:
#4837 yes I just assigned it to you

PR #4866, opened 2023-01-05
Show a warning when downloading an executable file

### LINKED ISSUE #4862: warn user and ask to confirm before downloading dangerous file
The idea of blocking FlowCrypt users from attaching potentially malicious files based on their file extension is the initial idea - but somehow, that isn't that great as other users can compose an encrypted message using other PGP-compatible client apps and bypass restrictions.

So we came up with the idea of blocking potentially malicious files based on their file extension at runtime when the attachment was about to render or the user interact with it thru the download button (or related actions).

We can refer to this link https://support.google.com/mail/answer/6590#zippy=%2Cmessages-that-have-attachments were a blacklisted file extension is collected. It's the same list used by Gmail when users try to block certain file types.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-security/issues/224#issuecomment-1369867665

martgil:
This PR adds a warning prompt when a user tries to download an executable file.

issue #4862  // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/attachment.ts
```diff
@@ -320,10 +320,71 @@ export class AttachmentDownloadView extends View {
...
+        '.wsf',
+        '.wsh',
+        '.xll',
+      ];
+      if (blacklistedFiles.some(badFileExtension => result.filename?.endsWith(badFileExtension))) {
+        const badFileExtensionWarning = 'This executable file was not checked for viruses, and may be dangerous to download or run. Proceed anyway?'; // xss-safe-value
+        BrowserMsg.send.showWarningForAttachmentDownload(this.parentTabId, { message: badFileExtensionWarning, attachment: attachmentForSave });
+        return;
+      }
+      Browser.saveToDownloads(attachmentForSave);
```
martgil, line 387: repeated codes - could be in a function somewhere in `extension/js/common/core/attachments`?

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -238,6 +239,11 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
     BrowserMsg.addListener('show_attachment_preview', async ({ iframeUrl }: Bm.ShowAttachmentPreview) => {
       await Ui.modal.attachmentPreview(iframeUrl);
     });
+    BrowserMsg.addListener('show_warning_for_attachment_download', async ({ message, attachment }: Bm.ShowWarningForAttachmentDownload) => {
+      if (await Ui.modal.confirm(message)) {
+        Browser.saveToDownloads(attachment);
```
martgil, line 244: For some unknown reason, this doesn't work. I think I need to put the confirmation modal into an `iframe` similar to what `showAttachmentPreview` does. Then add a test later for live Gmail and web pages.

martgil: the `attachment` variable inside `show_warning_for_attachment_download` (webmail) event listener shows limited property/fields -- compared to fields to attachment.js (common/core). It is the reason why `saveToDownloads()` fails since attachment.getData() is not present.
image.

I check the typescript definition in `browser_msg.ts` for my background message event handler and it is correct.

@rrrooommmaaa @ioanmo226 do you have any suggestions? Thanks in advance.

ioanmo226: You are right. Attachment variable has limited fields.
Please check this article.
https://stackoverflow.com/questions/32814553/sending-object-with-functions-via-chrome-runtime-sendmessage#:~:text=You%20can%27t%20pass,then%20utilize%20it.

rrrooommmaaa: Functions get stripped away when sending (serializing) the object

ioanmo226: Maybe we can copy-paste this getData function logic into `Browser.saveToDownloads` function?
@rrrooommmaaa What's your suggestion?

martgil: > Maybe we can copy-paste this getData function logic into Browser.saveToDownloads function?

I can also see it to be a possibility. Will wait for Roman's suggestion. 

Thank you very much Ioan and Roman.

rrrooommmaaa: As I understand, there is no need to call `Browser.saveToDownloads` from the background, only the confirmation?
So we could add a listener to show a confirmation and return `boolean`, that is, instead of `show_warning_for_attachment_download` simply `show_confirmation` with the message to show, without passing the attachment.
then we can get the result with `sendAwait` and do the `Browser.saveToDownloads` call from the calling thread. Would that work?

ioanmo226: Sounds okay to me.

extension/js/common/core/attachment.ts
```diff
@@ -64,6 +66,70 @@ export class Attachment {
...
+      '.vhd',
+      '.vxd',
+      '.wsc',
+      '.wsf',
+      '.wsh',
+      '.xll',
+    ];
+    const badFileExtensionWarning = 'This executable file was not checked for viruses, and may be dangerous to download or run. Proceed anyway?'; // xss-safe-value
+    if (blacklistedFiles.some(badFileExtension => attachment.name.endsWith(badFileExtension))) {
+      BrowserMsg.send.showWarningForAttachmentDownload(parentTabId, { message: badFileExtensionWarning, attachment });
```
rrrooommmaaa, line 127: Like this:
`if (! await BrowserMsg.send.showConfirmation(..........) {
  return;
}

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
rrrooommmaaa: oh this wouldn't work as the response for the message is required immediately

rrrooommmaaa: 1) One option is to implement a callback similar to what we have for `passphrase_dialog`.
This is a bit cumbersome but useful, so that we can easily show confirmation popups from other places when needed.
2) Another option is that we can split Browser.saveToDownloads into 2 methods. One method extracts data as Uint8Array and attachment's name and type and forms this intermediate object which we're able to send to background. And the other method takes this intermediate object and does the remaining work.
But I'm not happy with scattering `Browser.saveToDownloads` (or whatever the new sub-method may be called) across several modules -- this potentially creates a mess.

It's better to implement confirmation separately and do as less `Browser.saveToDownloads` calls as possible (or do these calls from a less amount of modules as possible).
3) So the third option would be to route ALL file download handling via the background (with the intermediate object), and do the file extension check in the background code.
Each option has its pros and cons.

martgil: @rrrooommmaaa I think I get your point about showing the prompt and waiting for the response from a browser message. I tried to make changes according to the description like simplification and try reading the browser message response in the right place.

I'm not getting a response after trying to return the data from `Ui.modal.confirm()`.

image

```
Error downloading file - Error: BrowserMsg(show_confirmation) sendRawResponse::Error: BrowserMsg.listen error: handler "show_confirmation" not set.
```

I'm getting this error but I can't identify where could I go wrong since the `show_confirmation` called in the setup-webmail and inbox.ts -- the pop-up confirmation appears.

We can use the test here: https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGrcFnCsQsNGKRQdVdHBbQKgRhL

rrrooommmaaa: Right, the messaging system is a bit complicated
-- as you can see in comments to `private static shouldRelayMsgToOtherPage`, the message can get to `setup-webmail-content-script.ts` by reaching the background page (relayed inside `bgListen()`), so the response callback is lost. (For now) we can only receive responses when sending messages to the background handler (with `BrowserMsg.sendAwait(undefined, ...`). Can you try and install the handler to the `background_page.ts` (instead of `inbox.ts` and `setup-webmail-content-script.ts`)? I vaguely remember that it is possible to show a modal from a background page handler, but not sure.

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -169,6 +169,9 @@ export class InboxView extends View {
     BrowserMsg.addListener('show_attachment_preview', async ({ iframeUrl }: Bm.ShowAttachmentPreview) => {
       await Ui.modal.attachmentPreview(iframeUrl);
     });
+    BrowserMsg.addListener('show_confirmation', async ({ message }: Bm.ShowConfirmation): Promise<Bm.Res.ShowConfirmationResult> => {
+      return Ui.modal.confirm(message);
```
rrrooommmaaa, line 173: `return await Ui.modal.confirm(message)`,
since this function is declared as `async`

extension/js/common/core/attachment.ts
```diff
@@ -64,6 +66,71 @@ export class Attachment {
     this.contentDescription = contentDescription || undefined;
   }
 
+  public static prepareFileAttachmentDownload = async (attachment: Attachment, parentTabId: string) => {
+    const blacklistedFiles = [
```
martgil, line 70: @rrrooommmaaa sir, I think I'm doing wrong to put this reusable function here since `js/common/core...` and `test/common/core../` codes are the same but have a conflict with paths - https://flowcrypt.semaphoreci.com/jobs/50cb70e7-f815-4e0b-9291-54bd864a5544. May I have your suggestion on this?

extension/chrome/settings/inbox/inbox.ts
martgil: right - thank you.

extension/js/common/core/attachment.ts
martgil: thank you - I just need to review the figure out how I can read the return value from the browserMsg with the help of your comment above.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
martgil: @rrrooommmaaa I tried to move the handlers to background pages and they don't seem to work. Although it is capable of returning plain data but not UI modal I think?

rrrooommmaaa: > Although it is capable of returning plain data like strings/boolean but not UI modal I think?

Maybe. Please add a test message with an executable file and a test to decrypt it. They'll be useful for debugging.

extension/js/common/core/attachment.ts
```diff
@@ -177,4 +177,62 @@ export class Attachment {
       return 'plainFile';
     }
   };
+
+  public isExecutableFile = () => {
+    return [
+      '.ade',
```
rrrooommmaaa, line 183: It'd probably make sense to not repeat the leading dot (.) for every item in this list, that is, simply have a list like `'ade', 'adp', ...`

test/source/tests/decrypt.ts
```diff
@@ -1286,5 +1286,42 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+        const attachmentFrame = await inboxPage.getFrame(['attachment.htm']);
+        await attachmentFrame.waitAll('@download-attachment');
+        const rawFileAttachment = await attachmentFrame.awaitDownloadTriggeredByClicking('@download-attachment');
+        const fileAttachment = Buf.fromUtfStr(rawFileAttachment[attachmentFilename]!.toString());
+        expect(fileAttachment.toString()).to.equal('# sample bat file\n');
+        await inboxPage.close();
+      })
+    );
+
+    test(
```
rrrooommmaaa, line 1306: I added a test to test the behavior from the mock `gmail` page, please finish it by emulating the pressing of "confirmation" button.
We also need to test the behavior of the attachment preview frame.
I think, it'd be quicker to test all these 4 scenarios in a single `test`

extension/js/common/core/attachment.ts
```diff
@@ -177,4 +177,62 @@ export class Attachment {
...
+      '.vb',
+      '.vbe',
+      '.vbs',
+      '.vhd',
+      '.vxd',
+      '.wsc',
+      '.wsf',
+      '.wsh',
+      '.xll',
+    ].some(badFileExtension => this.name.endsWith(badFileExtension));
```
rrrooommmaaa, line 236: that is `exeFileExtension => this.name.endsWith('.' + exeFileExtension)`

extension/chrome/elements/attachment.ts
```diff
@@ -320,10 +327,13 @@ export class AttachmentDownloadView extends View {
     });
     Xss.sanitizeRender(this.downloadButton, this.originalButtonHTML || '');
     if (result.success) {
+      const attachmentForSave = new Attachment({ name: result.filename, type: this.attachment.type, data: result.content });
       if (!result.filename || ['msg.txt', 'null'].includes(result.filename)) {
         result.filename = this.attachment.name;
       }
-      Browser.saveToDownloads(new Attachment({ name: result.filename, type: this.attachment.type, data: result.content }));
+      if (await AttachmentWarnings.confirmSaveToDownloadsIfNeeded(attachmentForSave, this)) {
```
rrrooommmaaa, line 334: Should we add this check to more places, e.g. `ManualDecryptView.decryptAndDownload`?

extension/chrome/elements/attachment_preview.ts
```diff
@@ -79,9 +80,11 @@ View.run(
           });
           $('#attachment-preview-download')
             .css('display', 'flex')
-            .on('click', e => {
+            .on('click', async e => {
               e.stopPropagation();
-              Browser.saveToDownloads(attachmentForSave);
+              if (await AttachmentWarnings.confirmSaveToDownloadsIfNeeded(attachmentForSave)) {
```
rrrooommmaaa, line 85: As `AttachmentPreview` takes the whole page, we don't need to use messaging here, a simple `Ui.modal.confirm` call will be done here

extension/chrome/settings/index.ts
```diff
@@ -132,9 +132,6 @@ View.run(
         Catch.setHandledTimeout(clear, 10000);
         $('.webmail_notifications').one('click', clear);
       });
-      BrowserMsg.addListener('open_google_auth_dialog', async ({ acctEmail, scopes }: Bm.OpenGoogleAuthDialog) => {
```
rrrooommmaaa, line 135: This message isn't sent by any code, I removed the handler.

extension/js/common/browser/browser-msg.ts
```diff
@@ -250,6 +256,7 @@ export class BrowserMsg {
           BrowserMsg.sendAwait(undefined, 'getLocalKeyExpiration', bm, true) as Promise<Bm.Res.GetLocalKeyExpiration>,
       },
     },
+    confirmationResult: (dest: Bm.Dest, bm: Bm.ConfirmationResult) => BrowserMsg.sendCatch(dest, 'confirmation_result', bm),
```
rrrooommmaaa, line 259: I added `confirmation_show` and `confirmation_result` messages to be able to relay confirmation display to the parent to make it "full-screen" and get the result back to the caller (similar to `passphrase_dialog` and `passphrase_entry`, only I used a promise instead of polling a structure with the result). It's not easy as the caller should have specific event listeners. For `AttachmentDownloadView` it works well with both webmail and `inbox.htm`.
What is your opinion, @sosnovsky ?

extension/chrome/elements/attachment.ts
sosnovsky: Yes, I think we should add this check for all places where user can download attachment.

What about adding parameter `shouldCheckAttachmentType` to `Browser.saveToDownloads`?

extension/js/common/browser/browser-msg.ts
sosnovsky: Current implementation looks good, cleaner that polling, and it works well. Probably it's possible in the future to remake `passphrase_entry` to use promise too?

sosnovsky:
I noticed that clicking download icon downloads attachment without confirmation:

Screenshot 2023-03-06 at 17 51 51

rrrooommmaaa:
> I noticed that clicking download icon downloads attachment without confirmation:
> 
> Screenshot 2023-03-06 at 17 51 51

Is this a newly-attached file in the compose window? So far I encountered and tested attachments in incoming messages that look like this
image

extension/chrome/elements/attachment.ts
rrrooommmaaa: I was thinking about it too, but this would make `saveToDownloads` `async` / return a promise. This may be not a problem, in some cases we can run it with `Catch.try`, but I'm a bit bothered by this piece of code
https://github.com/FlowCrypt/flowcrypt-browser/blob/14ffde42a5bc8ca13981c2138bff58c7f1aa7251/extension/js/common/settings.ts#L466

This
```
await Browser.saveToDownloads(new Attachment({ name, type: 'text/plain', data: Buf.fromUtfStr(backupText) }));
await Ui.delay(1000);
```
would look ugly (as the first `await` only awaits for the possible confirmation to complete, and the second `await` await the handling of the message -- that's counter-intuitive) 
Perhaps, we could include `await Ui.delay(1000);` into `Browser.saveToDownloads`, @sosnovsky ?

sosnovsky:
> Is this a newly-attached file in the compose window? 

No, it's attachment in decrypted message:

Screenshot 2023-03-07 at 11 19 11

Clicking on icon downloads file without any warning, while clicking on attachment name opens gmail preview and then shows warning when clicking on `Download` (correct behavior):

Screenshot 2023-03-07 at 11 22 04

rrrooommmaaa:
> No, it's attachment in decrypted message:

Well, this means there are different visual components for file attachments depending on type. @martgil please inspect

martgil:
I will look into it shortly. thank you!

DenBond7:
> @DenBond7 it is the attachment that is inline - one example of it is the message from the ECP page. I have generated one for you - https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGrcrwpjkXzwpdxNwJLZLgWvBCS

@martgil Was it really addressed to me? If yes, please clarify a question.

extension/chrome/elements/attachment.ts
sosnovsky: Having `Ui.delay` in `Browser.saveToDownloads` will affect all downloads, while it's needed only in this single case, so it's not the best solution.

What about moving this delay to `resetAccount` right after `await Settings.collectInfoAndDownloadBackupFile(acctEmail)`?
There we can also add comment `// wait before backup download is finished`

extension/js/common/browser/browser-msg.ts
sosnovsky: Yeah, looks like not worth to change `confirmation` implementation for now, as current one works well

sosnovsky:
> @DenBond7 it is the attachment that is inline - one example of it is the message from the ECP page. I have generated one for you - https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGrcrwpjkXzwpdxNwJLZLgWvBCS

Yes, I just re-checked - it was rich text message from Firefox. Works well for messages from Chrome.

martgil:
> Was it really addressed to me? If yes, please clarify a question.

@DenBond7 Sorry, I tagged the wrong person - should be for @rrrooommmaaa

rrrooommmaaa:
> > please inspect
> 
> it is the attachment that is inline - one example of it is the message from the ECP page. I have generated one for you - https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGrcrwpjkXzwpdxNwJLZLgWvBCS

Please add this message to the repo and include in tests, we need to fix and test `$('#attachments .download-attachment').on('click',` in `extension/chrome/elements/pgp_block_modules/pgp-block-attachmens-module.ts` and possibly other places

extension/chrome/elements/attachment.ts
rrrooommmaaa: My question was: should we make `Browser.saveToDownloads` `async` or not?
My point is: if it's not async, then it's clear that the function only initiates some process and we should (in some cases) give it some time to finish.
If it's async, and we await it, it's not evident that the promise is resolved "half-way through", and we need to wait "a bit more" still. This design is a bit obscure.

> Having Ui.delay in Browser.saveToDownloads will affect all downloads,

As I see it, with `Catch.try(() => Browser.saveToDownloads)` the resources will be released one second later, that's all.
Most of the calls are initiated by event handlers, so `Catch.try` isn't required there.

extension/js/common/core/attachment.ts
martgil: I have simplified this one as suggested. thanks!

extension/js/common/core/attachment.ts
martgil: done. thanks!

martgil:
@sosnovsky @rrrooommmaaa ready for review. thank you.

extension/chrome/elements/attachment_preview.ts
```diff
@@ -79,9 +80,15 @@ View.run(
           });
           $('#attachment-preview-download')
             .css('display', 'flex')
-            .on('click', e => {
+            .on('click', async e => {
               e.stopPropagation();
-              Browser.saveToDownloads(attachmentForSave);
+              if (!this.attachment.isExecutableFile()) {
```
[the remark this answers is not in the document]
martgil: thank you. I missed this one when self reviewing this PR. will update it now. thanks.

extension/chrome/elements/attachment_preview.ts
```diff
@@ -79,9 +80,11 @@ View.run(
           });
           $('#attachment-preview-download')
             .css('display', 'flex')
-            .on('click', e => {
+            .on('click', async e => {
               e.stopPropagation();
-              Browser.saveToDownloads(attachmentForSave);
+              if (!this.attachment.isExecutableFile() || (await Ui.modal.confirm(Lang.attachment.executableFileWarning))) {
```
rrrooommmaaa, line 85: Wouldn't it be simpler to call the helper method `AttachmentWarnings.confirmSaveToDownloadsIfNeeded`  without the `confirmationResultTracker` parameter? It will have the same effect

rrrooommmaaa: The way we did it in another place:
```
        if (await AttachmentWarnings.confirmSaveToDownloadsIfNeeded(attachment)) {
          Browser.saveToDownloads(attachment);
        }
```

extension/js/common/browser/browser-msg.ts
```diff
@@ -81,8 +82,12 @@ export namespace Bm {
   export type Ajax = { req: JQueryAjaxSettings; stack: string };
   export type AjaxGmailAttachmentGetChunk = { acctEmail: string; msgId: string; attachmentId: string };
   export type ShowAttachmentPreview = { iframeUrl: string };
+  export type ShowConfirmation = { text: string; isHTML: boolean; footer?: string };
   export type ReRenderRecipient = { email: string };
   export type SaveFetchedPubkeys = { email: string; pubkeys: string[] };
+  export type ProcessAndStoreKeysFromEkmLocally = { acctEmail: string; decryptedPrivateKeys: string[] };
```
rrrooommmaaa, line 88: We removed these messages from `master`.
The merge is incorrect

rrrooommmaaa: ProcessAndStoreKeysFromEkmLocally and GetLocalKeyExpiration

test/source/tests/decrypt.ts
```diff
@@ -1318,5 +1318,48 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+      'settings - download batch file attachment (should show a warning message)',
+      testWithBrowser('compatibility', async (t, browser) => {
+        const threadId = '187365d19ec9a10c';
+        const threadId2 = '18736a0687a8426b';
+        const acctEmail = 'flowcrypt.compatibility@gmail.com';
+        const expectedErrMsg = 'This executable file was not checked for viruses, and may be dangerous to download or run. Proceed anyway?';
+        const inboxPage = await browser.newExtensionPage(t, `chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId}`);
+        const pgpBlockPage = await inboxPage.getFrame(['pgp_block.htm']);
+        await pgpBlockPage.waitAndClick('@download-attachment-0'); // test for inline attachment
+        await inboxPage.waitAndRespondToModal('confirm', 'confirm', expectedErrMsg);
```
rrrooommmaaa, line 1332: need to test if the file is downloaded by using `awaitDownloadTriggeredByClicking`

test/source/tests/decrypt.ts
```diff
@@ -1318,5 +1318,48 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+        await pgpBlockPage.waitAndClick('@preview-attachment');
+        const attachmentPreviewPage = await inboxPage.getFrame(['attachment_preview.htm']);
+        await attachmentPreviewPage.waitAndClick('@attachment-preview-download');
+        await attachmentPreviewPage.waitAndRespondToModal('confirm', 'confirm', expectedErrMsg);
```
rrrooommmaaa, line 1336: need to test if the file is downloaded by using `awaitDownloadTriggeredByClicking`

test/source/tests/decrypt.ts
```diff
@@ -1318,5 +1318,48 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+        await inboxPage.close();
+        const inboxPage2 = await browser.newExtensionPage(t, `chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId2}`);
+        const pgpBlockPage2 = await inboxPage2.getFrame(['pgp_block.htm']);
+        await pgpBlockPage2.waitAndClick('@download-attachment-0');
+        await inboxPage2.waitAndRespondToModal('confirm', 'confirm', expectedErrMsg);
```
rrrooommmaaa, line 1341: need to test if the file is downloaded by using `awaitDownloadTriggeredByClicking`

test/source/tests/decrypt.ts
```diff
@@ -1318,5 +1318,48 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+        await inboxPage2.close();
+        const dbPage = await browser.newExtensionPage(t, 'chrome/dev/ci_unit_test.htm');
+        const accessToken = await BrowserRecipe.getGoogleAccessToken(dbPage, acctEmail);
```
rrrooommmaaa, line 1344: we can get the token via `inboxPage2`, save some time

test/source/tests/decrypt.ts
```diff
@@ -1318,5 +1318,48 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+        await dbPage.close();
+        const extraAuthHeaders = { Authorization: `Bearer ${accessToken}` }; // eslint-disable-line @typescript-eslint/naming-convention
+        const gmailPage = await browser.newPage(t, `${t.urls?.mockGmailUrl()}/${threadId}`, undefined, extraAuthHeaders);
+        await gmailPage.waitAll('iframe');
+        const pgpBlockPage3 = await gmailPage.getFrame(['pgp_block.htm']);
+        await pgpBlockPage3.waitAndClick('@download-attachment-0'); // test for inline attachment
+        await gmailPage.waitAndRespondToModal('confirm', 'confirm', expectedErrMsg);
```
rrrooommmaaa, line 1351: need to test if the file is downloaded by using `awaitDownloadTriggeredByClicking`

test/source/tests/decrypt.ts
```diff
@@ -1318,5 +1318,48 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+        const attachmentFrame = await gmailPage.getFrame(['attachment.htm']);
+        await attachmentFrame.waitAndClick('@attachment-container');
+        const attachmentPreviewPage2 = await gmailPage.getFrame(['attachment_preview.htm']);
+        await attachmentPreviewPage2.waitAndClick('@attachment-preview-download');
+        await attachmentPreviewPage2.waitAndRespondToModal('confirm', 'confirm', expectedErrMsg);
```
rrrooommmaaa, line 1356: need to test if the file is downloaded by using `awaitDownloadTriggeredByClicking`

extension/js/common/browser/browser-msg.ts
martgil: thank you - i have removed it now from this PR.

test/source/tests/decrypt.ts
```diff
@@ -1318,5 +1318,48 @@ d6Z36//MsmczN00Wd60t9T+qyLz0T4/UG2Y9lgf367f3d+kYPE0LS7mXuFmjlPXfw0nKyVsSeFiu
...
+        await gmailPage.close();
+        const gmailPage2 = await browser.newPage(t, `${t.urls?.mockGmailUrl()}/${threadId2}`, undefined, extraAuthHeaders);
+        const pgpBlockPage4 = await gmailPage2.getFrame(['pgp_block.htm']);
+        await pgpBlockPage4.waitAndClick('@download-attachment-0');
```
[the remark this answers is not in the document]
martgil: added the missing check. thanks!

test/source/tests/decrypt.ts
martgil: done - thanks!

test/source/tests/decrypt.ts
martgil: done - thanks!

test/source/tests/decrypt.ts
martgil: yup just removed the extra line. thanks!

test/source/tests/decrypt.ts
martgil: done - thanks!

test/source/tests/decrypt.ts
martgil: done - thanks!

test/source/tests/decrypt.ts
martgil: done - thanks!

martgil:
@sosnovsky I'm sorry to bother you, but I'm having an error on the tests I added which I don't experience in when running    `npm run test_local_chrome_consumer_mock`, `npm run test_local_chrome_enterprise_mock`, or  `run test_ci_chrome_consumer` in my PC. 

I look into the CI output but the error message is truncated.

Screenshot from 2023-04-11 15-00-25

I even tried to comment out part of the tests I added to check if it was caused by a programmatical error but it doesn't seem like as I get the same error message on different tests:

image

Do you maybe have an idea to solve the issue?

sosnovsky:
> @sosnovsky I'm sorry to bother you, but I'm having an error on the tests I added which I don't experience in when running npm run test_local_chrome_consumer_mock, npm run test_local_chrome_enterprise_mock, or  run test_ci_chrome_consumer in my PC.

Currently you can check more detailed error description by going to failed workflow, for example [consumer mock - standard test group](https://flowcrypt.semaphoreci.com/jobs/e0ca4b2a-5509-4725-bf21-c8da3978987a)), opening [Artifacts](https://flowcrypt.semaphoreci.com/jobs/e0ca4b2a-5509-4725-bf21-c8da3978987a/artifacts) tab and checking [debugHtmlAttachment-0.html](https://flowcrypt.semaphoreci.com/jobs/e0ca4b2a-5509-4725-bf21-c8da3978987a/artifacts/debugHtmlAttachment-0.html) file.

Feel free to ask any questions about found issues, so they won't block your progress :)

martgil:
> Currently you can check more detailed error description by going to failed workflow, for example consumer mock - standard test group), opening Artifacts tab and checking debugHtmlAttachment-0.html file.

@sosnovsky thanks - I can see it is helpful for analyzing network activities. I'd like to share what i noticed. Below we can see that the failing tests have the same error description. 

image

The test I added `download batch file attachment (should show a warning message)` does not return this error when running the test on my PC. I suspect that it could be with the test timeouts but I wasn't sure. It just came into my mind since the test `download batch file attachment (should show a warning message)` is doing a lot of checks, and the same error occasionally happens on other tests. Do you maybe have any suggestions to identify/fix the issue?

sosnovsky:
@martgil I tried to run this on mac and it fails there with the same error as on Semaphore:
```
not ok 1 - settings - download batch file attachment (should show a warning message)
#   (attempt 1 of 1) Failed:   ProtocolError: Protocol error (Page.setDownloadBehavior): Command can only be executed on top-level targets
```

I'll try to find how this can be fixed

sosnovsky:
@martgil I fixed test issue - download modal needed to be awaited in another frame, which caused previous timeout error.

Also can you please add comments for different parts of `settings - download batch file attachment (should show a warning message)` test, so it'll be easier to understand what is tested. As currently there are several `test for inline attachment` comments, a couple places where we test clicking download on attachment preview, and it's difficult to see difference between these similar parts of test.

martgil:
> @martgil I fixed test issue - download modal needed to be awaited in another frame, which caused previous timeout error.

Thanks for your help @sosnovsky. I also add comments in some parts of the test and improve the test title.

sosnovsky (approved):
Looks good for me 👍

@rrrooommmaaa is everything ok for you here?

rrrooommmaaa:
> @rrrooommmaaa is everything ok for you here?

Yes, I have approved and merged it.
I left the issue #4862 open as there is one use case left, I'm adding a comment there

PR #4875, opened 2023-01-10
Add external-service to allowed FES service values

### LINKED ISSUE #4874: add support for updated name external-service instead of enterprise-server
https://fes.flowcrypt.com/api response has changed `service` value from `enterprise-server` to `external-service`.

We need to check if this change affected web extension and update code to allow both values - `enterprise-server` and `external-service`

sosnovsky:
This PR adds `external-service` as possible value for `service` property of FES response 

close #4874

- Tests added or updated

### To be filled by reviewers


ioanmo226 (approved):
👍

PR #4881, opened 2023-01-11
#4855 Fix reply visual glitches

### LINKED ISSUE #4855: visual glitches in secure reply
I observed both on Firefox, may also affect Chrome - not tested

## 1) stretched icon

looks ok until I highlight the recipients

image

once I focus on recipients which will expand the area, the icon wrongly stretches

image

## 2) vertical size not properly recalculated

The inside of the frame can now be larger than the iframe size, causing the inside content to be cropped.

Sometimes it's just a few pixels (immediately after rendering) but sometimes half the frame is cropped (for example if I render the reply box first, and then open dev console on the right, which reduces space available for the iframe which shrinks, but the content doesn't shrink). The content also used to shrink.

In below screenshots, I manually scrolled the iframe left and right, using the scroller inside the iframe (which is not supposed to be there!) to produce these two looks

image

image

sosnovsky:
This PR fixes visual glitches which occurred when interacting with encrypted reply box.

close #4855

- Tests added or updated

### To be filled by reviewers


ioanmo226 (approved):
Looks good 👍 
Refactoring also looks good

PR #4888, opened 2023-01-17
#4868 fix: correctly parse draft saved from iOS app and get rid of tags

### LINKED ISSUE #4868: Extension doesn't recognize draft saved from iOS app
When I created draft in iOS app and opened draft in browser extension, it doesn't show `Open Draft`(which will open secure compose view with decrypted content)

image

ioanmo226:
This PR correctly parses draft saved from iOS app and get rid of tags

close #4868 // if this PR closes an issue

- Couldn't add test (Need to add live test for parsing existing drafts. However, implementing a live test would present a stability issue as drafts cannot be exported and may be discarded prior to the start of the live test or whatsoever reason)

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -3,7 +3,7 @@
 'use strict';
 
 import { GmailRes } from '../../../js/common/api/email-provider/gmail/gmail-parser.js';
-import { InvalidRecipientError } from '../../../js/common/api/email-provider/sendable-msg.js';
+import { InvalidRecipientError, SendableMsg } from '../../../js/common/api/email-provider/sendable-msg.js';
 import { AjaxErr, ApiErr } from '../../../js/common/api/shared/api-error.js';
 import { BrowserMsg } from '../../../js/common/browser/browser-msg.js';
 import { Env } from '../../../js/common/browser/env.js';
@@ -131,6 +131,7 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
           sendable.headers['In-Reply-To'] = this.view.replyParams.inReplyTo;
         }
         this.view.S.cached('send_btn_note').text('Saving');
+        this.draftSetPrefixIntoBody(sendable);
         const mimeMsg = await sendable.toMime();
         // If a draft was loaded from the local storage, once a user is back online, the local draft will be moved to the email provider
         if (!this.view.draftId || this.isLocalDraftId(this.view.draftId)) {
@@ -211,6 +212,13 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
       return localDraft;
     }
     return undefined;
+  };
+
+  private draftSetPrefixIntoBody = (sendable: SendableMsg) => {
+    if (!this.view.threadId && !this.view.draftId) {
+      const prefix = `(saving of this draft was interrupted - to decrypt it, send it to yourself)\n\n`;
+      sendable.body['text/plain'] = `${prefix}${sendable.body['text/plain'] || ''}`;
+    }
   };
 
   private doUploadDraftWithLocalStorageFallback = async (mimeMsg: string, uploadDraft: () => Promise<string>) => {
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -2,26 +2,25 @@
 
 'use strict';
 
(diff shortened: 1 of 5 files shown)


sosnovsky:
@ioanmo226 I found a bug here:
1) compose 2 drafts
2) open first draft
3) open second draft - compose modal still shows data from the first draft

ioanmo226:
Nice catch 👍 
Let me fix it.

sosnovsky:
Found another issue:
1) try to open legacy encrypted draft with `flowcrypt:link...` prefix
2) flowcrypt extension asks for passphrase
3) don't enter pasphrase and click `Close`
4) reload page
5) try to open this draft again, enter your passphrase
6) draft text is empty

Such situation should be quite rare, but will lead to losing customer's drafts which can be a big issue.

Also sometimes original gmail box doesn't disappear and is shown along with FlowCrypt encrypted compose box, like this:

Screenshot 2023-01-18 at 16 22 43

ioanmo226:
> @ioanmo226 I found a bug here:

this one is fixed.

ioanmo226:
> found a bug here:

However, this but is also on master branch and I suggest to fix it in another PR/branch.
This fix might contain more changes than expected.

ioanmo226:
FYI in this PR, I removed `Open Draft` button logic. (This extra step sounds bad to me)
Why do we need `Open Draft` button instead of replacing secure compose window directly?

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -295,29 +295,34 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
+      // For thread reply draft, we don't need to call openComposeWin
+      // Reply draft window will be replaced by secure compose by replaceStandardReplyBox function
+      const isComposeDraft = $(contenteditable).closest('table.aoP').find('img.ajn').length < 1;
+      if (PgpArmor.isEncryptedMsg(draftHtml) && isComposeDraft) {
+        draftId = String($(contenteditable).closest('.I5').find('input[name="draft"]')?.val())?.split(':')[1] ?? '';
+      }
+      if (draftId) {
         // close original draft window
-        const closeGmailComposeWindow = (target: HTMLElement) => {
+        const closeGmailComposeWindow = (target: HTMLElement | JQuery<HTMLElement>) => {
```
sosnovsky, line 318: We can use just `JQuery<HTMLElement>` here, as plain `HTMLElement` won't be passed anymore.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -295,29 +295,34 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
           const mouseUpEvent = document.createEvent('Event');
           mouseUpEvent.initEvent('mouseup', true, true); // Gmail listens for the mouseup event, not click
-          $(target).closest('.dw').find('.Ha')[0].dispatchEvent(mouseUpEvent); // jquery's trigger('mouseup') doesn't work for some reason
+          $(target).closest('.nH.Hd').find('.Ha')[0].dispatchEvent(mouseUpEvent); // jquery's trigger('mouseup') doesn't work for some reason
```
sosnovsky, line 321: After changing `target` type to `JQuery<HTMLElement>`, this can be changed to `target.closest('.nH.Hd')...`

extension/js/content_scripts/webmail/gmail-element-replacer.ts
ioanmo226: Sounds good :+1

extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -215,22 +213,6 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
...
-      // new message compose draft with known draftId
-      prefix = `[flowcrypt:link:draft_compose:${this.view.draftId}]\n\n`;
-    } else {
-      prefix = `(saving of this draft was interrupted - to decrypt it, send it to yourself)\n\n`;
-    }
-    if (sendable.body['text/plain']) {
-      sendable.body['text/plain'] = `${prefix}${sendable.body['text/plain'] || ''}`;
-    }
-  };
-
```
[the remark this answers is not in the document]
ioanmo226: Maybe you are right.
Let me revert this part.

sosnovsky:
> However, this bug is also on master branch and I suggest to fix it in another PR/branch. This fix might contain more changes than expected. What do you think @sosnovsky

Yes, that's okay, can be fixed in another PR

sosnovsky (approved):
👍

PR #4894, opened 2023-01-18
issue 4893 bring scripts in

### LINKED ISSUE #4893: bring scripts from flowcrypt-script into flowcrypt-browser + no sql statements
/update/downloads and /update/firefox should be static json files on flowcrypt-browser repo. Todo1: update browser extension to have these statically. Todo2: use these during flowcrypt-deploy to put them in appropriate place on the web.

tomholub:
This PR brings scripts from another repo that are only used here, so that I can close the other repo. Also, where there used to be SQL statements, we instead save the results to a file which flowcrypt-deploy will pick up and expose.

close #4893

**Tests**:
- will be tested manually

### To be filled by reviewers


tomholub:
I have brought in working scripts, which I refactored while doing so. The result is only partially tested, I'll test and fix it when there is a release to do.

PR #4914, opened 2023-01-30
prevent pasting large input on secure compose

martgil:
This PR adds a check when pasting a text input with a length greater than the limit and forbids pasting if it the condition is true.

close https://github.com/FlowCrypt/flowcrypt-browser/issues/4818

For manual test, please use: [55k.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/10533031/55k.txt) - paste it on intro and input body.

Difficult to test (https://github.com/FlowCrypt/flowcrypt-browser/issues/4914#issuecomment-1410801091) 

### To be filled by reviewers


martgil:
sir @rrrooommmaaa here is the change that works. currently, it doesn't have a test. I'd like to hear your suggestion regarding adding a test. is it necessary for us to inform the user about preventing the paste action via modal? If not, can I simply invoke the paste event in the input body and input intro then checks if both inputs received the data from the paste event?

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -78,11 +78,20 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
...
+      ev.preventDefault();
+    }
+  };
+
   private handlePaste = () => {
     this.squire.addEventListener('willPaste', (e: WillPasteEvent) => {
       const div = document.createElement('div');
       div.appendChild(e.fragment);
       const html = div.innerHTML;
+      this.checkInputLengthBeforePasting(div.innerText, this.view.S.cached('input_text').get(0), e);
```
rrrooommmaaa, line 94: Instead of copying `this.view.S.cached('input_text').get(0)` it'd be more clean to extract the element from the `squire` object itself. `SquireEditor` actually has a method `getRoot()` for this. It was missing in our `squire.d.ts` type definitions, so I added it. You can now use `this.squire.getRoot()`

Another remark: is there a good reason to use `div.innerText` for the check because what actually used is sanitized `div.innerHTML`. Can it be that we inserting a big load of tags with little text? Can we perform the check on the `sanitized` value instead?

Also, when you `preventDefault`, I guess, further manipulations on `e.fragment` serve no purpose? If so, they should be skipped. It'd be more convenient to have the check method to return a `boolean` value and then we can either `ev.preventDefault()` or do the rest steps. The method may be called `willInputLimitBeExceeded`

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -78,11 +78,20 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
     return this.view.sendBtnModule.popover.choices.richtext;
   };
 
+  public checkInputLengthBeforePasting = (textToPaste: string, targetInputField: HTMLElement, ev: ClipboardEvent) => {
+    const currentLength = targetInputField.innerText.length;
+    const limit = 50000;
+    if (textToPaste.length + currentLength > limit) {
```
rrrooommmaaa, line 84: Do we want to gracefully handle the situation when the paste (partially) replaces the existing content? This will happen if there is an active selection at the moment of pasting.

rrrooommmaaa:
> sir @rrrooommmaaa here is the change that works. currently, it doesn't have a test. I'd like to hear your suggestion regarding adding a test.

As I understand, clipboard-related tests are not easy to develop. I guess @tomholub will suggest to skip the test for this issue.

> is it necessary for us to inform the user about preventing the paste action via modal?

That's a question for @tomholub

tomholub:
I think we should indeed do a modal, and as Roman says, ok to skip test if difficult to do.

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: Thank you for all the remarks. I apply some changes according to your helpful insight.

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: Not sure, sir, since we only want to test the input limit. But if the process in this pr affects the performance significantly I think we could consider that.

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -78,13 +78,24 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
...
     this.squire.addEventListener('willPaste', (e: WillPasteEvent) => {
       const div = document.createElement('div');
       div.appendChild(e.fragment);
       const html = div.innerHTML;
       const sanitized = this.isRichText() ? Xss.htmlSanitizeKeepBasicTags(html, 'IMG-KEEP') : Xss.htmlSanitizeAndStripAllTags(html, '<br>', false);
       Xss.setElementContentDANGEROUSLY(div, sanitized); // xss-sanitized
+      if (this.willInputLimitBeExceeded(sanitized.trim(), this.squire.getRoot())) {
+        e.preventDefault();
+        return;
+      }
```
rrrooommmaaa, line 98: `Xss.setElementContentDANGEROUSLY(div, sanitized); // xss-sanitized`
can also be moved after `if {...}`, right?

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -78,13 +78,24 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
...
+  };
+
   private handlePaste = () => {
     this.squire.addEventListener('willPaste', (e: WillPasteEvent) => {
       const div = document.createElement('div');
       div.appendChild(e.fragment);
       const html = div.innerHTML;
       const sanitized = this.isRichText() ? Xss.htmlSanitizeKeepBasicTags(html, 'IMG-KEEP') : Xss.htmlSanitizeAndStripAllTags(html, '<br>', false);
       Xss.setElementContentDANGEROUSLY(div, sanitized); // xss-sanitized
+      if (this.willInputLimitBeExceeded(sanitized.trim(), this.squire.getRoot())) {
```
rrrooommmaaa, line 95: Do we need `trim()`? Why? `sanitized` is HTML. Did you notice any extra space appear during testing?

martgil: Yes, the secure compose has an extra space.

(before)
Screenshot 2023-02-16 at 5 43 23 PM

(after)
image

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: right - you are correct. So that `Xss.set...` will execute only when needed.

martgil:
@rrrooommmaaa ready for review. thank you!

extension/chrome/elements/compose-modules/compose-input-module.ts
rrrooommmaaa: We want to check the resulting element size, e.g. replacing a 40k text with another 40k text should trigger no error, right?
82782ace0c4e135b72b09ebda98e681b8f447e64 -- is it looking good?

rrrooommmaaa:
> I think we should indeed do a modal, and as Roman says, ok to skip test if difficult to do.

@martgil Please add a modal saying something like `The paste operation can't be completed because the resulting text size would exceed the allowed limit of 50K`.

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: the changes from https://github.com/FlowCrypt/flowcrypt-browser/commit/82782ace0c4e135b72b09ebda98e681b8f447e64 look good to me. I see that I missed that scenario for subtracting content. thank you.

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -78,12 +79,25 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
...
+
   private handlePaste = () => {
-    this.squire.addEventListener('willPaste', (e: WillPasteEvent) => {
+    this.squire.addEventListener('willPaste', async (e: WillPasteEvent) => {
       const div = document.createElement('div');
       div.appendChild(e.fragment);
       const html = div.innerHTML;
       const sanitized = this.isRichText() ? Xss.htmlSanitizeKeepBasicTags(html, 'IMG-KEEP') : Xss.htmlSanitizeAndStripAllTags(html, '<br>', false);
+      if (this.willInputLimitBeExceeded(sanitized, this.squire.getRoot(), () => this.squire.getSelectedText().length)) {
+        await Ui.modal.warning("The paste operation can't be completed because the resulting text size would exceed the allowed limit of 50K");
```
rrrooommmaaa, line 97: It's not good to make copies of the same message.
Please put the text into `Lang.compose`

rrrooommmaaa (changes requested):

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -78,12 +79,25 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
...
+        e.preventDefault();
```
rrrooommmaaa, line 98: preventDefault() should be called before `await`

extension/chrome/elements/compose.ts
```diff
@@ -222,6 +222,25 @@ export class ComposeView extends View {
...
+          const isInputLimitExceeded = this.inputModule.willInputLimitBeExceeded(clipboardEvent.clipboardData.getData('text/plain'), el, () => {
+            const selection = window.getSelection();
+            if (selection && selection.anchorNode === selection.focusNode && selection.anchorNode?.parentElement === el) {
+              return Math.abs(selection.anchorOffset - selection.focusOffset);
+            }
+            return 0;
+          });
+          if (isInputLimitExceeded) {
+            await Ui.modal.warning("The paste operation can't be completed because the resulting text size would exceed the allowed limit of 50K");
+            ev.preventDefault();
```
rrrooommmaaa, line 239: preventDefault() should be called before await

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: thank you - I'll keep this in mind.

extension/chrome/elements/compose.ts
martgil: will do - thank you!

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: got it - thanks.

rrrooommmaaa (approved):

PR #4927, opened 2023-02-01
#4917 fix: initialize external service at init

### LINKED ISSUE #4917: Set external service on AccountServer init
> Maybe it's possible to check `isCustomerUrlFesUsed(this.acctEmail)` on `AccountServer` init and set service url there, so we don't need to run `getExternalService` before each request?

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-browser/pull/4907#discussion_r1090509963_

ioanmo226:
This PR initialized external service at init in AccountServer

close #4917 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/js/common/api/account-server.ts
+++ b/extension/js/common/api/account-server.ts
@@ -27,6 +27,12 @@ export class AccountServer extends Api {
     this.externalService = new ExternalService(this.acctEmail);
   }
 
+  public static init = async (acctEmail: string) => {
+    const acctServer = new AccountServer(acctEmail);
+    await acctServer.initialize();
+    return acctServer;
+  };
+
   public initialize = async () => {
     if (!(await isCustomerUrlFesUsed(this.acctEmail))) {
       this.externalService.url = SHARED_TENANT_API_HOST;
--- a/extension/js/common/api/account-server.ts
+++ b/extension/js/common/api/account-server.ts
@@ -20,19 +20,21 @@ export type UploadedMessageData = {
  *   whether FES is deployed on the customer domain or not.
  */
 export class AccountServer extends Api {
-  private readonly potentialCustomerUrlFes: ExternalService;
-  private readonly sharedTenantFes: ExternalService;
+  private readonly externalService: ExternalService;
 
   public constructor(private acctEmail: string) {
     super();
-    this.potentialCustomerUrlFes = new ExternalService(this.acctEmail);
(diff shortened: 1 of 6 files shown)


extension/js/common/api/account-server.ts
```diff
@@ -20,19 +20,21 @@ export type UploadedMessageData = {
...
-    this.potentialCustomerUrlFes = new ExternalService(this.acctEmail);
-    this.sharedTenantFes = new ExternalService(this.acctEmail);
-    this.sharedTenantFes.url = SHARED_TENANT_API_HOST;
+    this.externalService = new ExternalService(this.acctEmail);
   }
 
+  public initialize = async () => {
+    if (!(await isCustomerUrlFesUsed(this.acctEmail))) {
+      this.externalService.url = SHARED_TENANT_API_HOST;
+    }
```
sosnovsky, line 33: Can we move this check to `constructor`? So we won't need to run it separately on each `AccountServer` initialisation. 

Then this code
```ts
const acctServer = new AccountServer(acctEmail);
await acctServer.initialize();
await acctServer.fetchAndSaveClientConfiguration();
```

Can be just 
```ts
await new AccountServer(acctEmail).fetchAndSaveClientConfiguration();
```
as it was before

ioanmo226: But as you know `initialize` is async functino and can't be placed in constructor.

sosnovsky: What about using static `init` function, like this
```ts
class AccountServer  {
  public static init = async (acctEmail: string) => {
    const acctServer = new AccountServer(acctEmail);
    await acctServer.initialize();
    return acctServer
  };
```

And then used this way
```ts
await AccountServer.init(acctEmail).fetchAndSaveClientConfiguration();
```

In this case we won't forget to run `initialize` for each `AccountServer` init

ioanmo226: It sounds good when AccountServier init was called from async funtion.
But in several places, we init AccountServer in constructor (like ComposeView consturctor and accountserver was used from other view modules).
For those places, we still need to call initalize in render function or so

sosnovsky: Got it 👍 Looks good now

sosnovsky (approved):

PR #4928, opened 2023-02-01
#4734 Use separate mock api for each test

### LINKED ISSUE #4734: one mock instance per test

sosnovsky:
This PR implements separate mock api for each test:
- nodejs finds available port for new mock api
- we create copy of extension build which uses new port
- then this port is passed to tests using `AvaContext`

close #4734

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -133,14 +132,12 @@ export class PgpBlockViewRenderModule {
...
+    w?.document.write(html);
     // Give some time for above dom to load in print dialog
     // https://stackoverflow.com/questions/31725373/google-chrome-not-showing-image-in-print-preview
     await Ui.time.sleep(250);
-    w!.window.print();
-    w!.document.close();
-    /* eslint-enable @typescript-eslint/no-non-null-assertion */
+    w?.window.print();
+    w?.document.close();
   };
```
tomholub, line 141: I generally recommend going with `!` instead of `?`, against the practice of many developers / teams.

Then things break, I like them to break explicitly. We have a render mechanism for errors that will explain the problem to the user so they can take a screenshot of the problem, and we have an exception logging endpoint to receive information about and fix most common exceptions.

That doesn't necessarily mean that this exact code needs to change back, I think it's fine as you've done it. I'm writing this comment to explain why you see a lot of `!` rather than `?` everywhere, and the rationale behind it.

On iOS, that would have crashed the app. Here, that would throw an exception, which we display, sometimes auto-report, and explain - so the stakes are different.

test/source/browser/browser-pool.ts
```diff
@@ -28,14 +26,13 @@ export class BrowserPool {
 
   public newBrowserHandle = async (t: AvaContext, closeInitialPage = true) => {
     await this.semaphore.acquire();
-    // ext frames in gmail: https://github.com/GoogleChrome/puppeteer/issues/2506 https://github.com/GoogleChrome/puppeteer/issues/2548
+    const extensionDir = t.extensionDir ?? this.extensionBuildDir;
     const args = [
       '--no-sandbox', // make it work in travis-ci
       '--disable-setuid-sandbox',
       '--kiosk-printing',
-      '--disable-features=site-per-process',
```
tomholub, line 36: If they actually fixed this `site-per-process` issue, that would mean we should be able to test the iframes directly where they are, for example, in Live tests we should be able to test them directly from the Gmail page, instead of copying the URL and opening a separate tab, and testing there. That would speed up the tests. Let's file that for another issue (unless you have already done exactly that).

test/source/browser/browser-pool.ts
```diff
@@ -148,7 +110,7 @@ export class BrowserPool {
       try {
         const browser = await withTimeouts(this.newBrowserHandle(t));
         try {
-          await withTimeouts(this.cbWithTimeout(async () => await cb(t, browser), consts.TIMEOUT_EACH_RETRY));
+          await withTimeouts(this.cbWithTimeout(async () => cb(t, browser), consts.TIMEOUT_EACH_RETRY));
```
tomholub, line 113: I suppose the `cb` returns a promise? If yes, it may seem like there is no point in awaiting the promise right here, and can just return the promise, and let the higher level function await it. But that's not entirely the same. By awaiting it explicitly, and always, we ensure that the stack trace will be complete. By passing the `Promise` around without awaiting it, we may lose out on some parts of the stack trace. That was the case on Node 12 and browsers of a few years ago, anyway. The situation I think somewhat improved, but I also think there are probably still gaps, so I generally recommend to never return a `Promise` without first awaiting it.

The function signature of course remains the same either way, it's just whether or not to include the `await` - which I strongly recommend to include.

test/source/test.ts
```diff
@@ -82,52 +75,67 @@ const testWithBrowser = (
...
         }
         t.log(`run time: ${Math.ceil((Date.now() - start) / 1000)}s`);
       },
       t,
       consts,
       flag
     );
+    if (t.closeMockApi) {
+      await t.closeMockApi();
+    }
```
tomholub, line 88: There should probably be a `try/finally` statement here, wrapping `withNewBrowserTimeoutAndRetry`. Then the `closeMockApi` would happen in `finally`. Then you can ensure it will run?

test/source/tests/tooling/index.ts
```diff
@@ -1,14 +1,18 @@
...
 
-export type AvaContext = ava.ExecutionContext<unknown> & {
+export type AvaContext = ExecutionContext<unknown> & {
   retry?: true;
   attemptNumber?: number;
   totalAttempts?: number;
   attemptText?: string;
+  extensionDir?: string;
+  urls?: TestUrls;
+  closeMockApi?: () => Promise<void>;
```
tomholub, line 15: Maybe `closeMockApi` wouldn't have to be on `AvaContext` if it was in try/finally above

tomholub (approved):
:+1: that's a lot of tests!

Looking good.

test/source/browser/browser-pool.ts
sosnovsky: Yes, it seems issue was fixed as it's closed now in puppeteer repo - https://github.com/puppeteer/puppeteer/issues/2548

I haven't changed test logic yet, just checked that all tests pass without `--disable-features=site-per-process` flag

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
sosnovsky: Thanks for explanation 👍 
You're right - after iOS code I've used to avoid `!`

test/source/browser/browser-pool.ts
sosnovsky: `cb` here returns `void`:
```ts
public withNewBrowserTimeoutAndRetry = async (cb: (t: AvaContext, browser: BrowserHandle) => void, ...
```

Also vscode notified that `await` isn't needed:
Screenshot 2023-02-01 at 15 09 06

test/source/test.ts
sosnovsky: Yeah, you're right, I added `try/finally` here

test/source/browser/browser-pool.ts
tomholub: Yup, if it returns void, then my comment is also void.

tomholub (approved):

test/source/mock/fes/shared-tenant-fes-endpoints.ts
```diff
@@ -125,3 +127,8 @@ const authenticate = (req: IncomingMessage, type: 'oidc' | 'fes'): string => {
   }
   return MockJwt.parseEmail(jwt);
 };
+
+const messageIdRegex = (req: IncomingMessage) => {
+  const port = parsePort(req);
+  return new RegExp(`{"emailGatewayMessageId":"<(.+)@standardsubdomainfes.localhost:${port}>"}`);
+};
```
ioanmo226, line 134: Maybe you can re-use messageIdRegex from customer-url-fes-endpoint.

ioanmo226 (approved):
Looks great 👍

test/source/mock/fes/shared-tenant-fes-endpoints.ts
sosnovsky: Good point, implemented it in another PR - https://github.com/FlowCrypt/flowcrypt-browser/pull/4932

sosnovsky:
> By the way, I can see that test sometimes fails with below error. Need to be investigated and fixed.

@ioanmo226 I think it happened once, test passes well now, but let's monitor if it'll fail in the future.
Sometimes there was an issue with test execution time limit, I increased it a bit, seems like it's enough for now.

package.json
```diff
@@ -59,29 +59,29 @@
...
-    "test_ci_chrome_enterprise": "npx ava --timeout=20m --verbose --concurrency=1 build/test/test/source/test.js -- ENTERPRISE-MOCK STANDARD-GROUP",
-    "test_ci_chrome_consumer_flaky": "npx ava --timeout=20m --verbose --concurrency=1 build/test/test/source/test.js -- CONSUMER-MOCK FLAKY-GROUP",
-    "dev_start_gmail_mock_api": "./build.sh && cd ./conf && node ../build/tooling/tsc-compiler --project tsconfig.test.json && cd .. && node ./build/test/test/source/mock.js",
+    "test_buf": "npx ava --timeout=20m --verbose --concurrency=10 build/test/test/source/buf.js",
+    "test_ci_chrome_consumer_live_gmail": "npx ava --timeout=20m --verbose --concurrency=10 build/test/test/source/test.js -- CONSUMER-LIVE-GMAIL STANDARD-GROUP",
+    "test_ci_chrome_consumer": "npx ava --timeout=20m --verbose --concurrency=10 build/test/test/source/test.js -- CONSUMER-MOCK STANDARD-GROUP",
+    "test_ci_chrome_enterprise": "npx ava --timeout=20m --verbose --concurrency=10 build/test/test/source/test.js -- ENTERPRISE-MOCK STANDARD-GROUP",
+    "test_ci_chrome_consumer_flaky": "npx ava --timeout=20m --verbose --concurrency=10 build/test/test/source/test.js -- CONSUMER-MOCK FLAKY-GROUP",
```
rrrooommmaaa, line 83: What is the purpose of the "flaky" group? There are resource-demanding or tests with an unstable performance. This group is meant for tests to be run consequently, one at a time

scripts/config-mock-build.sh
```diff
@@ -0,0 +1,6 @@
+# copy build to new folder and replace all usages of :8001 with appropriate mock port
+BUILD_PATH=./build/test/mock-builds/port-$2-$(openssl rand -hex 12)
+mkdir -p $BUILD_PATH
+cp -r ./$1/* $BUILD_PATH
+grep ":8001" $BUILD_PATH -lr | xargs sed -i.bak -e "s/\:8001/\:$2/g"
```
rrrooommmaaa, line 5: Can it be solved with a `[BUILD_REPLACEABLE_....]` placeholder?

package.json
tomholub: Some tests are inexplicably resource-demanding. When run in parallel with other tests, what can happen is that either that test will time out, or other tests will fail while the resource-intensive test hogs up resources.

Other reason that a test may end up in "flaky" section is if it needs to be executed serially for some other reason.

Generally speaking, these tests tend to cause reliability trouble  when executed in parallel, for one reason or another.

Of course, they could be debugged and fixed one by one and moved, possibly, if we had the time or willingness.

scripts/config-mock-build.sh
tomholub: I suppose it could.. we would still need to do two rounds of replacing? The step that normally replaces `[BUILD_REPLACEABLE_....]` and currently puts the `:8001` would still need to insert some more maybe `[TEST_REPLACEABLE_MOCK_PORT]` and then that would be replaced here.

It's true that as things are now, there is no meaning to the `:8001` in particular. So that should not be put it, to not be confusing. So `[TEST_REPLACEABLE_MOCK_PORT]` or similar would be a better choice here.

package.json
rrrooommmaaa: Does this mean we should set `--concurrency=1` instead of `--concurrency=10`?

sosnovsky: Yes, we can change it back to `--concurrency=1` as browser pool size is automatically set to 1 for `FLAKY-GROUP` - https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/util/index.ts#L31

tomholub: Shouldn't browser pool and concurrency always be exactly equal? Or is there a reason to have these two numbers diverge? I'd imagine, we could remove special code to choose that `FLAKY-GROUP` has 1 in the pool, and instead follow whatever the provided `--concurrency` ?

sosnovsky: Probably it can be simplified, as currently `--pool-size` parameter is used for determining `oneIfNotPooled`, which then also used for setting `ATTEMPTS` constant - https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/test.ts#L40

sosnovsky: https://github.com/FlowCrypt/flowcrypt-browser/issues/4955

PR #4937, opened 2023-02-06
#4922 Fix: attachment missing after switch to forward

### LINKED ISSUE #4922: switch to forward is missing files
1) click encrypted reply to a particular email
2) switch reply to forward using the dropdown on top right
3) I'm forwarding now but it's missing the attachments from previous email

ioanmo226:
This PR fixed attachment missing issue after switch to forward 

close #4922 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-render-module.ts
```diff
@@ -220,7 +220,7 @@ export class ComposeRenderModule extends ViewModule<ComposeView> {
     }
     this.view.recipientsModule.clearRecipients();
     if (option === 'a_forward') {
-      await this.view.quoteModule.addTripleDotQuoteExpandFooterAndQuoteBtn(this.view.replyMsgId, 'forward');
+      await this.view.quoteModule.addTripleDotQuoteExpandFooterAndQuoteBtn(this.view.replyMsgId, 'forward', true);
```
sosnovsky, line 223: I think we should force reload compose box when switching to reply too, as currently when you click `Secure Forward` and then switch to `Secure Reply` message includes attachment, but we doesn't need it for reply.

You can also set `forceReload` to true only when message includes attachment, because we need reload there only for attachments update.

ioanmo226: OK

ioanmo226: Please check new change

sosnovsky: Works well now 👍

test/source/tests/compose.ts
```diff
@@ -1684,6 +1684,22 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+        const appendUrl = 'threadId=18625ff9e8642033&skipClickPrompt=___cu_false___&replyMsgId=18625ff9e8642033';
+        const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility', {
+          appendUrl,
+          hasReplyPrompt: true,
+        });
+        await composePage.waitAndClick('@encrypted-reply');
+        await composePage.waitAndClick('@action-show-reply-options-popover');
+        await composePage.waitAndClick('@action-toggle-a_forward');
+        // Check attachment is still present after switch to forward
+        await composePage.waitForContent('.qq-upload-file', 'test.txt');
```
sosnovsky, line 1699: Let's also add testing switch back to reply here, to test that attachments are removed.

sosnovsky (approved):
👍

PR #4944, opened 2023-02-08
#4282 fix: duplicated contact search result

### LINKED ISSUE #4282: Duplicate contact search result
The test `compose - can load contact based on name` passes, but `human@flowcrypt.com` is displayed twice by the end of the test
image
I would expect it to appear only once

ioanmo226:
This PR fixed duplicated contact search result

close #4282 // if this PR closes an issue

- Test updated

### To be filled by reviewers

### DIFF
--- a/test/source/browser/controllable.ts
+++ b/test/source/browser/controllable.ts
@@ -523,6 +523,11 @@ abstract class ControllableBase {
     return files;
   };
 
+  public checkElementCount = async (selector: string, expectedCount: number) => {
+    const actualCount = await this.elementCount(selector);
+    expect(actualCount).to.equal(expectedCount);
+  };
+
   protected log = (msg: string) => {
     if (this.debugNamespace) {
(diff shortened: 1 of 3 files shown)


sosnovsky (commented):
Looks good, but let's also update ui test `compose - can load contact based on name` to ensure that search results list contains only 1 contact

test/source/browser/controllable.ts
```diff
@@ -523,6 +523,11 @@ abstract class ControllableBase {
     return files;
   };
 
+  public checkElementCount = async (selector: string, expectedCount: number) => {
```
sosnovsky, line 526: There is already [ensureElementCount](https://github.com/FlowCrypt/flowcrypt-browser/blob/feature/issue-4282/test/source/browser/controllable.ts#L464) method which performs the same check. Let's merge them into one to remove duplication

ioanmo226: You are right 👍 
Just merged it

sosnovsky (approved):
👍

PR #4946, opened 2023-02-09
Employ user verification mechanisms from OpenPGP v5

### LINKED ISSUE #4588: Revoked Uids are considered as valid
The attached key has 2 User IDs, one them is revoked (according to Kleopatra) as discovered in https://github.com/FlowCrypt/flowcrypt-enterprise-on-prem/issues/3291
[first-mock-test-0x358D46419AAB290AA05005C04A02B5C677B2716E0101-private.zip](https://github.com/FlowCrypt/flowcrypt-browser/files/9282727/first-mock-test-0x358D46419AAB290AA05005C04A02B5C677B2716E0101-private.zip)
However, in Flowcrypt Compatibilty Page this revoked UID `<user@example.com>`  is shown as perfectly valid.
Need to check the OpenPGP.js/key-handling code

rrrooommmaaa:
This PR employs user verification mechanisms from OpenPGP v5

On diagnostics page an unverified user id looks like this
image

As for Key parsing, all unverified user ids are omitted from the `Key.identities` property.

close #4588

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -87,8 +87,10 @@ export class Str {
     if (email.indexOf(' ') !== -1) {
       return false;
     }
-    email = email.replace(/\:8001$/, ''); // for MOCK tests until https://github.com/FlowCrypt/flowcrypt-browser/issues/4631
-    return /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/i.test(
+    // todo: `email-addresses` parser used by OpenPGP.js consider top-level domains to be valid for emails e.g. address@domain
+    // should we allow it too (or use `email-addresses` package when manipulating user ids in keys?
+    // For now, I'm explicitly `localhost` domain, which is perfectly legal for testing
+    return /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|localhost|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/i.test(
       email
     );
   };
--- a/extension/js/common/core/common.ts
+++ b/extension/js/common/core/common.ts
@@ -88,9 +88,7 @@ export class Str {
       return false;
     }
     email = email.replace(/\:8001$/, ''); // for MOCK tests, todo: remove from production
-    // todo: `email-addresses` parser used by OpenPGP.js consider top-level domains to be valid for emails e.g. address@domain
-    // should we allow it too (or use `email-addresses` package when manipulating user ids in keys?
-    // For now, I'm explicitly `localhost` domain, which is perfectly legal for testing
+    // `localhost` is a valid top-level domain for an email address, otherwise we require a second-level domain to be present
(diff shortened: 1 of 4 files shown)


extension/js/common/core/common.ts
```diff
@@ -87,8 +87,11 @@ export class Str {
     if (email.indexOf(' ') !== -1) {
       return false;
     }
-    email = email.replace(/\:8001$/, ''); // for MOCK tests until https://github.com/FlowCrypt/flowcrypt-browser/issues/4631
-    return /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/i.test(
+    email = email.replace(/\:8001$/, ''); // for MOCK tests, todo: remove from production
+    // todo: `email-addresses` parser used by OpenPGP.js consider top-level domains to be valid for emails e.g. address@domain
+    // should we allow it too (or use `email-addresses` package when manipulating user ids in keys?
+    // For now, I'm explicitly `localhost` domain, which is perfectly legal for testing
+    return /^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|localhost|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$/i.test(
```
rrrooommmaaa, line 94: I'm re-using `Str.parseEmail` method to re-extract email from userids verified by OpenPGP.js, as those verification methods only return userid string (and all valid userids are supposed to have email address embedded). 
OpenPGP.js is using (`email-addresses` NPM package)[`https://www.npmjs.com/package/email-addresses`] to extract email address from the userid.
So several questions/issues arise:
- Should we use `email-addresses` package or our own version for the re-extraction?
- If using own version, should we validate the email address? `email-addresses` validation is more lenient, as it allows a top level domain only, e.g. `address@domain`. As I understand, this type of validation isn't convenient for `compose` boxes, should we have a separate validation option when extracting emails from userids, like `VALIDATE-LENIENT"?

I had to explicitly add `localhost` as a valid domain into the regex

rrrooommmaaa: @tomholub ?

tomholub: I think it's good as you've done it. It would indeed be odd to allow `hello@yup` as a valid email when composing. Similarly, having two ways to verify based on context is unnecessary complication. I think it doesn't necessarily need 100% alignment with OpenPGP.js. If there is an edge case when one check passes and the other doesn't, the result is still a failed action. Actually there's currently three places that validate email: this, OpenPGP.js, Gmail API. So it will never be unified. As long as nobody reports a problem, I'd leave it.

tomholub: I'd remove the todo comment anyway.

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -305,9 +305,10 @@ export class OpenPGPKey {
     const key = await OpenPGPKey.extractExternalLibraryObjFromKey(pubkey);
     const result = new Map<string, string>();
     result.set(`Is Private?`, KeyUtil.formatResult(key.isPrivate()));
-    for (let i = 0; i < key.users.length; i++) {
+    const users = await OpenPGPKey.verifyAllUsers(key);
+    for (let i = 0; i < users.length; i++) {
       // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-      result.set(`User id ${i}`, key.users[i].userID!.userID);
+      result.set(`User id ${i}`, users[i].valid ? users[i].userID : '* REVOKED, INVALID OR MISSING SIGNATURE *');
```
tomholub, line 311: Could say `* REVOKED, INVALID OR MISSING SIGNATURE * ${users[i].userID}` for easier debugging

extension/js/common/core/common.ts
tomholub: As per above comment I'd remove the `todo` since I think it's ok as is

tomholub (approved):
👍

extension/js/common/core/common.ts
rrrooommmaaa: Done

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: Done

PR #4947, opened 2023-02-09
#4844 feat: renew id token when expires

### LINKED ISSUE #4844: Offer users to reconnect when id_token expires
The following error occurs when a JWT token has expired when requesting for `api/account/get`. 

I think one way to address this is by going to FlowCrypt settings -> resetting the account -> sign back in to refresh the id_token needed by the backend (not user friendly). 

The customer wonders if there's something we can do to simplify it. 

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGrbbxhKWXXkjFGnRLcMxlLkRQc

ioanmo226:
This PR renews id token when expires

close #4844 // if this PR closes an issue

- Difficult to test (Hard/impossible to reproduce in semaphoreci)

### To be filled by reviewers


extension/js/common/api/shared/api.ts
```diff
@@ -205,8 +206,23 @@ export class Api {
...
+    } catch (firstAttemptErr) {
+      const idToken = req.headers?.Authorization?.split(' ')[1];
+      if (ApiErr.isAuthErr(firstAttemptErr) && idToken) {
+        // force refresh token
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+        const { email } = GoogleAuth.parseIdToken(idToken);
+        if (email) {
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+          req.headers!.Authorization = await GoogleAuth.googleApiAuthHeader(email, true);
+          return await Api.ajax(req, Catch.stackTrace()) as RT;
```
ioanmo226, line 221: As `Api` class uses `GoogleAuth` and `GoogleAuth` uses `Api` class, circular dependency error occurs.
So I'm thinking to move common method(parseIdToken, googleApiAuthHeader, googleAuthRefreshToken, googleAuthSaveTokens) to new common class(such as GoogleAuthExtension)

@sosnovsky Do you have better suggestion?

sosnovsky: I think proposed solution should be good, we can call this new class `GoogleAuthHelper`

ioanmo226: Thanks

extension/js/common/api/email-provider/gmail/google-auth-helper.ts
```diff
@@ -0,0 +1,223 @@
...
+        crossDomain: true,
+        async: true,
+      },
+      Catch.stackTrace()
+    )) as unknown as GoogleAuthTokensResponse;
+  };
+
+  /// Duplicated code begin
+  // This logic was duplicated (It's exactly same as Api.ajax) but added here because we can't use Api.ajax here as it will cause circular dependency warning
+  private static ajax = async (req: JQueryAjaxSettings, stack: string): Promise<unknown | JQuery.jqXHR<unknown>> => {
```
ioanmo226, line 174: Let me know your thought about this block if you agree on duplicating logi or not.

ioanmo226: Sounds good.

PR #4949, opened 2023-02-10
#4844 feat: renew id token when expires

### LINKED ISSUE #4844: Offer users to reconnect when id_token expires

ioanmo226:
This PR <what does it do and why?>

close #4844 // if this PR closes an issue

- Difficult to test (Hard/impossible to reproduce in semaphoreci)

### To be filled by reviewers

### DIFF
--- a/extension/js/common/api/account-servers/external-service.ts
+++ b/extension/js/common/api/account-servers/external-service.ts
@@ -12,6 +12,7 @@ import { ParsedRecipients } from '../email-provider/email-provider-api.js';
 import { Buf } from '../../core/buf.js';
 import { ClientConfigurationError, ClientConfigurationJson } from '../../client-configuration.js';
 import { InMemoryStore } from '../../platform/store/in-memory-store.js';
+import { GoogleAuth } from '../email-provider/gmail/google-auth.js';
 
 // todo - decide which tags to use
 type EventTag = 'compose' | 'decrypt' | 'setup' | 'settings' | 'import-pub' | 'import-prv';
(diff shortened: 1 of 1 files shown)


extension/js/common/api/account-servers/external-service.ts
```diff
@@ -171,18 +172,45 @@ export class ExternalService extends Api {
...
+          ...(await this.authHdr()),
+        },
+        'json',
+        method
+      );
+    } catch (firstAttemptErr) {
+      const idToken = await InMemoryStore.get(this.acctEmail, InMemoryStoreKeys.ID_TOKEN);
+      if (ApiErr.isAuthErr(firstAttemptErr) && idToken) {
+        // force refresh token
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
sosnovsky, line 193: It seems `eslint-disable` not needed here

extension/js/common/api/account-servers/external-service.ts
```diff
@@ -171,18 +172,45 @@ export class ExternalService extends Api {
...
+        const { email } = GoogleAuth.parseIdToken(idToken);
+        if (email) {
+          // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
sosnovsky, line 196: And here too

sosnovsky (commented):
This solution is much simpler than the previous one, looks great 👍
Just need to remove a couple unnecessary `eslint-disable` comments

extension/js/common/api/account-servers/external-service.ts
ioanmo226: Ok

sosnovsky (approved):
Looks good 👍

I'll create another issue for testing id tokens refresh, maybe it's possible to reproduce with our mock apis.

Issue #4494, opened 2022-06-18
mock tests should have inputs configured in the test + one mock API per test


tomholub:
Currently, the test codebase has a single API mock for all tests, and some parts of the mock setup are meant for some tests while other parts for others, in non-obvious ways. The better example is when there is a comment like this:

```ts
      const sentMsgs = (await GoogleData.withInitializedData(acct)).searchMessagesBySubject(subject);
      expect(sentMsgs.length).to.equal(2);
      // this test is using PwdEncryptedMessageWithFesIdTokenTestStrategy to check sent result based on subject "PWD encrypted message with FES - ID TOKEN"
      // also see '/api/v1/message' in fes-endpoints.ts mock
```

The worse example is silently implied setup, like certain tests expect client configuration to be set up based on their domain, the example is here:

```ts
  ava.default('user2@standardsubdomainfes.test - PWD encrypted message with FES - Reply rendering', testWithBrowser(undefined, async (t, browser) => {
      const acct = 'user2@standardsubdomainfes.test'; // added port to trick extension into calling the mock
      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acct);
```

The mock API endpoints test for certain conditions etc - none of which is obvious from the test.

```ts
  '/api/v1/message': async ({ body }, req) => {
      // ...
      if (body.includes('"from":"user2@standardsubdomainfes.test"')) {
        // ...
      }
```

There's also some that may not apply to this particular test, and may be causing side effects:

```ts
    if (req.headers.host === standardFesUrl && req.url === `/api/v1/client-configuration?domain=standardsubdomainfes.test`) {
      return {
        clientConfiguration: { disallow_attester_search_for_domains: ['got.this@fromstandardfes.test'] },
      };
```

But some other test may rely on that, etc.. it really is spaghetti.

--------------

I'd like to follow what we do on iOS - one mock API per test, which is configured directly in the test definition with no side effects on other tests. It would look something like this:

```ts
  // testWithBrowser would create browser extension build that communicates to a particular port for mock api
  ava.default('some test', testWithBrowser(undefined, assignedPort, async (t, browser) => { 
    const mockApi = new MockApi();
    mockApi.fesConfig = { ... }; // particular client configuration for this test
    mockApi.ekmConfig = { returnKeys: [ekmKeySamples.e2eValidKey.prv] } // same
    mockApi.attesterConfig = { ... }; // configure Attester with particular public keys for this test 
[...]

PR #4952, opened 2023-02-12
issue #4494 Allow endpoint handlers configurations from inside a test

rrrooommmaaa:
This PR allows to configure handlers from inside a test
(also sets up one test in this way as a demo -- `user3@standardsubdomainfes.localhost:8001 - PWD encrypted message with FES web portal - pubkey recipient in bcc`

TODO: clear handlers after test finishes (if MockApi is reused)?
issue #4494

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/compose.ts
+++ b/test/source/tests/compose.ts
@@ -13,7 +13,7 @@ import { InboxPageRecipe } from './page-recipe/inbox-page-recipe';
 import { OauthPageRecipe } from './page-recipe/oauth-page-recipe';
 import { PageRecipe } from './page-recipe/abstract-page-recipe';
 import { SettingsPageRecipe } from './page-recipe/settings-page-recipe';
-import { protonMailCompatKey, somePubkey } from './../mock/attester/attester-endpoints';
+import { mockAttesterEndpoints, protonMailCompatKey, somePubkey } from './../mock/attester/attester-endpoints';
 import { TestVariant } from './../util';
 import { TestWithBrowser } from './../test';
 import { expect } from 'chai';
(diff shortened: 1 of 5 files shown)


rrrooommmaaa:
As I understand, we want something like this, @tomholub @sosnovsky ?

test/source/tests/compose.ts
```diff
@@ -2545,6 +2548,16 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+        t.mockApi?.setHandlers({
+          ...mockGoogleEndpoints,
+          // ...mockBackendEndpoints,
+          ...mockAttesterEndpoints,
+          ...mockKeyManagerEndpoints,
+          // ...mockWkdEndpoints,
+          // ...mockSksEndpoints,
+          ...mockCustomerUrlFesEndpoints,
+          // ...mockSharedTenantFesEndpoints,
+        });
```
tomholub, line 2560: I didn't mean this way, setting handlers.

I meant we would be setting configuration. Since now we have one MockApi per per test, we have an opportunity to configure mock api with different data (or expectations).

Like this:

```ts
    mockApi.fesConfig = { ... }; // particular client configuration for this test
    mockApi.ekmConfig = { returnKeys: [ekmKeySamples.e2eValidKey.prv] } // same
    mockApi.attesterConfig = { ... }; // configure Attester with particular public keys for this test 
    mockApi.googleConfig = {
      accounts: {
        'e2e.enterprise.test@flowcrypt.com': { // which account is allowed for Gmail login in this test
          messages: ['CC and BCC test'], // exported messages pre-loaded for this test, by subject
        }
      }
    }
```

notice it's receiving information on what to return (per service), not receiving the endpoints directly as done in the PR.

Have a look at any iOS mock tests, like these:
https://github.com/FlowCrypt/flowcrypt-ios/blob/master/appium/tests/specs/mock/composeEmail/CheckInvalidEmailRecipient.spec.ts
https://github.com/FlowCrypt/flowcrypt-ios/blob/master/appium/tests/specs/mock/composeEmail/CheckSendAsAlias.spec.ts

Notice how each test has some unique mock api configuration to work with, so it's clear which data the test needs.

tomholub: @sosnovsky

sosnovsky: Yes, currently it's not exposed, I was planning to add it to `AvaContext`:
```ts
t.mockApi = await startMockApiAndCopyBuild(t);
```
(here - https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/test.ts#L68)

And then use it in tests like this:
```ts
t.mockApi.fesConfig = { ... };
t.mockApi.ekmConfig = { ... };
```

tomholub:
I'll close this attempt - as it's quite different from the intended appraoch

rrrooommmaaa:
> I'll close this attempt - as it's quite different from the intended appraoch

I'm afraid we'll have to wait for half a year for the intended approach to be implemented -- and the tests behave very unstable right now

ioanmo226:
Okay

tomholub:
https://github.com/FlowCrypt/flowcrypt-browser/issues/5021

Issue #4906, opened 2023-01-24
add OpenPGP.js to content script, simplify


tomholub:
This is because last time we tried to use OpenPGP.js in the content script, it didn't work (on Firefox). Actually, it would be nice to load it there. Then we wouldn't have to make so many calls to background_script (which we have to eventually deprecate anyway because of Google Chrome extension manifest v3. https://github.com/FlowCrypt/flowcrypt-browser/issues/1509)

I'll file an issue to try to attempt running openpgp.js in content script again, at which point this comment could be removed, and some things simiplified, in particular:
 - background calls from content script to background page that are only done because OpenPGP.js can be removed, and instead appropriate methods (which rely on OpenPGP.js) can be called directly
 - pgp_block.htm/ts that today calls into background page can instead call the content script when it needs OpenPGP.js, and those background call handlers can be moved from background page to content script

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4725#discussion_r1085610857_

tomholub:
![image](https://user-images.githubusercontent.com/6306961/214399846-831d195c-11c9-4eb8-be07-43812a9bec5d.png)

Issue #2674, opened 2020-03-16
The secure message is rendered twice


limonte:
![image](https://user-images.githubusercontent.com/6059356/76258538-7359ce00-625c-11ea-9a1d-f22cae53a280.png)

limonte:
I added `console.trace()` to `XssSafeFactory.embeddedMsg()`

![image](https://user-images.githubusercontent.com/6059356/76618865-46463d80-6532-11ea-80d8-8852b788806d.png)

The unwanted `XssSafeFactory.embeddedMsg()` is getting called from `GmailElementReplacer.processAtts()`.

limonte:
@tomholub now I'm wondering how did you create the message which doubles pgp frames :)

When I'm sending a rich text message (e.g. to myself) it looks like this:

![image](https://user-images.githubusercontent.com/6059356/76642954-d4371e00-655c-11ea-9100-dfce72916008.png)

There are two files: `noname` and `encrypted.asc`, the message itself is empty.

On the other hand, your message is not empty and contains only `noname`:

![image](https://user-images.githubusercontent.com/6059356/76643048-034d8f80-655d-11ea-9cd8-a9312e2a07ea.png)

![image](https://user-images.githubusercontent.com/6059356/76643060-0b0d3400-655d-11ea-839f-4a89aeb0e8d1.png)

tomholub:
Can reliably reproduce on this message https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/WhctKJVzWGZJdQnRGCkWTXTFxRGmGWQwsdrDDnpmpbvvmHJgJxDstvPtJSktjNpDfSxWrFL

rrrooommmaaa:
Maybe PR #4958 will help fix it?

PR #4958, opened 2023-02-14
Retain block index information when rendering blocks found by MsgBlockParser

rrrooommmaaa:
This PR retain block index information when rendering blocks found by MsgBlockParser.
This will allow us to re-fetch msgId and find the right block by its index.

issue #4342

- Not worth testing
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -28,6 +28,9 @@ export class PgpBlockView extends View {
   public readonly isOutgoing: boolean;
   public readonly senderEmail: string;
   public readonly msgId: string | undefined;
+  // If the block was detected by MsgBlockParser.detectBlock,
+  // it may be useful to know its index in case we want to re-fetch the message by MsgId
+  public readonly blockIndex: number | undefined;
   public readonly encryptedMsgUrlParam: Buf | undefined;
   public readonly signature?: {
     // when parsedSignature is undefined, decryptModule will try to fetch the message
@@ -51,7 +54,18 @@ export class PgpBlockView extends View {
   public constructor() {
     super();
     Ui.event.protect();
-    const uncheckedUrlParams = Url.parse(['acctEmail', 'frameId', 'message', 'parentTabId', 'msgId', 'isOutgoing', 'senderEmail', 'signature', 'debug']);
+    const uncheckedUrlParams = Url.parse([
+      'acctEmail',
+      'frameId',
+      'message',
+      'parentTabId',
+      'msgId',
+      'blockIndex',
+      'isOutgoing',
+      'senderEmail',
+      'signature',
+      'debug',
+    ]);
     this.acctEmail = Assert.urlParamRequire.string(uncheckedUrlParams, 'acctEmail');
     this.parentTabId = Assert.urlParamRequire.string(uncheckedUrlParams, 'parentTabId');
     this.frameId = Assert.urlParamRequire.string(uncheckedUrlParams, 'frameId');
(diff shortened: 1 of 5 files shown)


rrrooommmaaa:
If we place several inline pgp blocks that need re-fetching. having `blockIndex` as a URL parameter would allow us to re-fetch correctly. I'm going to add a test that would fail otherwise.

However, for the issue #4342 I can implement a smarter and more flexible solution, although more complex: -- we can find the block by the signature in the cleartext message -- the signature is possible to find if I strip the all the whitespace from the signature section in the rendered html-to-text variant and from the extracted text/plain and compare them. Or is it too complex with little benefits, @tomholub ?

tomholub:
@sosnovsky ?

tomholub:
you can generally discuss similar questions with sosnovsky, he possibly has better intuition then me for these things. then if he's not sure, he can tag me in turn. or if i happen to see it first, i may weigh in.

sosnovsky:
But what about using code from Tom's proposal https://github.com/FlowCrypt/flowcrypt-browser/issues/4342#issuecomment-1428248809 ?

```ts
if (this.view.signature && !verifyRes.isErrFatal && this.view.decryptModule.canAndShouldFetchFromApi()) {
    // Sometimes the signed content is slightly modified when parsed from DOM,
    // so the message should be re-fetched straight from API to make sure we get the original signed data and verify again
    this.view.signature.parsedSignature = undefined; // force to re-parse
    await this.view.decryptModule.initialize(verificationPubs, true);
    return;
}
```

Should be simpler than retaining block indexes for possible future re-fetch.

rrrooommmaaa:
> But what about using code from Tom's proposal [#4342 (comment)](https://github.com/FlowCrypt/flowcrypt-browser/issues/4342#issuecomment-1428248809) ?

This is for pgp-signed messages with text in the `text/plain` part of a MIME message and a detached signature in the `application/pgp-signature` part, see message `17daefa0eb077da6` for example.

The issue with cleartext signed messages is a different story, they are simply inline blocks inside a MIME message, usually one block, but if a message arrives with more than one PGP blocks, we'll also need to re-fetch the data relevant for the particular block (as re-fetching is done inside `pgp_block.htm` iframe, hence, the block needs to have a way to find itself within the message body, either by index -- more general solution, or (for cleartext signed messages only) - by the signature part -- more complex but in a way more robust as it doesn't rely too much on `text/html` / `text/plain` correlation, i.e. the fact that running `MsgBlockParser` on the HTML part and on the text part will yield the same blocks in the same order.

sosnovsky:
@rrrooommmaaa got it, thanks for explanation! Finding by signature looks like a more solid solution, I think we should go with it

rrrooommmaaa:
I might have stumbled upon a couple of use cases where this index may be useful. I'll present my findings soon.
Or would you like to merge this now, @sosnovsky ?

rrrooommmaaa:
I linked #2674 -- it may be connected.
Also, as I understand, more use case examples were requested from me?
Currently multiple pgp blocks may be returned by `MsgBlockParser` and then passed to individual `pgp_block.htm` frames -- therefore whenever `msgId` is passed, `blockIndex` is also needed, otherwise it will glitch sooner or later.
After #4906 is done, we may think of refactoring `GmailElementReplacer` to do all the decryption work at once in the content script.

tomholub:
Therefore, you are saying it would be better to postpone this PR until #4906 is done?

rrrooommmaaa:
> Therefore, you are saying it would be better to postpone this PR until #4906 is done?

After #5022 is decided upon.

tomholub:
Ok to proceed with that. Would that be a separate PR first? That means the dependency is `4906 -> 5022 -> 4958 PR` or is it `4906 -> 5022 + 4958 PR` ?

rrrooommmaaa:
> Ok to proceed with that. Would that be a separate PR first? That means the dependency is `4906 -> 5022 -> 4958 PR` or is it `4906 -> 5022 + 4958 PR` ?

4096 -> 5022 -> 4958 PR. (if we remove `mgsId` parameter from `pgp_block.htm` frame, then this PR will no longer be needed -- can be abandoned)

rrrooommmaaa:
(meaning. no re-fetching in `pgp_block.htm` will ever happen)

sosnovsky:
> 4906 -> 5022 -> then 4958 PR not needed

This order looks good 👍

sosnovsky:
@rrrooommmaaa can we close this PR as https://github.com/FlowCrypt/flowcrypt-browser/issues/5022 is finished and merged?

rrrooommmaaa:
Yes, it's no longer relevant as `extractArmoredBlock` method no longer exists

PR #4960, opened 2023-02-17
#4950 re-enable viewing remote images in signed/encrypted messages

### LINKED ISSUE #4950: re-enable viewing remote images in signed/encrypted messages
... (edited - see below) ...

ioanmo226:
This PR re-enables viewing remote images in signed/encrypted messages

close #4950 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

chrome-capture-2023-1-21


sosnovsky (commented):
Looks good, just needs loader update

extension/js/common/platform/xss.ts
```diff
@@ -151,7 +151,9 @@ export class Xss {
             a.setAttribute('data-test', 'show-inline-image');
             Xss.replaceElementDANGEROUSLY(img, a.outerHTML); // xss-safe-value - "a" was build using dom node api
           } else {
-            Xss.replaceElementDANGEROUSLY(img, `<span>[Remote images are blocked due to security]</span>`); // xss-safe-value
+            img.setAttribute('data-src', img.getAttribute('src') ?? '');
+            img.classList.add('replace_to_base64_image');
+            img.setAttribute('src', '/img/loading.gif');
```
sosnovsky, line 156: let's use our standard loader `UI.spinner` - it's more lightweight as uses `svg`

sosnovsky (commented):
Looks good, just needs loader update

extension/js/common/platform/xss.ts
ioanmo226: Ok, Sounds good

sosnovsky (commented):
I found issue with this new implementation - images disappear when you try to forward message or reply to it

tomholub:
> I found issue with this new implementation - images disappear when you try to forward message or reply to it

That won't be very easy to fix, since sent messages use text/plain and not text/html. Let's file it for later?

sosnovsky:
Maybe it'll be easier to fix as we're converting remote images to base64. Or we can at least leave image urls there, as currently it's just empty space instead of images.

tomholub:
> Maybe it'll be easier to fix as we're converting remote images to base64. Or we can at least leave image urls there, as currently it's just empty space instead of images.

We can indeed start with image urls then - that's a good idea.

For the base64 images, you'd have to attach them as files, which would make sense for forwarding?

ioanmo226 (commented):
@tomholub , @sosnovsky 
In `htmlSanitizeAndStripAllTags` function, we convert html to platin text + remove img tags.
That's why images are removed when forwarding message.
So do we have to keep image tags in `htmlSanitizeAndStripAllTags` function?

https://github.com/FlowCrypt/flowcrypt-browser/blob/6249a4bcd5ebc6e22e7a61ea33554215a583dab3/extension/js/common/platform/xss.ts#L187

tomholub:
no, just replace remote images with urls to the image within the resulting plain text

ioanmo226:
You mean forwarded message should be `Test test.com/img message` when original message is `Test  message`

tomholub:
the first one. it's plain text, not html, no point putting html tags in there

ioanmo226:
ok

sosnovsky (approved):
Well done 👍

Issue #4929, opened 2023-02-01
Improve tests performance by testing iframes directly


sosnovsky:
If they actually fixed this `site-per-process` issue, that would mean we should be able to test the iframes directly where they are, for example, in Live tests we should be able to test them directly from the Gmail page, instead of copying the URL and opening a separate tab, and testing there. That would speed up the tests. Let's file that for another issue (unless you have already done exactly that).

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4928#discussion_r1093170239_

tomholub:
![image](https://user-images.githubusercontent.com/6306961/216057822-dfcbb34d-abfc-49bc-9cea-71952a807ded.png)

Issue #4810, opened 2022-11-24
Test (some of) GmailElementReplacer tests in mock environment


rrrooommmaaa:
For this we would need to extend our mock gmail endpoints, like `mail/u/0/#inbox/FMfcgzGrbHrBdFGBXqpFZvSkcQpKkvrM`
to return a page similar to what gmail once returned

_Originally posted by @rrrooommmaaa in https://github.com/FlowCrypt/flowcrypt-browser/pull/4807#discussion_r1031410524_

PR #4961, opened 2023-02-19
#4342 re-fetch cleartext signed message when HTML version is corrupted

### LINKED ISSUE #4342: signed message occasionally throws an error 'Message digest did not match'

rrrooommmaaa:
This PR re-fetches cleartext signed message from `text/plain` alternative part if signature verification fails

close #4342

issue #4810 (sets an example of how to test GmailElementReplacer)

issue #4929 (added a test helper method pgpBlockCheck for checking data in an existing pgp block)

- Tests added

### To be filled by reviewers


test/source/tests/decrypt.ts
```diff
@@ -892,6 +892,43 @@ XZ8r4OC6sguP/yozWlkG+7dDxsgKQVBENeG6Lw==
       })
     );
 
+    test(
+      'signature - cleartext signed messages from HTML are re-fetched when needed',
+      testWithBrowser('ci.tests.gmail', async (t, browser) => {
+        const acctEmail = 'ci.tests.gmail@flowcrypt.test';
+        const settingsPage = await browser.newExtensionSettingsPage(t, acctEmail);
+        const accessToken = await BrowserRecipe.getGoogleAccessToken(settingsPage, acctEmail); // todo: include in t?
```
rrrooommmaaa, line 900: I think we can simplify some tests if we can extract google access token from `t` -- we can put it there in a helper method like `setUpCommonAcct`, so we don't have to open a dummy page just to retrieve the access token.

sosnovsky: Yeah, it'll be nice to have access token available through `t`. Probably we can include such improvement when working on part 2 of #4494, there we can fetch access token when configuring test account.

sosnovsky (commented):
Well done 👍  Just a small improvement for `eslint-disable`

test/source/tests/decrypt.ts
sosnovsky: #4966

sosnovsky (approved):
👍

PR #4963, opened 2023-02-23
#4959 Show signature right away in secure compose window

### LINKED ISSUE #4959: Add signature in secure compose windows
Mail Subject: Signature missing in FC compose

The UX with regards to Gmail footer/signature should be the same as when composing with Gmail, that is, already show when new compose box is rendered (but don't show if this is a reply box, while still ensuring that it gets added appropriately for replies, and that it behaves well when sending out from a Draft too, whether new or reply, after this change).

ioanmo226:
This PR shows signature right away in secure compose window

close #4959 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


sosnovsky (approved):
👍

PR #4969, opened 2023-02-25
removed reference to Swal from background page

rrrooommmaaa:
This PR removes sweetalert reference from background page.
Was it ever used? Is it possible to use it?

(as part of research for issue #4866)

### To be filled by reviewers


rrrooommmaaa:
I'll mark this as ready after I do some tests on #4866

rrrooommmaaa:
It indeed doesn't make sense to call `Swal` from the background page

sosnovsky (approved):

PR #4971, opened 2023-02-25
Rely on OpenPGP.js to forbid using weak algos e.g. RSA1024 etc.

### LINKED ISSUE #4905: Use OpenPGP.js v6 native ways to reject algorithms
I suppose there was new functionality for this in OpenPGP.js v5, where we can simply configure which algorithms to use or forbid? Either in global config or when doing particular actions (forgot about their implementation). In which case we could stop using `removeWeakKeyPackets` / remove it. Can be another issue.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4725#discussion_r1085576518_

### LINKED ISSUE #5631: Missing RSA key bits reformatting when importing an RSA 2047 private key
Description:

When importing an RSA 2047 private key in the FlowCrypt browser extension, an error will occur saying it can be used for encryption and so the importing process will not be completed. As for encrypting message for using that 1 bit off rsa public key, it will throw an error complaining about the public key.

For comparison, the FlowCrypt android app smartly fixed that 1 bit off the 2047 key into a 2048. Ideally, it should also be the case for the FlowCrypt browser extensions for private key & contacts importing processes.

Test PGP key pair I used:
[2047.asc.txt](https://github.com/FlowCrypt/flowcrypt-browser/files/14504966/2047.asc.txt)

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGxRxKKzgrWRLpPCgKjfvdmTpDt

rrrooommmaaa:
This PR removes own protection as OpenPGP.js `config.minRSABits` implicitly has the default value of 2047.

close #4905
close #5631

- Unit tests are already in place:
`[KeyUtil.diagnose] decrypts and tests secure PK and insecure SK`
`[unit][OpenPGPKey.parse] sets usableForEncryption and usableForSigning to false for RSA key less than 2048`
`[unit][OpenPGPKey.parse] sets usableForEncryption to false and usableForSigning to true for 2048/RSA PK and 1024/RSA SK`
`[unit][OpenPGPKey.decrypt] sets usableForEncryption to false and usableForSigning to true for 2048/RSA PK and 1024/RSA SK`

### To be filled by reviewers


sosnovsky (approved):
looks good 👍

PR #4974, opened 2023-02-27
#4962 Public Key lookup from keys.openpgp.org

### LINKED ISSUE #4962: search public keys on keys.openpgp.org
I'd like to add another source of public keys in PubLookup, a class called `KeysOpenpgpOrg`.

So here we would have:

```diff
  public lookupEmail = async (email: string): Promise<PubkeysSearchResult> => {
    const wkdRes = await this.wkd.lookupEmail(email);
    if (wkdRes.pubkeys.length) {
      return wkdRes;
    }
    if (this.internalSks) {
      const res = await this.internalSks.lookupEmail(email);
      if (res.pubkey) {
        return { pubkeys: [res.pubkey] };
      }
    }
-    return await this.attester.lookupEmail(email);
+   const attesterRes = await this.attester.lookupEmail(email);
+   if(any results from attesterRes) {
+        return attesterRes;
+      }
+   const keysOpenpgpOrgRes = await this.keysOpenpgpOrg.lookupEmail(email);
+   if(any results from keysOpenpgpOrgRes) {
+        return keysOpenpgpOrgRes;
+      }      
+      return { pubkeys: [] };     
  };
```

Then we'll have a new set of client configuration that work similarly as `allow_attester_search_only_for_domains` and `disallow_attester_search_for_domains`, called  `allow_keys_openpgp_org_search_only_for_domains` and `disallow_keys_openpgp_org_search_for_domains`, and they will do work exactly the same as they do for Attester, except for this new key source. Tests for this should be copied and adjusted too, and a new set of test mock api endpoints will be needed.

As for the implementation:
 - here is the API docs https://keys.openpgp.org/about/api (please link as code comment), in particular we are calling `/vks/v1/by-email/<URI-ENCODED EMAIL-ADDRESS>`
 - Error `429` should be interpreted as error 404 - public key not found (because their rate limits are excessively strict, and you could run into them just by loading a conversation with 20 people in it a few times. Multiply that by 1000 workers in a building and I see a lot of 429s potentially happening. please document this as a comment).
 - `PubLookup.lookupEmail` should accept one more parameter: `{ skipOpenpgpOrg: boolean } = { skipOpenpgpOrg: false }`. All current usecases should use the default (do look up on there too) except when all of the below is true, in which case we will add `skipOpenpgpOrg: true`:
        -  we are loading public keys we already have just to keep them fresh, like we do in secure compose as its loading, or when verifying a message
        - we already have at least one valid key for that email address in local contacts

Again, the reason for the last requirement, is to call that server a lot less, trying to avoid these 429s. It will unfortunately make key lifecycle changes less robust, but it's the best we can do, I think. This should also be documented to make that clear.

ioanmo226:
This PR implemented public key lookup from keys.openpgp.org

close #4962 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-storage-module.ts
```diff
@@ -162,7 +162,8 @@ export class ComposeStorageModule extends ViewModule<ComposeView> {
       throw Error('Empty email');
     }
     try {
-      const lookupResult = await this.view.pubLookup.lookupEmail(email);
+      // Skip keys.openpgp.org search if there is at least one valid key
+      const lookupResult = await this.view.pubLookup.lookupEmail(email, storedPubkeys.length > 0);
```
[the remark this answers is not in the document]
ioanmo226: Maybe we can add condition something like this?

```ts
const shouldSkipOpenpgpOrg = storedPubkeys.some((pubkey) => pubkey.pubkey.usableForEncryption && pubkey.pubkey.usableForSigning)
```

ioanmo226:
@sosnovsky Ready for review

sosnovsky (approved):
👍

extension/chrome/elements/pgp_block_modules/pgp-block-signature-module.ts
```diff
@@ -42,7 +43,9 @@ export class PgpBlockViewSignatureModule {
         } else {
           $('#pgp_signature').addClass('gray_label').text('verifying signature...');
           try {
-            const { pubkeys } = await this.view.pubLookup.lookupEmail(signerEmail);
+            const storedContact = await ContactStore.getOneWithAllPubkeys(undefined, signerEmail);
+            const shouldSkipOpenpgpOrg = storedContact?.sortedPubkeys.some(pubkey => pubkey.pubkey.usableForEncryption);
+            const { pubkeys } = await this.view.pubLookup.lookupEmail(signerEmail, shouldSkipOpenpgpOrg);
```
tomholub, line 48: This will later need some improving:

1) `usableForEncryption` is not a useful metric when dealing with signature. Should use `usableForSigning`
2) Reviewing the code, it seems we only run this when we already failed verification, not every time that message is verified. Is that true? I suppose we could fetch from `keys.openpgp.org` unconditionally here.

That means, all in all, I think these 3 lines can be reverted.

extension/js/common/api/key-server/keys.openpgp.org.ts
```diff
@@ -0,0 +1,46 @@
+/* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
```
tomholub, line 1: This is really a nitpick but it doesn't seem common to use dots here, the style we use is `ClassName` that becomes `class-name.ts`

extension/js/common/api/pub-lookup.ts
```diff
@@ -20,19 +21,29 @@ export type PubkeysSearchResult = { pubkeys: string[] };
 export class PubLookup {
   public attester: Attester; // attester is a publicly available public key server
   public wkd: Wkd;
+  public keysOpenpgpOrg: KeysOpenpgpOrg; // keys.openpgp.org
   public keyManager: KeyManager | undefined; // key manager is a flowcrypt-provided internal company private and public key server
```
tomholub, line 25: unused, can remove

test/source/mock/backend/backend-data.ts
```diff
@@ -117,6 +119,19 @@ export class BackendData {
         allow_attester_search_only_for_domains: [],
       };
     }
+    if (domain === 'only-allow-some-domains-for-keys-openpgp-org-client-configuration.flowcrypt.test') {
+      return {
+        flags: [],
+        allow_keys_openpgp_org_search_only_for_domains: ['flowcrypt.com'],
```
tomholub, line 125: A test domain would be better, like `allowed-domain.test`

tomholub (commented):
I'm filing comments to address in a follow-up PR. Looks good overall

test/source/mock/backend/backend-data.ts
ioanmo226: Got it

extension/js/common/api/key-server/keys.openpgp.org.ts
ioanmo226: ok

tomholub:
>  is documented clearly and usefully, or doesn't need documentation

@ioanmo226 @sosnovsky this issue/PR includes two new client configuration properties, documentation for which is missing in https://flowcrypt.com/docs/technical/enterprise/configuration/client-configuration.html . I'll go ahead and add it now - this is something to be mindful of in future PRs

ioanmo226:
Sounds good. Well noted

sosnovsky:
Ok, I'll re-check configuration properties before approving in the future 👍

PR #4998, opened 2023-03-07
Allow to pass Uint8Array | string to MsgUtil functions

rrrooommmaaa:
This PR allows to pass Uint8Array | string to `MsgUtil` functions

### To be filled by reviewers


extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -36,12 +37,12 @@ export namespace PgpMsgMethod {
     export type Type = { data: Uint8Array | string };
     export type Decrypt = {
       kisWithPp: KeyInfoWithIdentityAndOptionalPp[];
-      encryptedData: Uint8Array;
+      encryptedData: Uint8Array | string;
```
rrrooommmaaa, line 40: I had an urge to refactor some `MsgUtil` functions to accept `Uint8Array | string` instead of `Uint8Array` to avoid unnecessary conversion in the calling pieces of code. These conversions are often not necessary.
This would also simplify a lot of mock and test code -- not done yet, if you approve, I'll finish the update.
@sosnovsky ?

sosnovsky: Sounds good, let's do it 👍

sosnovsky (approved):
Well done, looks much cleaner now 👍

Issue #4983, opened 2023-02-28
re-enable viewing remote images in signed/encrypted messages still not working as expected.


ioanmo226:
Showing remote images still don't work well yet.

ioanmo226:
@sosnovsky As I checked carefully, image is not displayed because image server is returning CORS error.
I think this issue should be fixed in customer side?

> Access to fetch at 'https://go.[clientserver].com/rs/222-DNK-376/images/[logo_name].png' from origin 'chrome-extension://mkefadjohdmonnjbidopnmjlepdomjmi' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

ioanmo226:
Also what error message do I have to show?

ioanmo226:
Will it fix CORS issue? I don't think it would solve cors issue

sosnovsky:
Checking it now..

sosnovsky:
> Will it fix CORS issue? I don't think it would solve cors issue

Yeah, seems like CORS issue will occur for all possible data fetches.
Then for now let's show error `Can't load remote image` and link below `Open image in new tab`

> Or maybe we can use background script to fetch resource?

Did you try this method?

ioanmo226:
Not yet. I'll try and i think this might work.
I have current in-progress [issue](https://github.com/FlowCrypt/flowcrypt-browser/issues/4962).
I'm gonna have a look at this issue after current issue.
Or do you wanna me to have a look at this one first?

tomholub:
> could it be that the cors error is the result of extension’s csp? maybe another approach to rendering could work around the problem. 

I suspect adjusting csp would fix the CORS issue. @rrrooommmaaa may know

sosnovsky:
> I suspect adjusting csp would fix the CORS issue. @rrrooommmaaa may know

I think CSP won't affect this issue, as this issue is on other server side - it doesn't allow external servers to fetch images.
We had similar issue with attachments stored on s3 - https://github.com/FlowCrypt/flowcrypt-browser/issues/3026.

tomholub:
Sure, it's missing CORS headers returned from the server. But in some contexts, extensions can still access such content. One such way is if we allow the extension to run access `*` (all websites) in manifest. Not that I want to do that. But there may be other ways still, for extensions.

[...]

PR #4994, opened 2023-03-06
#4983 Remote image load error handling

ioanmo226:
This PR implemented remote image load error handling

close #4983 // if this PR closes an issue

----------------------------------

**Tests** _(delete all except exactly one)_:
- Tests added or updated

--------------------------------

### To be filled by reviewers

I have reviewed that this PR... _(tick whichever items you personally focused on during this review)_:
- [ ] addresses the issue it closes (if any)
- [ ] code is readable and understandable
- [ ] is accompanied with tests, or tests are not needed
- [ ] is free of vulnerabilities
- [ ] is documented clearly and usefully, or doesn't need documentation

PR #4999, opened 2023-03-07
issue 4983 change CSP to allow loading remote images

tomholub:
This PR attempts to fix loading remote images.

issue #4983

- Tests will be added later

### To be filled by reviewers

### DIFF
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -78,5 +78,5 @@
     "/chrome/elements/oauth2.htm"
   ],
   "minimum_chrome_version": "96",
-  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; img-src 'self' https://*.googleusercontent.com data: blob:; frame-src 'self' blob:; worker-src 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; object-src 'none'; base-uri 'self'"
+  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; img-src 'self' https://* data: blob:; frame-src 'self' blob:; worker-src 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; object-src 'none'; base-uri 'self';"
 }
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -78,5 +78,5 @@
     "/chrome/elements/oauth2.htm"
   ],
   "minimum_chrome_version": "96",
-  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; img-src 'self' https://* data: blob:; frame-src 'self' blob:; worker-src 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; object-src 'none'; base-uri 'self';"
+  "content_security_policy": "upgrade-insecure-requests; script-src 'self'; frame-ancestors https://mail.google.com 'self'; img-src 'self' https://* data: blob:; frame-src 'self' blob:; worker-src 'self'; form-action 'none'; media-src 'none'; font-src 'none'; manifest-src 'none'; object-src 'none'; base-uri 'self'"
 }


tomholub:
I've tested this change by just manually adding an img tag. It didn't work on master, but works now.

@ioanmo226 could you please test it on the original customer sample?

image

ioanmo226:
Hmmm.. Just tested.
But still it doesn't work for customer sample. Still returns CORS error

ioanmo226:
Access to fetch at 'https://domain.com/logo.png' from origin 'chrome-extension://mkefadjohdmonnjbidopnmjlepdomjmi' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

ioanmo226:
I think loading image through our api won't trigger security if we don't store any data on our server?

tomholub:
Where can I find it? I tested this:

image

image

tomholub:
> I think loading image through our api won't trigger security/privacy if we don't store any data on our server?

Our users would have to trust us that we wouldn't have a peek at it, defying the whole point of our business.

ioanmo226:
> Where can I find it? I tested this:

Just sent you email (.eml file customer shared)

ioanmo226:
> > I think loading image through our api won't trigger security/privacy if we don't store any data on our server?
> 
> Our users would have to trust us that we wouldn't have a peek at it, defying the whole point of our product/business.

I see

tomholub:
How did you load the `.eml` email into your Gmail inbox? I used Thunderbird before, wonder if you have an easier way.

ioanmo226:
I'm currently using Apple mail app.
When I double-clicked .eml file, apple mail opened. 
And then I have to click `Move to inbox` button. (That's all)

image

sosnovsky:
> Seems to work when I replace an `img` `src` with the customer url `https://REDACTED/rs/REDACTED/images/REDACTED.png`

Yes, `CORS` error doesn't occur when image is loaded with `` tag.
It happens when we try to download image data and get `base64` from it.

tomholub:
> > Seems to work when I replace an `img` `src` with the customer url `https://REDACTED/rs/REDACTED/images/REDACTED.png`
> 
> Yes, `CORS` error doesn't occur when image is loaded with `` tag. It happens when we try to download image data and get `base64` from it.

Alright, so I suppose what we did was backwards after all.

We should stick to `` tags. Remove the csp restriction as I've just done, and all of the images will work, regardless of CORS.

ioanmo226:
Does it mean we have to revert https://github.com/FlowCrypt/flowcrypt-browser/pull/4994 and remove the CSP restrictions?
Are you okay with removing CSP restrictions?

tomholub:
I didn't entirely remove CSP of course, but I did adjust it to allow remote images from any source.

Whether we load images directly through `src` tags, or whether we download them from AJAX and load the data in, I don't see much of a security difference. So I propose to update the CSP as I've done here.

Yes, I'd revert it.

ioanmo226:
Okay, Yeah

tomholub:
I'd still like to continue to clearly frame images to make it clear they are framed, plus if the image is remote, I'd like a warning inside of that box/frame just above the image: `Authenticity of this remote image cannot be verified.`

tomholub:
@sosnovsky does this sound ok to you? Want to run this by you to make sure.

sosnovsky:
> @sosnovsky does this sound ok to you? Want to run this by you to make sure.

Yes, it's ok, as it seems no other way to make remote images work well in encrypted messages.

tomholub:
I'll let the customer know

ioanmo226:
What do you mean clearly frame images?
You mean add border to remote images?

> I'd still like to continue to clearly frame images to make them stand out

tomholub:
> What do you mean clearly frame images? You mean add border to remote images?
> 
> > I'd still like to continue to clearly frame images to make them stand out

I think we already border all images, with a css border.

I'd like something similar, like you have done for the error handling:

image

Except that instead of `Failed to load image` there would be the `Authenticity of this remote image cannot be verified.` and below that, there would be the actual image.

tomholub:
Sorry, I got confused. Apparently we no longer put a css border around images. I think we used to do that? Nevermind that.

Anyway, for remote images, please do as described here: https://github.com/FlowCrypt/flowcrypt-browser/pull/4999#issuecomment-1458420141 which you should be able to reuse existing code for.

PR #5002, opened 2023-03-07
#5001 Enable remote image loading

### LINKED ISSUE #5000: add an automated test for loading a cors-restricted remote image
Can use `` for the test.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4999_

### LINKED ISSUE #5001: load remote images directly in img tags, allow it in CSP, add warning around remote images
We should stick to `` tags. Remove the csp restriction as I've just done, and all of the images will work, regardless of CORS.

It seems to me to be the only sane way to support this functionality.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4999#issuecomment-1458392763_

ioanmo226:
This PR load remote images directly in img tags, allow it in CSP, add warning around remote images.

close #5000 // if this PR closes an issue
close #5001 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

Screenshot:
image


extension/css/cryptup.css
```diff
@@ -1375,6 +1353,23 @@ td {
   opacity: 0.75;
 }
 
+#pgp_block .remote_image_container {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  width: max-content;
```
sosnovsky, line 1360: With `width: max-content;` large images cause overflow and add scroll.
Let's replace it with `max-width: 100%`, so all images will fit into page width

extension/js/common/platform/xss.ts
```diff
@@ -141,24 +144,21 @@ export class Xss {
...
               'style',
               `text-decoration: none; background: #FAFAFA; padding: 4px; border: 1px dotted #CACACA; display: inline-block; ${heightWidth}`
             );
             a.setAttribute('data-test', 'show-inline-image');
             Xss.replaceElementDANGEROUSLY(img, a.outerHTML); // xss-safe-value - "a" was build using dom node api
           } else {
-            img.setAttribute('data-src', img.getAttribute('src') ?? '');
-            img.classList.add('replace_to_base64_image');
-            img.setAttribute('src', '/img/svgs/spinner-green-small.svg');
+            const remoteImgEl = `<div class="remote_image_container" data-src="${src}" data-size="${heightWidth}" data-test="remote-image-container"><span>Authenticity of this remote image cannot be verified.</span></div>`;
```
sosnovsky, line 154: I tested several images and they all have `height: auto; width: auto`, so maybe this calculation doesn't work for remote images?
Also setting original image size for large images will make them too big for email layout.
I think we won't need this `data-size` property here after setting `max-width: 100%` for `.remote_image_container`

ioanmo226: Well, here's the situation: 
let's say we have a logo.png image that is 300x300 pixels in size. If we style the image to have a width and height of 30 pixels, the rendered image may appear stretched or pixelated.

To avoid this, I think we should  include the original image's size attributes in the HTML code, so that the image is properly scaled down to the desired size.

extension/css/cryptup.css
ioanmo226: image

If we don't use `width: max-content`, warning text will overflow and look in multiple lines. (In cutomer email. Maybe it's because parent element is a tag though).
Are you fine with it?

sosnovsky: Ah, got it, then we probably should have both - `width: max-content` and `max-width: 90%`, so warning will fit in 1 line and large images won't have overflow.

ioanmo226: Sounds good

extension/js/common/platform/xss.ts
sosnovsky: Yes, it'll be nice to have, but as I mentioned I checked a couple remote images and they had `height: auto; width: auto`, which means `img.width ?? img.clientWidth` returns `null`. It seems like we can get dimensions of remote image only after it's loaded.
Also setting `max-width` should automatically resize images according to their proportions, so they won't be stretched or pixelated.

ioanmo226: image.clientWidth would be null but image.width would be correct value if width is set something like this (like customer's email) ``

But I agree. it won't be stretched badly.

sosnovsky (commented):
We still need `max-width: 100%` for `#pgp_block .remote_image_container img`, as currently large images overflow `.remote_image_container`:

Screenshot 2023-03-08 at 10 48 57

ioanmo226:
Oops!. You are right. Missed it when removing javascript variable.
Thanks for your notice

sosnovsky (approved):
👍

ioanmo226:
@sosnovsky We can merge this PR now?

PR #5013, opened 2023-03-14
use OpenPGP.js from web content script, simplify

### LINKED ISSUE #4906: add OpenPGP.js to content script, simplify

rrrooommmaaa:
This PR use OpenPGP.js from web content script, removing unnecessary calls to background

close #4906

- Not worth testing
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/background_page/background_page.ts
+++ b/extension/js/background_page/background_page.ts
@@ -7,7 +7,7 @@ import { Bm, BrowserMsg } from '../common/browser/browser-msg.js';
 import { emailKeyIndex } from '../common/core/common.js';
 import { VERSION } from '../common/core/const.js';
 import { ExpirationCache } from '../common/core/expiration-cache.js';
-import { processAndStoreKeysFromEkmLocally, getLocalKeyExpiration } from '../common/helpers.js';
+import { processAndStoreKeysFromEkmLocally } from '../common/helpers.js';
 import { Catch } from '../common/platform/catch.js';
 import { AcctStore } from '../common/platform/store/acct-store.js';
 import { ContactStore } from '../common/platform/store/contact-store.js';
@@ -61,7 +61,6 @@ console.info('background_process.js starting');
   BrowserMsg.bgAddListener('storeAcctGet', (r: Bm.StoreAcctGet) => AcctStore.get(r.acctEmail, r.keys));
   BrowserMsg.bgAddListener('storeAcctSet', (r: Bm.StoreAcctSet) => AcctStore.set(r.acctEmail, r.values));
   BrowserMsg.bgAddListener('processAndStoreKeysFromEkmLocally', processAndStoreKeysFromEkmLocally);
-  BrowserMsg.bgAddListener('getLocalKeyExpiration', getLocalKeyExpiration);
 
   // todo - when https://github.com/FlowCrypt/flowcrypt-browser/issues/2560
   //   is fixed, this can be moved to the gmail content script, and some may be removed
--- a/extension/js/common/browser/browser-msg.ts
+++ b/extension/js/common/browser/browser-msg.ts
@@ -84,7 +84,6 @@ export namespace Bm {
   export type ReRenderRecipient = { email: string };
   export type SaveFetchedPubkeys = { email: string; pubkeys: string[] };
   export type ProcessAndStoreKeysFromEkmLocally = { acctEmail: string; decryptedPrivateKeys: string[] };
-  export type GetLocalKeyExpiration = { acctEmail: string };
 
   export namespace Res {
     export type GetActiveTabInfo = {
@@ -107,7 +106,6 @@ export namespace Bm {
     export type AjaxGmailAttachmentGetChunk = { chunk: Buf };
     export type _tab_ = { tabId: string | null | undefined }; // eslint-disable-line @typescript-eslint/naming-convention
     export type SaveFetchedPubkeys = boolean;
-    export type GetLocalKeyExpiration = number | undefined;
     export type ProcessAndStoreKeysFromEkmLocally = {
       needPassphrase?: boolean;
       updateCount?: number;
@@ -135,8 +133,7 @@ export namespace Bm {
       | AjaxGmailAttachmentGetChunk
       | SaveFetchedPubkeys
       | ProcessAndStoreKeysFromEkmLocally
-      | PgpKeyBinaryToArmored
-      | GetLocalKeyExpiration;
+      | PgpKeyBinaryToArmored;
   }
 
   export type AnyRequest =
@@ -177,7 +174,6 @@ export namespace Bm {
     | ReRenderRecipient
     | SaveFetchedPubkeys
     | ProcessAndStoreKeysFromEkmLocally
-    | GetLocalKeyExpiration
     | PgpKeyBinaryToArmored
     | AuthWindowResult;
 
@@ -246,8 +242,6 @@ export class BrowserMsg {
           BrowserMsg.sendAwait(undefined, 'saveFetchedPubkeys', bm, true) as Promise<Bm.Res.SaveFetchedPubkeys>,
         processAndStoreKeysFromEkmLocally: (bm: Bm.ProcessAndStoreKeysFromEkmLocally) =>
           BrowserMsg.sendAwait(undefined, 'processAndStoreKeysFromEkmLocally', bm, true) as Promise<Bm.Res.ProcessAndStoreKeysFromEkmLocally>,
-        getLocalKeyExpiration: (bm: Bm.GetLocalKeyExpiration) =>
-          BrowserMsg.sendAwait(undefined, 'getLocalKeyExpiration', bm, true) as Promise<Bm.Res.GetLocalKeyExpiration>,
       },
     },
     passphraseEntry: (dest: Bm.Dest, bm: Bm.PassphraseEntry) => BrowserMsg.sendCatch(dest, 'passphrase_entry', bm),
--- a/extension/js/common/helpers.ts
+++ b/extension/js/common/helpers.ts
@@ -175,7 +175,7 @@ export const processAndStoreKeysFromEkmLocally = async ({
   };
 };
 
-export const getLocalKeyExpiration = async ({ acctEmail }: Bm.GetLocalKeyExpiration): Promise<Bm.Res.GetLocalKeyExpiration> => {
+export const getLocalKeyExpiration = async (acctEmail: string): Promise<number> => {
   const kis = await KeyStore.get(acctEmail);
   const expirations = await Promise.all(kis.map(async ki => (await KeyUtil.parse(ki.public))?.expiration ?? Number.MAX_SAFE_INTEGER));
   return Math.max(...expirations);
--- a/extension/js/content_scripts/webmail/setup-webmail-content-script.ts
+++ b/extension/js/content_scripts/webmail/setup-webmail-content-script.ts
@@ -14,6 +14,7 @@ import { Ui } from '../../common/browser/ui.js';
 import { ClientConfiguration, ClientConfigurationError } from '../../common/client-configuration.js';
 import { Str, Url } from '../../common/core/common.js';
 import { InMemoryStoreKeys, VERSION } from '../../common/core/const.js';
+import { getLocalKeyExpiration } from '../../common/helpers.js';
 import { Injector } from '../../common/inject.js';
 import { Lang } from '../../common/lang.js';
 import { Notifications } from '../../common/notifications.js';
@@ -382,7 +383,7 @@ export const contentScriptSetupIfVacant = async (webmailSpecific: WebmailSpecifi
   };
 
   const notifyExpiringKeys = async (acctEmail: string, clientConfiguration: ClientConfiguration, notifications: Notifications) => {
-    const expiration = await BrowserMsg.send.bg.await.getLocalKeyExpiration({ acctEmail });
+    const expiration = await getLocalKeyExpiration(acctEmail);
     if (expiration === undefined) {
       return;
     }
--- a/extension/manifest.json
+++ b/extension/manifest.json
@@ -40,27 +40,13 @@
   ],
   "content_scripts": [
     {
-      "matches": [
-        "https://mail.google.com/*"
-      ],
-      "css": [
-        "/css/webmail.css",
-        "/css/sweetalert2.css"
-      ],
-      "js": [
-        "/lib/purify.js",
-        "/lib/jquery.min.js",
-        "/lib/sweetalert2.js",
-        "/js/content_scripts/webmail_bundle.js"
-      ]
+      "matches": ["https://mail.google.com/*"],
+      "css": ["/css/webmail.css", "/css/sweetalert2.css"],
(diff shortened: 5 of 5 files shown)


extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -998,4 +999,27 @@ export class OpenPGPKey {
...
+          OpenPGPKey.fixUint8Arrays(x, processed);
+        }
+      }
+    }
+  };
+
+  private static patch = (opgpKey: OpenPGP.Key) => {
+    if (Catch.browser()?.name === 'firefox') {
+      // need to re-create Uint8Array from globalThis.Uint8Array or window.Uint8Array for Firefox content script
+      OpenPGPKey.fixUint8Arrays(opgpKey);
```
rrrooommmaaa, line 1022: The main problem with Firefox was that Uint8Array's created by OpenPGP.js were of `globalThis.Uint8Array` or `window.Uint8Array` types
and this fix https://github.com/openpgpjs/web-stream-tools/pull/23 didn't help because
calling a method of such an object (e.g. `stripped = bin.subarray(i);` inside OpenPGP.js) caused an exception `Error: Permission denied to access property "constructor"of such an object cause`
However, this type of workaround works: `new Uint8Array(bin).subarray(i)`
So this patch iterates over the tree of OpenPGP.Key object to re-initialize all Uint8Array's with `new Uint8Array`

Opinions, @sosnovsky, @tomholub?

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -269,6 +269,7 @@ export class OpenPGPKey {
         throw new Error(`Key ${key.id} is not a private key`);
       }
       opgpKey = (await opgp.revokeKey({ key: opgpKey, format: 'object' })).privateKey;
+      OpenPGPKey.patch(opgpKey);
```
rrrooommmaaa, line 272: Perhaps, it makes sense to call ``OpenPGP.patch` each time a new OpenPGP.Key object is created (by reading / fixing / generating etc.), and we also need to check message objects

tooling/build-types-and-manifests.ts
```diff
@@ -144,7 +144,20 @@ const makeLocalFesBuild = (sourceBuildType: string) => {
...
+    code.replace(/\/\* ----- [^\r\n]*\/content_scripts\/webmail\/.*}\)\(\);/s, `${testCode}\r\n\r\n})();`)
+  );
+};
+
 updateEnterpriseBuild();
 makeMockBuild(CHROME_CONSUMER);
 makeMockBuild(CHROME_ENTERPRISE);
 makeLocalFesBuild(CHROME_ENTERPRISE);
+makeContentScriptTestsBuild('chrome-consumer-mock');
+// makeContentScriptTestsBuild('firefox-consumer'); // for manual testing of content script in Firefox
```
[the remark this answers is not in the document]
rrrooommmaaa: Should we look into automating firefox tests?

tooling/build-types-and-manifests.ts
```diff
@@ -144,7 +144,20 @@ const makeLocalFesBuild = (sourceBuildType: string) => {
...
+  edit(`${buildDir(testBuildType)}/js/content_scripts/webmail_bundle.js`, code =>
+    code.replace(/\/\* ----- [^\r\n]*\/content_scripts\/webmail\/.*}\)\(\);/s, `${testCode}\r\n\r\n})();`)
+  );
+};
+
 updateEnterpriseBuild();
 makeMockBuild(CHROME_CONSUMER);
 makeMockBuild(CHROME_ENTERPRISE);
 makeLocalFesBuild(CHROME_ENTERPRISE);
+makeContentScriptTestsBuild('chrome-consumer-mock');
```
rrrooommmaaa, line 162: The build `chrome-consumer-content-script-tests-mock` is used in automated testing

test/source/tests/content-script.ts
```diff
@@ -0,0 +1,24 @@
...
+    'content script test',
+    testWithBrowser(undefined, async (t, browser) => {
+      const authorizationHeader = {
+        Authorization: 'Bearer just-to-load-and-run-content-scripts', // eslint-disable-line @typescript-eslint/naming-convention
+      };
+      const gmailPage = await browser.newMockGmailPage(t, authorizationHeader);
+      await gmailPage.waitAny('@content-script-test-result');
+      const allValues = (await gmailPage.readAll('@content-script-test-result')).map(el => el.innerText);
+      // multiple results appear in Firefox. The test is successful only if all divs have the word 'pass'
+      expect(Value.arr.unique(allValues)).to.eql(['pass']);
```
rrrooommmaaa, line 20: The test content script is called 3 times in Firefox for some unknown reason (while only once in Chrome) when opening the gmail page
image

test/source/tests/content-script-test.js
rrrooommmaaa, line 1: 1. Changes in this file are ignored on incremental builds for some reason.
2. I think it makes sense to test all possible OpenPGP.js use cases from the content script in Firefox

tooling/build-types-and-manifests.ts
```diff
@@ -144,7 +144,20 @@ const makeLocalFesBuild = (sourceBuildType: string) => {
   );
 };
 
+const makeContentScriptTestsBuild = (sourceBuildType: string) => {
+  const testCode = readFileSync('./test/source/tests/content-script-test.js').toString();
+  const testBuildType = sourceBuildType.endsWith('-mock')
+    ? sourceBuildType.slice(0, -5) + '-content-script-tests-mock'
+    : sourceBuildType + '-content-script-tests';
+  exec(`cp -r ${buildDir(sourceBuildType)} ${buildDir(testBuildType)}`);
+  edit(`${buildDir(testBuildType)}/js/content_scripts/webmail_bundle.js`, code =>
```
rrrooommmaaa, line 153: This basically wipes out the webMail code, only libraries are left and `[test/source/tests/content-script-test.js](https://github.com/FlowCrypt/flowcrypt-browser/pull/5013/files#diff-df3637596ba657872526c5b17037eaae42f0734b1ae951bdcf1ba7efbdb8f2fd)
immediately calls `OpenPGP.js` and `web-stream-tools` directly

test/source/tests/content-script.ts
rrrooommmaaa: This is an example of success
image

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: > > The main problem with Firefox was that Uint8Array's created by OpenPGP.js were of globalThis.Uint8Array or window.Uint8Array types
> 
> Does that mean they create sometimes one type and sometimes the other?
 
1. With regard to this particular test `if (x instanceof window.Uint8Array)` seems to be good enough, while `if (x instanceof globalThis.Uint8Array)` causes the test to fail.
2. The `window.openpgp` (where we take the instance of OpenPGP.js) was undefined when loading the OpenPGP.js bundle in Firefox, so I changed it to `globalThis.openpgp`. This works in both Chrome and Firefox. So `globalThis` in Firefox is a sandbox, while in Chrome it's a window.
3. And the issue (that I dealt with earlier) https://github.com/openpgpjs/web-stream-tools/pull/23/files is that OpenPGP.js doesn't recognize its "own" Uint8Array type when running in a Firefox sandbox.

I just had another look at OpenPGP.js code (and there is `native Web Cryptography API` involved) and have this hypothesis (not tested yet):
As the "Permission denied" error was encountered in `serializeParams` function, probably these "bad" `window.Uint8Array` objects are returned from this Web Cryptography API, while `globalThis.Uint8Array` (sandboxed) my be created by the OpenPGP.js code itself. And the "constructor" permission denied error actually means that `subarray` is trying to create `window.Uint8Array` (same as itself) but as it is called by a sandbox only `globalThis.Uint8Array` are allowed to be created, hence the cul-de-sac.  We can try and file this as a bug to Firefox.

> Option 1: Can the library be patched to always create the same "good" type?

The code itself only has simple `new Uint8Array` calls, so the fact that it creates `globalThis.Uint8Array` is a consequence that the code is running in a sandbox. So this question should be rephrased as -- can and should we make the code run not in a sandbox in Firefox?  If not, we can probably try and update the library to re-initialize all Uint8Array's coming from Web Cryptography API (that is, convert window.Uint8Array to sandboxed globalThis.Uint8Array), but how to unit-test this? We need to handle each place where Uint8Array arrives from the API and make sure we handle each new such place.

> Or are you saying that both `globalThis.Uint8Array` and `window.Uint8Array` are "bad" and we need just `Uint8Array`? Where does that one come from if not global or window?

Apparently this is due to the complex Firefox process model (that I don't fully understand yet) https://wiki.mozilla.org/Security/Sandbox/Process_model
Looks like `globalThis.Uint8Array` (sandbox) is bearable in a sense that it can be dealt with patches like https://github.com/openpgpjs/web-stream-tools/pull/23/files but there may be problems in the code consuming these objects should they try to test their types.
 
> Option 2: Could the library be patched to accept a `config.Uint8ArrayClass = Uint8Array` and then that would be used throughout the codebase? That would be a larger change, presumably.

It's not an issue of mis-configuration.

> Option 3: Alternatively, on the "broken" type, could we monkey patch it from outside the library like this? Right in `requireOpenpgpjs`
> 
> ```ts
> globalThis.Uint8Array.prototype.subarray = function (i) {
>   return new Uint8Array(bin).subarray(i);
> }
> 
> window.Uint8Array.prototype.subarray = function (i) {
>   return new Uint8Array(bin).subarray(i);
> }
> ```

I can try this.

test/source/tests/content-script.ts
rrrooommmaaa: Because of these 2 exceptions?
image

extension/js/common/core/crypto/pgp/openpgp-key.ts
rrrooommmaaa: This works in both Chrome and Firefox:
```
    if (window !== globalThis) {
        window.Uint8Array.prototype.subarray = function (i) {
            return new globalThis.Uint8Array(this).subarray(i);
        };
    }
```
and this too:
```
    if (window !== globalThis) {
        window.Uint8Array.prototype.subarray = function (i) {
            return new Uint8Array(this).subarray(i);
        };
    }
```
we can also add `Catch.browser().name === 'firefox'` into the condition.

The following results in an infinite recursion in Chrome:
```
        window.Uint8Array.prototype.subarray = function (i) {
            return new Uint8Array(this).subarray(i);
        };
```

rrrooommmaaa: But may this patch possibly affect other scripts on the page? Is `window` of the content script isolated from the `window` of the page itself?

tomholub: Yes, gmail code and content script code is isolated indeed.

tomholub: > we can also add Catch.browser().name === 'firefox' into the condition.

yes. This looks good. If that's the case, we don't have to test Firefox so thoroughly, since it's a global fix, and we also wouldn't have to set up Firefox automated tests at this time.

rrrooommmaaa: > we don't have to test Firefox so thoroughly

There potentially might be issues with other objects than `Uint8Array`. Also we might need to patch other methods that return an array, e.g. `slice` and `filter`?

rrrooommmaaa: I tested `slice` and `filter` -- `slice` failed so I patched it too, `filter` is working without modification.

rrrooommmaaa: I also tested the "bad" Uint8Array as input to a function like `readBinary()` or `openpgp.readKey{ binaryKey: ... }`
When unpatched, the permission denied exception happens.
When patched everything is ok.
My concern is whether our prototype patch applies to all potential sources of `Uint8Array`s such as network, background page etc.?
Should we pre-emptively re-initialize all Uint8Array arguments of, say, `KeyUtil` and `MsgUtil` methods?

tomholub: >  Should we pre-emptively re-initialize all Uint8Array arguments of, say, KeyUtil and MsgUtil methods?

I wouldn't. I'd just go with the patches that appear to work, do some manual testing on Firefox, release it, wait for any reports and then fix those (which should be coming from less common workflows).

We only probably have 5k users on Firefox, as opposed to hundreds of thousands on Chrome. Plus I treat Firefox as our beta channel. So I'm not too worried, and wouldn't want to over-focus on this.

extension/js/background_page/background_page.ts
```diff
@@ -60,8 +59,6 @@ console.info('background_process.js starting');
   BrowserMsg.bgAddListener('storeGlobalSet', (r: Bm.StoreGlobalSet) => GlobalStore.set(r.values));
   BrowserMsg.bgAddListener('storeAcctGet', (r: Bm.StoreAcctGet) => AcctStore.get(r.acctEmail, r.keys));
   BrowserMsg.bgAddListener('storeAcctSet', (r: Bm.StoreAcctSet) => AcctStore.set(r.acctEmail, r.values));
-  BrowserMsg.bgAddListener('processAndStoreKeysFromEkmLocally', processAndStoreKeysFromEkmLocally);
-  BrowserMsg.bgAddListener('getLocalKeyExpiration', getLocalKeyExpiration);
 
   // todo - when https://github.com/FlowCrypt/flowcrypt-browser/issues/2560
   //   is fixed, this can be moved to the gmail content script, and some may be removed
```
tomholub, line 64: Also `BrowserMsg.addPgpListeners();` could be removed from below, and added to content script message listener instead.

Then pgp_block can be updated to call content script instead of background script.

rrrooommmaaa: Should I do this in this PR?
I thought after #5022 we may no longer need some of these messages, e.g. `pgpMsgDecrypt` or `pgpMsgVerifyDetached`

tomholub: i see. then it doesn't have to be in this pr.

tomholub (commented):
Looks good on a high level, will wait for @sosnovsky to check

sosnovsky (approved):
Looks good for me too 👍

PR #5024, opened 2023-03-16
#5011 Fix: Image with CID (Content-ID) not displaying in pgp block

### LINKED ISSUE #5011: Image with CID (Content-ID) not displaying in pgp block
Here's image src format

\ 

image

ioanmo226:
This PR fixed image with CID (Content-ID) not displaying in pgp block issue

close #5011 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/js/common/platform/xss.ts
```diff
@@ -131,7 +131,7 @@ export class Xss {
           img.remove(); // just skip images
         } else if (!src) {
           img.remove(); // src that exists but is null is suspicious
-        } else if (imgHandling === 'IMG-TO-LINK') {
+        } else if (imgHandling === 'IMG-TO-LINK' && !img.classList.contains('cid_to_base64_img')) {
```
sosnovsky, line 134: I think we should have consistent behavior for inline images. As currently images with cid will be shown immediately on load, while other inline images will need to be clicked before showing (with `IMG-TO-LINK` image handling)

sosnovsky (commented):
Implementation looks good 👍

Need also test how this cid images are converted when creating reply or forwarding message (they should be replaced just with their names)

extension/js/common/platform/xss.ts
ioanmo226: I was also considering that. Should we display the 'show inline' button and then show the image when the user clicks on it? Or should we simply show the inline image immediately? I don't think there are any security issues with showing the inline image right away, though.

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -319,6 +321,50 @@ export class PgpBlockViewRenderModule {
...
+
+      // Check if the src attribute contains a CID
+      if (matches && matches[1]) {
+        const contentId = matches[1];
+        const contentIdAttachment = attachments.find(attachment => attachment.cid === `<${contentId}>`);
+
+        // Replace the src attribute with a base64 encoded string
+        if (contentIdAttachment) {
+          inlineCIDAttachments.push(contentIdAttachment);
+          node.setAttribute('src', `data:image/png;base64,${contentIdAttachment.getData().toBase64Str()}`);
```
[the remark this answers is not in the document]
ioanmo226: Sounds good 👍

extension/js/common/platform/xss.ts
sosnovsky: Yes, I think showing inline images immediately without additional button will be more convenient

martgil:
> I need a hand creating a test email for the UI testing, specifically one that includes an "img src" with a content ID.

@ioanmo226 sir, here are the email for test with CID images.

https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/KtbxLrjCNLMzfJBPGVJkSrFKTCGQfvFRkL - unencrypted email.
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGslbKpLhNpDphbRqtNRQgxkXNv - encrypted mime message.

extension/js/common/platform/xss.ts
ioanmo226: But then do we need `IMG-TO-LINK` type?
Do we have to remove this logic and update UI test too?

https://github.com/FlowCrypt/flowcrypt-browser/blob/6ff785cc55d3f892ffdea8609871628f8cfb2884/extension/js/common/platform/xss.ts#L9

sosnovsky: Seems like `IMG-TO-LINK` isn't needed after this change.

And we're already showing remote images immediately after load, so this update will make all images to behave the same - inline and remote.

ioanmo226: Yeah, I also agree

ioanmo226: But one thing I'm unsure is that `show image` button was customer request before then.

ioanmo226: Maybe @tomholub, I think you can confirm why we displayed `show image` button before.

sosnovsky: Yeah, I was thinking about it, but probably it was for security reasons (as for remote images).
But as we allow to load remote images now, then it won't be an issue to have inline images to be auto-showed too

ioanmo226: Makes sense

ioanmo226:
@sosnovsky Could you explain more?
For example, what should be reply/forward content for this sample email.
> Need also test how this cid images are converted when creating reply or forwarding message (they should be replaced just with their names)

https://mail.google.com/mail/u/0/#inbox/FMfcgzGslbKpLlkkQNMnMhHszzTxCRpk

ioanmo226:
You mean image name or cid link?

ioanmo226:
We show remote image url when replying/forwarding though for remote images.

sosnovsky:
It should look like other inline images - just image name (without url)

Screenshot 2023-03-17 at 20 52 50

ioanmo226:
Got it

ioanmo226:
@sosnovsky Ready for reveiw!

sosnovsky (approved):
Works well now 👍

PR #5042, opened 2023-03-27
#5017 Highlight development build in browser

### LINKED ISSUE #5017: Highlight development build in browser
from https://github.com/FlowCrypt/flowcrypt-browser/pull/5007#issuecomment-1469553844

It should be easier to find that browser currently uses development build:
- add `-dev` suffix to version in status bar

- check if we can have separate icon when using development build

ioanmo226:
This PR highlights development build in browser

close #5017 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/chrome/settings/index.ts
+++ b/extension/chrome/settings/index.ts
@@ -69,7 +69,8 @@ View.run(
     }
 
     public render = async () => {
-      $('#status-row #status_version').text(`v:${VERSION}`);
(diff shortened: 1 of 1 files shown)


ioanmo226:
As far as found, I think it's not possible to use separate icon for development build as manifest.json is static format.
> check if we can have separate icon when using development build

sosnovsky (approved):

PR #5044, opened 2023-03-28
Fix build on macOS

### LINKED ISSUE #5043: Build fails on macOS
After merging https://github.com/FlowCrypt/flowcrypt-browser/pull/5013 extension build fails on macOS with error
```
sed: 1: "s/^(\s*)return\x20Uint8 ...": bad flag in substitute command: 'm'
```

sosnovsky:
This PR fixes build on macOS

close #5043

### To be filled by reviewers

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -73,7 +73,7 @@ cp node_modules/@openpgp/web-stream-tools/lib/*.js $OUTDIR/lib/streams
 STREAMS_REGEX="s/'\.\/(streams|util|writer|reader|node-conversions)'/'\.\/\1\.js'/g"
 STREAMS_FILES=$OUTDIR/lib/streams/*
 # patch isUint8Array until https://github.com/openpgpjs/web-stream-tools/pull/23 is resolved
-ISUINT8ARRAY_REGEX="s/^(\s*)return\x20Uint8Array\.prototype\.isPrototypeOf\(input\);/\1return\x20Uint8Array\.prototype\.isPrototypeOf\(input\)\x20\|\|\x20globalThis\.Uint8Array\.prototype\.isPrototypeOf\(input\);/mg"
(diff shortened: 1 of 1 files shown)


scripts/build.sh
```diff
@@ -73,7 +73,7 @@ cp node_modules/@openpgp/web-stream-tools/lib/*.js $OUTDIR/lib/streams
 STREAMS_REGEX="s/'\.\/(streams|util|writer|reader|node-conversions)'/'\.\/\1\.js'/g"
 STREAMS_FILES=$OUTDIR/lib/streams/*
 # patch isUint8Array until https://github.com/openpgpjs/web-stream-tools/pull/23 is resolved
-ISUINT8ARRAY_REGEX="s/^(\s*)return\x20Uint8Array\.prototype\.isPrototypeOf\(input\);/\1return\x20Uint8Array\.prototype\.isPrototypeOf\(input\)\x20\|\|\x20globalThis\.Uint8Array\.prototype\.isPrototypeOf\(input\);/mg"
+ISUINT8ARRAY_REGEX="s/^(\s*)return\x20Uint8Array\.prototype\.isPrototypeOf\(input\);/\1return\x20Uint8Array\.prototype\.isPrototypeOf\(input\)\x20\|\|\x20globalThis\.Uint8Array\.prototype\.isPrototypeOf\(input\);/g"
```
sosnovsky, line 76: @rrrooommmaaa build on macOS failed because of `m` flag. I removed it and seems everything works well even without it

sosnovsky: I re-checked - files were not patched without `m` flag. 
But I updated `ISUINT8ARRAY_REGEX`, so it's not using `^`, and now files are patched.
What do you think about updated regex, is it ok?

rrrooommmaaa: I think it's ok to not use `^`,
It'd be nicer to leave "one or more whitespace" before `return` with `(\s+)` and `\1` in this case. 
What is wrong with `sed` on MacOS? Is it some old version? Can't it be updated?

PR #5045, opened 2023-03-28
#4605 Export test result to semaphoreci

### LINKED ISSUE #4605: Export test result to semaphoreci
I think we should implement `export test result to semaphoreci` feature like we did in iOS app.
Currently for browser extension, it's hard to find failed tests in semaphoreci if it fails.
I need to scroll up & down to check failed tests.
Here's reference.

https://docs.semaphoreci.com/essentials/test-summary/

cc: @tomholub

ioanmo226:
This PR implemented `Export test result to semaphoreci` feature

close #4605 // if this PR closes an issue

### To be filled by reviewers


package.json
```diff
@@ -79,7 +80,7 @@
     "test_async_stack": "node build/test/test/source/async-stack.js",
     "test_buf": "npx ava --timeout=3m --verbose --concurrency=10 build/test/test/source/buf.js",
     "test_ci_chrome_consumer_live_gmail": "npx ava --timeout=30m --verbose --concurrency=1 build/test/test/source/test.js -- CONSUMER-LIVE-GMAIL STANDARD-GROUP",
-    "test_ci_chrome_consumer": "npx ava --timeout=30m --verbose --concurrency=10 build/test/test/source/test.js -- CONSUMER-MOCK STANDARD-GROUP",
+    "test_ci_chrome_consumer": "npx ava --timeout=30m --verbose --concurrency=10 build/test/test/source/test.js -- CONSUMER-MOCK STANDARD-GROUP  --tap | tap-xunit > report.xml",
```
ioanmo226, line 83: @sosnovsky, hmmm... I'm currently investigating deeply why the --tap parameter isn't functioning.
Do you have any suggestions as to why this may be the case?
Your suggestion would be greatly appreciated.

sosnovsky: @ioanmo226 it worked for me when I moved `--tap` argument, like this:
`npx ava --timeout=30m --verbose --tap --concurrency=10 ...`

ioanmo226: Yes, that works!
I also attempted to move the `--tap` argument while including `| tap-xunit > report.xml`. That's why it failed on my end (tests all failed with this one).
Now, I might be able to generate an XML file from the TAP output and then send it to SemaphoreCI.

Thank you for your help.

ioanmo226 (commented):
https://github.com/FlowCrypt/flowcrypt-browser/blob/a5a6e78b5e947e4fdd7787a812b55bd43aded905/.semaphore/semaphore.yml#L46

Do we actually need this one? (I can see that it archives the chrome-extension.zip, but I can't find where we proceed with that file.)

@tomholub , @sosnovsky  Please check

ioanmo226:
Yeah, I agree

sosnovsky (commented):
Works well 👍

Can you please also add reporting to these unit tests:

Screenshot 2023-03-31 at 09 02 25

ioanmo226:
Sounds good. Sure

sosnovsky (approved):

PR #5067, opened 2023-04-05
#5021 Rewrite attester mock to use attester config

### LINKED ISSUE #5021: rewrite attester mock to use `mockapi.attesterConfig = LEGACY_GLOBAL_ATTESTER_MOCK_CONFIG`
> **Second wave of PRs**: allow local configuration of the mock APIs, but give every test the same configuration:
> 
> I would recommend to do this service by service, eg start with only Attester, then FES in another PR, etc.
> 
> ```ts
>   ava.default('some test', testWithBrowser(undefined, assignedPort, async (t, browser) => { 
>     const mockApi = new MockApi();
>     mockApi.fesConfig = LEGACY_GLOBAL_FES_MOCK_CONFIG;
>     mockApi.ekmConfig = LEGACY_GLOBAL_EKM_MOCK_CONFIG
>     mockApi.attesterConfig = LEGACY_GLOBAL_ATTESTER_MOCK_CONFIG;
>     mockApi.googleConfig = LEGACY_GLOBAL_GOOGLE_MOCK_CONFIG;
>     await mockApi.withMockedApis(assignedPort, async () => {
>       // run browser tests as before
> ```
> 
> This way, we ensure that all current schenanigans we do for tests can be configured using the new mechanism. But we don't yet do individual configuration per test, because it's a task on its own to figure out which tests need exactly what configuration.
>

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/4494#issuecomment-1471550025_

ioanmo226:
This PR rewrites attester mock to use attester config

close #5021 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


test/source/mock/attester/attester-endpoints.ts
```diff
@@ -435,3 +302,99 @@ NaqFqwp9jMCYEnjfM2VSYFTKFrQYm9zYUTIKnQxNtWC8Z85UKz2eMk/INukzWe7g
...
+  welcomeMessageEnabled?: boolean;
+}
+
+const LEGACY_GLOBAL_ATTESTER_MOCK_CONFIG: AttesterConfig = {
+  pubkeyLookup: {
+    'ci.tests.gmail@flowcrypt.test': {
+      pubkey: somePubkey,
+    },
+    'test.flowcrypt.pubkey.timeout@gmail.com': {
+      returnError: {
```
ioanmo226, line 323: @sosnovsky, to avoid further rewriting work, I would like to confirm with you regarding the AttesterConfig architecture.
I intend to use this architecture because we need to return the public key result for certain emails and return an error (like a timeout) for other emails. 
What do you think?
Let me know if you have better architecture in your mind.

```ts
pubkeyLookup?: Record<string, PubKeyLookUpResult>
interface PubKeyLookUpResult {
  pubkey?: string;
  returnError?: { code: number; message: string };
}

```

Instead of using this
```ts
pubkeyLookup?: Record<string, string>
returnError?: { code: number; message: string };

```

sosnovsky: Such architecture looks good 👍

ioanmo226: Sounds good!. Do you agree processing with above architecture?
Any other suggestions?

sosnovsky: Yes, let's use this one, looks simple and customizable for different test cases

ioanmo226:
I think this PR should be reviewed by everyone, as the current changes will affect future WKD, FES, and EKM refactors.

test/source/test.ts
```diff
@@ -66,8 +67,9 @@ const testWithBrowser = (
   return async (t: AvaContext) => {
     let closeMockApi: (() => Promise<void>) | undefined;
     if (isMock) {
-      const mockApi = await startMockApiAndCopyBuild(t);
-      closeMockApi = mockApi.close;
+      t.mockApi = await startMockApiAndCopyBuild(t);
+      t.mockApi.attesterConfig = ATTESTER_ACCOUNT_INIT_SETUP_MOCK_CONFIG;
```
tomholub, line 71: I would rather that we don't set it at all by default, and it gets set explicitly in tests where it's used.

If it's too much work at this stage to tailor the settings on each test, then that could be a follow-up PR. But This line should be somewhere in each `test(` every time it's needed, instead of here.

test/source/tests/compose.ts
```diff
@@ -2121,6 +2165,14 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       'auto-refresh expired key if newer version of the same key available',
       testWithBrowser('ci.tests.gmail', async (t, browser) => {
         // add an expired key manually
+        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
tomholub, line 2168: Is the non-null assertion needed here? Maybe we can modify the style to not use nulls and use undefined instead, then you won't need this line?

test/source/tests/compose.ts
```diff
@@ -2250,8 +2333,33 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+                code: Status.BAD_REQUEST,
+                message: 'Request timeout',
+              },
+            },
+            [recipients.cc]: {
+              pubkey: somePubkey,
+            },
+          },
+        };
+        const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
```
tomholub, line 2362: Love this, extremely clear imo.

tomholub (commented):
This looks great. Comments below.

test/source/tests/compose.ts
ioanmo226: `mockApi` is defined as optional here because we are extending `ExecutionContext`.
However, we should force unwrap `mockApi` because we know it won't be null as we set `mockApi` value in `testWithBrowser` function.
Therefore I think above eslint disable rule is required.

https://github.com/FlowCrypt/flowcrypt-browser/blob/aaabc42912f751a70a1f6537725571c13a465b34/test/source/tests/tooling/index.ts#L16

test/source/test.ts
ioanmo226: The `ATTESTER_ACCOUNT_INIT_SETUP_MOCK_CONFIG` configuration is useful/needed for setting up mock data for common accounts when we pass the `acct` variable to the `testWithBrowser` function, as the variable name suggests.

ioanmo226: This way we will have tons of this duplicated code in each test.
Is this fine?

`await BrowserRecipe.setUpCommonAcct(t, browser, acct);`

tomholub: Could be a follow-up PR, it's a larger change across the tests - just adding `const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);` into every test that needs it. Laborious but worth it imo, it will be consistent and explicit. And as you can see, more flexible for configuration. Attester is probably not the only config that will run into this issue, most mocks probably will. They will be needed during setup.

You actually wouldn't need to do this either yet:

```diff
const testWithBrowser = (
- acct: CommonAcct | undefined,
  cb: (t: AvaContext, browser: BrowserHandle) => Promise<void>,
  flag?: 'FAILING'
): Implementation<unknown[]> => {
```

Maybe just add a comment to `acct` that we plan to remove it. Then start passing `undefined` instead?

```diff
  test(
      'compose - expired can still send',
-     testWithBrowser('ci.tests.gmail', async (t, browser) => {
+     testWithBrowser(undefined, async (t, browser) => {
+     const acctEmail = '...';
        const expiredEmail = 'expired.on.attester@domain.com';
        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
        t.mockApi!.attesterConfig = {
          pubkeyLookup: {
+          // add required attester config here
            [expiredEmail]: {
              pubkey: expiredPubkey,
            },
          },
        };
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);
       const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
```

Up to you what you think will be less work. Can either rip the bandaid all at once (remove the `acct` argument and add `openSettingsLoginApprove` to all that used it), or do it gradually with `undefined`.

ioanmo226: Ok, maybe can remove the `acct` arg and add `await BrowserRecipe.setUpCommonAcct(t, browser, acct);`.

I think single `await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);` line isn't enough. Need to set passphrase depends on account.
Those all logics are already in `setUpCommonAcct` and I think we can just call `BrowserRecipe.setUpCommonAcct`

test/source/tests/compose.ts
tomholub: Alternatively, how about we just disable `no-non-null-assertion` for whole test folder?

test/source/test.ts
tomholub: Right.

tomholub: Can of course confirm with @sosnovsky in case he had some other idea or plans

test/source/tests/compose.ts
tomholub: It's a detail, in either case. Leave as is is also fine with me, just seems inconvenient to always have to worry about this.

test/source/mock/lib/api.ts
```diff
@@ -144,19 +146,24 @@ export class Api<REQ, RES> {
     if (!req.url) {
       throw new Error('no url');
     }
-    if (this.handlers[req.url]) {
+    const attesterHandler = getMockAttesterEndpoints(this.attesterConfig);
+    const newHandlers: Handlers<REQ, RES> = {
+      ...(attesterHandler as Handlers<REQ, RES>),
+      ...this.handlers,
+    };
```
sosnovsky, line 153: Can we move `getMockAttesterEndpoints` to [all-apis-mock.ts](https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/mock/all-apis-mock.ts#L29) where all mock endpoints are listed?

test/source/test.ts
sosnovsky: Removing `acct` parameter and running `await BrowserRecipe.setUpCommonAcct(t, browser, acct)` when needed sounds like a good choice for me 👍

test/source/tests/compose.ts
sosnovsky: I was thinking about possible improvement for this one too, probably we can move our properties from [AvaContext](https://github.com/FlowCrypt/flowcrypt-browser/blob/d85f2eac158df355a174e06d0689bf059b6fffc7/test/source/tests/tooling/index.ts#L8) to separate interface, like this:
```ts
interface AvaExecutionContext {
  retry?: true;
  attemptNumber?: number;
  totalAttempts?: number;
  attemptText?: string;
  extensionDir: string;
  urls: TestUrls;
  mockApi: Api<{ query: { [k: string]: string }; body?: unknown }, unknown>;
}

export type AvaContext = ExecutionContext<AvaExecutionContext>

// then it can be used in tests like this
async (t: AvaContext) => {
   t.context.mockApi.attesterConfig = { ... }
}
```

But need to check if it works

test/source/mock/lib/api.ts
ioanmo226: IMO, we can't as we have to use live `Api.attesterConfig` variable in `getMockAttesterEndpoints`

test/source/mock/keys-openpgp-org/keys-openpgp-org-endpoints.ts
```diff
@@ -1,7 +1,7 @@
 /* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
 
 import { HandlersDefinition } from '../all-apis-mock';
-import { somePubkey } from '../attester/attester-endpoints';
+import { somePubkey } from '../attester/attester-key-contstants';
```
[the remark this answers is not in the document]
ioanmo226: Yep!

test/source/mock/lib/api.ts
sosnovsky: Yeah, you're right 👍
But let's rename `newHandlers` to `allHandlers`, I think it'll be more appropriate name

tomholub (commented):
Ok with me, as long as I still get the type suggestions when typing

```ts
        t.mockApi.attesterConfig = {
          pubkeyLookup: ...
```

ioanmo226:
Yeah, auto-completion works well

test/source/mock/attester/attester-endpoints.ts
```diff
@@ -1,437 +1,87 @@
...
+          if (pubRes.delayInSeconds) {
+            await Util.sleep(pubRes.delayInSeconds);
+          }
+          return pubRes.pubkey;
+        }
+        throw new HttpClientErr('Pubkey not found', 404);
+      } else if (isPost(req)) {
+        oauth.checkAuthorizationForEmail(req.headers.authorization, emailOrLongid);
+        expect(body).to.contain('-----BEGIN PGP PUBLIC KEY BLOCK-----');
+        MOCK_ATTESTER_LAST_INSERTED_PUB[emailOrLongid] = body as string;
```
rrrooommmaaa, line 49: The tests will continue to interfere with each other via this global output constant.
Should the reference to the output be included in the config?

ioanmo226: Yes, I have also considered this issue but haven't found a suitable solution yet. At the moment, this POST method is only used in two places, and they are attempting with different email addresses, so it's not a problem for now. However, it could potentially become a problem in the future.

Could you please clarify what you mean by "reference to the output being included in the config"?

test/source/mock/lib/api.ts
```diff
@@ -144,19 +146,24 @@ export class Api<REQ, RES> {
     if (!req.url) {
       throw new Error('no url');
     }
-    if (this.handlers[req.url]) {
+    const attesterHandler = getMockAttesterEndpoints(this.attesterConfig);
+    const allHandlers: Handlers<REQ, RES> = {
+      ...(attesterHandler as Handlers<REQ, RES>),
```
rrrooommmaaa, line 151: Well, these `REQ` and `RES` keep popping up because this `Api` class was meant to be generic, but now this generic class is supposed to know about certain "attester"s. That's too specific and contradicts the initial design of this class.
So it'd be better for this class to query some other injected function or interface for additional handlers. This function or object can be injected in the constructor.

ioanmo226: If I understand correctly, you mean something like this?
Or you want something else?

```ts
interface AdditionalHandlerProvider<REQ, RES> {
  getHandlers(): Handlers<REQ, RES>;
}

public constructor(apiName: string, protected handlers: Handlers<REQ, RES>, protected urlPrefix = '', protected additionalHandlerProvider?: AdditionalHandlerProvider<REQ, RES>) {
}

// In chooseHandler funciton
  const additionalHandlers = this.additionalHandlerProvider?.getHandlers() || {};
  const allHandlers: Handlers<REQ, RES> = {
    ...additionalHandlers,
    ...this.handlers,
  };


class AttesterHandlerProvider<REQ, RES> implements AdditionalHandlerProvider<REQ, RES> {
  constructor(private attesterConfig: AttesterConfig) {}

  getHandlers(): Handlers<REQ, RES> {
    return getMockAttesterEndpoints(this.attesterConfig) as Handlers<REQ, RES>;
  }
}

```

rrrooommmaaa: Hmm... we can't pass it in the constructor, because the initialization is tucked too deep in `startAllApisMock`

It'd be better just to have one property e.g. `public configProvider: ConfigurationProviderInterface<REQ,  RES>` instead of `handlers`. This interface is what you described as `interface AdditionalHandlerProvider`

The class `ConfigurationProvider` (what you suggested as `AttesterHandlerProvider`) doesn't need to be generic, it should only support actual REQ and RES our handlers use, that is `{ query: { [k: string]: string }; body?: unknown }` and `unknown`, see https://github.com/FlowCrypt/flowcrypt-browser/pull/4952/files

We can initialize `ConfigurationProvider` with `AttesterConfig` etc. and wire it to the concrete instance of Api with `t.mockApi.configProvider = new ConfigurationProvider { attesterConfig: ... }` instead of current `t.mockApi.attesterConfig =`
Their types will match because the concrete `mockApi` types match our handlers and hence the ConfigurationProvider type too

test/source/mock/attester/attester-endpoints.ts
sosnovsky: What about not using separate `MOCK_ATTESTER_LAST_INSERTED_PUB` object, but add new keys directly to `attesterConfig.pubkeyLookup`, like this `attesterConfig.pubkeyLookup[emailOrLongid] = { pubkey: body as string };`?

Then in tests we can check value of `t.mockApi.attesterConfig.pubkeyLookup![recipient]`

test/source/mock/lib/api.ts
ioanmo226: In my opinion, I don't think this approach will be effective. If we only set up `mockApi.configProvider` during `Api` initialization, the `/attester/pub` and `/attester/ldap` endpoints will not receive live `attesterConfig` data.

For example, if a developer sets `t.mockApi.attesterConfig = x` at the beginning of a test, and then updates it to `t.mockApi.attesterConfig = y` during the test, the current configuration will not be reflected when the test calls `/attester/pub` or `/ldap`. Instead, these endpoints will only retrieve the previous value.

What do you think?

test/source/mock/attester/attester-endpoints.ts
ioanmo226: Pushed update. Please check

test/source/mock/lib/api.ts
rrrooommmaaa: I meant to not expose `t.mockApi.attesterConfig` any more.
Instead, 
`t.mockApi.configProvider = new ConfigProvider({ attesterConfig: x, ...}}`

`t.mockApi.configProvider = new ConfigProvider({ attesterConfig: y, ...}}`

or 
`t.mockApi.configProvider.attesterConfig = y;`

ioanmo226: Yep!. You are right. We don't need to expose `attesterConfig` anymore.

ioanmo226: I propose to do this in next PR as this PR is not small so that it will be easier for you to review.

ioanmo226:
Who is final reviewer?
I'll start work on #5077 and then #5076 after this PR is merged.

test/source/tests/compose.ts
sosnovsky: > Implemented this solution and pushed.
> Please check. Will revert changes if you are not okay

Let's revert it back to optional `mockApi` property, as it's not good practice to use `@ts-ignore`, can lead to other issues in the future. Later we can come back to this issue and try to find another solution.

test/source/test.ts
```diff
@@ -63,11 +64,14 @@ const testWithBrowser = (
   cb: (t: AvaContext, browser: BrowserHandle) => Promise<void>,
   flag?: 'FAILING'
 ): Implementation<unknown[]> => {
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
```
sosnovsky, line 68: let's revert `mockApi` back to optional property to not use `@ts-ignore`

test/source/tests/compose.ts
ioanmo226: Ok, how about we disable `no-non-null-assertion` rule? Adding `// eslint-disable-next-line @typescript-eslint/no-non-null-assertion` to every `t.mockApi.attesterConfig = ` doesn't seem like good idea too.

test/source/test.ts
```diff
@@ -199,6 +203,8 @@ test.after.always('send debug info if any', async t => {
   const debugHtmlAttachments = getDebugHtmlAtts(testId, mockApiLogs);
   if (debugHtmlAttachments.length) {
     console.info(`FAIL ID ${testId}`);
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
```
sosnovsky, line 207: let's revert `mockApi` back to optional property to not use `@ts-ignore`

test/source/mock/attester/attester-endpoints.ts
```diff
@@ -1,437 +1,88 @@
...
+          return pubRes.pubkey;
+        }
+        throw new HttpClientErr('Pubkey not found', 404);
+      } else if (isPost(req)) {
+        oauth.checkAuthorizationForEmail(req.headers.authorization, emailOrLongid);
+        expect(body).to.contain('-----BEGIN PGP PUBLIC KEY BLOCK-----');
+        if (attesterConfig.pubkeyLookup) {
+          attesterConfig.pubkeyLookup[emailOrLongid] = { pubkey: body as string };
+          return 'Saved'; // 200 OK
+        }
```
sosnovsky, line 50: probably it's possible that `attesterConfig.pubkeyLookup` will be `null`?
then we need `else`

```ts
if (attesterConfig.pubkeyLookup) {
   attesterConfig.pubkeyLookup[emailOrLongid] = { pubkey: body as string };
} else {
   attesterConfig.pubkeyLookup = { emailOrLongid: { pubkey: body as string }};
}
return 'Saved'; // 200 OK
```

test/source/test.ts
ioanmo226: > Ok, how about we disable no-non-null-assertion rule like @tomholub suggested? Adding // eslint-disable-next-line @typescript-eslint/no-non-null-assertion to every t.mockApi.attesterConfig =  doesn't seem like good idea too.

sosnovsky: Ok, let's use this solution for now

.eslintrc.js
```diff
@@ -89,6 +89,7 @@ module.exports = {
     '@typescript-eslint/typedef': 'off',
     '@typescript-eslint/unbound-method': 'error',
     '@typescript-eslint/unified-signatures': 'error',
+    '@typescript-eslint/no-non-null-assertion': 0,
```
sosnovsky, line 92: this will disable `no-non-null-assertion` for all our codebase, but we need to disable it only in `test/source` folder

sosnovsky (commented):
looks good, just small eslint change

.eslintrc.js
ioanmo226: Yep!. we don't need to disable `no-non-null-assertion` for all our codebase.
Just pushed fix

sosnovsky (approved):

rrrooommmaaa (commented):
Feel free to merge without my approval

Issue #2587, opened 2020-02-25
categorize: unknown err on firefox


tomholub:
Lately started seeing this on ff. If it affects chrome too, we should add a test

![image](https://user-images.githubusercontent.com/6306961/75272555-0ffc8480-57f6-11ea-9c25-4e182fa69f0a.png)

limonte:
I faced this issue, it was happening when typing symbols in the recipients input, the error was saying something about `dbIndex` being undefined, I thought it's related to my environment and didn't report it :/

I can't reproduce it anymore.

tomholub:
> typing

You can reproduce it by opening an email that has a public key attached to it as a file (in compatibility account, the `pubkey` label will get you some).

If you will have a time to look at it today, please do else I may look into it in a few hours.

tomholub:
I'll fix this, browser is blocking communication between content scripts and background page.

![image](https://user-images.githubusercontent.com/6306961/75344347-5b647080-5892-11ea-9567-15051c6f1da1.png)

To avoid converting data to and from blobs, I'll send it as string instead of a typed array.

PR #5069, opened 2023-04-05
GmailElementReplacer refactoring

### LINKED ISSUE #3944: Thunderbird signed + encrypted message - shows "Message Not Signed" in FlowCrypt [waiting for OpenPGP.js upgrade]
**UPDATE** Seems like openpgp.js issue. We will upgrade to v5 in the #3324.

While working on #3941 I sent 3 types of messages to `ci.tests.gmail@flowcrypt.dev`:

- signed html (all good)
- signed text (all good)
- signed + encrypted html (faced issues)

Here's how it looks like in Thunderbird:

CleanShot 2021-08-30 at 18 43 42@2x

And here's how it looks in FlowCrypt:

CleanShot 2021-08-30 at 18 44 22@2x

### LINKED ISSUE #4164: Re-fetching signed-only message from API doesn't refresh the fetched text
To be implemented after PR #4144 and possibly refactor the decrypt-render-signature-... module interaction

### LINKED ISSUE #4810: Test (some of) GmailElementReplacer tests in mock environment

### LINKED ISSUE #5022: Perform all the decryption/verification operations in the context of GmailElementReplacer
Don't rely on additional processing/message fetching inside `pgp_block.htm`

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-browser/issues/3448#issuecomment-1448967429_

rrrooommmaaa:
This PR is WIP

close #5022
close #3944
close #4164
close #4810

- Tests added or updated

### To be filled by reviewers


extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -520,7 +546,107 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
+      } else if (treatAs === 'privateKey') {
+        return await this.renderBackupFromFile(a, attachmentsContainerInner, msgElReference.msgEl);
+      } else if (treatAs === 'signature') {
+        // todo: generate frameId here, prevent frameId duplicates?
+        const { frameId, frameHtml: embeddedSignedMsgXssSafe } = this.factory.embeddedRenderMsg();
+        msgElReference.msgEl = this.updateMsgBodyEl_DANGEROUSLY(msgElReference.msgEl, 'set', embeddedSignedMsgXssSafe); // xss-safe-factory
+        const frameWindow = XssSafeFactory.getWindowOfEmbeddedMsg(frameId);
+        if (frameWindow) {
+          const renderModule = new RenderRelay(frameWindow);
+          this.processSignedMessage(msgId, renderModule, senderEmail).catch(Catch.reportErr); // todo: clear cached items
```
rrrooommmaaa, line 619: Here is a demo of the `GmailElementReplacer`-context signature verification I suggested.
It is currently triggered by an attachment that returns `treatAs()==='signature'` -- not an optimal way probably, but ok for the demo.

The idea is to insert a lightweight `pgp_render_block` frame instead of `pgp_block` that will only do rendering/printing, while decryption/verification is handled in a intermediate class `MessageRenderer` (probably should be renamed to `MessageProcessor`) that can be called by either `GmailElementReplacer` or `InboxActiveThreadModule`, so the logic from `PgpBlockViewDecryptModule` and `PgpBlockViewSignatureModule` etc. should be moved to `MessageRenderer`. I have done some work in this direction, but that's not easy as these modules are tightly coupled among other `PgpBlockView*Module`s.

Rendering -- The decryption/verification procedure sends notification to the iframe by means of `postMessage`:
- need to review security concerns (the receiving iframe can display a "signed" message sent by some other script, how dangerous is that?)
- probably need to stringify JSON and base64-encode with regard to #2587, test the message size limits
- messages may be lost if sent too early (before the iframe finished loading) -- so there is a need to queue messages until the iframe has signaled it's ready tor receive messages.

Are we on the same page here, @tomholub?
I can try and do `PgpBlockView*Module` decoupling in a separate PR to reduce the size of this one, if needed.
 cc @sosnovsky

sosnovsky: > need to review security concerns (the receiving iframe can display a "signed" message sent by some other script, how dangerous is that?)

what about adding some token with initial request, which will be sent back along with `RenderMessage`? then receiving frame will be able to check if received message is the right one.

overall direction looks good for me, new implementation looks more straightforward than the current one with multiple `pgp_block`

> I can try and do PgpBlockView*Module decoupling in a separate PR to reduce the size of this one, if needed.

It'll be good if possible and won't affect functionality in this PR

tomholub: as long as by "other scripts" you mean other pages of our own extension, then that's fine. doesn't need to be addressed. 

of course if it was handling requests sent by software we didn't write (eg other extensions) that would be a big problem.

tomholub: for the rest, i'd roma says it's good, then i'm good

rrrooommmaaa: > as long as by "other scripts" you mean other pages of our own extension, then that's fine. doesn't need to be addressed.
> 
> of course if it was handling requests sent by software we didn't write (eg other extensions) that would be a big problem.

The receiver can check `event.origin`, and it is something like `mail.google.com`, so other extensions could possibly "impersonate" our content script if only they send a message from their content script installed to `mail.google.com` as well.

rrrooommmaaa: > what about adding some token with initial request,

I guess the URL (query string) with which the frame was created is visible to every script having access to DOM, no?

It may be possible to include a sort of "public key" in the frame query string, so the frame can check the "signature" of messages from the content script, I'm wondering what overhead would be and how to accomplish this since we don't want to load OpenPGP.js into each frame.

tomholub: let's not over complicate this

if today the browser extension actually responds to messages sent from order extensions, then that's a HUGE problem and needs to be addressed by somehow forbidding that, in manifest or by always checking the origin. regardless of this PR. 

if today that does not happen, then none of this is a problem and we can proceed without further checks

test/source/tests/gmail.ts
```diff
@@ -144,23 +144,13 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
...
-        await gmailPage.page.setOfflineMode(true); // go offline mode
-        await gmailPage.press('Enter'); // open the message
-        const pgpBlockFrame = await gmailPage.getFrame(['pgp_block.htm']);
+        const expectedMessage = { content: ['this should decrypt even offline'], signature: 'signed', encryption: 'encrypted' };
+        await BrowserRecipe.pgpBlockCheck(t, await gmailPage.getFrame(['/chrome/elements/pgp_render_block.htm'], { timeout: 10000 }), expectedMessage);
+        await gotoGmailPage(gmailPage, '/FMfcgzGkbDRNgcQxLmkhBCKVSFwkfdvV'); // plain convo
         await gmailPage.page.setOfflineMode(true); // go offline mode
-        await pgpBlockFrame.frame.goto(pgpBlockFrame.frame.url()); // reload the frame
-        await pgpBlockFrame.waitForContent('@pgp-block-content', 'this should decrypt even offline');
+        await gmailPage.page.goBack({ waitUntil: 'load' });
```
rrrooommmaaa, line 152: What was the idea behind offline decryption?
To gracefully handle the situation when the page goes offline and the user presses `Back` button to see the previously loaded (cached) message?
I reworked the test for this scenario.
Right now it works ok as the message processing result is cached in some way in `GmailElementReplacer`, there was a leftover invisible pgp block iframe, so a little patch was required as the frameId was also inadvertently cached.
https://github.com/FlowCrypt/flowcrypt-browser/pull/5069/commits/52b13ee93ce099e5a044fd4c7ce28a2ccaf62b13

Should we support signed-only messages too (where we pull `raw` message from API)?
@tomholub @sosnovsky

tomholub: > What was the idea behind offline decryption?

Gmail webpage is actually sort of a web app that can actually work offline. Like, it will remember previous emails and you can open it on an airplane and view them, kind of like outlook.

No API calls of course, and no attachments.

So when the PGP message is rendered in the message body, we would like to still be able to render it even when offline.

test/source/tests/decrypt.ts
```diff
@@ -210,18 +230,22 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
...M%0AgFhhw%2FlxHig9iuNwO8GE0HBDmRv%2BHKxLXue0pHPWt9Ut%2FmY4r%2F%2BruXloxRsU8gjG%0AhzSamV5IdvfQy0xCog2bWDCL4rCngh1IkrFi0CtNNl1mPskhoZqMZjso290yNaUc%0AHeFIdyPyvjjxyoHd9K3BuXx6fPvYbZzFRz9YikMqHxz6AyHAiMJnl8OPFH6XOTki%0AO1liU9LI%2BMsLCmeDqGliNap9VMvpBkJK6lWoC0RDtqHM48sI4BqHBgW6nUwnGv5H%0AtKbDTgFfMZw5c%2BklOWIUHME4eFNyRej69uoofFyb2rNjXBqvKlL2g0dUXbm2nmYG%0AUW4JPHWria6djv6zg0h037c%2FP6%2BDhVdm2O8in8b%2BBgqsdr7ChYPp2jUX8rouFNwd%0AU4xAXYo7iLoDN7AXHUb%2BG19qrx3c%2FXBrb5msVllfjDfKspX9ftTBukl1%2FJv2QdE0%0AG0kEVzAVB3amt9KHNX%2BfMC28rBla60gfxmpEJ9Q7fZCAOqTJPPcS1D9AKVm3wpoh%0ANWNJXYstblVGNBGuYeJuvHyjcGfs23RPgy1PI%2FAqJJcumUCcGb8Aa4BufsyZLw4L%0ACkN6yaCuw5DdmeNklkm%2FNVDJJJpvkLYrTRr6V5VIeO1usvmTYwAg5301DjG%2FI6qP%0AVRJQ7GeUXB9G6r6g15KbNfIALz0SUt6wKrFn6H39aVlBXzFO5Y6EmmD%2BYapfJH0o%0AhOHiIbWHXqU9rPToLC2Pn9WDq9FIUMx%2B0wL0En172e2%2BUOfEoPrcoykVFXemGCVp%0AIbB9HsIUVrsYyqvs7HLQfMdR%2FSjw%2BsFilKzIIBFjgNWHiyOwshVMkHCUHohcw6mJ%0A3cx01xWgJaG42ggGSYvkb4BvZEfgKmslwIV79pbwQxuayKNNqCpPwHuqA1fdlZ2E%0AhR8Dn%2BShq2eP5lStlK0V%2BGYT9fBcfuqApdUKNmJA1Pks%2Bj9h0CzseWYA%0A%3DHRui%0A-----END%20PGP%20MESSAGE-----&msgId=1600f39127880eed&senderEmail=&isOutgoing=___cu_false___&acctEmail=flowcrypt.compatibility%40gmail.com',
```
rrrooommmaaa, line 220: With the refactoring done in this PR it no longer makes sense to test `pgp_block` with the encrypted message passed in the query string, so I refactored the tests to actually load the message into the mock gmail page. However, some messages are missing: `1600f39127880eed` and a number of others -- I temporarily commented out tests for them. I guess they may be somewhere in the old archives, like the ones we used to have in `flowcrypt-bin`.
@tomholub is it possible to get these files?

test/source/tests/decrypt.ts
```diff
@@ -610,15 +654,19 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
...
+        const { authHdr } = await BrowserRecipe.setUpCommonAcct(t, browser, 'compatibility');
+        await BrowserRecipe.pgpBlockVerifyDecryptedContent(
+          t,
+          browser,
+          '1622eaa286f90737',
+          {
+            content: ['This will will have a larger attachment below', 'image-large.jpg'],
+            encryption: 'encrypted',
+            signature: 'not signed',
+            expectPercentageProgress: true,
```
rrrooommmaaa, line 666: As decryption now happens in the content script context, the ajax request to download the large attachment is handed over to the background page as per #1597
Thus displaying progress will be more complicated. The background page may send progress callbacks in the form of browser messages. They may be handled by the pgp_block frame directly.
That is, the content script creates a pgp_block iframe (already implemented),
initiates data download in background and receives data (already implemented),
meanwhile the pgp block receives messages from the background to display progress updates. The content script should pass additional data to the background, like frameId.
Not too complicated, @tomholub ?

extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -172,12 +47,15 @@ export class PgpBlockViewRenderModule {
     });
   };
 
-  public renderContent = async (htmlContent: string, isErr: boolean) => {
+  public renderContent = (htmlContent: string, isErr: boolean) => {
+    /*
+    todo:
```
rrrooommmaaa, line 52: do we really need this flag: `successfully_received_at_leat_one_message`, @tomholub ?
it is used in some debug diagnostics?

tomholub: if you don't see clear use, then you can remove it.

test/source/tests/decrypt.ts
tomholub: i think it is what it is, seems ok

test/source/tests/decrypt.ts
tomholub: flowcrypt.compatibility email account will have them, probably

tomholub:
How is this looking?

rrrooommmaaa:
> How is this looking?

It's mostly done. Finishing the tests and final refactorings.

extension/js/content_scripts/webmail/setup-webmail-content-script.ts
```diff
@@ -433,7 +435,14 @@
         ppEvent,
         Catch.try(() => notifyExpiringKeys(acctEmail, clientConfiguration, notifications))
       );
-      await webmailSpecific.start(acctEmail, clientConfiguration, inject, notifications, factory, notifyMurdered);
+      const relayManager = new RelayManager();
+      window.addEventListener('message', e => {
```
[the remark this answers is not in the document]
rrrooommmaaa: fixed

extension/chrome/elements/pgp_render_block.ts
```diff
@@ -0,0 +1,66 @@
...
+  public setHandlers = () => {
+    /*
+    $('.pgp_print_button').on(
+      'click',
+      this.setHandler(() => this.renderModule.printPGPBlock())
+    );
+    */
+  };
+
+  private handleMessage = (event: MessageEvent<unknown>) => {
```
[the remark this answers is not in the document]
rrrooommmaaa: fixed

extension/chrome/elements/pgp_block_modules/pgp-block-error-module.ts
```diff
@@ -2,27 +2,24 @@
...
-  public constructor(private view: PgpBlockView) {}
+  public constructor(private view: PgpBaseBlockView) {}
 
-  public renderErr = async (errBoxContent: string, renderRawMsg: string | undefined, errMsg?: string) => {
+  public renderErr = (errBoxContent: string, renderRawMsg: string | undefined, errMsg?: string) => {
     this.view.renderModule.setFrameColor('red');
     this.view.renderModule.renderErrorStatus(errMsg || 'decrypt error');
     const showRawMsgPrompt = renderRawMsg ? '<a href="#" class="action_show_raw_pgp_block">show original message</a>' : '';
-    await this.view.renderModule.renderContent(`<div class="error">${errBoxContent.replace(/\n/g, '<br>')}</div>${showRawMsgPrompt}`, true);
+    this.view.renderModule.renderContent(`<div class="error">${errBoxContent.replace(/\n/g, '<br>')}</div>${showRawMsgPrompt}`, true);
```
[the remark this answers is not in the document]
rrrooommmaaa: This code accepts html code with possible links and `<button>` tags. The allowed origins are extension's URL and `mail.google.com` (that is the origin of the content script) -- so potentially another extension installed as content script to `mail.google.com` may send it.
Adding complex messages not containing injectable html would make the code much more complex. Should we ignore this alert? @sosnovsky @tomholub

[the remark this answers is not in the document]
rrrooommmaaa: (this is `.renderErr` handler previously called as `errorModule.renderErr(...)` )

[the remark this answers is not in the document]
sosnovsky: So another extension can send some message to FlowCrypt, which will trigger error alert with html code possibly containing some link?

extension/js/common/message-renderer.ts
```diff
@@ -0,0 +1,782 @@
...
+  ) => {
+    const { frameId, frameXssSafe } = loaderContext.factory.embeddedRenderMsg(type);
+    loaderContext.setMsgBody(frameXssSafe, 'set');
+    await this.relayAndStartProcessing(this.relayManager, loaderContext, loaderContext.factory, frameId, printMailInfo, senderEmail, cb);
+  };
+
+  private processMessageWithDetachedSignature = async (msgId: string, renderModule: RenderInterface, senderEmail: string | undefined) => {
+    try {
+      renderModule.renderText('Loading signed message...');
+      const raw = await this.msgGetRaw(msgId); // todo: can we try to restore the content attachment from 'full'?
```
rrrooommmaaa, line 685: It's not easy to implement the previous "Retry" feature (when failing to get 'raw' message for PGP/MIME signed message)
image

Should I implement or can we replace with a simpler UI?
@tomholub

extension/chrome/elements/pgp_block_modules/pgp-block-error-module.ts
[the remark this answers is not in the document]
rrrooommmaaa: I guess, arbitrary Javascript can also be injected too, e.g. as a handler for injected button

extension/js/common/message-renderer.ts
rrrooommmaaa: Or we can create pgp_block after loading 'raw' first and display error above the native gmail attachments like this
image

extension/js/common/message-renderer.ts
```diff
@@ -0,0 +1,782 @@
...
+          // todo: queue into a dictionary?
+          await PassphraseStore.waitUntilPassphraseChanged(this.acctEmail, result.needPassphrase);
+          renderModule.clearErrorStatus();
+          renderModule.renderText('Decrypting...');
+          result = await cb(renderModule, frameId);
+        }
+        // todo: wait for renderModule completion: the queue has been actually flushed,
+        // and then remove the frame from relayManager.frames
+        // Something like:
+        // await relayManager.waitForCompletion(renderModule)
```
rrrooommmaaa, line 543: I think it'll make sense to clean up resources connected with a msgId as soon as all pgp_block frames are rendered (setTestState('ready') message is queued and queue is flushed).
We don't want the content-script internal variables to bloat, do we?

extension/js/common/message-renderer.ts
tomholub: Second option is ok, but while loading, there should be progress visible. It may take a very long time for a message to load, someone may otherwise thing the extension is broken.

extension/js/common/message-renderer.ts
tomholub: Bloat would indeed be problematic. It happened in the past, some people leave the tab open for months.

If this is difficult to implement, then a simple alternative would be to delete resources that are > 24 hours old.

extension/js/common/message-renderer.ts
```diff
@@ -0,0 +1,782 @@
...
+        undefined
+      );
+    }
+  };
+
+  private static btnHtml = (text: string, addClasses: string) => {
+    return `<button class="button long ${addClasses}" style="margin:30px 0;" target="cryptup">${text}</button>`;
+  };
+
+  public renderMsg = ({ senderEmail, blocks }: { blocks: MsgBlock[]; senderEmail?: string }, showOriginal: boolean) => {
```
rrrooommmaaa, line 222: This class has the entry points for rendering to be re-used by both GmailElementReplacer and extensions's own inbox page.
Only a few methods are public:
- renderMsg
- processAttachment
- processInlineBlocks
- msgGetProcessed.

Some other methods may be grouped in separate classes, I believe.

Also, some methods from `Mime` class, I believe, should be moved here or into a separate static class, namely `processBody` and `processAttachments` (former `processDecoded`) as they go beyond Mime-related logic.

extension/js/common/message-renderer.ts
```diff
@@ -0,0 +1,782 @@
...
+import { ApiErr, AjaxErr } from './api/shared/api-error.js';
+import { isCustomerUrlFesUsed } from './helpers.js';
+
+export type ProccesedMsg = MimeProccesedMsg;
+
+export class MessageRenderer {
+  public readonly downloader: Downloader;
+  private constructor(
+    private readonly acctEmail: string,
+    private readonly gmail: Gmail,
```
rrrooommmaaa, line 42: I wonder if we should avoid using `Gmail` object directly, but delegate all download operations to `Downloader`.
`Downloader` provides some extra features e.g. caching.

extension/js/common/message-renderer.ts
rrrooommmaaa: > But while loading, there should be progress visible. It may take a very long time for a message to load, someone may otherwise think the extension is broken.

By means of a spinner?
I see it is currently broken. I posted #5196

tomholub: Some sort of indication that a message is loading, be it a spinner, percentage, or just a "Loading message..." text at a minimum.

extension/chrome/elements/pgp_block_modules/pgp-block-error-module.ts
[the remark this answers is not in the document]
sosnovsky: Sounds like a potential security issue, is there no way to disable messages from other extensions?
Or maybe we can allow html code only for our extension URL, while `mail.google.com` can send only plain text?

extension/js/common/message-renderer.ts
sosnovsky: It'll be nice to have all download logic in one `Downloader` class, can be improvement for another PR

extension/js/common/message-renderer.ts
rrrooommmaaa: > By the way, is there a lot of functionality left to be refactored in this PR?

No, all the current tests succeed, I'm looking at scenarios that weren't tested yet to see if anything is broken.
Mainly doing simplifications and adding tests.

extension/js/common/message-renderer.ts
rrrooommmaaa: Implemented like this:
image
This status is rendered for all attachments when they are being analyzed. The status remains unchanged when 'raw' message is fetched to process this signature.

Then, if network failure happens at this stage, the displayed status would be `Categorize: net err` (I wonder why such strange text was already in place)

extension/chrome/elements/pgp_block_modules/pgp-block-error-module.ts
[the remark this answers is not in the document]
tomholub: I still can't bring myself to believe that the browser extension would blindly accept messages from another extension. Have you tested this? I would expect this to have to be explicitly allowed, eg in the manifest.

If it is indeed the case, I find it very unlikely that this cannot be somehow limited when setting up the listener, or in manifest, etc.

test/source/tests/flaky.ts
```diff
@@ -94,7 +94,7 @@ export const defineFlakyTests = (testVariant: TestVariant, testWithBrowser: Test
     test(
       'setup - create key - with backup to inbox',
       testWithBrowser(async (t, browser) => {
-        const acctEmail = 'flowcrypt.test.key.new.manual@gmail.com';
+        const acctEmail = 'flowcrypt.test.key.new.manual.1@gmail.com';
```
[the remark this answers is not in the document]
tomholub: >  We should also configure "test strategy" behavior explicitly via the configuration in the beginning of each test (currently done in TestBySubjectStrategyContext)

Oh, definitely. And we don't need to call that `strategy` either. Some more sane name would do.

> As this PR enhances the mock google storage in a way that sent messages become visible to the extension, these tests start to interfere with each other if run under the same account -- the backup message is visible to other tests. > We should isolate the sent messages between the tests (and, if I understand @tomholub idea correctly), we should generally hide messages not related to the test.
> So is this part of https://github.com/FlowCrypt/flowcrypt-browser/issues/5142 or a separate task?

Generally, tests should be independent of one another.

So, of each test has their own configuration and their own mock, then each test should have their own mock data. Mock data of one mock instance should be independent from another mock instance.

Under those conditions, sent messages can be automatically visible, no problem imo.

I wouldn't fix it in this PR, unless this issue is preventing you from finishing the PR.

If Roma has any better idea, we could go with that instead.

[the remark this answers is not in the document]
sosnovsky: For sure, tests should have independent data, here we need to replace current `send strategy` implementation with something similar to mock tests implementation on iOS - https://github.com/FlowCrypt/flowcrypt-ios/issues/1704

It should be in separate PR, as the last part of https://github.com/FlowCrypt/flowcrypt-browser/issues/4494

extension/chrome/elements/pgp_block_modules/pgp-block-error-module.ts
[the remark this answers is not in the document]
sosnovsky: I found this Chrome documentation on extension messaging - https://developer.chrome.com/docs/extensions/mv3/messaging/#external

It says that for receiving messages from another extensions another method should be used - called `chrome.runtime.onMessageExternal.addListener`, so probably our current implementation with `chrome.runtime.onMessage.addListener` should get messages sent only from FlowCrypt extension?
Needs to be tested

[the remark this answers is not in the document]
rrrooommmaaa: `chrome.runtime` messaging can't be used to send a message from content script to another part of extension.
The current codebase is using the background page as a relay.
Since the background page is scheduled to be removed (in favor of service workers) as of #1509, this way of communication won't exist in the future.
So in this PR direct communication is implemented using basic Javascript `postMessage` mechanism, which only allows to check `event.origin`. And `event.origin` of a message sent from the content scrip is `https://mail.google.com`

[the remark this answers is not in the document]
rrrooommmaaa: Right, perhaps I was confused by a previous scenario, when messages were sent from iframes inside the content script:
```
When sending message from iframe within content script page (mail.google.com), the parent will NOT get such message directly, and it will only be delivered to background page, from where we have to relay it around.
```
We can indeed try sending messages from the content script to its inner iframes.
I'll test it shortly.

extension/chrome/settings/inbox/inbox-modules/inbox-active-thread-module.ts
```diff
@@ -5,24 +5,49 @@
...
 import { Xss } from '../../../../js/common/platform/xss.js';
 import { Browser } from '../../../../js/common/browser/browser.js';
 import { Attachment } from '../../../../js/common/core/attachment.js';
+import { LoaderContextInterface } from '../../../../js/common/loader-context-interface.js';
+
+class LoaderContext implements LoaderContextInterface {
+  public constructor(private readonly factory: XssSafeFactory, public renderedMessageXssSafe: string | undefined, public renderedAttachments: string[]) {}
+
+  public renderPlainAttachment = (a: Attachment) => {
+    // todo: render error argument
```
rrrooommmaaa, line 27: We can do this in a separate issue, as the `settings/inbox` page isn't a priority now.

extension/js/common/message-renderer.ts
```diff
@@ -0,0 +1,795 @@
...
+        // actual encrypted attachment - show it
+        loaderContext.prependEncryptedAttachment(a);
+        return 'replaced'; // native should be hidden, custom should appear instead
+      } else if (treatAs === 'encryptedMsg') {
+        await this.setMsgBodyAndStartProcessing(loaderContext, treatAs, messageInfo.printMailInfo, messageInfo.from?.email, (renderModule, frameId) =>
+          this.processCryptoMessage(a, renderModule, frameId, messageInfo.from?.email, messageInfo.isPwdMsgBasedOnMsgSnippet)
+        );
+        return 'hidden'; // native attachment should be hidden, the "attachment" goes to the message container
+      } else if (treatAs === 'publicKey') {
+        // todo - pubkey should be fetched in pgp_pubkey.js
```
rrrooommmaaa, line 331: @tomholub This comment is about 5 years old.
We still pass public keys `embeddedPubkey` and private keys `embeddedBackup` in the query string.
Should we do something about it?

tomholub: I don't really see any issue, I'm unsure why the comment.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -783,6 +607,25 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
 
+  // loaderEl is a loader reference in case we're processing an attachment
+  // todo: we could also re-use a common method like this in Inbox
+  private handleException = (e: unknown, loaderEl?: JQueryEl) => {
+    if (ApiErr.isAuthErr(e)) {
+      this.notifications.showAuthPopupNeeded(this.acctEmail);
+      loaderEl?.text('Auth needed');
+    } else if (ApiErr.isNetErr(e)) {
+      loaderEl?.text('Network error');
+      // todo:    } else if (ApiErr.isInPrivateMode(e)) {
```
rrrooommmaaa, line 618: Need to figure out where to put this `isInPrivateMode` detection, if still relevant

extension/js/common/message-renderer.ts
rrrooommmaaa: I removed it

extension/js/common/message-renderer.ts
rrrooommmaaa: Implemented 24 hours expiration

extension/js/common/message-renderer.ts
```diff
@@ -0,0 +1,817 @@
...
+    senderEmail: string | undefined,
+    cb: (renderModule: RenderInterface, frameId: string) => Promise<{ publicKeys?: string[]; needPassphrase?: string[] }>
+  ): Promise<{ processor: Promise<unknown> }> => {
+    const renderModule = relayManager.createRelay(frameId);
+    if (printMailInfo) {
+      renderModule.setPrintMailInfo(printMailInfo);
+    }
+    const processor = cb(renderModule, frameId)
+      .then(async result => {
+        const appendAfter = $(`iframe#${frameId}`); // todo: review inbox-active-thread -- may fail
```
rrrooommmaaa, line 580: Currently, `settings/inbox` rendering differs from `webmail` in a way that it first accumulates HTML strings and then adds them to the DOM. So if this is called by MessageRenderer.processAttachment, `iframe#${frameId}` may not yet exist in the DOM.
We should refactor inbox to act more like webmail, that is, first adding some placeholders to the page, and then modifying/appending to them.
Since `inbox` isn't a priority in this issue, this can be done in a separate PR.

sosnovsky:
@rrrooommmaaa I built extension and now can't open any encrypted message, got such errors:
```
Error: BrowserMsg(db) sendRawResponse::TypeError: Cannot read properties of undefined (reading 'readKeys')`
```
```
Could not open this message with FlowCrypt.
other: TypeCannot read properties of undefined (reading 'readMessage')
```

rrrooommmaaa:
> Could not open this message with FlowCrypt.
> other: Typeopgp is undefined

What browser is it?
Have you re-built in full? (with `npm install`)?

sosnovsky:
> Could not open this message with FlowCrypt.
> other: Typeopgp is undefined

This error happened in Firefox.
I tried in both - Chrome and Firefox, there were different errors for different messages. 
You can check attached logs from console - https://github.com/FlowCrypt/flowcrypt-browser/pull/5069#issuecomment-1597155051 from Firefox, https://github.com/FlowCrypt/flowcrypt-browser/pull/5069#issuecomment-1597157060 from Chrome

> What browser is it? Have you re-built in full? (with `npm install`)?

Yes, I ran `npm install` and then also tried to re-build with `npm run build`

sosnovsky:
I checked console and there is error even without opening any email, just after loading gmail page:
```js
SyntaxError: expected expression, got keyword 'return' [openpgp.js:4378:7](moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/lib/openpgp.js)
Loading FlowCrypt 8.4.8 [webmail_bundle.js:13150:29](moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js)
Error: Some keys could not be parsed
    parseAndCheckPrivateKeys moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:10025
    # Catch.onErrorInternalHandler moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:85
    # Catch.reportErr moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:101
    # processKeysFromEkm moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:13364
[webmail_bundle.js:68:13](moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js)
Some keys could not be parsed
parseAndCheckPrivateKeys@moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:10025:19
[webmail_bundle.js:72:13](moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js)
TypeError: opgp is undefined
    parseMany moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:4561
    parseMany moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:3903
    parse moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:3893
    expirations moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:10126
    getLocalKeyExpiration moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:10126
    # Catch.onErrorInternalHandler moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:85
    # Catch.reportErr moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:101
[webmail_bundle.js:68:13](moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js)
opgp is undefined
OpenPGPKey.parseMany@moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:4561:24
KeyUtil.parseMany@moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:3903:33
KeyUtil.parse@moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:3893:32
getLocalKeyExpiration/expirations<@moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:10126:80
getLocalKeyExpiration@moz-extension://cc209f92-c631-4136-896d-fd5c240338aa/js/content_scripts/webmail_bundle.js:10126:47
```

sosnovsky:
Line with error in openpgp.js

```js
4377  function is_bytes(a) {
4378 (     return a instanceof Uint8Array ||      return a instanceof globalThis.Uint8Array);
4379  }
```

sosnovsky:
> So these are both on Mac?

Yes, that's on Mac, but there wasn't any errors during build. I can re-check later today which regexes from build script may fail on Mac

rrrooommmaaa:
That's because of greediness.

Should be `return (a instanceof Uint8Array || a instanceof globalThis.Uint8Array);` instead of `(     return a instanceof Uint8Array ||      return a instanceof globalThis.Uint8Array);`

rrrooommmaaa:
> That's because of greediness.

I push the lazy-matching fix. Can you build it again on Mac?

rrrooommmaaa:
This means the lazy modificator isn't supported. I guess, `/U` isn't supported either.
I'm a bit confused by this behavior as I specified the space character to be excluded with `[^\(\)\x20]`
We can try to replace `\x20` with `\s` if it helps.
But wouldn't be better to use a package with the same behavior on all platforms,
some package installable with `npm`?  e.g. https://www.npmjs.com/package/replace

sosnovsky:
I updated `ISUINT8ARRAY_REGEX3` to be 
```sh
ISUINT8ARRAY_REGEX3="s/return\x20([^\(\)\x20]+)\x20instanceof\x20Uint8Array;/return\x20\(\1\x20instanceof\x20Uint8Array\x20\|\|\x20\1\x20instanceof\x20globalThis\.Uint8Array\);/g"
```
and it worked well, all occurrences in `openpgp.js` were replaced. Maybe let's just use this regex?

rrrooommmaaa:
> Maybe let's just use this regex?

Yes, we can make do with this regex.

sosnovsky (approved):
Great job, I haven't found any issues - everything works well. Maybe I missed something, as list of changes is very big, but if anything will be found - it can be fixed in another PR, this one can be merged so there won't be merge conflicts with other PRs.

Issue #5076, opened 2023-04-11
Remove `acct` varaiable from `testWithBrowser` and move setup account logic to each `tests`


ioanmo226:
Could be a follow-up PR, it's a larger change across the tests - just adding `const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);` into every test that needs it. Laborious but worth it imo, it will be consistent and explicit. And as you can see, more flexible for configuration. Attester is probably not the only config that will run into this issue, most mocks probably will. They will be needed during setup.

You actually wouldn't need to do this either yet:

```diff
const testWithBrowser = (
- acct: CommonAcct | undefined,
  cb: (t: AvaContext, browser: BrowserHandle) => Promise<void>,
  flag?: 'FAILING'
): Implementation<unknown[]> => {
```

Maybe just add a comment to `acct` that we plan to remove it. Then start passing `undefined` instead?

```diff
  test(
      'compose - expired can still send',
-     testWithBrowser('ci.tests.gmail', async (t, browser) => {
+     testWithBrowser(undefined, async (t, browser) => {
+     const acctEmail = '...';
        const expiredEmail = 'expired.on.attester@domain.com';
        // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
        t.mockApi!.attesterConfig = {
          pubkeyLookup: {
+          // add required attester config here
            [expiredEmail]: {
              pubkey: expiredPubkey,
            },
          },
        };
+      const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acctEmail);
       const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
```

Up to you what you think will be less work. Can either rip the bandaid all at once (remove the `acct` argument and add `openSettingsLoginApprove` to all that used it), or do it gradually with `undefined`.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/5067#discussion_r1162549426_

PR #5087, opened 2023-04-15
Removed global oauth variable #5086

### LINKED ISSUE #5086: Isolate OauthMock data for each test
Looks like tests interfere with each other by re-using the same global `oauth` object, possibly overwriting each other's tokens
image

rrrooommmaaa:
This PR removes the global `oauth` variable for better test isolation

close #5086 

- Internal test framework change

### To be filled by reviewers

### DIFF
--- a/test/source/tests/compose.ts
+++ b/test/source/tests/compose.ts
@@ -1352,26 +1352,20 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       'compose - expired can still send',
       testWithBrowser(async (t, browser) => {
         const acct = 'ci.tests.gmail@flowcrypt.test';
+        const expiredEmail = 'expired.on.attester@domain.com';
         t.mockApi!.configProvider = new ConfigurationProvider({
           attester: {
             pubkeyLookup: {
               [acct]: {
                 pubkey: somePubkey,
               },
-            },
-          },
-        });
-        await BrowserRecipe.setUpCommonAcct(t, browser, 'ci.tests.gmail');
-        const expiredEmail = 'expired.on.attester@domain.com';
-        t.mockApi!.configProvider = new ConfigurationProvider({
-          attester: {
-            pubkeyLookup: {
               [expiredEmail]: {
                 pubkey: expiredPubkey,
               },
             },
           },
         });
+        await BrowserRecipe.setUpCommonAcct(t, browser, 'ci.tests.gmail');
         const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compose');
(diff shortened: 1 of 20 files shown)


rrrooommmaaa:
I tucked `oauth` instance into `ConfigurationProvider` class to nicely wire handlers.
However, the same `ConfigurationProvider` object should be used throughout the test to not lose the tokens.
So I modified the tests that use log in provided by `testWithBroswer` until #5076 is fixed.

test/source/tests/decrypt.ts
```diff
@@ -1366,15 +1347,13 @@ XZ8r4OC6sguP/yozWlkG+7dDxsgKQVBENeG6Lw==
     test(
       `verify - sha1 shows error`,
       testWithBrowser('compatibility', async (t, browser) => {
-        t.mockApi!.configProvider = new ConfigurationProvider({
-          attester: {
-            pubkeyLookup: {
-              'sha1@sign.com': {
-                pubkey: sha1signpubkey,
-              },
+        t.mockApi!.configProvider!.config!.attester = {
```
ioanmo226, line 1350: @rrrooommmaaa, do you prefer setting options like this?
I prefer using the `ConfigurationProvider` constructor and passing the config there.

This is because in the near future, we will also have to pass ekm, wkd, and other configs to configProvider, and if 
we don't pass it to constructor we would end up force-unwrapping `configProvider!.config!.objectToConfig` for each config, which is not ideal.

test/source/tests/compose.ts
```diff
@@ -2296,30 +2288,28 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
-        t.mockApi!.configProvider = new ConfigurationProvider({
-          attester: {
-            pubkeyLookup: {
-              [recipients.to]: {
-                pubkey: somePubkey,
-              },
-              [recipients.bcc]: {
-                pubkey: somePubkey,
-              },
+        t.mockApi!.configProvider!.config!.attester = {
```
ioanmo226, line 2291: I propose merging this pull request after #5076.
#5076 is expected to be ready today, and once it is merged, we can revert the changes made in this pull request (https://github.com/FlowCrypt/flowcrypt-browser/pull/5087/commits/37c2fa400800cc1a318cc1dde19432d2919c1026) and close it.

sosnovsky: Sounds reasonable, let's wait for finishing https://github.com/FlowCrypt/flowcrypt-browser/pull/5083, as it part of our bigger test refactoring, then it probably should fix this issue too.

ioanmo226 (approved):
Looks good on my side. @sosnovsky You may merge PR if you also think it's good.

sosnovsky (approved):
Good for me too 👍

Issue #4955, opened 2023-02-13
Use --concurrency param for browser pool


sosnovsky:
Shouldn't browser pool and concurrency always be exactly equal? Or is there a reason to have these two numbers diverge? I'd imagine, we could remove special code to choose that `FLAKY-GROUP` has 1 in the pool, and instead follow whatever the provided `--concurrency` ?

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4928#discussion_r1104402026_

sosnovsky:
It seems we can remove `--pool-size` parameter and just use `--concurrency` for setting browser pool size

tomholub:
provided it works well, yes :-)

sosnovsky:
I think it should work well, as currently each test uses separate mock api, so they shouldn't interfere.
We can also try to increase concurrency value for flaky group, to check if they'll fail or still pass.

sosnovsky:
Looks like we'll need to use separate values for `--concurrency` and `--pool-size`, as `--concurrency` param of `ava` means how many test files (not tests) will be executed simultaneously
(found by @IvanPizhenko at https://github.com/FlowCrypt/enterprise-server/issues/4807)

sosnovsky:
Hi @ioanatflowcrypt, can you please check if this issue is relevant or can be closed, thanks!

PR #5088, opened 2023-04-16
Speed up and improve live tests

rrrooommmaaa:
This PR speeds up live gmail tests by testing iframes directly,
also makes more strict checks that no leftover original attachment remain

issue #4929 

- Internal test improvements

### To be filled by reviewers

### DIFF
--- a/test/source/tests/gmail.ts
+++ b/test/source/tests/gmail.ts
@@ -179,14 +179,14 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
         await gotoGmailPage(gmailPage, '/QgrcJHrtqfgLGKqwChjKsHKzZQpwRHMBqpG');
         const urls = await gmailPage.getFramesUrls(['/chrome/elements/pgp_block.htm'], { sleep: 10, appearIn: 25 });
         expect(urls.length).to.equal(1);
-        const params = urls[0].split('/chrome/elements/pgp_block.htm')[1];
-        await BrowserRecipe.pgpBlockVerifyDecryptedContent(t, browser, {
-          params,
+        const pgpBlockFrame = await gmailPage.getFrame(['pgp_block.htm']);
+        await BrowserRecipe.pgpBlockCheck(t, pgpBlockFrame, {
           content: ['test that content from msg.asc renders'],
           encryption: 'encrypted',
           signature: 'not signed',
         });
         await pageHasSecureReplyContainer(t, browser, gmailPage);
+        expect(await gmailPage.isElementVisible('div.aQH')).to.equal(false); // original attachment container(s) should be hidden
       })
     );
 
@@ -201,9 +201,8 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
           appearIn: 25,
         });
         expect(pgpBlockUrls.length).to.equal(1);
-        const url = pgpBlockUrls[0].split('/chrome/elements/pgp_block.htm')[1];
-        await BrowserRecipe.pgpBlockVerifyDecryptedContent(t, browser, {
-          params: url,
+        const pgpBlockFrame = await gmailPage.getFrame(['pgp_block.htm']);
+        await BrowserRecipe.pgpBlockCheck(t, pgpBlockFrame, {
           content: ['1234'],
           encryption: 'not encrypted',
           signature: 'signed',
@@ -213,6 +212,7 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
         await testMinimumElementHeight(gmailPage, '.pgp_block.publicKey', 120);
         const pubkeyPage = await gmailPage.getFrame(['/chrome/elements/pgp_pubkey.htm']);
         await pubkeyPage.waitForContent('@container-pgp-pubkey', 'Fingerprint: 50B7 A032 B5E1 FBAB 24BA B205 B362 45FD AC2F BF3D');
+        expect(await gmailPage.isElementVisible('div.aQH')).to.equal(false); // original attachment container(s) should be hidden
       })
     );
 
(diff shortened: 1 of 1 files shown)


rrrooommmaaa:
@sosnovsky Why sometimes there is only one artefact when 2 tests are failed?
image

image

It generally looks as if gmail live tests run in parallel (with another instance of test suite?)
Strange interference that never happens when a single test is run.
This primarily concerns drafts, as drafts are deleted at the beginning of almost each test

sosnovsky:
> @sosnovsky Why sometimes there is only one artefact when 2 tests are failed?

Looks like we're storing only 1 file in Artifacts folder even when multiple are written:
Screenshot 2023-04-25 at 10 33 03

Should be fixed too, here is code in `semaphore.yml` - https://github.com/FlowCrypt/flowcrypt-browser/blob/master/.semaphore/semaphore.yml#L90

sosnovsky:
> It generally looks as if gmail live tests run in parallel (with another instance of test suite?)

It shouldn't be possible, I think, as live tests run only in `master` and they can't interfere.
What about running drafts test first and then all other tests?

sosnovsky:
> Do you know why we have pool of 3 browsers for gmail live tests? Does this make sense?

Not sure, maybe previously we ran live tests simultaneously, but for now we don't need these 3 browsers, just 1 will be enough.
There is a related task, where pool size should equal to `--concurrency` param value  - https://github.com/FlowCrypt/flowcrypt-browser/issues/4955

rrrooommmaaa:
> Not sure, maybe previously we ran live tests simultaneously, but for now we don't need these 3 browsers, just 1 will be enough.

Changing this value to 1 dramatically increased the run time to over an hour

sosnovsky:
> > Not sure, maybe previously we ran live tests simultaneously, but for now we don't need these 3 browsers, just 1 will be enough.
> 
> Changing this value to 1 dramatically increased the run time to over an hour

Oh, interesting, did you find what cause such time increase?

rrrooommmaaa:
> Oh, interesting, did you find what cause such time increase?

Apparently, the tests were run concurrently. I wonder why there is still a failure.
https://github.com/FlowCrypt/flowcrypt-browser/issues/5095#issuecomment-1523832506

sosnovsky:
@rrrooommmaaa I converted this PR to draft, as there were no activity here for a couple weeks.
Do you plan to work on this one after https://github.com/FlowCrypt/flowcrypt-browser/pull/5069?

rrrooommmaaa:
I think most of the updates are already incorporated into PR #5069.
After completion of it, I'll review this one, yes.

sosnovsky:
@rrrooommmaaa probably now you can work on this one, as live tests are still failing

rrrooommmaaa:
We can also use `getFrame` directly (#4929) here if we don't need `debug` flag

https://github.com/FlowCrypt/flowcrypt-browser/blob/cbb2aff53d2f894e41e7c22371103cb546eb172b/test/source/tests/page-recipe/gmail-page-recipe.ts#L14-L16

test/source/tests/gmail.ts
```diff
@@ -322,15 +345,16 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       )
     );
 
-    test(
+    // draft-sensitive test
+    test.serial(
```
rrrooommmaaa, line 349: It'll be run serially anyway. I can leave this `serial` as precaution or delete it

test/source/tests/gmail.ts
```diff
@@ -289,11 +311,12 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       )
     );
 
-    test(
+    // convo-sensitive, draft-sensitive test
+    test.serial(
```
rrrooommmaaa, line 315: It'll be run serially anyway. I can leave this `serial` as precaution or delete it

test/source/tests/gmail.ts
```diff
@@ -251,10 +271,12 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
         await pageHasSecureReplyContainer(t, browser, gmailPage);
         const pubkeyPage = await gmailPage.getFrame(['/chrome/elements/pgp_pubkey.htm']);
         await pubkeyPage.waitForContent('@container-pgp-pubkey', 'Fingerprint: 50B7 A032 B5E1 FBAB 24BA B205 B362 45FD AC2F BF3D');
+        expect(await gmailPage.isElementVisible('.aQH')).to.equal(false); // original attachment container(s) should be hidden
       })
     );
 
-    test(
+    // draft-sensitive test
+    test.serial(
```
rrrooommmaaa, line 279: It'll be run serially anyway. I can leave this `serial` as precaution or delete it

test/source/tests/gmail.ts
```diff
@@ -433,7 +457,8 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
       })
     );
 
-    test(
+    // convo-sensitive, draft-sensitive test
+    test.serial(
```
rrrooommmaaa, line 461: It'll be run serially anyway. I can leave this `serial` as precaution or delete it

.semaphore/semaphore.yml
```diff
@@ -7,6 +7,8 @@ agent:
 auto_cancel:
   running:
     when: branch != 'master'
+execution_time_limit:
+  hours: 1
```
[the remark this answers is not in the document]
sosnovsky: Let's leave it here, so it'll be easier to find in the future if we'll need to change this limit

test/source/tests/gmail.ts
sosnovsky: Is there another config somewhere for running these 4 tests serially?

sosnovsky:
> We can also use `getFrame` directly (#4929) here if we don't need `debug` flag
> 
> https://github.com/FlowCrypt/flowcrypt-browser/blob/cbb2aff53d2f894e41e7c22371103cb546eb172b/test/source/tests/page-recipe/gmail-page-recipe.ts#L14-L16

Is this `debug` flag used somewhere in tests or we can remove it without breaking tests? It'll be better to switch to using direct `getFrame` in #4929

test/source/tests/gmail.ts
rrrooommmaaa: all gmail live tests are run serially: --concurrency=1
https://github.com/FlowCrypt/flowcrypt-browser/blob/cbb2aff53d2f894e41e7c22371103cb546eb172b/package.json#L84

rrrooommmaaa:
> Is this `debug` flag used somewhere in tests or we can remove it without breaking tests?

Actually, as richtext option is disabled in Chrome, the `debug` flag is used to allow it in the test environment.
So I renamed the method to more descriptive `openSecureComposeWithRichTextWorkaround`

test/source/tests/gmail.ts
sosnovsky: We found that `--concurrency=1` is applied only to number of concurrent test files, but there can be more than 1 concurrent test - https://github.com/FlowCrypt/enterprise-server/issues/4807. So it'll be better to leave `test.serial`

sosnovsky:
> Actually, as richtext option is disabled in Chrome, the `debug` flag is used to allow it in the test environment. So I renamed the method to more descriptive `openSecureComposeWithRichTextWorkaround`

Then let's leave it for now, until we'll find another way to enable richtext option on Chrome for tests (or it'll be ready to be enabled by default on Chrome too)

PR #5097, opened 2023-04-18
#5076 Remove `acct` variable from `testWithBrowser` and move setup account logic to each tests

### LINKED ISSUE #5076: Remove `acct` varaiable from `testWithBrowser` and move setup account logic to each `tests`

ioanmo226:
This PR removed `acct` variable from `testWithBrowser` and move setup account logic to each tests

close #5076 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


sosnovsky (approved):
Well done 👍

Possible improvement - in most cases we just need 1 recipient with pubkey, but it needs full attester structure, like this:
```ts
t.mockApi!.configProvider = new ConfigurationProvider({
  attester: {
    pubkeyLookup: {
      [acct]: {
        pubkey: somePubkey,
      },
    },
  },
});
```

What about adding helper method:
```ts
const singlePubKeyAttesterConfig = (email: string, pubKey: string): AttesterConfig => {
  return {
    pubkeyLookup: {
      [email]: {
        pubkey: pubKey,
      },
    },
  };
};

// then it can be used like this
t.mockApi!.configProvider = new ConfigurationProvider({
  attester: singlePubKeyAttesterConfig(senderEmail, somePubkey),
});
```
Can be done in separate PR

ioanmo226:
Sounds good 👍

PR #5100, opened 2023-04-18
#5098 Simplify AttesterConfig for single pub key

### LINKED ISSUE #5098: Simplify AttesterConfig for single pub key
Possible improvement - in most cases we just need 1 recipient with pubkey, but it needs full attester structure, like this:
```ts
t.mockApi!.configProvider = new ConfigurationProvider({
  attester: {
    pubkeyLookup: {
      [acct]: {
        pubkey: somePubkey,
      },
    },
  },
});
```

What about adding helper method:
```ts
const singlePubKeyAttesterConfig = (email: string, pubKey: string): AttesterConfig => {
  return {
    pubkeyLookup: {
      [email]: {
        pubkey: pubKey,
      },
    },
  };
};

// then it can be used like this
t.mockApi!.configProvider = new ConfigurationProvider({
  attester: singlePubKeyAttesterConfig(senderEmail, somePubkey),
});
```
Can be done in separate PR

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-browser/pull/5097#pullrequestreview-1389404310_

ioanmo226:
This PR simplified AttesterConfig for single pub key

close #5098 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/tests/settings.ts
+++ b/test/source/tests/settings.ts
@@ -25,7 +25,7 @@ import { AvaContext } from './tooling';
 import { mockBackendData } from '../mock/backend/backend-endpoints';
 import { ClientConfiguration, keyManagerAutogenRules } from '../mock/backend/backend-data';
 import { ConfigurationProvider, HttpClientErr, Status } from '../mock/lib/api';
-import { somePubkey, testMatchPubKey } from '../mock/attester/attester-key-constants';
+import { singlePubKeyAttesterConfig, somePubkey, testMatchPubKey } from '../mock/attester/attester-key-constants';
 import { emailKeyIndex } from '../core/common';
 
 export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: TestWithBrowser) => {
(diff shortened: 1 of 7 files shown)


sosnovsky (approved):
Nice improvement 👍

Screenshot 2023-04-18 at 21 23 00

test/source/tests/settings.ts
```diff
@@ -60,13 +54,7 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
         const acct = 'flowcrypt.compatibility@gmail.com';
         t.mockApi!.configProvider = new ConfigurationProvider({
-          attester: {
-            pubkeyLookup: {
-              [acct]: {
-                pubkey: somePubkey,
-              },
-            },
-          },
+          attester: singlePubKeyAttesterConfig(acct, somePubkey),
```
rrrooommmaaa, line 57: `somePubkey` isn't related to `flowcrypt.compatibilty@gmail.com` account (or 'ci.tests.gmail` account), but it affects the tests because it may be looked up for signature verification. This complicates tests by adding an unnecessary pubkey into testing.
I see the pubkey returned by the attester in this (and a dozen of other) test only serves as a flag (yes/no). Still it'd be better if a pubkey related to the account is returned.
This isn't hard to implement. we can do a helper method like `setupCommonAcctWithAttester` -- it'll work in a similar way as the current version of `setUpCommonAcct`, but it can also parse the private key, convert it into pubkey with `KeyUtil.asPublicKey` and wire it to the attester config. Even less code in tests.
@sosnovsky ?

sosnovsky: Sounds good, thanks for proposal 👍

PR #5107, opened 2023-04-21
#5105 Migrate `FES` to use `ConfigurationProvider`

### LINKED ISSUE #5105: Rewrite `FES` mock to use `ConfigurationProvider`

ioanmo226:
This PR migrates `FES` to use `ConfigurationProvider`

close #5105 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


test/source/mock/fes/shared-tenant-fes-endpoints.ts
```diff
@@ -5,109 +5,154 @@ import { IncomingMessage } from 'http';
...
+      }
       return {
-        vendor: 'Mock',
-        service: 'external-service',
-        orgId: 'standardsubdomainfes.test',
-        version: 'MOCK',
-        apiVersion: 'v1',
+        clientConfiguration: {
+          flags: [],
+        },
```
ioanmo226, line 87: @sosnovsky, In `Attester`, we returned an error when `config` was not defined, as shown below.
However, for `FES`, I think it would be better to return an empty client configuration instead of throwing an error. Otherwise, we would have to add `clientConfiguration: { flags: [] }` to every test, which would be duplicative and not provide much benefit.

What are your thoughts on this?

```ts
throw new HttpClientErr(`Not implemented: ${req.method}`);
```

sosnovsky: Sounds good, let's do it this way here

ioanmo226: OK

sosnovsky (commented):
Nice improvement 👍

sosnovsky (approved):
👍

sosnovsky (approved):

PR #5125, opened 2023-04-30
issue #4862 add warning modal when downloading unencrypted attachment

### LINKED ISSUE #4862: warn user and ask to confirm before downloading dangerous file

martgil:
This PR adds a warning modal when downloading unencrypted file attachments on the FlowCrypt web extension page.

close #4862

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/attachment.ts
+++ b/extension/chrome/elements/attachment.ts
@@ -289,7 +289,9 @@ export class AttachmentDownloadView extends View {
       await this.recoverMissingAttachmentIdIfNeeded();
       await this.downloadDataIfNeeded();
       if (!this.isEncrypted) {
-        Browser.saveToDownloads(this.attachment);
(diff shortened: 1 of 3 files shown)


sosnovsky (approved):
Looks like all cases for dangerous attachments are handled now 👍

PR #5130, opened 2023-05-02
#5123 fix: get google token info issue

### LINKED ISSUE #5123: `Enable Google Contact search` is shown even after granting permissions
Customer reported that `Enable Google Contact search` button appears even when contact permissions are granted.
And after clicking on it contacts search still returns only FlowCrypt contacts, but not contacts from Google.

I reproduced this issue too - just try to enter contact name which doesn't exist (to have empty search results) and then click `Enable Google Contact search`

It seems this bug appeared in the latest 8.4.6 release.
Also should add test for this issue, so it won't reappear in the future.

ioanmo226:
This PR fixed get google token info issue

close #5123 // if this PR closes an issue

- Difficult to test (Mock test is already implemented at `compose - load contacts through API`. However this issue happened because https://gmail.googleapis.com endpoint is no longer available and it's difficult to simulate this situation in live test)

### To be filled by reviewers

### DIFF
--- a/tooling/build-types-and-manifests.ts
+++ b/tooling/build-types-and-manifests.ts
@@ -91,12 +91,6 @@ const updateEnterpriseBuild = () => {
       pattern: /const FLAVOR = 'consumer';/g,
       replacement: `const FLAVOR = 'enterprise';`,
     },
-    {
-      // for now we use www.googleapis.com on consumer until CORS resolved to use gmail.googleapis.com
-      // (on enterprise we already use gmail.googleapis.com)
-      pattern: /const GMAIL_GOOGLE_API_HOST = '[^']+';/g,
-      replacement: `const GMAIL_GOOGLE_API_HOST = 'https://gmail.googleapis.com';`,
-    },
   ];
   const constFilepaths = [`${buildDir(CHROME_ENTERPRISE)}/js/common/core/const.js`, `${buildDir(CHROME_ENTERPRISE)}/js/content_scripts/webmail_bundle.js`];
   for (const constFilepath of constFilepaths) {
--- a/extension/js/common/api/email-provider/gmail/google-auth.ts
+++ b/extension/js/common/api/email-provider/gmail/google-auth.ts
@@ -85,7 +85,7 @@ export class GoogleAuth {
   public static getTokenInfo = async (accessToken: string): Promise<GoogleTokenInfo> => {
     return (await Api.ajax(
       {
-        url: `${GMAIL_GOOGLE_API_HOST}/oauth2/v1/tokeninfo?access_token=${accessToken}`,
+        url: `${GMAIL_GOOGLE_API_HOST}/oauth2/v3/tokeninfo?access_token=${accessToken}`,
         timeout: 10000,
       },
       Catch.stackTrace()


tooling/build-types-and-manifests.ts
```diff
@@ -91,12 +91,6 @@ const updateEnterpriseBuild = () => {
       pattern: /const FLAVOR = 'consumer';/g,
       replacement: `const FLAVOR = 'enterprise';`,
     },
-    {
-      // for now we use www.googleapis.com on consumer until CORS resolved to use gmail.googleapis.com
-      // (on enterprise we already use gmail.googleapis.com)
-      pattern: /const GMAIL_GOOGLE_API_HOST = '[^']+';/g,
-      replacement: `const GMAIL_GOOGLE_API_HOST = 'https://gmail.googleapis.com';`,
```
ioanmo226, line 98: @tomholub @rrrooommmaaa 
Who can explain why we used `https://gmail.googleapis.com` instead of `googleapis.com` for enterprise builds?

tomholub: customer white listed the gmail subdomain in particular. 

not available? that sounds like a problem. roma should discuss this with our main customer

sosnovsky: I'm checking this one - it seems there is https://oauth2.googleapis.com/tokeninfo endpoint which can be used for checking oauth token, then customer will need to add just this domain, not full `googleapis.com`

sosnovsky: > wouldn't that mean our enterprise releases would be completely broken? unlikely, no? can you build and try the enterprise version locally? the one from master?

it seems only auth token endpoint is unavailable, as other functionality works well, and it affects only contacts search as FlowCrypt thinks that user didn't give permission to search contacts

tomholub: That means it was using the wrong endpoint subdomain the whole time for permissions.

Therefore this PR needs to be implemented to use the oauth2 subdomain for this call (at least on enterprise).

sosnovsky: Yes, `oauth2.googleapis.com` is already used in our code, so it should be whitelisted on customer side. 
Then we need just update our code to use https://oauth2.googleapis.com/tokeninfo endpoint for token check.
And I think we should use it for both - customer and enterprise.

sosnovsky: > // for now we use www.googleapis.com on consumer until CORS resolved to use gmail.googleapis.com

Not sure what CORS issue we've had there, but I think we should try `gmail.googleapis.com` for customer releases too, so endpoints will be consistent across customer and enterprise

tomholub: You can try. Last time it required a manifest update, which disables all extensions that are already installed.

sosnovsky: CORS error is still there and Google support says its expected behavior - https://issuetracker.google.com/issues/157312473#comment15, so it'll be ok to leave `googleapis.com` for customer releases for now.

ioanmo226: For this search contact`scope` issue, we may use `https://oauth2.googleapis.com/tokeninfo`.
But how about those?


https://github.com/FlowCrypt/flowcrypt-browser/blob/eac5ce47d5b9b59948dcbed134fb7dd1e525df23/extension/js/common/api/email-provider/gmail/gmail.ts#L192
https://github.com/FlowCrypt/flowcrypt-browser/blob/481f25d5c58bf46af657a96f1f37f0d4f30fa9ab/extension/js/common/api/email-provider/gmail/google.ts#L29
https://github.com/FlowCrypt/flowcrypt-browser/blob/481f25d5c58bf46af657a96f1f37f0d4f30fa9ab/extension/js/common/api/email-provider/gmail/google.ts#L32

ioanmo226: Hmmm... I can see that only `getTokenInfo` api is broken on `gmail.googleapis.com`

sosnovsky: Yes, other endpoints work well, only `getTokenInfo` request was affected

ioanmo226:
@sosnovsky Ready for review

PR #5132, opened 2023-05-03
#5131 fix: contact search result loading

### LINKED ISSUE #5131: Contacts search results sometimes not reliable
For example, in `flowcrypt.compatibility@gmail.com` account try to enter recipient starting with `vict` - google compose returns 2 contacts, while FlowCrypt search shows `No Contacts found`

Also sometimes searching for public keys takes too long and FlowCrypt search section shows `No Contacts found` while keys are searching:
1) I have `some@test.com` as contact
2) entering `some` as recipient in google compose returns `some@test.com` as search result
3) entering `some` as recipient in FlowCrypt compose shows `No Contacts found` as result, but I checked console and there were some requests searching for public key (https://keys.openpgp.org/vks/v1/by-email/some@test.com, https://flowcrypt.com/attester/pub/some@test.com ...). After they finished search results changed to `some@test.com`, but it took 7-8 seconds, I think.

So it seems search results are rendered only when public keys search is finished, but it can be confusing to users when search takes longer than usual. For this case let's render search results instantly and show some loader instead of `lock` icon, while public keys are searched.

ioanmo226:
This PR fixed contact search result loading for PGP key loading status

close #5131 // if this PR closes an issue

- Tests added or updated

chrome-capture-2023-4-4

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-recipients-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-recipients-module.ts
@@ -781,7 +781,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       for (const contact of renderableContacts) {
         ulHtml += `<li class="select_contact" email="${Xss.escape(contact.email.replace(/<\/?b>/g, ''))}">`;
         if (contact.pgpLoading) {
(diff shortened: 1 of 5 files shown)


sosnovsky (approved):
Nice improvement, works much better now 👍

PR #5143, opened 2023-05-08
#5136 Fix parsing multiple public keys from LDAP

### LINKED ISSUE #5136: Fix private key import with multiple LDAP public keys
> The ldap-relay returns two keys (one of them expired - not sure if relevant). The user tries to import a key during setup, it gets cross-checked in browser extension but it seems to cross-check the manually imported key only against one LDAP key of the two. Then it fails.
> 
> But it should successfully proceed if manually imported PRV key matches ANY of the LDAP public keys. Not just the first one, or not just any randomly chosen one - matching any of the LDAP keys must succeed.
> 
> Customer has SETUP_ENSURE_IMPORTED_PRV_MATCH_LDAP_PUB client configuration .

sosnovsky:
This PR fixes parsing of multiple public keys returned from LDAP.

close #5136

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/settings/setup.ts
+++ b/extension/chrome/settings/setup.ts
@@ -7,7 +7,7 @@ import { asyncSome, Url } from '../../js/common/core/common.js';
 import { ApiErr } from '../../js/common/api/shared/api-error.js';
 import { Assert } from '../../js/common/assert.js';
 import { Catch, CompanyLdapKeyMismatchError } from '../../js/common/platform/catch.js';
-import { KeyInfoWithIdentity, KeyUtil } from '../../js/common/core/crypto/key.js';
+import { Key, KeyInfoWithIdentity, KeyUtil } from '../../js/common/core/crypto/key.js';
 import { Gmail } from '../../js/common/api/email-provider/gmail/gmail.js';
 import { Google } from '../../js/common/api/email-provider/gmail/google.js';
 import { KeyImportUi } from '../../js/common/ui/key-import-ui.js';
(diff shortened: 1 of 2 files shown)


extension/chrome/settings/setup.ts
```diff
@@ -415,7 +415,10 @@ export class SetupView extends View {
       const result = await this.pubLookup.attester.doLookupLdap(this.acctEmail);
       if (result.pubkeys.length) {
         const prvs = await KeyStoreUtil.parse(await KeyStore.getRequired(this.acctEmail));
-        const parsedPubKeys = await KeyUtil.parseMany(result.pubkeys.join('\n'));
+        const parsedPubKeys: Key[] = [];
+        for (const pubKey of result.pubkeys) {
+          parsedPubKeys.push(...(await KeyUtil.parseMany(pubKey)));
+        }
```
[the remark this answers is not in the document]
sosnovsky: I thought about using `Promise.all`, but current code looks simpler and gives the same result.

test/source/tests/setup.ts
```diff
@@ -1009,6 +1009,44 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        t.mockApi!.configProvider = new ConfigurationProvider({
+          attester: {
+            ldapRelay: {
+              [acct]: {
+                pubkey: [expiredPubkey, hasPubKey].join('\n'),
+              },
+            },
+          },
+          fes: flowcryptTestClientConfiguration,
+        });
```
tomholub, line 1025: Love how clear this is

tomholub (approved):

test/source/tests/setup.ts
sosnovsky: Yeah, writing tests is much easier now

PR #5145, opened 2023-05-09
#5144 fix: can't open encryptd draft issue after passphrase expire

### LINKED ISSUE #5144: Can't open encrypted draft for non-encrypted thread
1) open thread with non-encrypted messages
2) click `Secure reply`
3) enter some reply text and save this draft
4) restart browser
5) open this thread
6) instead of draft there is `Waiting for pass phrase to open draft...` label

There should be `Enter passphrase` button below this label, so users will be able to enter passphrase for decrypting draft.

It happens only for cases when passphrase is stored only during current session (for example, for `@flowcrypt.com` users)

ioanmo226:
This PR fixed encrypted draft open issue when passphrase expired

close #5144 // if this PR closes an issue

- Difficult to test (Testing the process of opening an encrypted draft within a thread cannot be simulated in a mock test and should only be done in live tests. However, it can be challenging to reproduce the behavior of an expired passphrase during live testing. Additionally, testing with drafts within a thread can be unstable since they may be discarded at any time.)

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -363,11 +363,7 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
 
   private renderPPDialogAndWaitWhenPPEntered = async (longids: string[]) => {
     if (this.view.isReplyBox) {
-      const promptHtml = `
-        <div id="pgp_background" class="pgp_insecure">
-          <div id="pgp_error" data-test="pgp-error" class="pgp_badge red_label short" style="">pass phrase needed</div>
-          <div id="pgp_block" data-test="pgp-block-content"><div class="error"><a data-test="action-show-passphrase-dialog" class="action_open_passphrase_dialog" href="#">Enter passphrase</a> to open this draft.</div></div>
-        </div>`;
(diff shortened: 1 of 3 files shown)


sosnovsky (commented):
Looks nice 👍

Can we add check and show it just when there is no other message with `Enter passphrase` block? As currently we have almost identical blocks for encrypted message and draft:

Screenshot 2023-05-09 at 14 18 36

So it should work like this:
- if there is no other message requiring passphrase - show `Enter passphrase` button in draft block
- if there is encrypted message requiring passphrase, then show `Enter passphrase` button in message block, and `Waiting for pass phrase to open draft...` in draft block

ioanmo226:
Upon checking, it seems difficult to determine if there are other PGP messages in the thread. Since both messages and drafts are rendered simultaneously, attempting to check the PGP iframe in the draft render code may return 0 results, even if there are actually PGP messages present.

sosnovsky:
> Upon checking, it seems difficult to determine if there are other PGP messages in the thread. Since both messages and drafts are rendered simultaneously, attempting to check the PGP iframe in the draft render code may return 0 results, even if there are actually PGP messages present.

then let's simplify UI for draft, so it'll be a bit different from message `Enter passphrase` block:

Screenshot 2023-05-09 at 22 40 29

sosnovsky (approved):
👍

PR #5148, opened 2023-05-11
#5142 Migrate google mock endpoint to `ConfigurationProvider`

### LINKED ISSUE #5142: Rewrite `Google` mock to use `ConfigurationProvider`

ioanmo226:
This PR migrated google mock endpoint to `ConfigurationProvider`

close #5142 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/mock/google/google-endpoints.ts
+++ b/test/source/mock/google/google-endpoints.ts
@@ -60,6 +60,28 @@ export interface GoogleConfig {
   aliases?: Record<string, MockUserAlias[]>;
 }
 
+export const multipleEmailAliasList: MockUserAlias[] = [
+  {
+    sendAsEmail: 'alias1@example.com',
+    displayName: 'An Alias1',
+    replyToAddress: 'alias2@example.com',
+    signature: '',
+    isDefault: false,
+    isPrimary: false,
+    treatAsAlias: false,
+    verificationStatus: 'accepted',
+  },
+  {
+    sendAsEmail: 'alias2@example.com',
+    displayName: 'An Alias1',
+    replyToAddress: 'alias2@example.com',
+    signature: '',
+    isDefault: false,
+    isPrimary: false,
+    treatAsAlias: false,
+    verificationStatus: 'accepted',
+  },
+];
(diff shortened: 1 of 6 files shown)


test/source/mock/google/google-endpoints.ts
```diff
@@ -138,76 +150,52 @@ export const getMockGoogleEndpoints = (oauth: OauthMock): HandlersDefinition =>
...
+        isDefault: true,
+        isPrimary: true,
+        treatAsAlias: false,
+        verificationStatus: 'accepted',
+      };
+      // If the account is a compatibility account, return specific aliases.
+      // This account is used in hundreds of tests, so we handle it separately to avoid duplicating the code.
+      if (acct === 'flowcrypt.compatibility@gmail.com') {
+        const alias = 'flowcryptcompatibility@gmail.com';
+        return {
```
ioanmo226, line 172: @sosnovsky As you can see, when acct is `flowcrypt.compatibility@gmail.com`, I returned specific HARD_CODED aliases.
Do you agree with this or want me to duplicate below same configuration to every compatibility tests?

sosnovsky: It's ok to have it hardcoded here, no need for duplicated code

test/source/mock/google/google-endpoints.ts
```diff
@@ -81,55 +120,28 @@ export const getMockGoogleEndpoints = (oauth: OauthMock): HandlersDefinition =>
...
-              { person: { emailAddresses: [{ metadata: { primary: true }, value: 'testsearchorder6@flowcrypt.com' }] } },
-            ],
-          };
-        } else {
-          return empty;
-        }
-      } else {
-        return empty;
+      if (!config.contacts) {
+        return {};
```
sosnovsky, line 127: What if we'll return `{}` even when config isn't set? (and for `/othercontacts` too)
Then we won't need to declare empty config like we have in some tests `{ google: { contacts: [] } }`

sosnovsky (commented):
Well done 👍

sosnovsky (approved):
Nice 👍

PR #5155, opened 2023-05-15
#5146 fix: import multiple public keys issue

### LINKED ISSUE #5026: Failed to import keys with image data
Description: The FlowCrypt browser extension raises an error when importing a key with image data. Image data are stored in key users' attributes. 

The error shown is `TypeError: Cannot read properties of null (reading 'userID')` because the image data's userID is indeed null while the software is trying to read it.

image

sample public key with image data can be found at https://flowcrypt.com/attester/pub/A62F3CDFECC0926C

output from gpg:
```
pub  rsa4096/A62F3CDFECC0926C
     created: 2016-06-15  expires: never       usage: SC  
     trust: unknown       validity: unknown
sub  rsa2048/D14BB2C566698205
     created: 2016-06-15  expires: 2024-06-13  usage: SE  
sub  rsa2048/2D2EA188B5988036
     created: 2016-06-15  expires: 2024-06-13  usage: SA  
[ unknown] (1). User <user@example.com>
[ unknown] (2)  User <user@example.com>
[ revoked] (3)  User <user@example.com>
[ revoked] (4)  User <user@example.com>
[ unknown] (5)  [jpeg image of size 5473]
[ revoked] (6)  User <user@example.com>
```

Notice the 5th item from the users list. It is an image added via `addphoto` command by gpg and utilized by software like GPG Keychain for Mac - https://superuser.com/questions/336894/how-can-i-add-my-picture-to-my-public-key-using-gnupg

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGslbKpfTcJlrrsQzJnWJzgHQzn

### LINKED ISSUE #5146: Error during importing Public keys from a file
I'm working on https://github.com/FlowCrypt/flowcrypt-android/issues/1692. After trying to import the provided file with public keys I saw the following
image

This file contains a lot of public keys. Maybe we have to allow importing `good` keys and show a warning for `bad` keys.
@sosnovsky I will send you the file.

ioanmo226:
This PR fixed importing multiple public keys (which contains some good keys and some bad keys) issue 

close #5146 // if this PR closes an issue
close #5026 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


sosnovsky:
> @sosnovsky OpenPGP v5.9.0 is now live, and it addresses the issue of `Can't read properties of null reading userId`.

Great! Let's also include https://github.com/FlowCrypt/flowcrypt-browser/issues/5026 in this PR, as it should be addressed by this fix too.

ioanmo226:
OK

martgil:
Hello @ioanmo226, sure - please try this one: https://github.com/openpgpjs/openpgpjs/files/11319793/sample.pub.txt

ioanmo226:
Thank you @martgil

ioanmo226:
@sosnovsky As you can see from below comment we return empty `emails` and `identities` when `emails.length !== identities.length`.
As a result, `emails` value become empty array. I think we should remove this condition?

https://github.com/FlowCrypt/flowcrypt-browser/blob/994bf8d0a8b5ead3c1f713901de5ec99f67af757/extension/js/common/core/crypto/pgp/openpgp-key.ts#L550

sosnovsky:
> @sosnovsky As you can see from below comment we return empty `emails` and `identities` when `emails.length !== identities.length`. As a result, `emails` value become empty array. I think we should remove this condition?

As for me, it's ok to remove this condition. Let's also check with @rrrooommmaaa, as he wrote this code recently

ioanmo226:
@rrrooommmaaa You mean you want me to update logic something like this?
```ts
if (emails.length === identities.length || !key.isPrivate()) {
  // OpenPGP.js uses RFC 5322 `email-addresses` parser, so we expect all identities to contain a valid e-mail address
  return { emails, identities };
}
```

sosnovsky (approved):
Works well now 👍

PR #5158, opened 2023-05-17
#5157 Fix live test with inline signature

### LINKED ISSUE #5157: `mail.google.com - pubkey gets rendered with new signed and encrypted Thunderbird signature` live test fails

sosnovsky:
This PR updates test with inline signature - previously FlowCrypt didn't parse inline signatures, so it was expected to have `not signed` badge in this case. Now inline signatures are parsed and message have `signed` badge with valid inline signature.

close #5157

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -9,56 +9,56 @@ auto_cancel:
     when: branch != 'master'
 
 blocks:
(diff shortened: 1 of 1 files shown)


ioanmo226 (approved):
👍

PR #5160, opened 2023-05-18
issue #5159 use alternative html injection point for container

### LINKED ISSUE #5159: Gmelius rendering issue compatibility issue when using FlowCrypt simultaneously
Description: A Gmelius user reported that the software failed to render when using FlowCrypt simultaneously. I investigated the issue and found that Gmelius and FlowCrypt both tried to access the same element. Gmelius complained about not having the expected element because the FlowCrypt extension had taken over the slot for that element.

### Steps to reproduce the issue:

1. Install Gmelius and configure it for your Gmail account.
2. Go to Gmail and notice a new section:
image
3. Install FlowCrypt and configure it for your Gmail account.
4. Go back to Gmail, refresh the page, and notice that the new section is gone and an error on Web Console appears where it js complains about a problem in accessing an element.

One solution is to use another element container to inject the FlowCrypt secure compose button without compromising the UI itself. However, we may find the solution I have proposed to be inadequate in terms of its placement in the HTML structure.

I will publish a PR regarding this to visualize it.

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGsmhTPdszCbbJjXhGNWqcZFfTg

martgil:
This PR uses an alternative injection point for the FlowCrypt secure compose button in Gmail (mail.google.com) to fix the compatibility issue with Gmelius.

close #5159

- Not worth testing
- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/inject.ts
+++ b/extension/js/common/inject.ts
@@ -25,7 +25,7 @@ export class Injector {
   private S: SelCache;
   private container: { [key: string]: Host } = {
     composeBtnSel: {
-      gmail: 'div.aeN, div.aBO', // .aeN for normal look, .aBO for new look https://github.com/FlowCrypt/flowcrypt-browser/issues/4099
+      gmail: 'div.W8, div.aBO', // .aeN for normal look, .aBO for new look https://github.com/FlowCrypt/flowcrypt-browser/issues/4099
       outlook: 'div._fce_b',
       settings: '#does_not_have',
     },


sosnovsky:
@martgil after this change `Secure Compose` button isn't visible for me in Chrome and Firefox:

Screenshot 2023-05-18 at 18 40 36

sosnovsky:
I tried to install Gmelius extension and it works well for me along with FlowCrypt (on `master` branch)

Screenshot 2023-05-18 at 18 50 30

martgil:
Thanks, @sosnovsky looks like the issue isn't present on this particular Gmail UI. That means we can recommend the user to switch to this one which would let them use FlowCrypt and Gmelius simultaneously. Thanks for your help!

PR #5161, opened 2023-05-19
#5037 fix: incorrect modification detection when saving drafts

### LINKED ISSUE #5037: Incorrect modification detection when saving drafts
The failing use case that I saw:
1) open a new secure compose window
2) quickly type/paste something into the body before the first `draftSave()` is activated (`SAVE_DRAFT_FREQUENCY = 3000` ms)
3) the draft will not be saved

This causes some tests to periodically fail (e.g. `mail.google.com - saving and rendering compose drafts when offline`) and generally bad UX.

Reason: `ComposeDraftModule.hasBodyChanged()` relies on `lastDraftBody` property which is initialized as `undefined`
and it ignores the first updates (returns false)

ioanmo226:
This PR fixed incorrect modification detection when saving drafts issue

close #5037 // if this PR closes an issue

- Difficult to test (Couldn't always reproduce this issue because in automated test, it's hard to type text to email body in less than 3 seconds after frame load)

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -118,9 +118,9 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
     }
   };
 
-  public setLastDraftBody(draftBody: string) {
+  public setLastDraftBody = (draftBody: string): void => {
     this.lastDraftBody = draftBody;
-  }
+  };
 
   public draftSave = async (forceSave = false): Promise<void> => {
     if (this.disableSendingDrafts) {
--- a/extension/chrome/elements/compose-modules/compose-draft-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-draft-module.ts
@@ -29,7 +29,7 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
   private currentlySavingDraft = false;
   private disableSendingDrafts = false;
   private saveDraftInterval?: number;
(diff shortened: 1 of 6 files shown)


extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -337,11 +341,6 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
   };
 
   private hasBodyChanged = (msgBody: string) => {
-    if (this.lastDraftBody === undefined) {
-      // first check
-      this.lastDraftBody = msgBody;
-      return false;
-    }
```
sosnovsky, line 344: This check was used for existing drafts, after it's removing existing drafts are re-saved even without changes:
1) create and save draft
2) go to Drafts folder
3) open this draft
4) wait for 3 seconds (without any changes)
5) draft is saved and `last edit` time is changed

Existing draft shouldn't be re-saved when no changes applied.
Probably it's possible to create such ui test?

ioanmo226: I see. I thought adding this line of code would also cover this case but I added in wrong way. Just fixed it.
Let me add UI test.
https://github.com/FlowCrypt/flowcrypt-browser/blob/8da995814e20e971f61b211155b5dcf43ddf8a17/extension/chrome/elements/compose-modules/compose-input-module.ts#L47

sosnovsky (approved):
Works well now 👍

PR #5167, opened 2023-05-22
#5123 fix: get id token issue in acct store

### LINKED ISSUE #5123: `Enable Google Contact search` is shown even after granting permissions

ioanmo226:
This PR  fixed getting id token issue in acct store

close #5123 // if this PR closes an issue

- Difficult to test (This should be simulated in live test and hard to reproduce it)

### To be filled by reviewers

### DIFF
--- a/extension/js/common/platform/store/acct-store.ts
+++ b/extension/js/common/platform/store/acct-store.ts
@@ -142,7 +142,7 @@ export class AcctStore extends AbstractStore {
   };
 
   public static getScopes = async (acctEmail: string): Promise<Scopes> => {
-    const accessToken = await InMemoryStore.get(acctEmail, InMemoryStoreKeys.GOOGLE_TOKEN_ACCESS);
+    const accessToken = await this.getIdTokenUntilAvailable(acctEmail);
     // const { google_token_scopes } = await AcctStore.get(acctEmail, ['google_token_scopes']);
     const result: { [key in GoogleAuthScopesNames]: boolean } = {
       email: false,
@@ -175,6 +175,17 @@ export class AcctStore extends AbstractStore {
       }
     }
     return result;
+  };
+
+  private static getIdTokenUntilAvailable = async (acctEmail: string): Promise<string> => {
+    for (let i = 0; i < 20; i++) {
+      const accessToken = await InMemoryStore.get(acctEmail, InMemoryStoreKeys.GOOGLE_TOKEN_ACCESS);
+      if (accessToken) {
+        return accessToken;
+      }
+      await Time.sleep(300);
+    }
+    throw new Error('Id Token not available');
   };
 
   private static fixAcctStorageResult = (acctEmail: string, acctStore: AcctStoreDict, keys: AccountIndex[]): AcctStoreDict => {


extension/js/common/platform/store/acct-store.ts
```diff
@@ -142,7 +142,7 @@ export class AcctStore extends AbstractStore {
   };
 
   public static getScopes = async (acctEmail: string): Promise<Scopes> => {
-    const accessToken = await InMemoryStore.get(acctEmail, InMemoryStoreKeys.GOOGLE_TOKEN_ACCESS);
+    const accessToken = await this.getIdTokenUntilAvailable(acctEmail);
```
ioanmo226, line 145: @sosnovsky #5123 issue has occurred due to the null value of the `accessToken` in the `InMemoryStore` when attempting to retrieve it in the `getScopes` function. This problem arises because the `getScopes` function and the `Google.gmailCall` function, which retrieves and stores the access token in the `InMemoryStore`, are being executed simultaneously.

To resolve this issue, we have two potential solutions:
Current draft solution and another is to retrieve accessToken from Google if not available: In this alternative approach, we would attempt to fetch the accessToken from Google if it is not available in the `InMemoryStore`, similar to how it is done in the `gmailCall` function.

Both methods have their pros and cons:

1st method cons: If the user has a slow network speed, the `getIdTokenUntilAvailable` method may throw an error after a specific timeout. Moreover, the `getIdTokenUntilAvailable` method must guarantee that at least one API call is made to obtain the accessToken.

2nd method cons: This approach would result in duplicated API calls to retrieve the access token. Additionally, it would require duplicating the code logic for obtaining the access token.

Which method would you prefer to address this issue?

extension/js/common/platform/store/acct-store.ts
```diff
@@ -142,7 +142,7 @@ export class AcctStore extends AbstractStore {
   };
 
   public static getScopes = async (acctEmail: string): Promise<Scopes> => {
-    const accessToken = await InMemoryStore.get(acctEmail, InMemoryStoreKeys.GOOGLE_TOKEN_ACCESS);
+    const accessToken = await this.getAccessTokenUntilAvailable(acctEmail);
```
ioanmo226, line 145: @sosnovsky https://github.com/FlowCrypt/flowcrypt-browser/issues/5123 issue has occurred due to the null value of the `accessToken` in the `InMemoryStore` when attempting to retrieve it in the `getScopes` function. This problem arises because the `getScopes` function and the `Google.gmailCall` function, which retrieves and stores the access token in the `InMemoryStore`, are being executed simultaneously.

To resolve this issue, we have two potential solutions:
Current draft solution and another is to retrieve `accessToken` from Google if not available: In this alternative approach, we would attempt to fetch the `accessToken` from Google if it is not available in the `InMemoryStore`, similar to how it is done in the `gmailCall` function.

Both methods have their pros and cons:

1st method cons: If the user has a slow network speed, the `getAccessTokenUntilAvailable` method may throw an error after a specific timeout. Moreover, the `getAccessTokenUntilAvailable` method must guarantee that at least one API call is made to obtain the accessToken.

2nd method cons: This approach would result in duplicated API calls to retrieve the access token. Additionally, it would require duplicating the code logic for obtaining the access token.

Which method would you prefer to address this issue?

extension/js/common/platform/store/acct-store.ts
sosnovsky: First method looks better for me - it should be very bad connection when user won't be able to fetch token in a few seconds, I think we should use it and check that call for fetching tokens was executed.

extension/js/common/platform/store/acct-store.ts
sosnovsky: First method looks better for me - it should be very bad connection when user won't be able to fetch token in a few seconds, I think we should use it and check that call for fetching tokens was executed.

PR #5170, opened 2023-05-22
issue #5168 increase recipient text field limit to 8000

### LINKED ISSUE #5168: Increase recipients textfield maxlength
Currently we have `maxlength=256` for recipients textfields, but customer reported that it's now impossible to send message for 100+ recipients, while in earlier versions it worked well.

I think we can increase `maxlength` to 8000 for recipients textfields, as it worked well before and didn't cause extension crash.

martgil:
This PR updates recipient text field input limit from 256 to 8000.

close #5168 

### To be filled by reviewers


sosnovsky (approved):
Thanks for quick implementation 👍

PR #5175, opened 2023-05-24
#5169 fix: disable contacts search when pasting long string

### LINKED ISSUE #5169: Disable contacts search when pasting long string
I pasted string with 5 recipients into `To` input, after switching to `Subject` input I got error `Error searching contacts: Error:
Query length must be less than or equal to 100.`:

Screenshot 2023-05-22 at 14 00 44

I think we should just disable search when pasting string with multiple recipients, as it works well only for single recipient

ioanmo226:
This PR disables contacts search when pasting long string

close #5169 // if this PR closes an issue

- Not worth testing

### To be filled by reviewers


sosnovsky (approved):
Works well 👍

PR #5192, opened 2023-05-26
#5150 fix: apostrophe file name encoding

### LINKED ISSUE #5150: Fix attachments encoding with apostrophe in file name
Looks like google doesn't like this apostrophe in the middle of the file name (between `what` and `s_up`
```
 filename*0*=utf-8''what's_up%25253F.txt.pgp
```
Probably this is why extra double quotes appear
image
```
        {
          "partId": "1",
          "mimeType": "application/octet-stream",
          "filename": "\"what\"'s_up%25253F.txt.pgp",
```

According to https://datatracker.ietf.org/doc/html/rfc2231#section-3
the apostrophe has special meaning like in this example
```
 title*0*=us-ascii'en'This%20is%20even%20more%20
```
we have the 2 required apostrophes right after `utf-8` denoting unspecified language, that's ok, but we should probably escape the apostrophe with %27, so we can take the filename directly from Gmail API (when decoding).
If the filename was encoded like this by the browser extension, we should also fix the encoding part.
@sosnovsky ?

_Originally posted by @rrrooommmaaa in https://github.com/FlowCrypt/flowcrypt-browser/pull/4537#discussion_r1192634117_

ioanmo226:
This PR fixed apostrophe file name encoding issue

close #5150 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/lib/emailjs/emailjs-mime-codec.js
```diff
@@ -768,7 +768,7 @@
                     // unencoded lines: {name}*{part}
                     // if any line needs to be encoded then the first line (part==0) is always encoded
                     key: key + '*' + i + (item.encoded ? '*' : ''),
-                    value: /[\s";=]/.test(item.line) ? '"' + item.line + '"' : item.line
+                    value: /[\s";=']/.test(item.line) ? '"' + item.line + '"' : item.line,
```
ioanmo226, line 771: Do you agree with this approach @sosnovsky?
AFAIK as per the RC standard, single quote should be enclosed in double quotes.

With this way, current issue will be resolved but unit tests are complaining about this change and I need to update unit test.

sosnovsky: Approach is ok for me, issue is fixed with it 👍

test/source/tests/compose.ts
```diff
@@ -634,6 +634,25 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       })
     );
 
+    test(
+      'compose - check apostrophe file name encoding',
+      testWithBrowser(async (t, browser) => {
+        const acct = 'ci.tests.gmail@flowcrypt.test';
+        const subject = 'check apostrophe file name encoding';
+        const filename = `what's_up?.txt`;
```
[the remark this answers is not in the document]
sosnovsky: let's just remove question mark from file name, so it'll be `what's_up.txt`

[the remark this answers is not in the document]
ioanmo226: Ohh. yes. That sounds okay.
As our aim is to test single quote encoding.

test/source/tests/compose.ts
```diff
@@ -634,6 +634,25 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       })
     );
 
+    test(
+      'compose - check apostrophe file name encoding',
+      testWithBrowser(async (t, browser) => {
```
sosnovsky, line 639: I tried to run this test in `master` branch and it still passes, so probably something is not correct in attachments check here

ioanmo226: Hmm... let me check

ioanmo226: Hmm.. I tried to run test in master branch and it fails. Could you try again?

sosnovsky: For `master` branch I mean taking `emailjs-mime-codec.js` from `master` branch, so it doesn't include fix.
I tried it again - still passes, can you try to run test this way?

sosnovsky (approved):
Well done 👍

rrrooommmaaa:
@ioanmo226 Have you tested uploading the fixed MIME to gmail and it looked ok?

If so, please replace `message-export-182263bf9f105adf.json` file with the correctly formed message.
When exported from gmail, I expect there won't be irregularities like these we currently have:
image

ioanmo226:
@rrrooommmaaa which MIME message?
And what do you mean uploading MIME to gmail? Is there way to upload MIME to gmail?

> Have you tested uploading the fixed MIME to gmail and it looked ok?

rrrooommmaaa:
The message `message-export-182263bf9f105adf.json` used for tests. It was supposed to have an encrypted file name like this `what's_up%253F.txt.pgp`, but gmail provides something different when downloaded in `full` format (and on screen too)
The actual message in MIME format is in this field:
image
If you Base64-decode this, you'll be able to see its structure and save it to `.eml` file.
And yes, you can upload it with Thunderbird (after corrections), for example, to any folder on the server. Of course, the other fields (in 'full' format) will be regenerated by gmail and a new `id` may be given.
But it'd be better if you actually send a similar message with the fixed version of Flowcrypt extension and then export it from Gmail to see if we get the correct file name in `full` format.
And replace the old message with the new one to fix this line in test:
```
await checkIfFileDownloadsCorrectly(t, browser, '182263bf9f105adf', "_what_'s_up%25253F.txt.pgp"); // todo: #5150, should be "what's_up%253F.txt.pgp"
```

ioanmo226:
https://github.com/FlowCrypt/flowcrypt-browser/pull/5069/commits/4afd8f8299ac2ad7eb5e0b7a6d42c12f806bfeb1

I pushed to this PR because it might conflict changes.

PR #5210, opened 2023-06-06
#5209 Make Build Script More Robust

### LINKED ISSUE #5209: Make Build Script More Robust
In the readme, we have a `Note for Mac OS users` section which suggests upgrading `bash` to v4 and installing the `GNU` `cp` utility. I think this is likely due to the use of the `--parents` option with the `cp` command, which is not supported by the `BSD` version of `cp` that comes as default with macOS.

To make our build script more robust and user-friendly, especially for macOS users, we can consider replacing `cp --parents` with `rsync -R`. The `-R` option in `rsync` has similar functionality to `--parents` in `cp`, maintaining the directory structure while copying. 

This change would eliminate the need for macOS users to upgrade bash or install the GNU cp utility, making the setup process more straightforward. Importantly, it will make it easier for developers using macOS to get started with the project, potentially broadening the community of contributors and speeding up development.

cc: @sosnovsky

ioanmo226:
This PR makes build script more robust

close #5209 // if this PR closes an issue

### To be filled by reviewers


sosnovsky (approved):
Works great 👍

PR #5213, opened 2023-06-08
#5004 Fix: . cryptup_reply div should be hidden in FES messages

### LINKED ISSUE #5004: . cryptup_reply div should be hidden in FES messages
1. Send encrypted message to 2 recipients - 1 with pubkey (so it'll be sent encrypted) and 1 without pubkey (it'll be sent password-protected using FES)
2. Open password-protected message on FES and send `reply-all` message
3. When opening this reply message in Gmail `.cryptup_reply` div code is visible, but it should be hidden

Screenshot 2023-03-09 at 12 51 48

This bug doesn't happen if you choose `reply` option, instead of `reply-all`. Please check if this issue on extension side, or something should be changed on FES.

ioanmo226:
This PR fixed  `.cryptup_reply div should be hidden in FES messages` issue

close #5004 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
```diff
@@ -178,7 +178,7 @@ export class PgpBlockViewRenderModule {
       // eslint-disable-next-line @typescript-eslint/naming-convention
       await AcctStore.set(this.view.acctEmail, { successfully_received_at_leat_one_message: true });
     }
-    const contentWithLink = linkifyHtml(htmlContent);
+    const contentWithLink = Xss.convertHtmlEntities(linkifyHtml(htmlContent));
```
ioanmo226, line 181: @sosnovsky Are you happy with this approach?
I converted `&lt;` to correct `<` so that html can be rendered correctly

sosnovsky: Hmm, it fixes issue but breaks keeping tags in plain text (as failed test shows).
Did you find any difference in message text when sending simple `reply` versus `reply all`? As `reply` message doesn't show `.cryptup_reply` div

ioanmo226: or how about just convert `&lt;` to `<` and `&gt;` to `>` using regex or something like that in above function?

ioanmo226: OK

sosnovsky: Looks like it happens because FES sends reply messages with `text/plain` body, if I add `text/html` then `div. crypt_up` is hidden, but `cryptup-data` property disappear. Can you please check if we remove such properties when decrypting messages?

sosnovsky: No, `cryptup-data` property should be there. If message body is `text/plain` then it's present, but if I change it to `text/html` on FES side, then browser extension removes `cryptup-data` property from `div. crypt_up` and it's not possible to parse reply recipients.
So I thought maybe we remove html properties when decrypting messages for security reasons, but not sure

ioanmo226: Just finished checking. We indeed remove such properties when showing the `pgp_block`.

However, when the user clicks `reply/reply all` and tries to retrieve the `fcReplyToken`, it attempts to get it from the plain decrypted content.
I thin it should work fine when you return `text/html` on the `FES` side.

sosnovsky: I tested it with `text/html` on FES side, but clicking `reply` can't fetch recipients list as decrypted content doesn't include `cryptup-data`:

Screenshot 2023-06-09 at 12 58 53

Then we should fix this issue on both sides:
- on FES we should use `text/html` instead of `text/plain`
- in browser extension we should add exception for `div.cryptup_reply`, so it's `cryptup-data` property won't be removed

What do you think?

ioanmo226: Sounds good.
However, I think this one should be considered after FES deploy.

sosnovsky: Hmm, but then users won't be able to reply to FES messages until they update browser extension to the latest version, as `cryptup-data` will be empty. Need some backwards-compatible solution

sosnovsky: Maybe we can extract `cryptup-data` before rendering message and then remove `div.cryptup-reply`?
Is it possible or we parse `cryptup-data` when clicking `reply` button?

ioanmo226: Yep, You are right.
I think we can. Let me implement solution

ioanmo226: @sosnovsky, @martgil,

I'm not sure how to generate a test email for this case since the current `FES` returns the `text/plain` type.

Could someone please assist me in generating a test email for the above case? The email should have a message body in `text/html` format and include the `cryptup-data` attribute(Message which might be generated by using the "reply all" function to an FES message)

sosnovsky: I mean we should extract `cryptup-data` from current `text/plain` message. Because if we change message body to `text/html` then users with older extension builds (without this fix) won't be able to reply to these messages, as `cryptup-data` property is removed during decryption for `text/html`.

ioanmo226: Aha, I see.
That makes sense

ioanmo226: @sosnovsky, I was confused about the issue of going backward and forward.
However, the current functionality of extracting `cryptup-data` from a `text/plain` message works well.
The problem lies in the visibility of the `cryptup_reply` div due to the presence of `&lt;` and `&gt;` instead of `<` and `>`.

I believe a simple solution would be to replace those two strings with `<` and `>`, respectively. By doing so, the issue should be resolved. What are your thoughts on this?

sosnovsky: Yeah, it'll be simpler solution than changing logic of message decryption. But it should be just temporary fix, as correct one will be changing FES message body type to `text/html` (as it contains html tags). We can't change type now, as it'll break replying to FES messages in current browser extension version (+ need to check it in mobile apps).

So I propose such steps:
- in this PR implement replacing `&lt;` and `&gt;` with `<` and `>` (maybe just for `div.cryptup_reply`, if it's not too complicated, so it won't affect other messages)
- in this PR update decryption logic of `text/html` to leave `cryptup-data` property in decrypted message (as currently all properties are removed)
- I'll file an issue for FES to change password message replies body type to `text/html`, but it should be implemented after several releases, when all users will have browser extension versions with correct support for `cryptup-data` property

As a result, we'll have a quick fix for this issue, and will prepare browser extension for correct fix in the future.

ioanmo226: Sounds good!

ioanmo226: No. @martgil 
Thank you

ioanmo226:
@sosnovsky In case you missed this.

sosnovsky:
> @sosnovsky In case you missed this.

I checked it yesterday, but `cryptup-data` property is still removed from `text/html` body, and I want to try to send you such message from local FES, so you'll be able to check it on your side.

ioanmo226:
ok

sosnovsky:
@ioanmo226 did you receive reply-all message in "Test message for #5004" thread?
As I'm not sure if email sending is allowed for all recipients in local environment, or only for approved ones.

ioanmo226:
Yeah. I received it.

sosnovsky:
> Yeah. I received it.

Great, can you please check why `cryptup-data` property is still removed.

ioanmo226:
Yeah, I can confirm that `cryptup-data` is also removed in my side.
Let me check.

ioanmo226:
Fixed. Please check

sosnovsky:
`cryptup-data` property is ok now, but after clicking `Encrypted reply` recipients list is empty in compose box.
It should be prefilled with recipients parsed from `cryptup-data`

ioanmo226:
Hmm.. when clicking `reply` or `reply all` displays correct list of recipients in my side.
Maybe you tested in another way?
image

sosnovsky:
> Hmm.. when clicking `reply` or `reply all` displays correct list of recipients in my side.
> Maybe you tested in another way?

It loads only original sender for message from FES. But it should load also recipients from `cryptup-data` - it's just base64 string, you can decode it and see array of recipients.

By the way, this original sender shouldn't be added to recipients list - you can add it to `FLOWCRYPT_REPLY_EMAIL_ADDRESSES` at https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/api/email-provider/gmail/gmail-parser.ts#L13 for local debug.

sosnovsky:
I sent you another message from production FES, you can try to click `Secure reply` there and then correct recipients list will be loaded.

ioanmo226:
Sorry but I'm still uncertain about what's problematic with the current implementation. 

When you click `reply` to `Re: Test message for #5004`, the recipient `tests.only@flowcrypt.dev` is loaded.

And, when you click `reply all`, the recipients `tests.only@flowcrypt.dev`, `roma...@flowcrypt.com`, and `no.publickey@flowcrypt.com` are loaded. I think those behaviors are correct.

Here is the base64 decoding of `cryptup-data`.
image

sosnovsky:
> And, when you click reply all, the recipients tests.only@flowcrypt.dev, roma...@flowcrypt.com, and no.publickey@flowcrypt.com are loaded. I think those behaviors are correct.

Yes, looks like these buttons work well, the only issue is with this button:

Screenshot 2023-06-14 at 16 46 53

sosnovsky:
> Behavior is different different for local FES and deployed FES because test.only is not in `FLOWCRYPT_REPLY_EMAIL_ADDRESSES`. Just added it. Please check

Yeah, you're right - just re-checked and everything works well, I thought it shouldn't affect recipients list, sorry for missing it :)

sosnovsky (approved):

PR #5214, opened 2023-06-09
issue #5211 Add public key lookup in compatibility test page

### LINKED ISSUE #5211: [Feature suggestion] fetch and load public key to key compatibility checker page
Description: The OpenPGP key compatibility check page is super helpful for evaluating users' public keys. We use it quite often to have an overview of the available user public key in the attester.

I feel that adding an input field that accepts user email and a button for loading the public key above the ASCII armored key textarea would make it even more useful and versatile. This would improve repetitive tasks such as looking up a user's public key from the attester, copying the public key, and pasting it into the key compatibility checker.

(kindly ignore the error message) 
image

martgil:
This PR Adds a public key lookup in the compatibility test extension page.

close #5211

Screenshots:

Screenshot from 2023-06-09 13-24-10
Screenshot from 2023-06-09 13-23-55

### To be filled by reviewers

### DIFF
--- a/extension/chrome/settings/modules/compatibility.htm
+++ b/extension/chrome/settings/modules/compatibility.htm
@@ -17,6 +17,14 @@
       <div class="line">Locally test any public or private key.</div>
 
       <div class="line">
+        <label
+          >Search for public key on FlowCrypt Attester
+          <input type="text" class="input_email" placeholder="Email address" maxlength="256" />
+        </label>
+        <button class="button green action_load_public_key">load public key</button>
+      </div>
+
+      <div class="line">
         <textarea
           placeholder="ASCII Armored Public or Private Key....."
           class="input_key"
--- a/extension/chrome/settings/modules/compatibility.htm
+++ b/extension/chrome/settings/modules/compatibility.htm
@@ -19,7 +19,7 @@
       <div class="line">
         <label
           >Search for public key on FlowCrypt Attester
-          <input type="text" class="input_email" placeholder="Email address" maxlength="256" />
+          <input class="input_email" type="text" placeholder="Email address" style="height: 37px !important" maxlength="256" />
         </label>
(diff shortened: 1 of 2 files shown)


martgil:
ready for review. thank you!

extension/chrome/settings/modules/compatibility.htm
```diff
@@ -16,6 +16,14 @@ <h1>OpenPGP key compatibility test</h1>
 
       <div class="line">Locally test any public or private key.</div>
 
+      <div class="line">
+        <label
+          >Search for public key on FlowCrypt Attester
+          <input type="text" class="input_email" placeholder="Email address" maxlength="256" />
```
sosnovsky, line 22: let's add `style="height: 37px !important"`, so it'll be the same height as button and passphrase input

sosnovsky (commented):
Works great, just a small UI improvement

extension/chrome/settings/modules/compatibility.htm
martgil: Thank you very much.

image

extension/chrome/settings/modules/compatibility.htm
```diff
@@ -16,6 +16,14 @@ <h1>OpenPGP key compatibility test</h1>
 
       <div class="line">Locally test any public or private key.</div>
 
+      <div class="line">
+        <label
+          >Search for public key on FlowCrypt Attester
```
tomholub, line 21: Since this uses PubLookup, which is indeed the better choice, it is looking up public keys on other places as well. It would be good to change this to `Search for public key on key servers`

Further, this page is now probably in need of some instruction on top. "You may use this page to ..." and "You can either paste the public key you want to check, or you can have one loaded from key servers by entering the email" and similar.

PR #5255, opened 2023-06-26
#5222 Mock definition leftovers

### LINKED ISSUE #5222: a few more mock items that should be done through configuration
I was going through the mock definitions and it's greatly improved, just the below leftovers that I noticed are still being configured through in the mock endpoints directly and not through configuration, which would be much preferred. The goal of doing it through configuration is clarity and no surprises, even if a bit more verbose.

At this point, when 95% of things are actually set through configuration, it's also a matter of consistency.

----------------------------

`getMockSharedTenantFesEndpoints` is using `FesConfig`, but `mockCustomerUrlFesEndpoints` isn't. It should be using the same `FesConfig` structure to get configured as well. I suppose the `FesConfig` type would have to be extended to allow for more types of configuration, for example, here we still have values in the mock:

```ts
    if (req.headers.host === fesUrl && req.method === 'POST' && typeof body === 'string') {
      // test: `compose - user@standardsubdomainfes.localhost:8001 - PWD encrypted message with FES web portal`
      authenticate(req, 'oidc');
      if (body.includes(`"from":"user@standardsubdomainfes.localhost:${port}"`)) {
        return await processMessageFromUser(body, fesUrl);
      }
      if (body.includes(`"from":"user2@standardsubdomainfes.localhost:${port}"`)) {
        return await processMessageFromUser2(body, fesUrl);
      }
      if (body.includes(`"from":"user3@standardsubdomainfes.localhost:${port}"`)) {
        return await processMessageFromUser3(body, fesUrl);
      }
      if (body.includes(`"from":"user4@standardsubdomainfes.localhost:${port}"`)) {
        return await processMessageFromUser4(body, fesUrl);
      }
    }
```

They could instead be configured, for example, it would be conceivable to configure a validation method like 

```diff
export interface FesConfig {
  returnError?: HttpClientErr;
  clientConfiguration?: FesClientConfiguration;
+  messagePostValidator?: (body, fesUrl) => Promise<...>
}
```

So when not configured, it will just throw error. If configured, it will do the checks. Then we don't have to have different users for each of the 4 tests (ok to retain different users if easier, but the point is that next time we need to write a test, it can be easily copy-pasted without having to change the user).

------------------------------------

I would rather not have an exception for this:

```ts
      // If the account is a compatibility account, return specific aliases.
      // This account is used in hundreds of tests, so we handle it separately to avoid duplicating the code.
      if (acct === 'flowcrypt.compatibility@gmail.com') {
        const alias = 'flowcryptcompatibility@gmail.com';
        return {
          sendAs: [
            {
              ...primarySendAs,
              signature:
                '<div dir="ltr">flowcrypt.compatibility test footer with an img<br><img src="https://flowcrypt.com/assets/imgs/svgs/flowcrypt-logo.svg" alt="Image result for small image"><br></div>',
            },
            {
              sendAsEmail: alias,
              displayName: 'An Alias',
              replyToAddress: alias,
              signature: '',
              isDefault: false,
              isPrimary: false,
              treatAsAlias: false,
              verificationStatus: 'accepted',
            },
          ],
        };
      }
```

Do hundreds of those tests actually need this exact setup? I'm guessing that 95% of them would do fine with just this

```ts
      // If no aliases are defined in the config, return only the primary send-as object
      if (!config?.aliases) {
        return { sendAs: [primarySendAs] };
      }
```

And then there will maybe be a few odd tests that will need the full setup. It should be taken out of here imo.

------------------------------------------

This should be done through configuration, it should be easy to isolate the test by searching for that thread id.

```ts
          const statusCode = id === '16841ce0ce5cb74d' ? 404 : 400; // intentionally testing missing thread
          throw new HttpClientErr(`MOCK thread not found for ${acct}: ${id}`, statusCode);
```

This should be done through configuration:

```ts
        const data = await GoogleData.withInitializedData(acct);
        if (mimeMsg.subject?.includes('saving and rendering a draft with image')) {
          data.addDraft('draft_with_image', raw, mimeMsg);
        }
        if (mimeMsg.subject?.includes('check existing draft not saved without changes')) {
          data.addDraft('check_existing_draft_save', raw, mimeMsg);
        }
        if (mimeMsg.subject?.includes('RTL')) {
          data.addDraft(`draft_with_rtl_text_${mimeMsg.subject?.includes('rich text') ? 'rich' : 'plain'}`, raw, mimeMsg);
        }
```

--------------------------------------------

There should be no emails in the mock definitions, these should be set through configuration

```ts
          if (acctEmail === 'put.key@key-manager-autogen.flowcrypt.test') {
...
          }
          if (acctEmail === 'expire@key-manager-keygen-expiration.flowcrypt.test') {
...
          }
          if (acctEmail.includes('updating.key')) {
...
            return {};
          }
```

-----------------------------

ioanmo226:
This PR implemented mock definition leftovers

close #5222 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/mock/key-manager/key-manager-endpoints.ts
+++ b/test/source/mock/key-manager/key-manager-endpoints.ts
@@ -10,6 +10,13 @@ import { OauthMock } from '../lib/oauth';
 
 export interface KeyManagerConfig {
   keys?: string[];
+  putKeysExpectation?: Record<
+    string,
+    {
+      identity?: string;
+      expirationExists?: boolean;
+    }
+  >;
   returnError?: HttpClientErr;
   putReturnError?: HttpClientErr;
 }
@@ -32,39 +39,23 @@ export const getMockKeyManagerEndpoints = (oauth: OauthMock, config: KeyManagerC
           if (config.putReturnError) {
             throw config.putReturnError;
           }
-          const { privateKey } = body as Dict<string>;
-          if (acctEmail === 'put.key@key-manager-autogen.flowcrypt.test') {
+          if (config.putKeysExpectation && config.putKeysExpectation[acctEmail]) {
+            const expectation = config.putKeysExpectation[acctEmail];
+            const { privateKey } = body as Dict<string>;
             const prv = await KeyUtil.parseMany(privateKey);
             expect(prv).to.have.length(1);
             expect(prv[0].algo.bits).to.equal(2048);
             expect(prv[0].identities).to.have.length(1);
-            expect(prv[0].identities[0]).to.equal('First Last <put.key@key-manager-autogen.flowcrypt.test>');
+            if (expectation.identity) {
+              expect(prv[0].identities[0]).to.equal(expectation.identity);
+            }
             expect(prv[0].isPrivate).to.be.true;
             expect(prv[0].fullyDecrypted).to.be.true;
-            expect(prv[0].expiration).to.not.exist;
-            config.keys = [privateKey];
-            return {};
-          }
-          if (acctEmail === 'expire@key-manager-keygen-expiration.flowcrypt.test') {
-            const prv = await KeyUtil.parseMany(privateKey);
-            expect(prv).to.have.length(1);
-            expect(prv[0].algo.bits).to.equal(2048);
-            expect(prv[0].identities).to.have.length(1);
-            expect(prv[0].identities[0]).to.equal('First Last <expire@key-manager-keygen-expiration.flowcrypt.test>');
-            expect(prv[0].isPrivate).to.be.true;
-            expect(prv[0].fullyDecrypted).to.be.true;
-            expect(prv[0].expiration).to.exist;
-            config.keys = [privateKey];
-            return {};
-          }
-          if (acctEmail.includes('updating.key')) {
-            const prv = await KeyUtil.parseMany(privateKey);
-            expect(prv).to.have.length(1);
-            expect(prv[0].algo.bits).to.equal(2048);
-            expect(prv[0].identities).to.have.length(1);
-            expect(prv[0].isPrivate).to.be.true;
-            expect(prv[0].fullyDecrypted).to.be.true;
-            expect(prv[0].expiration).to.not.exist;
+            if (expectation.expirationExists) {
+              expect(prv[0].expiration).to.exist;
+            } else {
+              expect(prv[0].expiration).to.not.exist;
+            }
             config.keys = [privateKey];
             return {};
           }
--- a/test/source/tests/setup.ts
+++ b/test/source/tests/setup.ts
@@ -833,6 +833,11 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
           },
           ekm: {
             keys: [negativeExpirationKey.privateKey],
+            putKeysExpectation: {
+              [acctEmail]: {
+                expirationExists: false,
+              },
+            },
           },
           fes: {
             clientConfiguration: getKeyManagerAutogenRules(t.urls!.port!),
@@ -1423,18 +1428,23 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
     test(
       'get.updating.key@key-manager-choose-passphrase-forbid-storing.flowcrypt.test - automatic update of key found on key manager',
       testWithBrowser(async (t, browser) => {
+        const acct = 'get.updating.key@key-manager-choose-passphrase-forbid-storing.flowcrypt.test';
         t.mockApi!.configProvider = new ConfigurationProvider({
           attester: {
             pubkeyLookup: {},
           },
           ekm: {
             keys: [testConstants.updatingPrv],
+            putKeysExpectation: {
+              [acct]: {
+                expirationExists: false,
+              },
+            },
           },
           fes: {
             clientConfiguration: getKeyManagerChoosePassphraseForbidStoringRules(t.urls!.port!),
           },
         });
-        const acct = 'get.updating.key@key-manager-choose-passphrase-forbid-storing.flowcrypt.test';
         const settingsPage = await BrowserRecipe.openSettingsLoginApprove(t, browser, acct);
         const passphrase = 'long enough to suit requirements';
         await SetupPageRecipe.autoSetupWithEKM(settingsPage, {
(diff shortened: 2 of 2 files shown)


test/source/mock/google/google-endpoints.ts
```diff
@@ -58,7 +58,13 @@ export type MockUserAlias = {
...
   contacts?: string[];
   othercontacts?: string[];
-  aliases?: Dict<MockUserAlias[]>;
+  aliases?: Record<
+    string,
+    {
+      primarySignature?: string;
+      list?: MockUserAlias[];
+    }
+  >;
```
ioanmo226, line 67: Tom mentioned in issue that he doesn't want to have exception for `flowcrypt.compatiblity@gmail.com` aliases.
So I changed aliases type to like this.

@sosnovsky Do you agree with this change?
Or you have better idea?
> I would rather not have an exception for this:
> 
>       // If the account is a compatibility account, return specific aliases.
>       // This account is used in hundreds of tests, so we handle it separately to avoid duplicating the code.
>       if (acct === 'flowcrypt.compatibility@gmail.com') {
>

sosnovsky: Maybe it's possible to make `primarySignature` property of `GoogleConfig` and then `aliases` type will be `Dict<MockUserAlias[]>` as before?

ioanmo226: Yeah, That sounds better

test/source/tests/settings.ts
```diff
@@ -191,7 +220,21 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+              },
+            },
+          },
+          google: {
+            aliases: {
+              [acct]: flowcryptCompatibilityAliasList,
+            },
+          },
+        });
+        await BrowserRecipe.setUpCommonAcct(t, browser, 'compatibility');
```
sosnovsky, line 237: Maybe it's possible just add optional `aliases` parameter to `setupCommonAcctWithAttester`?

ioanmo226: Sounds good

test/source/mock/fes/customer-url-fes-endpoints.ts
```diff
@@ -2,303 +2,124 @@
...
-    throw new HttpClientErr('Not Found', 404);
-  },
-  '/api/v1/message/FES-MOCK-EXTERNAL-FOR-GATEWAYFAILURE@EXAMPLE.COM-ID/gateway': async () => {
-    // test: `user4@standardsubdomainfes.localhost:8001 - PWD encrypted message with FES web portal - a send fails with gateway update error`
-    throw new HttpClientErr(`Test error`, Status.BAD_REQUEST);
-  },
+      const port = parsePort(req);
+      if (req.headers.host === standardFesUrl(port) && req.url === `/api/v1/client-configuration?domain=standardsubdomainfes.localhost:${port}`) {
+        return {
+          clientConfiguration: { flags: [], disallow_attester_search_for_domains: ['got.this@fromstandardfes.com'] }, // eslint-disable-line @typescript-eslint/naming-convention
```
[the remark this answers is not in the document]
ioanmo226: Yeah

test/source/mock/fes/customer-url-fes-endpoints.ts
```diff
@@ -2,303 +2,124 @@
...
+        // test `status404 does not return any fesUrl` uses this
+        // this makes enterprise version tolerate missing FES - explicit 404
+        throw new HttpClientErr(`Not found`, 404);
       }
-      if (body.includes(`"from":"user3@standardsubdomainfes.localhost:${port}"`)) {
-        return await processMessageFromUser3(body, fesUrl);
+      if (req.headers.host === `fes.google.mock.localhost:${port}`) {
+        // test `compose - auto include pubkey is inactive when our key is available on Wkd` uses this
+        // this makes enterprise version tolerate missing FES - explicit 404
+        throw new HttpClientErr(`Not found`, 404);
```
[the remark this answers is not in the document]
ioanmo226: Yep

test/source/mock/fes/customer-url-fes-endpoints.ts
```diff
@@ -2,303 +2,124 @@
...
+          service: 'external-service',
+          orgId: 'standardsubdomainfes.test',
+          version: 'MOCK',
+          apiVersion: 'v1',
+        };
       }
-      if (body.includes(`"from":"user2@standardsubdomainfes.localhost:${port}"`)) {
-        return await processMessageFromUser2(body, fesUrl);
+      if (req.headers.host === `fes.localhost:${port}`) {
+        // test `status404 does not return any fesUrl` uses this
```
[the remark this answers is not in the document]
ioanmo226: I agree.

ioanmo226:
@sosnovsky All reviews are addressed.
Also further refactored mock definitions.
Please check

test/source/tests/tooling/browser-recipe.ts
```diff
@@ -16,7 +16,29 @@ import { GmailPageRecipe } from '../page-recipe/gmail-page-recipe';
...
+interface CommonAcctConfig {
+  google?: {
+    primarySignature?: string;
+    aliases?: MockUserAlias[];
+    draftIdToSave?: string;
+    contacts?: string[];
+    threadNotFoundError?: Record<string, number>;
+    htmlRenderer?: (msgId: string, prerendered?: string) => string | undefined;
+    getMsg?: Dict<Dict<{ error: Error } | { msg: GmailMsg }>>;
+  };
```
sosnovsky, line 35: Can we replace it with just `google?: GoogleConfig`?

ioanmo226: I thought about it too.
But didn't want to pass `primarySignature: Record<string, string>` and `aliases: Record<string, MockerUserAliases[]?`
This way `google: { primarySignature: flowcryptCompatibilityPrimarySignature },` should be converted to `google: { [acctEmail]: {primarySignature: flowcryptCompatibilityPrimarySignature }},` which seems useless

sosnovsky: What about this?
```ts
interface CommonAcctConfig {
  google?: GoogleConfig & {
    primarySignature?: string;
    aliases?: MockUserAlias[];
  };
  attester?: AttesterConfig & {
    includeHumanKey?: boolean;
    includeFlowcryptCompatibilityKey?: boolean;
  };
}
```

ioanmo226: Sounds good!

PR #5261, opened 2023-06-30
#5254 Fix: Different sender name issue

### LINKED ISSUE #5254: Different sender name is displayed for encrypted emails sent by the extension
Description: We received a bug report in which the receiver of an encrypted message sent using FlowCrypt sees a sender name that is different from what it should be.

For example, I send an encrypted email from my test account to my corporate account. The profile name of the test account is "Test Account", which was formerly known as "Mart Robles".

When the receiver opens the email, they see that the sender name is set to the former name instead of the current one.

Screenshot 2023-06-26 at 1 34 39 PM

Screenshot 2023-06-26 at 1 35 07 PM

The reporting user also shares a test email which confirms the following issue: https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGsnLGRgMHPqpKsCBjgJGTDMqSB

They also added that the problem does not occur when sending an original email sent by Gmail, which I have also confirmed.

Reference:
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGsnLCzRmkRSSxcGFVqLrKhMSGd

ioanmo226:
This PR fixed different sender name issue

close #5254 // if this PR closes an issue

- Difficult to test (Hard to reproduce in test environment)

### To be filled by reviewers

### DIFF
--- a/extension/js/common/api/email-provider/gmail/gmail-parser.ts
+++ b/extension/js/common/api/email-provider/gmail/gmail-parser.ts
@@ -124,16 +124,12 @@ export namespace GmailRes {
   };
 
   export type GoogleUserProfile = {
-    names: GoogleName[];
-  };
-
-  type GoogleName = {
-    metadata: GoogleMetadata;
-    displayName: string;
-  };
-
-  type GoogleMetadata = {
-    primary?: boolean;
+    names: {
+      metadata: {
+        primary?: boolean;
+      };
+      displayName: string;
+    }[];
   };
 }
 
--- a/extension/js/common/api/email-provider/gmail/gmail-parser.ts
+++ b/extension/js/common/api/email-provider/gmail/gmail-parser.ts
@@ -122,6 +122,19 @@ export namespace GmailRes {
       };
     }[];
   };
+
+  export type GoogleUserProfile = {
+    names: GoogleName[];
+  };
+
+  type GoogleName = {
+    metadata: GoogleMetadata;
+    displayName: string;
+  };
+
+  type GoogleMetadata = {
+    primary?: boolean;
+  };
 }
 
 export class GmailParser {
--- a/extension/js/common/api/email-provider/gmail/google.ts
+++ b/extension/js/common/api/email-provider/gmail/google.ts
@@ -10,6 +10,32 @@ import { GmailRes } from './gmail-parser.js';
 import { GoogleAuth } from './google-auth.js';
 import { Serializable } from '../../../platform/store/abstract-store.js';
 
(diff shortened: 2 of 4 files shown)


ioanmo226:
@sosnovsky In case you missed this

extension/js/common/api/email-provider/gmail/google.ts
```diff
@@ -10,6 +10,32 @@ import { GmailRes } from './gmail-parser.js';
 import { GoogleAuth } from './google-auth.js';
 import { Serializable } from '../../../platform/store/abstract-store.js';
 
+type GoogleUserProfile = {
+  resourceName: string;
+  etag: string;
+  names: GoogleName[];
+};
```
sosnovsky, line 17: We probably don't need this full object, let's use only needed fields and move it to [GmailRes](https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/api/email-provider/gmail/gmail-parser.ts#L15):

```js
export type GoogleProfileNames = {
    names: {
      metadata: {
        primary: boolean;
      };
      displayName: string;
    };
  };
```

test/source/mock/google/google-endpoints.ts
```diff
@@ -154,6 +154,27 @@ export const getMockGoogleEndpoints = (oauth: OauthMock, config: GoogleConfig |
...
+        resourceName: 'people/resourceId',
+        etag: '%EgUBAi43PRoEAQIFByIMSFMydTZETy9IEFE9',
+        names: [
+          {
+            metadata: {
+              primary: true,
+            },
+            displayName: 'First Last',
+            familyName: 'First',
+            givenName: 'LLast',
```
sosnovsky, line 171: `LLast` -> `Last`

sosnovsky (commented):
Works well 👍

extension/js/common/api/email-provider/gmail/gmail-parser.ts
```diff
@@ -122,6 +122,19 @@ export namespace GmailRes {
...
+  };
+
+  type GoogleName = {
+    metadata: GoogleMetadata;
+    displayName: string;
+  };
+
+  type GoogleMetadata = {
+    primary?: boolean;
+  };
```
sosnovsky, line 137: let's merge them into single `type`:
```js
export type GoogleUserProfile = {
    names: {
      metadata: {
        primary?: boolean;
      };
      displayName: string;
    };
  };
```

sosnovsky (approved):
👍

PR #5264, opened 2023-07-01
fix timeout race

### LINKED ISSUE #5263: race with timeoutPromise sometimes fails (the process gets stuck)
A worker promise may not give a turn to other promises, regardless having `await` statements in a loop.

I discovered this in `awaitDownloadTriggeredByClicking`, other cases need to be reviewed too.

rrrooommmaaa:
This PR fixed timeout race in awaitDownloadTriggeredByClicking

close #5263

- Does not need tests (internal test infrastructure fix)

### To be filled by reviewers


test/source/browser/controllable.ts
```diff
@@ -529,8 +529,15 @@ abstract class ControllableBase {
...
-          files[filename] = fs.readFileSync(path.resolve(downloadPath, filename));
+        if (newFilenames.length) {
+          for (const filename of newFilenames) {
+            const filepath = path.resolve(downloadPath, filename);
+            files[filename] = fs.readFileSync(filepath);
+            fs.unlinkSync(filepath);
+          }
+        } else {
+          // Give a turn to other promises, including timeoutPromise
+          await new Promise(resolve => setImmediate(resolve));
```
rrrooommmaaa, line 540: This is a hack that is guaranteed to give a turn to timeoutPromise.
Though in this particular case, we'd better use `await Util.sleep`, as the code is going to read the directory without pauses

@sosnovsky ?

sosnovsky: Where we should use `await Util.sleep` here, instead of `await new Promise(resolve => setImmediate(resolve))`?

rrrooommmaaa: We should use `sleep` because otherwise the filesystem will be queried unnecessary often (pausing only for the period of execution of some async Javascript code in the testing process).
In the current `master` it is queried in a loop (if no `.crdownload` files are found) without even allowing other async Javascript code to run. The hack solves the timeout issue, but a little pause would make sense, I guess.
The question is: how many millis?

sosnovsky: Ah, I see now, thanks, what about `500ms`?

sosnovsky (approved):
👍

PR #5266, opened 2023-07-01
send render messages to tab and check selector uniqueness in tests

### LINKED ISSUE #5265: pgp_block_render message is handled twice if broadcast
Most messages are not visible, but `renderInnerAttachments` leads to attachment duplication
image

rrrooommmaaa:
This PR sends `pgp_render_message` to tab
This avoids duplicate handling of the message (resulting in duplicate inner attachments)
The tests are now stricter, they expect only a single element to match a selector to be clicked, typed into or focused,
otherwise an exception is thrown.
Also, nicer test logs, test name and status is included in the log
image

close #5265
close #5293

- Test infrastructure updated

### To be filled by reviewers


sosnovsky (approved):
Looks good 👍

PR #5284, opened 2023-07-07
issue #5280 do not report err 400 when submitting to Attester

### LINKED ISSUE #5280: do not report this particular error to the backend
image

image

martgil:
This PR updates the `IGNORE_ERR_MSG` variable to include the error message "could not find valid self-signature".

close #5280

- Difficult to test: I'm still looking into how I could trigger the following error mentioned in #5280 to ensure that the report should skip reporting if its all about `could not find valid self-signature`.

### To be filled by reviewers


tomholub:
This kind of simple fix doesn't need a test, the downside is small. I just don't want to see that many errors crowding out other errors when it's not the extension's fault this happens.

martgil:
@tomholub, was it sufficient for your requirements, or would you prefer to receive no reports regarding the error message "This key does not appear valid"?

tomholub:
I've updated the error string to only apply to submissions to Attester.

tomholub (approved):

PR #5292, opened 2023-07-11
#5282 Strip query strings on error reports

### LINKED ISSUE #5282: when reporting errors, do not send any url parameters in url field
For example: `url: chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/compose.htm?frameId=frame_KHukGj`

Anything after `?` should be left out. Due to the variability inside the URL, the errors are not recognized as the same error, so there is a new record with count=1 for each of these errors on the backend, when it should have been a single record with count=many.

martgil:
This PR strips query strings on reported errors.

close #5282 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/dev/ci_unit_test.ts
+++ b/extension/chrome/dev/ci_unit_test.ts
@@ -23,11 +23,6 @@ import { PgpHash } from '../../js/common/core/crypto/pgp/pgp-hash.js';
 import { PgpArmor } from '../../js/common/core/crypto/pgp/pgp-armor.js';
 import { Xss } from '../../js/common/platform/xss.js';
 
-/* Used for test "Catcher does not include query string on report" */
-const urlParamSize = new URL(window.location.href).searchParams.size;
-if (urlParamSize <= 0) {
-  window.location.href = '?test1=test_value_1&test2=test_value_2';
-}
 /**
  * importing all libs that are tested in ci tests
  * add lib name below, let the IDE resolve the actual import
--- a/extension/chrome/dev/ci_unit_test.ts
+++ b/extension/chrome/dev/ci_unit_test.ts
@@ -23,6 +23,11 @@ import { PgpHash } from '../../js/common/core/crypto/pgp/pgp-hash.js';
 import { PgpArmor } from '../../js/common/core/crypto/pgp/pgp-armor.js';
 import { Xss } from '../../js/common/platform/xss.js';
 
+/* Used for test "Catcher does not include query string on report" */
+const urlParamSize = new URL(window.location.href).searchParams.size;
+if (urlParamSize <= 0) {
+  window.location.href = '?test1=test_value_1&test2=test_value_2';
+}
 /**
  * importing all libs that are tested in ci tests
  * add lib name below, let the IDE resolve the actual import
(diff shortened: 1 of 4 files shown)


martgil:
@sosnovsky Good afternoon. I would like to add a test for this one but unfortunately I don't have sufficient idea on how I could test/evaluate for the value of `ErrorReport` before it gets sent to the shared-tenant-fes's error collector.

I can see the following below but I don't know how I could populate the `reportedErrors` in the test: 
https://github.com/FlowCrypt/flowcrypt-browser/blob/9c1ff8a7e0867a287df825e1da90651daf133914/test/source/test.ts#L141-L195

If its not too much trouble, may I hear any of your suggestion. thank you.

sosnovsky:
> I can see the following below but I don't know how I could populate the `reportedErrors` in the test:

For adding error you should trigger some error in browser extension - as example you can check comments on different `.filter` expressions at https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/test.ts#L150, there you can find names of tests which trigger errors.

martgil:
@sosnovsky thank you very much. I get it now. The error I used to test was not typically reproducible in a normal basis. For example, I'm able to trigger the error reporting by submitting a private key in attester via the extension's correct public key record -> change the public key -> invalid private key. 

I think its difficult to test since it request a real error to trigger? That makes me wonder if should there be a deliberate button to trigger a sample error for example like in the android app where there is an experimental feature to deliberately crash the android app and that sends a error reports.

sosnovsky:
I think it'll be easier to implement unit test for testing this functionality - https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/tests/browser-unit-tests/unit-Catch.js

martgil:
Hello @sosnovsky - thank you very much for your help. I added a test but I have a little problem. The page where the error triggers is at `ci_unit_test.htm` which does not have a query parameter so the test was ineffective.

https://github.com/FlowCrypt/flowcrypt-browser/pull/5292/files#diff-74ef3cf27375d014484496aa98d4736d0780e1415766b97a5f92c3294d283750R295

 I come up with the idea of adding a test query parameter to `ci_unit_test.htm` as a resolution.

sosnovsky:
> Hello @sosnovsky - thank you very much for your help. I added a test but I have a little problem. The page where the error triggers is at `ci_unit_test.htm` which does not have a query parameter so the test was ineffective.
> 
> https://github.com/FlowCrypt/flowcrypt-browser/pull/5292/files#diff-74ef3cf27375d014484496aa98d4736d0780e1415766b97a5f92c3294d283750R295
> 
> I come up with the idea of adding a test query parameter to `ci_unit_test.htm` as a resolution.

Good solution, maybe we can just add some test parameter at https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/tests/unit-browser.ts#L64, so it'll be `chrome/dev/ci_unit_test.htm?test=1`?

martgil:
> Good solution, maybe we can just add some test parameter at https://github.com/FlowCrypt/flowcrypt-browser/blob/master/test/source/tests/unit-browser.ts#L64, so it'll be `chrome/dev/ci_unit_test.htm?test=1`?

Thanks - I'm for something like this. This must be the best way option.

sosnovsky (approved):
Works well 👍

PR #5294, opened 2023-07-13
allow to update a private key from file

### LINKED ISSUE #5291: Add ability to update private key from file
Currently users can update private key only by directly pasting armored key.
We need option to update private key by loading it from file (there was customer feedback that it's easier for users to load file than open it in text editor and then copy armored text)

Adding new private key has both options, so we probably can reuse UI from it.

rrrooommmaaa:
This PR allows to update a key from file

close #5291

- Tests added or updated

### To be filled by reviewers


extension/chrome/settings/modules/my_key_update.ts
```diff
@@ -101,11 +103,14 @@ View.run(
...
     private updatePrivateKeyHandler = async () => {
-      const updatedKey = await KeyUtil.parse(String(this.inputPrivateKey.val()));
       const updatedKeyEncrypted = await KeyUtil.parse(String(this.inputPrivateKey.val()));
+      const updatedKey = await KeyUtil.parse(KeyUtil.armor(updatedKeyEncrypted)); // create a "cloned" copy to decrypt later
       const updatedKeyPassphrase = String($('.input_passphrase').val());
+      KeyImportUi.allowReselect();
       if (typeof updatedKey === 'undefined') {
         await Ui.modal.warning(Lang.setup.keyFormattedWell(this.prvHeaders.begin, String(this.prvHeaders.end)), Ui.testCompatibilityLink);
+      } else if (updatedKeyEncrypted.identities.length === 0) {
+        await Ui.modal.error(Lang.setup.prvHasUseridIssue);
```
rrrooommmaaa, line 113: This page is now using `KeyImportUi` for some visual handling, but it's not using `KeyImportUi.checkPrv` as it is not easy to adapt it here (we can try this in a separate issue)

Also, what should the message be if there is something wrong with UserIds (we currently don't support UserIds that don't contain email address in them)
I added this text
image

rrrooommmaaa: @sosnovsky is there a vision on whether we should support keys with uids without emails?

rrrooommmaaa: We currently support `identities` missing email address for public keys of recipients, but `compose` module assumes `emails[0]` matches `identities[0]` but that's not necessarily true. I suggest to refactor this part. Instead of "synchronized" arrays we can have an array of objects of the type returned by `Str.parseEmail`, it has `{ email, name, full }` -- everything that's required. `full` is the original `identity`

sosnovsky: > @sosnovsky is there a vision on whether we should support keys with uids without emails?

I think it's quite rare case and we didn't get any requests for such functionality, so we can leave it as is for now. Then we'll probably also need to update our mobile apps to support such keys.

> We currently support identities missing email address for public keys of recipients, but compose module assumes emails[0] matches identities[0] but that's not necessarily true. I suggest to refactor this part. 

That sounds good 👍

sosnovsky (approved):
Well done 👍

PR #5299, opened 2023-07-15
frame-level dest for confirmation_result and pgp_block_render messages

### LINKED ISSUE #5298: Firefox pgp_block rendering is broken after #5266
The implemented messaging mechanism only works in Chrome.
In Firefox, the pgp_block simply contains `Loading...` and never gets updated

rrrooommmaaa:
This PR specifies tabId:frameId in messaging when sending a message from a content script to an iframe within it.

close #5298

issue #1509 (part of refactoring)

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/attachment.ts
+++ b/extension/chrome/elements/attachment.ts
@@ -25,8 +25,8 @@ import { AttachmentWarnings } from './shared/attachment_warnings.js';
 export class AttachmentDownloadView extends View {
   public fesUrl?: string;
   public confirmationResultResolver?: (confirm: boolean) => void;
+  public readonly parentTabId: string;
   protected readonly acctEmail: string;
-  protected readonly parentTabId: string;
   protected readonly frameId: string;
   protected readonly origNameBasedOnFilename: string;
   protected readonly isEncrypted: boolean;
@@ -85,8 +85,8 @@ export class AttachmentDownloadView extends View {
     this.gmail = new Gmail(this.acctEmail);
   }
 
-  public getParentTabId = () => {
-    return this.parentTabId;
+  public getDest = () => {
+    return this.tabId;
   };
 
   public render = async () => {
--- a/extension/chrome/elements/pgp_block.ts
+++ b/extension/chrome/elements/pgp_block.ts
@@ -4,7 +4,7 @@
 
 import { Url } from '../../js/common/core/common.js';
 import { Assert } from '../../js/common/assert.js';
-import { RenderMessage, RenderMessageWithFrameId } from '../../js/common/render-message.js';
+import { RenderMessage } from '../../js/common/render-message.js';
 import { Attachment } from '../../js/common/core/attachment.js';
 import { Xss } from '../../js/common/platform/xss.js';
 import { PgpBlockViewAttachmentsModule } from './pgp_block_modules/pgp-block-attachmens-module.js';
@@ -12,9 +12,9 @@ import { PgpBlockViewErrorModule } from './pgp_block_modules/pgp-block-error-mod
 import { PgpBlockViewPrintModule } from './pgp_block_modules/pgp-block-print-module.js';
 import { PgpBlockViewQuoteModule } from './pgp_block_modules/pgp-block-quote-module.js';
 import { PgpBlockViewRenderModule } from './pgp_block_modules/pgp-block-render-module.js';
-import { Ui } from '../../js/common/browser/ui.js';
+import { CommonHandlers, Ui } from '../../js/common/browser/ui.js';
 import { View } from '../../js/common/view.js';
-import { Bm } from '../../js/common/browser/browser-msg.js';
+import { BrowserMsg } from '../../js/common/browser/browser-msg.js';
 
 export class PgpBlockView extends View {
   public readonly acctEmail: string; // needed for attachment decryption, probably should be refactored out
@@ -27,6 +27,7 @@ export class PgpBlockView extends View {
   public readonly errorModule: PgpBlockViewErrorModule;
   public readonly renderModule: PgpBlockViewRenderModule;
   public readonly printModule = new PgpBlockViewPrintModule();
+  private tabId!: string;
 
   public constructor() {
     super();
@@ -41,21 +42,14 @@ export class PgpBlockView extends View {
     this.quoteModule = new PgpBlockViewQuoteModule(this);
     this.errorModule = new PgpBlockViewErrorModule(this);
     this.renderModule = new PgpBlockViewRenderModule(this);
-    chrome.runtime.onMessage.addListener((message: Bm.Raw) => {
-      if (message.name === 'pgp_block_render') {
-        const msg = message.data.bm as RenderMessageWithFrameId;
-        if (msg.frameId === this.frameId) {
-          this.processMessage(msg);
-          return true;
-        }
-      }
-      return false;
-    });
-    window.addEventListener('load', () => window.parent.postMessage({ readyToReceive: this.frameId }, '*'));
   }
 
+  public getDest = () => {
+    return this.tabId;
+  };
+
   public render = async () => {
-    //
+    this.tabId = await BrowserMsg.requiredTabId();
   };
 
   public setHandlers = () => {
@@ -63,6 +57,12 @@ export class PgpBlockView extends View {
       'click',
       this.setHandler(() => this.printModule.printPGPBlock())
     );
+    BrowserMsg.addListener('pgp_block_render', async (msg: RenderMessage) => {
+      this.processMessage(msg);
+    });
+    BrowserMsg.addListener('confirmation_result', CommonHandlers.createConfirmationResultHandler(this));
+    BrowserMsg.listen(this.tabId);
+    BrowserMsg.send.pgpBlockReady(this.parentTabId, { frameId: this.frameId, tabId: this.tabId });
   };
 
   private processMessage = (data: RenderMessage) => {
--- a/extension/chrome/elements/pgp_block_modules/pgp-block-attachmens-module.ts
+++ b/extension/chrome/elements/pgp_block_modules/pgp-block-attachmens-module.ts
@@ -7,7 +7,7 @@ import { Attachment } from '../../../js/common/core/attachment.js';
 import { Browser } from '../../../js/common/browser/browser.js';
 import { BrowserMsg } from '../../../js/common/browser/browser-msg.js';
 import { PgpBlockView } from '../pgp_block.js';
-import { CommonHandlers, Ui } from '../../../js/common/browser/ui.js';
+import { Ui } from '../../../js/common/browser/ui.js';
 import { Xss } from '../../../js/common/platform/xss.js';
 import { KeyStore } from '../../../js/common/platform/store/key-store.js';
 import { XssSafeFactory } from '../../../js/common/xss-safe-factory.js';
@@ -20,10 +20,6 @@ export class PgpBlockViewAttachmentsModule {
   public includedAttachments: Attachment[] = [];
 
   public constructor(private view: PgpBlockView) {}
-
-  public getParentTabId = () => {
-    return this.view.parentTabId;
-  };
 
   public renderInnerAttachments = (attachments: Attachment[], isEncrypted: boolean) => {
     Xss.sanitizeAppend('#pgp_block', '<div id="attachments"></div>');
@@ -78,8 +74,6 @@ export class PgpBlockViewAttachmentsModule {
         }
       })
     );
-    BrowserMsg.addListener('confirmation_result', CommonHandlers.createConfirmationResultHandler(this));
-    BrowserMsg.listen(this.view.parentTabId);
   };
 
   private previewAttachmentClickedHandler = async (attachment: Attachment) => {
@@ -102,7 +96,7 @@ export class PgpBlockViewAttachmentsModule {
         type: encrypted.type,
         data: decrypted.content,
       });
-      if (await AttachmentWarnings.confirmSaveToDownloadsIfNeeded(attachment, this)) {
+      if (await AttachmentWarnings.confirmSaveToDownloadsIfNeeded(attachment, this.view)) {
         Browser.saveToDownloads(attachment);
       }
     } else {
--- a/extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
+++ b/extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts
@@ -145,7 +145,7 @@ export class PgpBlockViewRenderModule {
   public renderSignatureOffline = () => {
     this.renderSignatureStatus('error verifying signature: offline, click to retry').on(
       'click',
-      this.view.setHandler(() => window.parent.postMessage({ retry: this.view.frameId }, '*'))
+      this.view.setHandler(() => BrowserMsg.send.pgpBlockRetry(this.view.getDest(), { frameId: this.view.frameId }))
     );
   };
 }
--- a/extension/chrome/settings/inbox/inbox.ts
+++ b/extension/chrome/settings/inbox/inbox.ts
@@ -24,7 +24,6 @@ import { XssSafeFactory } from '../../../js/common/xss-safe-factory.js';
 import { AcctStore } from '../../../js/common/platform/store/acct-store.js';
 import { RelayManager } from '../../../js/common/relay-manager.js';
 import { MessageRenderer } from '../../../js/common/message-renderer.js';
-import { Env } from '../../../js/common/browser/env.js';
 
 export class InboxView extends View {
   public readonly inboxMenuModule: InboxMenuModule;
@@ -62,16 +61,11 @@ export class InboxView extends View {
     this.inboxNotificationModule = new InboxNotificationModule(this);
     this.inboxActiveThreadModule = new InboxActiveThreadModule(this);
     this.inboxListThreadsModule = new InboxListThreadsModule(this);
-    window.addEventListener('message', e => {
-      if (e.origin === Env.getExtensionOrigin()) {
-        this.relayManager.handleMessageFromFrame(e.data);
-      }
-    });
   }
 
   public render = async () => {
     this.tabId = await BrowserMsg.requiredTabId();
-    this.relayManager = new RelayManager(this.tabId, this.debug);
+    this.relayManager = new RelayManager(this.debug);
     this.factory = new XssSafeFactory(this.acctEmail, this.tabId);
     this.injector = new Injector('settings', undefined, this.factory);
     this.webmailCommon = new WebmailCommon(this.acctEmail, this.injector);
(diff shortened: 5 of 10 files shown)


sosnovsky (commented):
It works well for me in Firefox, but now stucks on `Loading...` in Chrome and then works well after reloading Gmail page. Also it seems to happen only with account which stores passphrase in memory and needs to render `Enter passphrase` block before decrypting message. Can you please check if the same issue happens for you?

rrrooommmaaa:
> click 'Finish session' the block get stuck in `Loading...`

In gmail page the "reload" of pgp_block frames is not implemented -- I totally missed it.
So we can implement this in a separate PR.
I refactored `RelayManager` a bit, so that its `MutationObserver` can find a reloaded `frameId` -- a frame present in both `removedNodes` and `addedNodes`. This can be the mechanism to restart the frame processing, but for this a reference to the processing function should be stored somewhere -- need to think it through and implement.
For now I disabled the test of this case for the gmail page. We can also simply display something like `Please reload the page` there.

sosnovsky:
> Rendering Enter passphrase works ok, however when I click 'Finish session' the block get stuck in Loading...
I caught that in this test code
and going to TDD it.
Is this your case?

No, for me it stucks when loading `Enter passphrase` block:

https://github.com/FlowCrypt/flowcrypt-browser/assets/1062093/586059ac-0267-4e3e-b63d-d19eee9e884f

It renders well after I manually reload page.

sosnovsky:
> In gmail page the "reload" of pgp_block frames is not implemented -- I totally missed it.
So we can implement this in a separate PR.
I refactored RelayManager a bit, so that its MutationObserver can find a reloaded frameId -- a frame present in both removedNodes and addedNodes. This can be the mechanism to restart the frame processing, but for this a reference to the processing function should be stored somewhere -- need to think it through and implement.
For now I disabled the test of this case for the gmail page. We can also simply display something like Please reload the page there.

Can we do it in the current PR? So this issue won't be in `master`

sosnovsky:
Also strange that this issue doesn't appear in Firefox, `Enter passphrase` block is rendered there without page reload

rrrooommmaaa:
> Also strange that this issue doesn't appear in Firefox, `Enter passphrase` block is rendered there without page reload

Are you saying that `Loading...` isn't stuck if you open a signed-only message.
So this is the tab opened with the older version of Flowcrypt installed, and it behaves like this after you update Flowcrypt?

sosnovsky:
> Are you saying that Loading... isn't stuck if you open a signed-only message?

I just checked - it's the same for signed-only and for encrypted messages, stuck on `Loading...`

> So this is the tab opened with the older version of Flowcrypt installed, and it behaves like this after you update Flowcrypt?

No, it's the version with the latest changes from these PR. I closed chrome, built latest code, started chrome and tried to open encrypted message.

P.S. I just was able to open message without stucking - it happens when I click message on inbox page very fast after opening mail.google.com, probably before full page is loaded, then message is rendered without issue. Maybe it's caused by added `addEventListener('load')`?

rrrooommmaaa:
> > In gmail page the "reload" of pgp_block frames is not implemented -- I totally missed it.
> > So we can implement this in a separate PR.
> > I refactored RelayManager a bit, so that its MutationObserver can find a reloaded frameId -- a frame present in both removedNodes and addedNodes. This can be the mechanism to restart the frame processing, but for this a reference to the processing function should be stored somewhere -- need to think it through and implement.
> > For now I disabled the test of this case for the gmail page. We can also simply display something like Please reload the page there.
> 
> Can we do it in the current PR? So this issue won't be in `master`

This issue is already in `master`. Solving this in this PR will make the PR bigger

rrrooommmaaa:
> Maybe it's caused by added `addEventListener('load')`?

Only if `load` event never happens in the pgp_block iframe, and that's weird.



https://github.com/FlowCrypt/flowcrypt-browser/assets/23335944/de94c23a-8bbe-45ee-8dea-0e08025582b2

I'm not able to reproduce it. Here is my video. Starting Chrome with ci.test.gmail mail page and clicking on a message.

It'd be great if you are able to create a failing test for this issue.

sosnovsky:
> I'm not able to reproduce it. Here is my video. Starting Chrome with ci.test.gmail mail page and clicking on a message.

Here message is rendered without entering passphrase, it works for me in this case too.
It stucks on `Loading...` only when I need to enter passphrase to decrypt message. Can you try to reproduce it with your `@flowcrypt.com` account?

rrrooommmaaa:
> It stucks on `Loading...` only when I need to enter passphrase to decrypt message.

You previously said a signed-only message also failed

rrrooommmaaa:
> Can we do it in the current PR? So this issue won't be in `master`

I implemented restarts of pgp_blocks on gmail page with the test c1ecc12c9e6196a98b53244018b618006676d88c

sosnovsky:
> You previously said a signed-only message also failed

Yes, it fails too, but only for account who doesn't store passphrase locally. I can try to create failing test for this case tomorrow

sosnovsky:
> You previously said a signed-only message also failed

Yes, it fails too, but only for account who doesn't store passphrase locally. I can try to create failing test for this case tomorrow

sosnovsky:
> I implemented restarts of pgp_blocks on gmail page with the test https://github.com/FlowCrypt/flowcrypt-browser/commit/c1ecc12c9e6196a98b53244018b618006676d88c

I tested with the latest changes - it still stucks on `Loading...`. Test passes because it uses custom FlowCrypt inbox page, it works well for me there too, but stucks on Gmail inbox page. Trying to add test with this behavior

sosnovsky:
Adding such test seems to be complicated, as our test infrastructure currently doesn't support browser restarts.

Here my steps to reproduce this bug:
1) configure FlowCrypt extension to not store passphrase locally
2) restart browser
3) go to gmail inbox
4) try to open encrypted or signed message - it'll stuck on `Loading...`
5) if I reload page - stuck is gone, `Enter passphrase` block is rendered for encrypted message, or decrypted text for signed message

rrrooommmaaa:
> Can you try to reproduce it with your `@flowcrypt.com` account?

The messages are looking fine on my Flowcrypt account. I have `Enter passphrase` invitations all right. I don't think the issue is related to passphrase, especially given that signed-only messages also fail in your setup.
Are you using Mac there? Are there any interesting errors in the browser console?
I can add some additional error output, maybe it'll help.

sosnovsky:
> The messages are looking fine on my Flowcrypt account. I have Enter passphrase invitations all right. I don't think the issue is related to passphrase, especially given that signed-only messages also fail in your setup.

I sent video with stuck issue to your email, may be you'll notice something there.

> Are you using Mac there? Are there any interesting errors in the browser console?
I can add some additional error output, maybe it'll help.

I checked console logs and there were no errors, it'll be helpful if you can add some additional logging

rrrooommmaaa:
So it looks like the content script gets into a faulty state on each start of browser on Mac, and no matter how many messages you open, they all are stuck. Until reload. Is this correct?

rrrooommmaaa:
> > So it looks like the content script gets into a faulty state on each start of browser on Mac, and no matter how many messages you open, they all are stuck. Until reload. Is this correct?
> 
> Yes, correct, after page reload it works well. Also message loads without stuck if I open it with direct link, like https://mail.google.com/mail/u/0/#inbox/FMfcgzGtwCwm...

Are these symptoms on `master` branch too?

sosnovsky:
> Are these symptoms on master branch too?

No, it works well on `master`, that's why I asked to fix this issue in current PR

rrrooommmaaa:
@sosnovsky Can you check if the issue still persists on this branch. If so, can you check `pgp_block.htm` frames urls, namely `parentTabId` parameter.
Can you also see if `issue-1509-window-messaging-from-iframe-to-parent-live-test` branch works better -- there are some additional refactorings on top of `issue-5298-tab-and-frame-dest-messaging-live-test` -- I was planning to merge it later.

sosnovsky:
> Can you also see if `issue-1509-window-messaging-from-iframe-to-parent-live-test` branch works better -- there are some additional refactorings on top of `issue-5298-tab-and-frame-dest-messaging-live-test` -- I was planning to merge it later.

I checked this branch - stuck issue happens there too.

> Can you check if the issue still persists on this branch. If so, can you check `pgp_block.htm` frames urls, namely `parentTabId` parameter.

I added some logging and found where this bug happens - https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-5298-tab-and-frame-dest-messaging-live-test/extension/js/common/relay-manager.ts#L121
Here `frameData` doesn't have `tabId` property for cases when loading stucks:

correct `frameData`:
Screenshot 2023-07-20 at 23 47 52

stuck `frameData`:
Screenshot 2023-07-20 at 23 47 37

sosnovsky:
> Every `frameData` starts like this, then upon receiving a `pgp_block_ready` message from the `pgp_block`, it is "activated" and has `tabId`.

I logged `sendCatch` in `browser-msg.ts`, `tabId` is set there:

```
sendCatch pgp_block_ready to 496727799:6, bm: {"frameId":"frame_OPEUZwzufc","tabId":"496727799:13"}
```

> Have you checked `parentTabId` URL parameter of the failing `pgp_block.htm` frame? Is it looking correct?

Yes, I logged these values in `setHandlers` of `pgp_block.ts`:
```
parentTabId = 496727799:6, frameId = frame_OPEUZwzufc, tabId = 496727799:13
```

rrrooommmaaa:
> parentTabId = 496727799:6,

This looks unusual, the main page usually ends with `:0`.
After reload (when everything works fine), does `parentTabId` look like `...:0`?

rrrooommmaaa:
If the gmail page's frameId is non-zero for some reason, we should modify this line
https://github.com/FlowCrypt/flowcrypt-browser/blob/035092eb11f98ff8759df880c26956603eeb03f4/extension/js/common/browser/browser-msg.ts#L442
so that it checks again the registry of content scripts (the content script can identify itself as a content script with an additional parameter in the `_tab_` message) instead of comparing to `0`

rrrooommmaaa:
Does the issue still exist after the update?

sosnovsky:
Yes, issue is still there. 

I re-checked logs and your assumption is correct - messages load well when `parentFrameId` ends with `:0`, when it has different value - loading is stuck. I found another place where we use fixed `:0` - https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-5298-tab-and-frame-dest-messaging-live-test/extension/js/background_page/bg-handlers.ts#L40, probably we should make it dynamic there too?

> If the gmail page's frameId is non-zero for some reason, we should modify this line

I added logging to `shouldRelayMsgToOtherPage`, but looks like it isn't called when rendering message (for both, successful and stuck loading)

rrrooommmaaa:
> I found another place where we use fixed `:0` - https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-5298-tab-and-frame-dest-messaging-live-test/extension/js/background_page/bg-handlers.ts#L40, probably we should make it dynamic there too?

We can, but I think this only affects `ajax_progress` handlers, so not critical.
Interesting, does a real top-level `:0` frame exist at this moment and why it isn't trying to wire the handlers, or is it trying to do this at the same time?
It'd be interesting to try this: if `initInternalVars` received a non-null frameId in `requireTabId`, assume that this is a wrong frame to install `GmailElementReplacer` and stop further actions, etc.
There are probably several frames competing to install the handlers.

sosnovsky:
I tested with your latest commit and now it doesn't render even `Loading...` label when frameId doesn't end with `...:0`.
It shows raw encrypted message, like this:
```
-----BEGIN PGP MESSAGE-----
Version: FlowCrypt Email Encryption 8.4.8

wV4D3QcbXeD33qgSAQdAnht5aOkRM2/5xt4om4qpCWQhpEhSlciMgne1pYd
QDswalUndLjeSAAikj3X923W/rpA9zSloc0LWVkru5p2X+ARqZwlTpeDqn7
....
```

rrrooommmaaa:
well, this means that the main google page is somehow not a top-level frame.
what could be the reason for this?
How does the top-level frame look like in this case? is there some frame containing this `:6` gmail frame?

rrrooommmaaa:
Can we display a notification saying something like "The mail page is loaded in a nested frame, please reload the page"?

Fixing the issue right now may not be a good idea because the messaging system is going to be re-worked as part of #1509 -- the background page will be removed, most of the methods like `shouldRelayMsgToOtherPage` will also be either removed or seriously modified, so it's not very clear where we'll end up.

As for the failing test for it -- I think we can implement it in mock environment by loading a page that has the `gmail.localhost` page inside a frame.

sosnovsky:
> How does the top-level frame look like in this case? is there some frame containing this :6 gmail frame?

I haven't found such frame, and didn't notice any difference with the page which has `...:0` frame. But I'll recheck it tomorrow

> Can we display a notification saying something like "The mail page is loaded in a nested frame, please reload the page"?

No, it won't be a nice user experience when they'll need to reload page to make decryption work. And it won't be just 1 time, but a regular reload, probably every day.

> Fixing the issue right now may not be a good idea because the messaging system is going to be re-worked as part of https://github.com/FlowCrypt/flowcrypt-browser/issues/1509 -- the background page will be removed, most of the methods like shouldRelayMsgToOtherPage will also be either removed or seriously modified, so it's not very clear where we'll end up.

Yeah, it'll be remade, but it can take a long time, as migrating to manifest v3 will be quite complex transition and won't take 1 week. Also currently everything works well in Chrome in `master`, but fails in Firefox, and vice versa in this PR, maybe we can just find some solution which will fix Firefox stuck while not breaking Chrome?

rrrooommmaaa:
@tomholub I think we can save some costs not implementing the fix on the framework that's about to be phased out

tomholub:
Meaning, the fix for this won't matter once we stop using background page?

Would it be possible then to instead try to work on a PR that transfers this particular bit of broken functionality to content script instead?

rrrooommmaaa:
> Meaning, the fix for this won't matter once we stop using background page?

Tabs learn their addresses via sending a `_tab_` message to `background` page, and the `background` page responds with a message containing the sender `tabId` and `frameId`. When the `background` page is gone, we'll need some substitution for this mechanism or seriously re-organize messaging, including the address system.

However, in PR #5320 there are additional changes on top of this PR -- `pgp_block_ready` message is posted via `postMessage` to the parent window, so I guess, we should at least get the `tabId` set in `frameData`. @sosnovsky please confirm

tomholub:
> > Meaning, the fix for this won't matter once we stop using background page?
> 
> Tabs learn their addresses via sending a `_tab_` message to `background` page, and the `background` page responds with a message containing the sender `tabId` and `frameId`. When the `background` page is gone, we'll need some substitution for this mechanism or seriously re-organize messaging, including the address system.
> 
> However, in PR #5320 there are additional changes on top of this PR -- `pgp_block_ready` message is posted via `postMessage` to the parent window, so I guess, we should at least get the `tabId` set in `frameData`. @sosnovsky please confirm

I think the rough idea of how this is used, from memory, is:
1) content script -> background page: what's my id? bg -> cs: it's 123
2) content script creates a frame (like a message frame) with a parentTabId=123
3) frame wants to communicate with parent tab, so it sends a message to:123

Alternative A: isn't there a browser messaging mechanism that just sends to:parent natively? That would be cool

Alternative B: Each page could choose their own ID randomly. Then all messages could be broadcast. In the payload of the message could be id of intended recipient tab. Tab's listening code can filter incoming messages by that id. With that, the sequence is:

1) content script chooses id 123
2) creates a frame with parentTabId=123
3) frame sends to: broadcast, {toTabId: 123, content:{hello: "there"}}
4) all frames receive the message, but only the parent actually processes it / responds

sosnovsky:
@rrrooommmaaa it seems we're mixing `frameId` and `tabId`, just to make it clear:
page loads well - `parentTabId = 496729555:0, frameId = frame_EenpbuUGFZ, tabId = 496729555:15`
loading stuck - `parentTabId = 496729600:28, frameId = frame_EDtrJKOgNP, tabId = 496729600:35`
I logged it at [setHandlers](https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-5298-tab-and-frame-dest-messaging-live-test/extension/chrome/elements/pgp_block.ts#L65)

so `frameId` doesn't have any suffix like `:0` in the end, it's added only to tab ids.
also chatgpt gave me such information about tab ids in chrome:

> In Chrome, each tab and extension runs in a separate process called a "renderer process." When you have multiple tabs open, each tab is assigned a unique process ID (PID) for its renderer process. The renderer process is responsible for rendering the web page content and executing the JavaScript code for that specific tab.
> 
> The PID is essentially a numerical identifier that distinguishes one process from another in the operating system. When you see tab IDs in Chrome, they are often represented as a combination of the process ID (PID) and the "tab index" within that process. The format of this ID is usually PID:Tab_Index.
> 
> For example:
> 
> Tab ID for the first tab in the process with PID 123 might be: 123:0
> Tab ID for the second tab in the same process might be: 123:1
> The reason you see different numbers after the colon (:) is that when a new tab is opened, Chrome may choose to reuse an existing renderer process for resource optimization, especially if the new tab is related to the same domain or shares some resources with an existing tab. As a result, the new tab gets assigned the next available tab index within that process.
> 
> So, if you observe different numbers after the colon in the tab ID (e.g., :0, :6, etc.), it means that those tabs are running in different renderer processes, and their unique numbers are the tab indexes within those respective processes. The process IDs are usually allocated based on availability and resource management by the Chrome browser.

rrrooommmaaa:
> page loads well - `parentTabId = 496729555:0, frameId = frame_EenpbuUGFZ, tabId = 496729555:15`
> loading stuck - `parentTabId = 496729600:28, frameId = frame_EDtrJKOgNP, tabId = 496729600:35`

`BrowserMsg.tabId` and `BrowserMsg.requiredTabId` actually return `tab.id:frameId` from `MessageSender`: https://developer.chrome.com/docs/extensions/reference/runtime/#type-MessageSender
and, as stated here, frameId is ` 0 for top-level frames, positive for child frames.` 

So that's unexpected to have frameId=28 for the top level frame.

`frameId = frame_EDtrJKOgNP` -- this is totally different. It's an identifier assigned before the actual frame is created.
This is why we need to associate it with `tabId`

rrrooommmaaa:
> Alternative A: isn't there a browser messaging mechanism that just sends to:parent natively? That would be cool

We can send non-sensitive data this way.

> Alternative B: Each page could choose their own ID randomly. Then all messages could be broadcast.

We can also consider this. The question is -- how should we split this into issues/PRs?
This PR was meant to fix Firefox (it's broken on `master`), now it's getting bigger and bigger.

tomholub:
> > Alternative A: isn't there a browser messaging mechanism that just sends to:parent natively? That would be cool
> 
> We can send non-sensitive data this way.
> 
> > Alternative B: Each page could choose their own ID randomly. Then all messages could be broadcast.
> 
> We can also consider this. The question is -- how should we split this into issues/PRs? This PR was meant to fix Firefox (it's broken on `master`), now it's getting bigger and bigger.

after this PR gets in master in some form, I think it would be appropriate to consider Alternative B for the next PR against master, that sounds like a relatively clean refactor just alone. Even if handlers otherwise remain in bg after that PR?

After that, maybe move handers from bg to content script. If it's easy and direct, then a bunch. If it's difficult and requires more changes, then I suppose one by one in separate PRs. You can use judgement to make PRs more bite sized when you see the opportunity and when it doesn't add too much work to do so.

PR #5305, opened 2023-07-17
rendering after load event

rrrooommmaaa:
This PR ensures that view rendering happens after 'load' event

### To be filled by reviewers

### DIFF
--- a/extension/js/common/view.ts
+++ b/extension/js/common/view.ts
@@ -11,11 +11,15 @@ export abstract class View {
   public static run<VIEW extends View>(viewClass: new () => VIEW) {
     try {
       const view = new viewClass();
-      (async () => {
-        await view.render();
-        await Promise.resolve(view.setHandlers()); // allow both sync and async
-        View.setTestViewStateLoaded();
-      })().catch(View.reportAndRenderErr);
+      window.addEventListener('load', async () => {
+        try {
+          await view.render();
+          await Promise.resolve(view.setHandlers()); // allow both sync and async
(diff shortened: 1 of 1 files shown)


extension/js/common/view.ts
```diff
@@ -11,11 +11,15 @@ export abstract class View {
   public static run<VIEW extends View>(viewClass: new () => VIEW) {
     try {
       const view = new viewClass();
-      (async () => {
-        await view.render();
-        await Promise.resolve(view.setHandlers()); // allow both sync and async
-        View.setTestViewStateLoaded();
-      })().catch(View.reportAndRenderErr);
+      window.addEventListener('load', async () => {
```
rrrooommmaaa, line 14: We sometimes might want to run some actions after the `load` event (e.g. in `setHandlers`).
Looks like this event isn't tracked anywhere  in `View` class, is it?
In most cases the view is loaded by the time `setHandlers` is executed -- I think it probably happens during `requiredTabId` execution -- but it's not explicitly checked and not all `View` implementations use it.
Should we be on the safe side by modifying the app like this for all view, @sosnovsky?

sosnovsky: Looks good, I tested and haven't found any issues after adding `addEventListener`, so we can use it for all views to be sure that everything is loaded before running `setHandlers`.

Did you notice any issues caused by missing check for `load` event before? As I didn't remember any problems with it

sosnovsky (approved):

PR #5320, opened 2023-07-21
Issue 1509 window messaging from iframe to parent live test

rrrooommmaaa:
This PR uses postMessage from child frame to parent (on top of #5299)

issue #1509

- Not worth testing
- Tests added or updated

### To be filled by reviewers


rrrooommmaaa:
@sosnovsky `frameData.tabId` should be set even for non-zero top-level frameId

sosnovsky:
> @sosnovsky `frameData.tabId` should be set even for non-zero top-level frameId

I tested this issue and loading stuck is still there.

Further logging showed that `pgp_block_ready` msg isn't received at https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-1509-window-messaging-from-iframe-to-parent-live-test/extension/js/content_scripts/webmail/setup-webmail-content-script.ts#L223, but it's sent from https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-1509-window-messaging-from-iframe-to-parent-live-test/extension/chrome/elements/pgp_block.ts#L65:

```
DID SET HANDLERS: parentFrameId = 496731892:19, frameId = frame_HnOQqmYrJa, tabId = 496731892:26
SEND pgp_block_ready msg to 496731892:19 with {"frameId":"frame_HnOQqmYrJa","tabId":"496731892:26"}
```

sosnovsky:
Hmm, seems like when there are 2 opened Gmail tabs - first tab receives `pgp_block_ready` messages sent from second tab, that's why loading is stuck on second tab:
1) I opened tab 1 with mail.google.com
2) opened encrypted message
3) it showed `Enter passphrase` block
4) I opened tab 2 with mail.google.com
5) opened encrypted message
6) `Loading...` stuck on tab 2
7) I checked console on tab 1 - it received `pgp_block_ready` message which was sent to tab 2

sosnovsky:
I changed `BrowserMsg.send.pgpBlockReady(this.parentTabId, ...)` to `BrowserMsg.send.pgpBlockReady('broadcast',  ...)` at https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-1509-window-messaging-from-iframe-to-parent-live-test/extension/chrome/elements/pgp_block.ts#L65 and now loading stuck is gone, but clicking `Enter passphrase` works only on first tab (probably will work after changing receiver to `broadcast` too).

Will changing `this.parentTabId` to `'broadcast'` cause any side issues or it can be an appropriate solution?

I also checked Firefox - everything works well there, even with multiple tabs.

rrrooommmaaa:
> Further logging showed that `pgp_block_ready` msg isn't received at https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-1509-window-messaging-from-iframe-to-parent-live-test/extension/js/content_scripts/webmail/setup-webmail-content-script.ts#L223, but it's sent from https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-1509-window-messaging-from-iframe-to-parent-live-test/extension/chrome/elements/pgp_block.ts#L65

Oh I didn't finish the wiring via `BrowserMsg.sendToParentWindow`, my bad. Should be fixed now. This PR was initially an experiment based on `confirmation_show`-`confirmation_result` communication with response sent as a new message. And also using `postMessage` for sending non-sensitive data.

rrrooommmaaa:
> I checked console on tab 1 - it received pgp_block_ready message which was sent to tab 2

Really don't understand how this is possible. Really messy Chrome implementation on Mac with a non-zero top-level `frameId` to begin with.

rrrooommmaaa:
> Will changing `this.parentTabId` to `'broadcast'` cause any side issues or it can be an appropriate solution?
> I also checked Firefox - everything works well there, even with multiple tabs.

These messages are relayed by the `background` page, as far as I can see. Such as #4264

rrrooommmaaa:
So should we merge it?

sosnovsky (approved):

extension/js/common/browser/browser-msg.ts
```diff
@@ -242,7 +255,7 @@ export class BrowserMsg {
     notificationShow: (dest: Bm.Dest, bm: Bm.NotificationShow) => BrowserMsg.sendCatch(dest, 'notification_show', bm),
     notificationShowAuthPopupNeeded: (dest: Bm.Dest, bm: Bm.NotificationShowAuthPopupNeeded) =>
       BrowserMsg.sendCatch(dest, 'notification_show_auth_popup_needed', bm),
-    showConfirmation: (dest: Bm.Dest, bm: Bm.ShowConfirmation) => BrowserMsg.sendCatch(dest, 'confirmation_show', bm),
+    showConfirmation: (bm: Bm.ShowConfirmation) => BrowserMsg.sendToParentWindow('confirmation_show', bm, 'confirmation_result'),
```
tomholub, line 258: Let's rename this to `sendToParentWindowUnsafely` ?

Is this truly unsafe? What is the issue - does host page js receive this too?

extension/js/common/browser/browser-msg.ts
```diff
@@ -325,39 +341,28 @@ export class BrowserMsg {
...
-            // sometimes received events get duplicated
-            // while first event is being processed, second even will arrive
-            // that's why we generate a unique id of each request (uid) and filter them above to identify truly unique requests
-            // if we got here, that means we are handing a duplicate request
-            // we'll indicate will respond = true, so that the processing of the actual request is not negatively affected
-            // leaving it at "false" would respond with null, which would throw an error back to the original BrowserMsg sender:
-            // "Error: BrowserMsg.sendAwait(pgpMsgDiagnosePubkeys) returned(null) with lastError: (no lastError)"
-            // the duplication is likely caused by our routing mechanism. Sometimes browser will deliver the message directly as well as through bg
-            return true;
+  // this is a means to get messages from child iframes
```
tomholub, line 344: Add comment about being insecure and why and how?

extension/js/common/browser/browser-msg.ts
```diff
@@ -425,6 +430,47 @@ export class BrowserMsg {
     });
   };
 
+  private static sendToParentWindow = (name: string, bm: Dict<unknown> & { messageSender: Bm.Dest }, responseName?: string) => {
```
tomholub, line 433: Add comment about being insecure, how and why.

sosnovsky:
> So should we merge it?

Yes, looks like both issues, on firefox and chrome, are fixed, so this PR is ok to be merged.
I'll let you know if these issues will appear, but hope they won't, currently works great 👍

extension/js/common/browser/browser-msg.ts
rrrooommmaaa: yes, a handler from the host js or another extension installed to this page will be able to see the contents of this message

tomholub: I wonder if we could just encrypt & sign these messages, and then use this as the messaging without worry? Both receiving frame and sending frame will have access to local storage, where they can find the same symmetric password, let's say.

Then we can brutally simplify all of that browser messaging dispatch spaghetti.

PR #5321, opened 2023-07-23
#5308 Fix the error when sending password-protected emails that contain emojis

### LINKED ISSUE #5308: Unable to send password messages if the contact name retrieved from contacts.google.com includes emoji
Error message:
```
--------- BUG REPORT ----------

[name]
composer: send: bad request (errMsg: Bad recipient)

[stack]
    at BrowserExtension.prepareBugReport (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/js/common/browser/browser-extension.js:8:44)
    at ComposeErrModule.handleSendErr (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/compose-modules/compose-err-module.js:115:61)
    at async ComposeSendBtnModule.extractProcessSendMsg (chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/compose-modules/compose-send-btn-module.js:117:21)

url: chrome-extension://bnjglocicdkmhmoohhfkfkbbkejdhdgc/chrome/elements/compose.htm?frameId=frame_FyyyfGeQfc&acctEmail=[SCRUBBED]&parentTabId=241918943%3A0


[error]
{
  "status": 400,
  "url": "https://flowcrypt.com/shared-tenant-fes/api/v1/message",
  "responseText": "{\"code\":400,\"message\":\"Bad recipient\",\"details\":\"Invalid RFC 5322 mailbox User Name⭐️ Real Name <user@corp.test>\"}",
  "statusText": "Bad Request",
  "resMsg": "Bad recipient",
  "resDetails": "Invalid RFC 5322 mailbox User Name⭐️ Real Name <user@corp.test>"
}

[details]
{}
```
Steps to reproduce:
1. Create a contact at contacts.google.com and name it `Example ⭐️ Name`, fill in the email address field and save.
2. Use FlowCrypt's secure compose and compose a password protected message for `Example ⭐️ Name`. Click enable google contact search and grant the required permission.
3. Finish composing the secure email, set a password, and send the email and you'll get the following error:
image
4. Click 'ok' and you'll get the following stack trace above when you are about to report the issue.

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGtwDDWqBNzHzXpZhWdNvJlBXNs

martgil:
This PR adds `.stripEmojis()` to strip out emojis of the recipient contact's name.

close #5308 

- Tests added or updated

### To be filled by reviewers


martgil:
Manual test:

[Screencast from 2023-07-29 14-57-12.webm](https://github.com/FlowCrypt/flowcrypt-browser/assets/46025304/b8502b30-7bcc-41c2-b361-28be767d670f)

martgil:
@sosnovsky ready for review. thank you.

extension/js/common/platform/xss.ts
```diff
@@ -230,6 +230,15 @@ export class Xss {
     return str.replace(/&/g, '&amp;').replace(/"/g, '&quot;').replace(/'/g, '&#39;').replace(/</g, '&lt;').replace(/>/g, '&gt;').replace(/\//g, '&#x2F;');
   };
 
+  public static stripEmojis = (str: string) => {
+    const emojiRegex = /[\u{1F300}-\u{1F64F}\u{2700}-\u{27B0}\u{23E9}-\u{23FA}\u{2194}-\u{21AA}\u261D\u26A1\u263A-\uD83E\uD83E]/gu;
```
sosnovsky, line 234: It'll be better to use https://github.com/mathiasbynens/emoji-regex for emoji regex, as periodically some new emojis are added and our static regex won't include them, while this lib is regularly updated to include all existing emojis.
For example, this regex doesn't detect 🥴 emoji

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -245,13 +245,13 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
       /* eslint-disable @typescript-eslint/no-non-null-assertion */
       switch (recipient.sendingType) {
         case 'to':
-          result.to!.push({ email: recipient.email, name: recipient.name });
+          result.to!.push({ email: recipient.email, name: Xss.stripEmojis(recipient.name || '') });
```
sosnovsky, line 248: let's move recipient object creation outside of `switch` statement, so we won't need to duplicate code for `to`, `cc` and `bcc`:
```ts
const recipientObj = { email: recipient.email, name: Xss.stripEmojis(recipient.name || '') };
switch (recipient.sendingType) {
    case 'to':
        result.to!.push(recipientObj);
        break;
    case 'cc':
...
```

sosnovsky (commented):
Works well 👍 Just a couple small improvements needed

extension/js/common/platform/xss.ts
martgil: Thank you for the remarks. Yes, I actually have problems with the ranges of emojis to remove - I'll check the repo you've shared and implement it here.

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: thank you - will do.

extension/js/common/platform/xss.ts
martgil: @sosnovsky After updating the `tsconfig.json` and including this line `"emoji-regex": ["../node_modules/emoji-regex/index.d.ts", "../node_modules/emoji-regex/index.js", "COMMENT"]` in the compilerOption -> paths, I'm still getting the following error: 
> Cannot find module 'emoji-regex'. Did you mean to set the 'moduleResolution' option to 'nodenext', or to add aliases to the 'paths' option?

I have checked `node_modules/emoji-regex` and those files are present. I'm sorry, may I hear your suggestions on how I could proceed further?

martgil: I think I have found it out. Thanks!

extension/js/common/platform/xss.ts
```diff
@@ -235,6 +235,7 @@ export class Xss {
   };
 
   public static stripEmojis = (str: string) => {
+    console.log('ci test');
```
martgil: just added and remove this temporarily to somehow fixed CI test not working correctly aftering clicking 'rerun' which doesn't produce the desired effect.

martgil:
ready for review again. thank you.

martgil:
marking back to draft state a better solution can be discussed at https://github.com/FlowCrypt/flowcrypt-browser/issues/5308

martgil:

extension/js/common/platform/xss.ts
```diff
@@ -230,6 +232,15 @@ export class Xss {
...
   };
 
+  public static stripEmojis = (str: string) => {
+    if (Intl !== undefined && typeof Intl.Segmenter === 'function') {
+      str = [...new Intl.Segmenter().segment(str)].map(x => x?.segment).join('');
+    } else {
+      str = str.replace(Xss.EMOJI_REGEX, '');
+    }
+    return this.escape(str);
+  };
```
sosnovsky, line 242: Sorry for confusion - we don't need to use `Intl.Segmenter` here, as in `emailjs-mime-codec.js` it's used only for separating emojis from other characters (as simple string `split` doesn't work for emojis).

So here it's not needed and this method can be simplified to just:
```js
const stringWithoutEmojis = str.replace(Xss.EMOJI_REGEX, '');
return this.escape(stringWithoutEmojis);
```

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -243,15 +243,16 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
     const result: ParsedRecipients = { to: [], cc: [], bcc: [] };
     for (const recipient of recipients) {
       /* eslint-disable @typescript-eslint/no-non-null-assertion */
+      const recipientObj = { email: recipient.email, name: Xss.stripEmojis(recipient.name || '') };
```
sosnovsky, line 246: It seems here we remove emojis even for recipients who have encryption (so we won't use FES for sending messages), and emoji in recipient name should work for them.

Probably it'll be enough to remove emojis just on creating password-protected message before sending it to FES.

test/source/tests/compose.ts
```diff
@@ -66,6 +66,23 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+        await BrowserRecipe.setupCommonAcctWithAttester(t, browser, 'compatibility', {
+          google: { acctAliases: flowcryptCompatibilityAliasList },
+        });
+        const recipientEmail = 'User ⭐ Name <test@email.com>';
+        const msgPwd = 'super hard password for the message';
+        const subject = 'Strip emojis in display name format';
+        const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
+        await ComposePageRecipe.selectFromOption(composePage, acct);
+        await ComposePageRecipe.fillMsg(composePage, { to: recipientEmail }, subject);
+        await ComposePageRecipe.sendAndClose(composePage, { password: msgPwd });
```
sosnovsky, line 82: This test doesn't check if emojis are stripped (as it doesn't fail when we use `Intl.Segmenter`, which doesn't remove emojis), because initial error was on FES side and here we use mock endpoints which doesn't return error when recipient name contains emoji.

I think we should check that name of recipient in sent message doesn't contain emoji (by checking it in `GoogleData` - [example](https://github.com/FlowCrypt/flowcrypt-browser/blob/ab31c96d09b3a2f86f82cfbbbd55fbd5c28cc172/test/source/tests/compose.ts#L1883))

extension/chrome/elements/compose-modules/compose-input-module.ts
martgil: Upon retesting, the emoji needs to be stripped even in this phase otherwise the error will occur when the email is about to be sent and reaches the FES server. I'll try to look up on other solution.

extension/js/common/platform/xss.ts
martgil: Ah, yes -- I will remove the usage of `Intl.Segmenter` and will simplify the function as suggested. thanks again.

test/source/tests/compose.ts
martgil: Right - that approach will be much more sensible unlike the one I wrote which only confirmed that the issue is fixed when no error occurs during email sent. I will update the test.

extension/chrome/elements/compose-modules/compose-input-module.ts
sosnovsky: I think emojis should be stripped when formatting password messages in [formatSendablePwdMsgs](https://github.com/FlowCrypt/flowcrypt-browser/blob/ab31c96d09b3a2f86f82cfbbbd55fbd5c28cc172/extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts#L94)

martgil: Ah yes, that is right - I have updated the code and manipulated the `newMsg` object inside the formatSendablePwdMsgs function. Thank you!

martgil:

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -346,7 +346,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       const recipient = orderedRecipients[processed];
       const escapedTitle = Xss.escape(recipient.element.getAttribute('title') || '');
       const nameOrEmail = recipient.name || recipient.email || recipient.invalid || '';
-      const emailHtml = `<span class="email_address ${recipient.element.className}" title="${escapedTitle}">${Xss.escape(nameOrEmail)}</span>`;
+      const emailHtml = `<span class="email_address ${recipient.element.className}" title="${escapedTitle}">${Xss.stripEmojis(nameOrEmail)}</span>`;
```
sosnovsky, line 349: in initial review I meant that we don't need too use `stripEmojis` each time when recipient is rendered in UI. as issue happens only when sending message to recipient with emoji using FES.

so in UI we can leave emojis and strip them only when formatting password-protected message for FES. it'll be simpler, but will fix original issue

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -96,6 +96,16 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
...
+    const { to, cc, bcc } = newMsg.recipients;
+    const updateContactName = (recipient: { email: string; name: string }) => {
+      return {
+        ...recipient,
+        name: Xss.stripEmojis(recipient.name),
+      };
+    };
+    newMsg.recipients.to = to?.map(updateContactName);
+    newMsg.recipients.cc = cc?.map(updateContactName);
+    newMsg.recipients.bcc = bcc?.map(updateContactName);
```
sosnovsky, line 108: I checked code for uploading messages to FES, and we already apply some formatting to recipients before upload: https://github.com/FlowCrypt/flowcrypt-browser/blob/c0cb65990e174951b31464bf71e863bef4640685/extension/js/common/api/account-servers/external-service.ts#L146

Let's move `stripEmojis` there and apply it to formatted email:
```ts
to: (recipients.to || []).map(Str.formatEmailWithOptionalName).map(Xss.stripEmojis),
...
```

test/source/tests/compose.ts
```diff
@@ -66,6 +68,27 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
...
+        const msgPwd = 'super hard password for the message';
+        const subject = 'Test Sending Message With Recipient Name Contains Emoji';
+        const composePage = await ComposePageRecipe.openStandalone(t, browser, 'compatibility');
+        await ComposePageRecipe.selectFromOption(composePage, acct);
+        await ComposePageRecipe.fillMsg(composePage, { to: recipientEmail }, subject);
+        await ComposePageRecipe.sendAndClose(composePage, { password: msgPwd });
+        const sentMsg = (await GoogleData.withInitializedData(acct)).searchMessagesBySubject(subject)[0];
+        const rawMessage = await Parse.convertBase64ToMimeMsg(sentMsg.raw!);
+        const toHeader = rawMessage.headers.get('to') as AddressObject;
+        expect(!toHeader.text.includes('⭐')).to.equal(true);
```
sosnovsky, line 88: let's make it simpler:
```ts
expect(toHeader.text).to.not.include('⭐');
```

extension/chrome/elements/compose-modules/compose-recipients-module.ts
martgil: Oh, okay - I just thought it would be great to have a consistent UI but yes, I'm reverting that change. thank you.

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
martgil: Thanks - adding .map() that would perform the `Xss.stripEmojis` callback function is a better solution and with less code to write.

test/source/tests/compose.ts
martgil: Will do - thank you. In addition to this, I have to write the test in the mock endpoint for shared-tenant-fes `/shared-tenant-fes/api/v1/message` since now it will be in the mock endpoint's request body. thanks again.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
sosnovsky, line 349: no need for `stripEmojis` here, let's change it back to `escape`

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -808,7 +808,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
         }
         displayEmail = '<div class="select_contact_email" data-test="action-select-contact-email">' + Xss.escape(displayEmail) + '</div>';
         if (contact.name) {
-          ulHtml += '<div class="select_contact_name" data-test="action-select-contact-name">' + Xss.escape(contact.name) + displayEmail + '</div>';
+          ulHtml += '<div class="select_contact_name" data-test="action-select-contact-name">' + Xss.stripEmojis(contact.name) + displayEmail + '</div>';
```
sosnovsky, line 811: no need for `stripEmojis` here, let's change it back to `escape`

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -905,7 +905,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       const recipientId = this.generateRecipientId();
       const recipientsHtml =
         `<span tabindex="0" id="${recipientId}" data-test="${recipientId}">` +
-        `<span class="recipient-name">${Xss.escape(name || '')}</span>` +
+        `<span class="recipient-name">${name || ''}</span>` +
```
sosnovsky, line 908: `Xss.escape` is needed here

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -1056,7 +1056,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
     } else if (info && info.sortedPubkeys.length) {
       if (info.info.name) {
         recipient.name = info.info.name;
-        $(el).find('.recipient-name').text(Xss.escape(info.info.name));
+        $(el).find('.recipient-name').text(Xss.stripEmojis(info.info.name));
```
sosnovsky, line 1059: no need for `stripEmojis` here, let's change it back to `escape`

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -1097,7 +1097,7 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
       $(el).addClass('no_pgp');
       if (info?.info.name) {
         recipient.name = info.info.name;
-        $(el).find('.recipient-name').text(Xss.escape(info.info.name));
+        $(el).find('.recipient-name').text(Xss.stripEmojis(info.info.name));
```
sosnovsky, line 1100: no need for `stripEmojis` here, let's change it back to `escape`

extension/js/common/api/account-servers/external-service.ts
```diff
@@ -143,9 +144,9 @@ export class ExternalService extends Api {
         JSON.stringify({
           associateReplyToken,
           from,
-          to: (recipients.to || []).map(Str.formatEmailWithOptionalName),
-          cc: (recipients.cc || []).map(Str.formatEmailWithOptionalName),
-          bcc: (recipients.bcc || []).map(Str.formatEmailWithOptionalName),
+          to: (recipients.to || []).map(Str.formatEmailWithOptionalName).map(Xss.stripEmojis),
+          cc: (recipients.cc || []).map(Str.formatEmailWithOptionalName).map(Xss.stripEmojis),
+          bcc: (recipients.bcc || []).map(Str.formatEmailWithOptionalName).map(Xss.stripEmojis),
```
sosnovsky, line 149: looks nice 👍

extension/chrome/elements/compose-modules/compose-recipients-module.ts
martgil: will do - thanks!

extension/chrome/elements/compose-modules/compose-recipients-module.ts
martgil: ok - will do. thanks.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
martgil: ok, will do - thanks.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
martgil: ok, will do - thanks.

extension/chrome/elements/compose-modules/compose-recipients-module.ts
martgil: updated - thanks.

extension/js/common/api/account-servers/external-service.ts
martgil: yes, it is. thank you.

test/source/mock/google/strategies/send-message-strategy.ts
```diff
@@ -472,6 +472,8 @@ export class TestBySubjectStrategyContext {
       this.strategy = new SaveMessageInStorageStrategy();
     } else if (subject.includes('Test Sending Message With Attachment Which Contains Emoji in Filename')) {
       this.strategy = new SaveMessageInStorageStrategy();
+    } else if (subject.includes('Test Sending Message With Recipient Name Contains Emoji')) {
+      this.strategy = new SaveMessageInStorageStrategy();
```
sosnovsky, line 476: Probably this one not needed, as now we're checking recipient in shared-tenant endpoint

martgil: Aha - yes, it is not needed now. I'll remove it.

martgil:
@sosnovsky This should be all good and ready to merge. thanks!

sosnovsky (approved):
Yeah, all good now 👍

Issue #5329, opened 2023-07-26
don't rely on browser's tab.id and frameId in messaging


rrrooommmaaa:
> > Alternative B: Each page could choose their own ID randomly. Then all messages could be broadcast.
> 

after this PR gets in master in some form, I think it would be appropriate to consider Alternative B for the next PR against master, that sounds like a relatively clean refactor just alone.

rrrooommmaaa:
Closed by #5356

PR #5330, opened 2023-07-26
updates related to #1509

### LINKED ISSUE #5373: Canceling passphrase dialog sometimes leads to glitches
I rarely see an unexpected decryption error like this
```
----- mockBackendData Catch.reportErr -----21:30
name: OperationError21:30
message: 21:30
url: https://gmail.localhost:33675/gmail/17daefa0eb077da621:30
trace: Error21:30
    at <async> decrypt (/extension/js/common/symmetric-message-encryption.ts:42:26)21:30
    at <async> <anonymous> (/extension/js/common/browser/browser-msg.ts:349:17)21:30
### Catch.reportErr calling stack ###21:30
#     at Catch.formatExceptionForReport (chrome-extension://kimlbdocogdofnhjfioniflhebimmknh/js/content_scripts/webmail_bundle.js:279:90)21:30
#     at Catch.onErrorInternalHandler (chrome-extension://kimlbdocogdofnhjfioniflhebimmknh/js/content_scripts/webmail_bundle.js:87:29)21:30
#     at Catch.reportErr (chrome-extension://kimlbdocogdofnhjfioniflhebimmknh/js/content_scripts/webmail_bundle.js:103:18)21:30
# 21:30
# url: https://gmail.localhost:33675/gmail/17daefa0eb077da621:30
# 21:30
######################
```
It may be somehow connected with `PassphraseView.closeDialog` sending two messages at once.

rrrooommmaaa:
Trying to fix #5373 (to be merged after PR #5372)

- Tests added or updated

### To be filled by reviewers


rrrooommmaaa:
@tomholub 

These issues are encountered:
1) broadcast messages sent with `chrome.runtime.sendMessage` don't reach content script's handler
2) in Firefox, broadcast messages sent with `chrome.runtime.sendMessage` from the content script don't reach any frames (only background)

tomholub:
who are the senders of broadcast messages? i imagine a pass phrase frame, as an example. that one could send two messages, one broadcast and one message to parent.

tomholub:
deleted my comment. what if we used the insecure messaging everywhere, does it easily reach where we need it? 

then we encrypt & sign, decrypt & verify.

rrrooommmaaa:
Somewhat easier, I think.
You said to take the symmetric key for encryption from the local storage.
There might be problems addressing local storage from the content script, not sure.
We have this comment:
https://github.com/FlowCrypt/flowcrypt-browser/blob/6d6a3404cc94c273b6f46e829bb58197813f22e0/extension/js/common/platform/store/acct-store.ts#L88-L90

Also, how should we keep session passphrases? Keeping them in the content script would mean that opening several gmail tabs would require entering passphrases into each of them independently, right?

tomholub:
> Also, how should we keep session passphrases? Keeping them in the content script would mean that opening several gmail tabs would require entering passphrases into each of them independently, right?

I'm ok with that limitation. We could then reintroduce a background worker later, if it turns out a problem.

>    // extension storage can be disallowed in rare cases for content scripts throwing 'Error: Access to extension API denied.' 
   // go through bg script to avoid such errors 

Could have been fixed on recent versions? We could just try it and see what happens. Alternatively, we could try `window.localStorage` instead, in case it can be adequately accessed from everywhere. Worst case, each content script would make a worker, and access the extension storage through its worker.

rrrooommmaaa:
> then we encrypt & sign, decrypt & verify.

You were previously talking about using a symmetric key. Can you elaborate on `sign/verify` part?

tomholub:
I don't have anything that specific in mind, but surely window.crypto has some way to encrypt things in a manner that can be verified to be authentic when decrypting.

Roughly we could send something like `{destination: [list of receiving frames], content: encryptedData}`

Where the encryptedData also contains the destination. That way, receiving frames don't have to decrypt it to decide if they should drop the message or not, while at the same time they can still verify that the message and destination is authentic and matching once they do decrypt it.

Then the content would just be an encrypted/serialised json of whatever today is in these messages.

Just a rough idea.

rrrooommmaaa:
some API is not available for the content script. e.g. `chrome.tabs`, so some features aren't possible without the background page, e.g. to open a settings page from this notification
image
What should we do about it?

tomholub:
> some API is not available for the content script. e.g. `chrome.tabs`, so some features aren't possible without the background page, e.g. to open a settings page from this notification
> image
> What should we do about it?

I could just make it a regular `a href target=_blank` ? I forgot why we have to do this through background page.

rrrooommmaaa:
> I forgot why we do this through background page.

To disallow opening `settings` tab multiple times?

tomholub:
> > I forgot why we do this through background page.
> 
> To disallow opening `settings` tab multiple times?

Sounds not so important, ok to not worry about that

rrrooommmaaa:
We also use `chrome.windows.create` to create Oauth popup.
What should we do? Embed it in a frame?

rrrooommmaaa:
this mainly involves `reconnectAcctAuthPopup` process initiated by the content  script

tomholub:
> We also use `chrome.windows.create` to create Oauth popup.
> What should we do? Embed it in a frame?

> this mainly involves `reconnectAcctAuthPopup` process initiated by the content script

This will not be easy, imo it warrants a separate PR. So I'd leave that as is for now, from background script.

sosnovsky:
Maybe then it'll be better to implement migration to native `fetch` in another PR?
So in this PR you'll implement fix for original issue https://github.com/FlowCrypt/flowcrypt-browser/issues/5329 and then we can find solution for this `fetch` migration (or just leave it for later).

It's possible to allow inline scripts using nonce or hashes - https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_inline_script, but not sure if it's possible in our case.

tomholub:
I agree, if moving fetch() is giving you trouble, it would be wise to leave it where it is for now, and take it one mergeable step at a time.

On firefox it kind of looks like the type of problem that could be resolved by adjusting the manifest, but I'm not 100% sure.

ioanmo226:
@sosnovsky Is this PR still relevant?

sosnovsky:
> @sosnovsky Is this PR still relevant?

I think it can be closed, as initial issue is not reproducible anymore.

PR #5332, opened 2023-07-27
#5331 Address localStorage from content script directly

### LINKED ISSUE #5331: Direct access of local storage from content script
>    // extension storage can be disallowed in rare cases for content scripts throwing 'Error: Access to extension API denied.' 
   // go through bg script to avoid such errors 

Could have been fixed on recent browser versions? That's probably from 6 years ago. We could just try it and see what happens.

rrrooommmaaa:
This PR is preparation to address localStorage directly, without using the background page

close #5331

### To be filled by reviewers


rrrooommmaaa:
Should we merge this PR separately?

sosnovsky (approved):
I didn't notice any issues while testing, will let you know if something works incorrectly

PR #5341, opened 2023-07-31
#5340 Check if attachment is decrypted before downloading

### LINKED ISSUE #5340: Fix error when downloading attachment from PGP/MIME message
When trying to download attachment from PGP/MIME message there is an error message shown:

> There was a problem decrypting this file (format: Error: Error during parsing. This message / key probably does not conform to a valid OpenPGP format.). Downloading encrypted original.

But attachment is downloaded decrypted. It happens because extension thinks that attachment is encrypted, and tries to decrypt it again.

sosnovsky:
This PR adds check if attachment is decrypted before trying to download it, to fix error message which appeared when extension tried to decrypt already decrypted attachment.

close #5340

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/pgp_block_modules/pgp-block-attachments-module.ts
+++ b/extension/chrome/elements/pgp_block_modules/pgp-block-attachments-module.ts
@@ -60,7 +60,7 @@ export class PgpBlockViewAttachmentsModule {
         event.stopPropagation();
         const attachment = this.includedAttachments[Number($(target).attr('index'))];
         if (attachment.hasData()) {
-          await this.decryptAndSaveAttachmentToDownloads(attachment);
+          await this.downloadAttachment(attachment);
         } else {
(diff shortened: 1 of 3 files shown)


extension/chrome/elements/pgp_block_modules/pgp-block-attachments-module.ts
```diff
@@ -82,6 +82,16 @@ export class PgpBlockViewAttachmentsModule {
     BrowserMsg.send.showAttachmentPreview(this.view.parentTabId, { iframeUrl });
   };
 
+  private downloadAttachment = async (attachment: Attachment) => {
```
tomholub, line 85: Let's call it `decryptIfNeededAndDownloadAttachment`

tomholub (approved):

PR #5369, opened 2023-08-18
#5368 Remove trailing hyphens on signature

### LINKED ISSUE #5368: Don't add "--" separator if user signature begins with "--"

martgil:
This PR removes the trailing slash from the email signature.

close #5368 

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-footer-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-footer-module.ts
@@ -33,7 +33,7 @@ export class ComposeFooterModule extends ViewModule<ComposeView> {
 
   public createFooterHtml = (footer: string) => {
     const sanitizedPlainFooter = Xss.htmlSanitizeAndStripAllTags(footer, '\n');
-    const sanitizedHtmlFooter = sanitizedPlainFooter.replace(/\n/g, '<br>');
+    let sanitizedHtmlFooter = sanitizedPlainFooter.replace(/\n/g, '<br>');
     const footerFirstLine = sanitizedPlainFooter.split('\n')[0];
     if (!footerFirstLine) {
       return '';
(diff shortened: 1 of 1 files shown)


extension/chrome/elements/compose-modules/compose-footer-module.ts
```diff
@@ -33,14 +33,16 @@ export class ComposeFooterModule extends ViewModule<ComposeView> {
...
+    let sanitizedHtmlFooter = sanitizedPlainFooter.replace(/\n/g, '<br>');
     const footerFirstLine = sanitizedPlainFooter.split('\n')[0];
     if (!footerFirstLine) {
       return '';
     }
     if (/^[*\-_=+#~ ]+$/.test(footerFirstLine)) {
       return sanitizedHtmlFooter; // first line of footer is already a footer separator, made of special characters
     }
+    sanitizedHtmlFooter =
+      sanitizedHtmlFooter.startsWith('&nbsp;--') || sanitizedHtmlFooter.startsWith('--') ? sanitizedHtmlFooter.replace('--', '') : sanitizedHtmlFooter;
```
sosnovsky, line 45: Looks like we already checking if footer starts with separator at line 41, but for some reason this regex didn't work in user's case.

Let's just remove `&nbsp;` from `footerFirstLine`, so it'll look like:
```js
const footerFirstLine = sanitizedPlainFooter.split('\n')[0]?.replace('&nbsp;', '');
```

martgil: @sosnovsky thank you very much. I have tested the updated code and the regex works after the removal of the `&nbsp;` and I get the desired result.

sosnovsky: Were you able to reproduce original issue reported by user? As for me it works well even with additional newline on the beginning of signature

martgil: Yes, the new line doesn't seem to interfere with my testing case as well. Perhaps there is a non-breaking space lying around somewhere there?

sosnovsky: On screenshots it looks like plain signature, without any non-standard characters

PR #5371, opened 2023-08-18
#5317 Add custom authentication config to local store

### LINKED ISSUE #5317: add custom authentication config to local store
- [x] A new type should be created `DomainConfiguration = { authentication: clientConfigurationJson: ClientConfigurationJson }` and also `AuthenticationConfiguration = { oauth: {  clientId: string, clientSecret: string?, redirectUrl: string, authCodeUrl: string, tokensUrl: string, } }`. 
 - [x] `ExternalService.fetchAndSaveClientConfiguration` to be updated: It should first call `/api/v1/client-configuration/authentication?domain=${domain}` (NOT authenticated, add a parameter to `request`), and save the whole API response (which can be empty on default deployments, or contain the `{oauth: ...}` type) into storage with `await AcctStore.set(this.acctEmail, { authentication: r });`. And after that proceed with the rest of the method
 - [x] Update FES mock by adding this `/authenticate` endpoint, which always returns an empty json, except when otherwise configured for a particular test. 
 - [x] Add a test which will have some specific made up values configured in the FES mock, and verify these values from extension settings page -> `local_store` link on the bottom, which should newly also display the `authentication` contents

martgil:
This PR Adds a custom authentication data configuration to a local store.

close #5317 

- Tests added or updated

### To be filled by reviewers


martgil:
@sosnovsky ready for review. thank you.

test/source/tests/setup.ts
```diff
@@ -2479,9 +2479,56 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+        const debugFrame = await SettingsPageRecipe.awaitNewPageFrame(settingsPage, '@action-show-local-store-contents', ['debug_api.htm']);
+        await debugFrame.waitForContent('@container-pre', 'authentication');
+        const key = `cryptup_${emailKeyIndex(acctEmail, 'authentication')}`;
+        const auth = (await settingsPage.getFromLocalStorage([key]))[key];
+        const { oauth } = auth as FesAuthenticationConfiguration;
+        expect(oauth.clientId).to.be.not.empty;
+        expect(oauth.clientSecret).to.be.not.empty;
+        expect(oauth.redirectUrl).to.be.not.empty;
+        expect(oauth.authCodeUrl).to.be.not.empty;
+        expect(oauth.tokensUrl).to.be.not.empty;
```
sosnovsky, line 2511: I think it'll be better to compare these properties with real values, not just check if they're not empty. We can even move oauth config to separate variable and then compare with it:
```ts
const oauthConfig = {
  redirectUrl: 'https://example.com/redirect',
  clientId: 'your-client-id',
  clientSecret: 'your-client-secret',
  authCodeUrl: 'https://example.com/auth',
  tokensUrl: 'https://example.com/tokens',
};
t.context.mockApi!.configProvider = new ConfigurationProvider({
           fes: {
             authenticationConfiguration: {
               oauth: oauthConfig
             },
           },
         });
...
const { oauth } = auth as FesAuthenticationConfiguration;
expect(oauth).to.eq(oauthConfig);
```

sosnovsky (commented):
Well done 👍

test/source/tests/setup.ts
martgil: thank you very much - the one you have suggested is by far the best way to check JSON values instead of the way I tried to test them. Thanks again - I learned a lot.

sosnovsky: You're welcome :)

sosnovsky (dismissed):

test/source/tests/setup.ts
sosnovsky: Looks like `to.eq` doesn't work for objects, just for simple values. Let's try to use `to.deep.equal`

martgil: Thanks - `to.deep.equal` seems to be used across the codebase and worked for checking objects.

martgil:
@sosnovsky ready for review again. thank you.

sosnovsky (approved):
👍

PR #5372, opened 2023-08-20
#1509 fetch() implementation

rrrooommmaaa:
This PR replaces JQuery ajax calls with fetch()
Modifies the mock API to use HTTP/2 -- required to display fetch upload progress.

issue #1509

### To be filled by reviewers


rrrooommmaaa:
https://github.com/FlowCrypt/flowcrypt-browser/pull/5330#issuecomment-1672077204

@tomholub This seems to work in Chrome, but Firefox still doesn't allow to call `fetch` from the content script.
How should we wrap this PR up?

tomholub:
Is there a change in manifest file that could allow this?

rrrooommmaaa:
> Is there a change in manifest file that could allow this?

I think no.

rrrooommmaaa:
I tried it now and, yes, looks like the message is successfully fetched from `googleapis` by `webmail_bundle.js` -- though I can't see these requests in Network log for some reason.
To reproduce we need a message with encryption-related data in an attachment, e.g. PGP/MIME message.
I'm also getting errors when trying to fetch a private key from EKM from `webmail_bundle.js` , are you getting them?
```
error: -1 when GET-ing https://ekm.flowcrypt.com/v1/keys/private (no body):  -> (no standard err msg)


provided ajax call stack:
KeyManager/this.getPrivateKeys@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:7720:34
contentScriptSetupIfVacant/startPullingKeysFromEkm/<@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:15664:78
Catch.try/<@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:118:23
setTimeout handler*Catch.setHandledTimeout@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:238:19
startPullingKeysFromEkm@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:15661:31
async*entrypoint@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:15767:27
async*contentScriptSetupIfVacant@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:15836:23
gmailWebmailStartup@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:15980:23
@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:15995:15
Catch.try/<@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:118:23
@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:16000:3
@moz-extension://b71981f4-f479-4233-94ba-ee2af0217449/js/content_scripts/webmail_bundle.js:16003:3


url: https://mail.google.com/mail/u/0/#inbox/FMfcgzGtwgpbwjXsBqVBZlGDRWlxCtMR
```

rrrooommmaaa:
@sosnovsky `api.js` file name tells us that the request was originated by an extension page or frame, it should be `webmail_bundle.js` instead when originated by the content script -- though these requests seem to be missing entirely from the Network log.
The `startPullingKeysFromEkm()` call is supposed to happen when you open `mail.google.com` -- does it give errors.
If it gives an error, it can probably be fixed by adjusting EKM headers, so we can merge this PR as it is, if everyone else is happy with it.

extension/js/common/api/account-servers/external-service.ts
```diff
@@ -150,50 +149,52 @@ export class ExternalService extends Api {
       ),
     });
     const multipartBody = { content, details };
-    return await this.request<FesRes.MessageUpload>('POST', `/api/${this.apiVersion}/message`, {}, multipartBody, { upload: progressCb });
+    return await this.request<FesRes.MessageUpload>(`/api/${this.apiVersion}/message`, { fmt: 'FORM', data: multipartBody }, { upload: progressCb });
```
rrrooommmaaa, line 152: If upload progress needs to be supported for this call, should convert the body to 'JSON' the same way as in `Gmail.msgSend` and make sure FES supports HTTP/2

extension/js/common/api/shared/api.ts
```diff
@@ -60,37 +83,152 @@ export class Api {
...
+            body = JSON.stringify(req.data);
+            headersInit.push(['Content-Type', 'application/json; charset=UTF-8']);
+          } else if (req.dataType === 'TEXT') {
+            if (supportsRequestStreams && req.progress?.upload) {
+              const upload = req.progress?.upload;
+              const transformStream = new TransformStream();
+              uploadPromise = async () => {
+                const transformWriter = transformStream.writable.getWriter();
+                for (let offset = 0; offset < req.data.length; ) {
+                  const chunkSize = Math.min(1000, req.data.length - offset);
```
rrrooommmaaa, line 121: Perhaps, it makes sense increase the upload chunk size?

sosnovsky:
> The startPullingKeysFromEkm() call is supposed to happen when you open mail.google.com -- does it give errors.
If it gives an error, it can probably be fixed by adjusting EKM headers, so we can merge this PR as it is, if everyone else is happy with it.

I reproduced error from your issue - it didn't appear for me if I just open gmail.com and browse messages, but if I open encrypted message and then reload page - `/keys/private` error appears in logs.
Which EKM headers should be changed to fix this error? I can try to change them locally and test if error is gone

extension/js/common/api/account-servers/external-service.ts
sosnovsky: Looks like currently FES supports only HTTP/1.1:

Screenshot 2023-08-23 at 22 39 39

rrrooommmaaa:
I can move the ajax call back to the background page.
@tomholub ?

sosnovsky:
> I can move the ajax call back to the background page.

But then we'll still need to move it to the content script, as background page will be removed after migrating to manifest v3?

extension/js/common/api/account-servers/external-service.ts
tomholub: Isn't upload progress achievable regardless of http version? We know how much data we intend to send, and we surely know how much data we've sent so far.

extension/js/common/api/shared/api.ts
```diff
@@ -60,37 +83,152 @@ export class Api {
...
+      headers: headersInit,
+      body,
+      duplex,
+      mode: 'cors',
+      signal: abortController.signal,
+    };
+    let readyState = 1; // OPENED
+    const reqContext = { url: req.url, method: req.method, data: body, stack: req.stack };
+    try {
+      const fetchPromise = fetch(url, requestInit);
```
tomholub, line 154: what's the benefit of `fetch()` over `$.ajax` ?

tomholub:
> > I can move the ajax call back to the background page.
> 
> But then we'll still need to move it to the content script, as background page will be removed after migrating to manifest v3?

Does the same problem happen when using `$.ajax`? The fetching code is quite complicated. HTTP requests are complicated. It could be that the code is wrong, but it only manifests on Firefox, and only sometimes. Has an attempt been made to see exactly where and how the requests are falling through with regards to that fetching code?

If we are sure it's not our code, then it's a Firefox bug. Can you still reproduce it in Firefox nightly or similar latest build? If yes, it should be packaged into a minimal demo code, and submitted to bugzilla.

extension/js/common/api/shared/api.ts
rrrooommmaaa: This was done per this migration instruction [Replace XMLHttpRequest() with global fetch()](https://developer.chrome.com/docs/extensions/migrating/to-service-workers/#replace-xmlhttprequest)

extension/js/common/api/account-servers/external-service.ts
rrrooommmaaa: fetch upload streaming currently has only limited support, only in Chrome > 105 and HTTP/2

rrrooommmaaa:
> Does the same problem happen when using `$.ajax`? 

`$.ajax` doesn't allow CORS request from the content script, this is why we have `ajax` handler in the background page
https://github.com/FlowCrypt/flowcrypt-browser/blob/c4cc7985bc01f483796c57d25281b0feb79571b0/extension/js/common/api/shared/api.ts#L64-L66

rrrooommmaaa:
> Has an attempt been made to see exactly where and how the requests are falling through with regards to that fetching code?

I can't see any requests from the content script in Network log in Firefox for some reason....

extension/js/common/api/shared/api.ts
tomholub: huh. i see. there could be a small library available that encapsulates the more complicated parts of the fetch handling? perhaps there's a bug there, and it could simplify it.

tomholub:
I mean, by debugging the code itself

tomholub:
How is this looking at the moment?

sosnovsky:
I fixed CORS issue in Firefox by running ajax requests through background page (as proposed by Roman).

But fixing upload progress is more complicated, as current implementation won't work in Firefox because it doesn't support upload progress streaming (https://bugzilla.mozilla.org/show_bug.cgi?id=1387483, https://stackoverflow.com/a/69400632/3091318) and doesn't have plans for implementing. Also to make it work in Chrome we should support HTTP/2 on backend, and customers backends will need to support it too for showing progress.

Possible solution for now will be to use `XMLHTTPRequest` for requests which need upload progress, and `fetch` for other requests. And when there will be some solution which will work for both Chrome and Firefox - migrate to using `fetch` everywhere. 

Mandatory migration to Manifest V3 in Chrome Web Store isn't planned yet (https://groups.google.com/a/chromium.org/g/chromium-extensions/c/zQ77HkGmK9E/m/HjaaCIG-BQAJ), probably next year, but can be postponed. So mentioned `XMLHTTPRequest` solution should work well for us for some time.

tomholub:
ok, let's go with the partial solution then, and your comment here will become an issue to keep track

tomholub:
How much work before this could be merged? Or are you running into more problems?

sosnovsky:
> How much work before this could be merged? Or are you running into more problems?

I think it should be ready by the end of this week. During testing I noticed some other small issues - like incorrect download progress percent when downloading large messages (it goes up to 150%).
After finishing implementation I'll retest fetching and uploading messages functionality in both Chrome and Firefox, to be sure that nothing is broken after these changes.

sosnovsky:
@ioanmo226 this issue is part of bigger refactoring to implement [Manifest V3](https://developer.chrome.com/docs/extensions/mv3/intro/) requirements, which should be required in the future for all extensions. In this PR we switch to using `fetch` instead of `Jquery.Ajax`, as in V3 it's not possible to use `XMLHTTPRequest` (which is used by `Jquery.Ajax`).

Roman worked on this switch, but there were some issues in loading messages in Firefox, as currently it doesn't support some features of `fetch`. So I worked on implementing fallbacks which use `XMLHTTPRequest`, but can be removed and switched to `fetch` when all needed functionality will be implemented there.

.semaphore/semaphore.yml
```diff
@@ -58,8 +58,8 @@ blocks:
 
   - name: Live UI Gmail tests
     dependencies: []
-    run:
-      when: branch = 'master' OR branch =~ 'live-test'  OR branch =~ 'gmail-test'
+    # run:
+    #   when: branch = 'master' OR branch =~ 'live-test'  OR branch =~ 'gmail-test'
```
ioanmo226, line 62: Maybe we can now enable live test?

ioanmo226: Or you plan to now run live test every time?

sosnovsky: Oh, right, I forgot to rollback this change, thanks for noticing!

extension/js/common/api/email-provider/gmail/google.ts
```diff
@@ -17,69 +18,73 @@ export class Google {
...
   ): Promise<RT> => {
     progress = progress || {};
-    let data, url;
-    if ('upload' in progress) {
+    let url;
+    let dataPart:
+      | { method: 'POST' | 'PUT'; data: Dict<Serializable>; dataType: 'JSON' }
+      | { method: 'POST'; data: string; contentType: string; dataType: 'TEXT' }
+      | { method: 'GET'; data?: UrlParams }
+      | { method: 'DELETE' };
```
ioanmo226, line 47: How about we declear types like below so that we can make code easy to read and simple and avoid code duplication?

```ts
type JsonRequestParams = {
  method: 'POST' | 'PUT';
  data: Dict<Serializable>;
};

type TextRequestParams = {
  method: 'POST';
  data: string;
  contentType: string;
};

type GetRequestParams = {
  method: 'GET';
  data?: UrlParams;
};

type DeleteRequestParams = {
  method: 'DELETE';
};

params?: JsonRequestParams | TextRequestParams | GetRequestParams | DeleteRequestParams
```

extension/js/common/api/account-servers/external-service.ts
```diff
@@ -151,50 +150,52 @@ export class ExternalService extends Api {
...
+          fmt: 'FORM';
+          method: 'POST';
+        }
+      | { data: Dict<Serializable>; fmt: 'JSON'; method: 'POST' }
+      | undefined = vals
+      ? {
+          ...vals,
+          method: 'POST',
+        }
+      : undefined;
```
ioanmo226, line 196: This one too.

extension/js/common/api/email-provider/gmail/google.ts
```diff
@@ -17,69 +18,73 @@ export class Google {
...
-      'operationId' in progress
-        ? { operationId: progress.operationId, expectedTransferSize: progress.expectedTransferSize, frameId: progress.frameId }
-        : undefined;
-    const xhr = Api.getAjaxProgressXhrFactory('download' in progress || 'upload' in progress ? progress : {});
-    const request = { xhr, context, url, method, data, headers, crossDomain: true, contentType, async: true };
-    return (await GoogleOAuth.apiGoogleCallRetryAuthErrorOneTime(acctEmail, request)) as RT;
+    const authorization = await GoogleOAuth.googleApiAuthHeader(acctEmail);
+    // todo: const context = 'operationId' in progress ? { operationId: progress.operationId, expectedTransferSize: progress.expectedTransferSize, tabId: progress.tabId } : undefined;
+    const progressCbs = 'download' in progress || 'upload' in progress ? progress : undefined;
+    const request: Ajax = { progress: progressCbs, url, ...dataPart, headers: { authorization }, stack: Catch.stackTrace() };
```
ioanmo226, line 66: How about we like below?

```ts
  const baseUrl = `${GMAIL_GOOGLE_API_HOST}/gmail/v1/users/me/${path}`;
  const url = params?.method === 'POST' && 'contentType' in params
    ? `${baseUrl}?uploadType=multipart`
    : baseUrl;

  // eslint-disable-next-line @typescript-eslint/naming-convention
  const headers = { authorization: await GoogleOAuth.googleApiAuthHeader(acctEmail) };

const progressCbs = 'download' in progress || 'upload' in progress ? progress : undefined;
  // Construct request based on method and provided params
  const request: Partial<Ajax> = {
    url,
    method: params?.method ?? 'GET',
    headers,
    ...('data' in params && { data: params.data }),
    ...('contentType' in params && { contentType: params.contentType }),
    stack: Catch.stackTrace(),
    // Progress callbacks if provided
   progress: progressCbs,
  };

```

package.json
```diff
@@ -79,7 +79,7 @@
     "test_patterns": "node build/test/test/source/patterns.js",
     "test_async_stack": "node build/test/test/source/async-stack.js",
     "test_buf": "npx ava --timeout=3m --verbose --concurrency=10 build/test/test/source/buf.js",
-    "test_ci_chrome_consumer_live_gmail": "npx ava --timeout=45m --verbose --tap --concurrency=1 build/test/test/source/test.js -- CONSUMER-LIVE-GMAIL STANDARD-GROUP | npx tap-xunit > report.xml",
+    "test_ci_chrome_consumer_live_gmail": "npx ava --timeout=45m --verbose --concurrency=1 build/test/test/source/test.js -- CONSUMER-LIVE-GMAIL STANDARD-GROUP",
```
ioanmo226, line 82: should re-enable tap?

extension/js/common/api/shared/api.ts
```diff
@@ -3,34 +3,69 @@
...
+  ['api-version']?: string;
+};
+export type Ajax = {
+  url: string;
+  headers?: AjaxHeaders;
+  progress?: ProgressCbs;
+  timeout?: number; // todo: implement
+  stack: string;
+} & (
+  | { method: 'GET' | 'DELETE'; data?: UrlParams }
```
ioanmo226, line 35: This one too

extension/js/common/api/shared/api.ts
```diff
@@ -105,7 +316,81 @@ export class Api {
...
+    headers?: AjaxHeaders,
+    resFmt?: T
+  ): Promise<FetchResult<T, RT>> => {
+    progress = progress || ({} as ProgressCbs);
+    let formattedData: FormData | string | undefined;
+    let dataPart:
+      | { method: 'GET' }
+      | { method: 'POST' | 'PUT'; data: Dict<Serializable>; dataType: 'JSON' }
+      | { method: 'POST' | 'PUT'; data: string; dataType: 'TEXT' }
+      | { method: 'POST' | 'PUT'; data: FormData; dataType: 'FORM' };
```
ioanmo226, line 358: This one too

ioanmo226 (commented):
Works great. small comments

package.json
ioanmo226: > How about we declear types like below so that we can make code easy to read and simple and avoid code duplication?

extension/js/common/api/shared/api.ts
ioanmo226: > How about we declear types like below so that we can make code easy to read and simple and avoid code duplication?

extension/js/common/api/shared/api.ts
ioanmo226: > How about we declear types like below so that we can make code easy to read and simple and avoid code duplication?

ioanmo226:
By the way, also better to add another reviewer as this seems breaking changes?
Like @tomholub or @rrrooommmaaa ?

extension/js/common/api/email-provider/gmail/google.ts
sosnovsky: It'll be a nice improvement, I tried to create separate types this way but looks like it'll require too many changes. For example in [apiCall](https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-1509-fetch/extension/js/common/api/shared/api.ts#L328) in `values` parameter we use `fmt` property, which then is converted to `dataType` in `dataPart` object. Also for `apiCall` method `method` parameter is optional, but it's required for [gmailCall](https://github.com/FlowCrypt/flowcrypt-browser/blob/issue-1509-fetch/extension/js/common/api/email-provider/gmail/google.ts#L19). So we'll need different types with optional and required properties, I think we can leave it as is in the current PR  and file another issue for using such types.

package.json
sosnovsky: yeah, missed this one too, thanks

extension/js/common/api/email-provider/gmail/google.ts
sosnovsky: I implemented improvements for `url` variable, but removing `dataPart` and using `data` and `contentType` directly leads to another compiler errors. Also `Partial<Ajax>` isn't compatible with `Ajax` type on `.ajax` method, will need further changes, probably can be included in https://github.com/FlowCrypt/flowcrypt-browser/pull/5372#discussion_r1385589300

sosnovsky:
> By the way, might better to add another reviewer as this seems breaking changes and I might have missed something? Like @tomholub or @rrrooommmaaa ?

Unfortunately Roman doesn't work on FlowCrypt anymore, that's why I was finalising this PR.
Tom checked this PR before I started working on it, looked good for him. I worked mostly on fixing issues found during previous reviews, so I think it shouldn't break anything as currently all tests are passing without issues.

extension/js/common/api/email-provider/gmail/google.ts
ioanmo226: Okay, got it

ioanmo226:
Got it.

ioanmo226:
Please check test

sosnovsky:
> Please check test

We've missed that Google API urls have different base urls - `${GMAIL_GOOGLE_API_HOST}/gmail/v1/users/me` and `${GMAIL_GOOGLE_API_HOST}/upload/gmail/v1/users/me`, so it caused test fails.
Reverted this change, tests should pass now

ioanmo226 (approved):

PR #5381, opened 2023-08-23
#5342 Test handling of error when updating private key

### LINKED ISSUE #5342: test for handling of errors when adding private key
All good, but you've probably missed my comment on previous review - please also add test for checking if errors are handled correctly on private key update, thanks!

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/flowcrypt-browser/pull/5327#pullrequestreview-1558670801_

martgil:
This PR adds a test for the private key update error handling.

close #5342 

- Tests added or updated

### To be filled by reviewers


martgil:
@sosnovsky Good day! May I ask for some help? I'm struggling with satisfying the two paths where errors should be handled, namely:

1.) https://github.com/FlowCrypt/flowcrypt-browser/blob/c4cc7985bc01f483796c57d25281b0feb79571b0/extension/chrome/elements/shared/key_errors.ts#L28

2.) https://github.com/FlowCrypt/flowcrypt-browser/blob/c4cc7985bc01f483796c57d25281b0feb79571b0/extension/chrome/elements/shared/key_errors.ts#L26

For case 2.), look at https://github.com/FlowCrypt/flowcrypt-browser/blob/c4cc7985bc01f483796c57d25281b0feb79571b0/extension/js/common/core/crypto/pgp/openpgp-key.ts#L162 but unfortunately, i do not have a sufficient idea on how I could generate such key that would trigger the said error.

martgil:
> Here you probably can use key without user email, as we require a valid email address for user key 

Thank you very much for your help, Roma. I'm now able to trigger that during setup:

image

> I think you can skip test for this error, as it should be quite rare.

Oh, okay - I understood clearly.

martgil:

test/source/tests/settings.ts
```diff
@@ -706,6 +706,45 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+        const invalidKeyFormat = 'non-valid-private-key-format';
+        const invalidPrivateKeyBlock = '-----BEGIN PGP PRIVATE KEY BLOCK-----\r\n\r\ninvalid-key\r\n-----END PGP PRIVATE KEY BLOCK-----';
+        await myKeyPage.waitAndType('@input-prv-key', invalidKeyFormat);
+        await myKeyPage.waitAndClick('@action-update-key');
+        await myKeyPage.waitAndRespondToModal('warning', 'confirm', 'This does not appear to be a validly formatted key.');
+        await myKeyPage.waitAndType('@input-prv-key', invalidPrivateKeyBlock);
+        await myKeyPage.waitAndClick('@action-update-key');
+        await myKeyPage.waitAndRespondToModal('error', 'confirm', 'An error happened when processing the key');
+        await myKeyPage.waitAndType('@input-prv-key', keyCanBeFixedTestKey);
+        await myKeyPage.waitAndClick('@action-update-key');
```
sosnovsky, line 743: on this step we get error, but don't check it:

Screenshot 2023-09-19 at 14 34 13

test/source/tests/settings.ts
```diff
@@ -706,6 +706,45 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
...
+        await browser.newExtensionPage(t, 'chrome/settings/setup.htm');
+        await settingsPage.close();
```
sosnovsky, line 745: why do we need to open another page here, probably these 2 lines can be deleted?

test/source/tests/settings.ts
martgil: Ah, yes. This particular error is handle differently in the `my_key_update.htm`. For instance, during setup, the FlowCrypt browser extension offer the user to correct their key by updating the user id wherein during update it will throw this error. 

My opinion here is 50/50 -- the key can be updated here, though if they use a "fixable" key during setup, then it will be fixed in that point where they have to update the key because the updated key during setup can be downloadable.

test/source/tests/settings.ts
martgil: ah yes - it can be deleted. thanks!

test/source/tests/settings.ts
sosnovsky: I think we can leave current implementation for now, as this PR is aimed at adding tests of errors handling, so here we should just check correct error message.

martgil: Okay - thanks, I have also checked for the correct error message.

sosnovsky: Yeah, looks good now 👍

sosnovsky (approved):
👍

PR #5388, opened 2023-08-28
#5122 Strip BCC emails from regular recipients in password messages

### LINKED ISSUE #5122: another BCC leakage issue
detailed information about the issue can be found at https://github.com/FlowCrypt/flowcrypt-security/issues/258.

martgil:
This PR strips the emails specified in the `bcc` fields that was later being specified as regular recipients in password-protected messages.

close #5122 

- Tests added or updated

### To be filled by reviewers


martgil:
@sosnovsky Good afternoon, i'm currently writing a test and I can't seem to find a reference case where we actually decrypt a password protected message from a shared-tenant-fes (for a gmail consumer). I'm thinking of  decrypting a pwd message from the shared tenant fes -- would that be okay or a shared-tenant-fes mock is required (but the path I'm going to take is not so clear for me just yet).

martgil:
@sosnovsky Thank you very much. I think I have seen the place where I should do the password message testing. thank you.

martgil:

extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -253,6 +253,11 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
     newMsgData: NewMsgData
   ): Promise<{ bodyWithReplyToken: SendableMsgBody; replyToken: string }> => {
     const recipients = getUniqueRecipientEmails(newMsgData.recipients);
+    for (const validRecipientElement of this.view.recipientsModule.getValidRecipients()) {
+      if (validRecipientElement.sendingType === 'bcc') {
+        recipients.splice(1, recipients.indexOf(validRecipientElement.email));
+      }
+    }
```
sosnovsky, line 260: let's make it simpler - we can remove bcc recipients before parsing unique emails:
```ts
const recipientsWithoutBcc = { ...newMsgData.recipients, bcc: [] }; // remove bcc recipients, so they're not included in reply data
const recipients = getUniqueRecipientEmails(newMsgData.recipients);
```

martgil: thank you - I didn't know about this before. So basically, the `bcc` array from the de-sturured `newMsgData.recipients` object gets overridden by an empty array?

sosnovsky: Correct, so you can replace any object property with another value

martgil: thank you very much. super helpful as always.

test/source/tests/compose.ts
```diff
@@ -3140,6 +3141,39 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       })
     );
 
+    test(
+      'user5@standardsubdomainfes.localhost:8001 - PWD encrypted message with FES web portal - recipient in bcc should not be included in cryptup-data recipients',
```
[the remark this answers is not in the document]
martgil: Will do - thanks.

test/source/mock/google/strategies/send-message-strategy.ts
```diff
@@ -473,6 +473,8 @@ export class TestBySubjectStrategyContext {
       this.strategy = new SmimeSignedMessageStrategy();
     } else if (GMAIL_RECOVERY_EMAIL_SUBJECTS.includes(subject)) {
       this.strategy = new SaveMessageInStorageStrategy();
+    } else if (subject.includes('The PWD encrypted message does not include the BCC recipient')) {
```
sosnovsky, line 476: this subject isn't used anymore, let's remove it

test/source/mock/fes/fes-constants.ts
```diff
@@ -141,7 +142,15 @@ export const processMessageFromUser3 = async (body: string, fesUrl: string) => {
...
-  const decryptedMimeMsg = decrypted.content?.toUtfStr();
+  const decryptedMimeMsg = decrypted.content!.toUtfStr();
+  const regex = /cryptup-data="([^"]*)"/;
+  const match = decryptedMimeMsg.match(regex);
+  if (match && match.length > 1) {
+    const cryptupData = base64decode(match[1]);
+    const jsonObject = JSON.parse(cryptupData);
+    expect(jsonObject.recipient).to.not.include('flowcrypt.compatibility@gmail.com');
+    expect(cryptupData).to.contains('to@example.com');
+  }
```
sosnovsky, line 153: If there will be some bug and `cryptup-data` won't be found by regex, then this `if` part will be ignored, but test won't fail. That's why we should check values with `expect` instead of `if`. And let's also rename variables, instead of generic `regex` and `match`:
```ts
const cryptupDataRegex = /cryptup-data="([^"]*)"/;
const cryptUpDataMatch = decryptedMimeMsg.match(cryptupDataRegex);
expect(cryptUpDataMatch).to.not.be.undefined;
expect(cryptUpDataMatch.length).to.be.greaterThan(1);
const cryptupData = base64decode(match[1]);
...
expect(jsonObject.recipient).to.not.include('flowcrypt.compatibility@gmail.com');
expect(jsonObject.recipient).to.include('to@example.com');
```

test/source/mock/google/strategies/send-message-strategy.ts
martgil: yes - i'll remove it.

test/source/mock/fes/fes-constants.ts
martgil: it's helpful to know - I will be considering that when testing data to ensure that the error will persist for variables that could potentially be `undefined` in case of an immediate bug appearing in code.

martgil:
@sosnovsky ready for review again. thank you!

test/source/mock/fes/fes-constants.ts
```diff
@@ -142,6 +143,16 @@ export const processMessageFromUser3 = async (body: string, fesUrl: string) => {
   });
   expect(decrypted.success).to.equal(true);
   const decryptedMimeMsg = decrypted.content?.toUtfStr();
+  const cryptupDataRegex = /cryptup-data="([^"]*)"/;
+  const cryptupDataMatch = decryptedMimeMsg?.match(cryptupDataRegex);
+  expect(cryptupDataMatch).to.not.be.undefined;
+  expect(cryptupDataMatch && cryptupDataMatch.length).to.be.greaterThan(1);
```
sosnovsky, line 149: let's simplify `cryptupDataMatch && cryptupDataMatch.length` to `cryptupDataMatch?.length`

test/source/mock/fes/fes-constants.ts
```diff
@@ -142,6 +143,16 @@ export const processMessageFromUser3 = async (body: string, fesUrl: string) => {
...
+  if (cryptupDataMatch) {
+    const cryptupData = base64decode(cryptupDataMatch[1]);
+    const jsonObject = JSON.parse(cryptupData);
```
sosnovsky, line 152: let's remove `if (cryptupDataMatch) {` and use `base64decode(cryptupDataMatch![1])` instead.

it's ok to use `!` in tests, as having `undefined` value will cause test to fail and it'll need to be fixed. while using `!` in extension code can cause some unexpected failures, which won't be handled.

test/source/mock/fes/fes-constants.ts
martgil: will simplify the code. thank you.

test/source/mock/fes/fes-constants.ts
martgil: Thank you - I think I get the point. It's sensible to use `!` since the value I'm testing with is guaranteed to have value, unlike in extension, which is prone to unexpected error.

sosnovsky: Exactly👍

sosnovsky (approved):
👍

PR #5398, opened 2023-09-05
#5383 Improve parsing of emails with key backups

### LINKED ISSUE #5383: Design issue when opening the email containing backup, hint not shown
Description: When opening the "Your FlowCrypt Backup" email that contains the private key backup, the hint indicating whether the private key has been imported or how to import it directly from there is not displayed. Instead, a message stating "Unknown OpenPGP Format" is shown to describe the private key.

image

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#archive/FMfcgzGtwgpbLMRtxzgpRtbdDFtZJkdD

martgil:
This PR fixes the ui design issue wherein the private key import hint is not shown for a FlowCrypt backup file attachment.

close #5383 

- Tests added or updated

### To be filled by reviewers


extension/js/common/core/attachment.ts
```diff
@@ -199,10 +199,10 @@ export class Attachment {
       // ends with one of .gpg, .pgp, .???.asc, .????.asc
       return 'encryptedFile';
       // todo: after #4906 is done we should "decrypt" the encryptedFile here to see if it's a binary 'publicKey' (as in message 1869220e0c8f16dd)
+    } else if (this.name.match(/flowcrypt-backup-[a-z0-9]+-[A-F0-9]{40}\.(key|asc)$/g)) {
+      return 'privateKey';
     } else if (this.isPublicKey()) {
       return 'publicKey';
-    } else if (this.name.match(/(cryptup|flowcrypt)-backup-[a-z0-9]+\.(key|asc)$/g)) {
```
martgil, line 204: I updated the regex name as the FlowCrypt backup name is no longer having the prefix `cryptup`.

martgil:

extension/js/common/core/attachment.ts
sosnovsky: But some old backups still has `cryptup` in their names, so correct parsing will stop working for them.

Also I have my key backup from 2021 with name `flowcrypt-backup-<MY_EMAIL>.key` and parsing stopped working for it, as new regex expects key fingerprint after `<MY_EMAIL>`. Regex should work for cases when fingerprint isn't included in backup name

test/source/tests/decrypt.ts
```diff
@@ -68,7 +68,7 @@ export const defineDecryptTests = (testVariant: TestVariant, testWithBrowser: Te
     test(
       `decrypt - backup message rendering`,
       testWithBrowser(async (t, browser) => {
-        const threadId = '15f84afa553d8a83';
+        const threadId = '188923a75165a3c8';
```
sosnovsky, line 71: let's not just replace old message with the new one, as some users have backups with old format and backup message rendering should work for this cases too.

we can rename this test to `decrypt - old backup message rendering` and add another one which will check rendering of new backup format

martgil: Thank you - I'm a bit slow and didn't realize that earlier. I will check both old and new backups.

extension/js/common/core/attachment.ts
martgil: Ah, yes. I will update the regex so that it can cover both old and new regex. Thanks again for your help.

extension/js/common/core/attachment.ts
```diff
@@ -199,10 +199,10 @@ export class Attachment {
       // ends with one of .gpg, .pgp, .???.asc, .????.asc
       return 'encryptedFile';
       // todo: after #4906 is done we should "decrypt" the encryptedFile here to see if it's a binary 'publicKey' (as in message 1869220e0c8f16dd)
+    } else if (this.name.match(/(cryptup|flowcrypt)-backup-([a-z0-9]+(?:\-[A-F0-9]{40})?)\.(key|asc)$/g)) {
```
sosnovsky, line 202: please move this private key check below public key check (as it was before this PR), as comment on line 201 is related to public keys parsing

extension/js/common/core/attachment.ts
```diff
@@ -136,13 +136,22 @@ export class Attachment {
       return this.treatAsValue === 'publicKey';
     }
     return (
-      this.type === 'application/pgp-keys' ||
```
martgil, line 139: This always return `true` even with private key attachments so an additional checks is added.

extension/js/common/core/attachment.ts
martgil: will do. I also updated the private key detection a bit.

sosnovsky (approved):
Well done 👍

PR #5406, opened 2023-09-11
Update ESLint config

sosnovsky:
This PR updates ESLint config - I noticed that `@typescript-eslint/no-non-null-assertion` rule wasn't active and found that `typescript-eslint` changed rules sets a bit - this `no-non-null-assertion` rule was moved to `@typescript-eslint/strict` set and some other rules were separated to `@typescript-eslint/stylistic` set. I also removed unused `eslint-disable` comments

### To be filled by reviewers


ioanmo226 (approved):
New eslint config looks good 👍

PR #5411, opened 2023-09-13
#5368 Remove non-printable space characters on email footer

### LINKED ISSUE #5368: Don't add "--" separator if user signature begins with "--"

martgil:
This PR removes non-printable Unicode characters on the email footer that result in duplicated `--` in the user's email footer.

close #5368 

### To be filled by reviewers


martgil:
I am closing this PR as the changes are quite extensive. However, I wanted to provide a helpful resource for invisible unicode characters, just in case we need it in the future. Here is the link: [invisible-characters.com](https://invisible-characters.com/)

PR #5412, opened 2023-09-15
#5368 Remove non-breaking space character on footer

### LINKED ISSUE #5368: Don't add "--" separator if user signature begins with "--"

martgil:
This PR removes the non-breaking space character in the email footer/signature that causes a duplicate footer separator.

close #5368 

- Test added or updated.

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-footer-module.ts
```diff
@@ -34,7 +34,7 @@ export class ComposeFooterModule extends ViewModule<ComposeView> {
   public createFooterHtml = (footer: string) => {
     const sanitizedPlainFooter = Xss.htmlSanitizeAndStripAllTags(footer, '\n');
     const sanitizedHtmlFooter = sanitizedPlainFooter.replace(/\n/g, '<br>');
-    const footerFirstLine = sanitizedPlainFooter.split('\n')[0];
+    const footerFirstLine = sanitizedPlainFooter.split('\n')[0].replace(/&nbsp;/g, '');
```
sosnovsky, line 37: let's replace it with common space, so it'll be ` ` instead of empty string

sosnovsky (commented):
Well done, it fixes issue on my side👍

And let's add a test which will check that there is no double separator in this case. I think it should be a simple one - you can set user's signature by using `configProvider.google.aliases` config

extension/chrome/elements/compose-modules/compose-footer-module.ts
martgil: Thank you - will do.

> And let's add a test which will check that there is no double separator in this case. I think it should be a simple one - you can set user's signature by using configProvider.google.aliases config

I agree - I'll be writing the test now. thank you.

martgil: Roma, for me the `configProvider.google.aliases` doesn't fit very well on the test so I try to re-use the idea of directly putting a signature with `configProvider.google.acctPrimarySignature` which is great for the test. I hope that is also an acceptable approach.

martgil:
ready for review. thank you!

sosnovsky (approved):
Well done 👍

PR #5440, opened 2023-10-03
#5428 Fix: "Secure reply" button not inserted for non-English Gmail UI

### LINKED ISSUE #5428: "Secure reply" button not inserted for non-English Gmail UI
Currently we insert `secure reply` button after standard Gmail reply button, which [is found using](https://github.com/FlowCrypt/flowcrypt-browser/blob/ab31c96d09b3a2f86f82cfbbbd55fbd5c28cc172/extension/js/content_scripts/webmail/gmail-element-replacer.ts#L256) `aria-label="Reply"`.
But it's correct only for English UI, while in other languages `aria-label` property has different value.
So standard reply button can't be found and `secure reply` button isn't inserted.

We should use another selector which is consistent across different languages.

Screenshot 2023-09-26 at 14 01 55

ioanmo226:
This PR fixed `Secure reply` button not inserted for non-English Gmail UI issue

close #5428 // if this PR closes an issue

- Difficult to test (Hard to simulate gmail language change)

### To be filled by reviewers


sosnovsky (approved):
Well done 👍

PR #5445, opened 2023-10-05
#5422 Fix noname attachment detected as encryptedMsg

### LINKED ISSUE #5422: Non-encrypted HTML page tries to render by the extension as encrypted results to unreadable regular emails
Description: The FlowCrypt browser extension tries to render an HTML email as encrypted message which results to the following error message: "Problem extracting armored message". The email turns into a what seems to be a PNG file based on its chunked contents displayed after clicking the "show original message".

image

Test email shared by the user:
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzGtxKSjkJdCkwlFDhfgdPdWRXnH

Affected users:
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGtwztzFgmFHWJxmltxxtSztfTm
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGtwzspsmJwqdjqXbtKZPQVmtdF

### LINKED ISSUE #5450: regular png triggering flowcrypt wrongly
> There is this case where the 'noname' file with the content type of 'application/octet-stream' but with a bytes of png and indeed was a png file when opening it. they were using this image as an cid image embedded in a regular emails.

_Originally posted by @martgil in https://github.com/FlowCrypt/flowcrypt-browser/issues/5422#issuecomment-1748283821_

martgil:
This PR adds another fix for catching ambiguous "noname" file which has an `application/octet-stream` content type but with bytes of an PNG (based on its magic headers) that the extension mistakenly recognize as `encryptedMsg`. 

close #5450 
related previous issue #5422 

- Tests added or updated

### To be filled by reviewers


martgil:

extension/js/common/core/attachment.ts
```diff
@@ -219,7 +219,7 @@ export class Attachment {
     } else {
       // && !Attachment.encryptedMsgNames.includes(this.name) -- already checked above
       const isAmbiguousAscFile = /\.asc$/.test(this.name); // ambiguous .asc name
-      const isAmbiguousNonameFile = !this.name || this.name === 'noname'; // may not even be OpenPGP related
+      const isAmbiguousNonameFile = !this.name || (this.name === 'noname' && this.type !== 'application/octet-stream'); // may not even be OpenPGP related
```
martgil, line 222: I updated this particular test as a file named "noname" with the type of "application/octet-stream". It is challenging to determine the appropriate characteristics of an ambiguous "noname" file.

test/source/tests/decrypt.ts
```diff
@@ -1979,16 +1979,23 @@ XZ8r4OC6sguP/yozWlkG+7dDxsgKQVBENeG6Lw==
...
-        await attachmentFrame.waitForContent('@container-attachment-header', 'PLAIN FILE');
+        const inboxPage1 = await browser.newExtensionPage(t, `chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId1}`);
+        await inboxPage1.waitAll('iframe');
+        const attachmentFrame1 = await inboxPage1.getFrame(['attachment.htm']);
+        await attachmentFrame1.waitForSelTestState('ready');
+        await attachmentFrame1.waitForContent('@attachment-name', 'noname');
+        await attachmentFrame1.waitForContent('@container-attachment-header', 'PLAIN FILE');
+        const inboxPage2 = await browser.newExtensionPage(t, `chrome/settings/inbox/inbox.htm?acctEmail=${acctEmail}&threadId=${threadId2}`);
+        await inboxPage2.waitAll('iframe');
+        const attachmentFrame2 = await inboxPage1.getFrame(['attachment.htm']);
```
sosnovsky, line 1995: here you extract frame from `inboxPage1`, but it should be from `inboxPage2`.

on `inboxPage2` message fails to decrypt with `Problem extracting armored message` message

martgil: Oh, my bad - I have updated the condition checks in the `attachment.ts` accordingly which passes the tests.

martgil:
@sosnovsky It looks good to me at this point. Is it the same on your end?

sosnovsky (commented):
Looks good for me now too, just a small improvement - let's don't add `Unknown OpenPGP format` label in this case, as it can be confusing for users. As it's non-encrypted plain message it'll be better to just show default Gmail UI - `One attachment. Scanned by Gmail`.

Without FlowCrypt:
Screenshot 2023-10-11 at 12 04 12

With FlowCrypt (should be the same as without FlowCrypt):
Screenshot 2023-10-11 at 12 03 58

martgil:
> Looks good for me now too, just a small improvement - let's don't add `Unknown OpenPGP format` label in this case, as it can be confusing for users. As it's non-encrypted plain message it'll be better to just show default Gmail UI - `One attachment. Scanned by Gmail`.

Thanks, Roma - I didn't noticed this one. I will also check the other case behaves as you described here.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -454,7 +454,6 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
     const loaderContext = new GmailLoaderContext(this.factory, msgEl, attachmentsContainerInner);
     attachmentsContainerInner = $(attachmentsContainerInner);
-    attachmentsContainerInner.parent().find(this.sel.numberOfAttachments).hide();
```
martgil, line 457: I had to remove this line for the Gmail notes about scanned attachments to show up even with FlowCrypt enabled:

image

martgil:

test/source/tests/decrypt.ts
sosnovsky: You still get frame from `inboxPage1`, but it should be `const attachmentFrame2 = await inboxPage2.getFrame(['attachment.htm']);`

sosnovsky (commented):
Works well now, just need a small test fix

martgil:
Thanks @sosnovsky -- all good now.

PR #5453, opened 2023-10-11
#5438 Fix "see original" button not added to encrypted threads in Gmail UI

### LINKED ISSUE #5438: FlowCrypt's "see original" button is missing on Gmail
Description: I noticed that the FlowCrypt's "see original" button is content-injected and is now missing when opening the detected encrypted emails. I can recall that is present whether the encrypted email is readable or not.

Current behavior:
image

Expected behavior:
image

Current behavior on the web extension page:
image

Steps to reproduce:
1. Using the FlowCrypt browser extension, compose and send a public key encrypted message/email.
2. Check the sent email from the 'sent' folder of Gmail.
3. Notice that there is no "see original" button visible where it is typically present.

martgil:
This PR will fix the issue where the "see original" button is not showing when opening an encrypted email.

Proposed fix:
image

Current behaviour:
image

close #5438 

- Tests added or updated

### To be filled by reviewers


martgil:
@sosnovsky As far as I'm aware, the original message should work by showing the original message itself. In this case, most likely these results would expect a PGP block, is it? My intuition is that the page should work like before where a popup window will appear and to simplify it, I'd like to reuse the "print page" except that the print-content section will be updated and show the PGP block (or the original message). 



Is it correct?

sosnovsky:
As I understand, we already have implementation for `show original` functionality and just need to update css selector for adding `see original` button?
We also have [renderMsg](https://github.com/FlowCrypt/flowcrypt-browser/blob/768ebbb9cb1f77eb19385f3308682c5b6e233cfa/extension/js/common/message-renderer.ts#L254) method with `showOriginal` parameter, may be it was used for rendering original message?

martgil:
Thank you Roma. Unfortunately, even after updating the CSS selector to where the button is injected, clicking the button still doesn't do anything. The function `addfcConvoIcon` accepts a callback function but in our current code, the callback function specified was set to trigger a click event against the target element where the button is injected but has no actual event handler there as of now. In comparison, the `see original` functionality in the web extension page of viewing the message, simply swaps out messages from showing an original message to showing a rendered message (and vice versa), which I'm planning to do upon your confirmation.

martgil:
I also did an alternative which would simply show the original message in a new popup window upon clicking the `see original` button -- the same as how it worked before.

sosnovsky:
> I also did an alternative implementation in my current changes which would simply show the original message in a new popup window upon clicking the see original button -- the same as how it worked before.

This one sounds better. As for me, it should work the similar way as Gmail shows full content for clipped messages - you can check it by going to inbox of `flowcrypt.compatibility` account, opening `FES message for #5458` thread, scrolling to the bottom of the message and clicking `View entire message`.

Maybe we can just reuse URL of this newly opened tab and open it when user clicks `See original`?

martgil:
Thanks for the suggestion, it is helpful. Yes, I think that would be possible. It's just a little bit tricky to do in the content script side -- I just need to pull the data from Gmail instead of parsing information from Gmail web UI as it may not be practical to do so.

martgil:
Hi @sosnovsky -- I'm able to reuse the `view original message`of Gmail which will open upon clicking the `see original` button. Is it correct, that in this case, I might've need to add a live gmail test to test it out?

sosnovsky:
> Hi @sosnovsky -- I'm able to reuse the `view original message`of Gmail which will open upon clicking the `see original` button. Is it correct, that in this case, I might've need to add a live gmail test to test it out?

Yes, here we'll need to have a live gmail test, so in the future tests will fail if `see original` button disappears.
But let's not add a separate test for it and just add testing of `see original` button to some existing test.

martgil:
@sosnovsky Ready for review. Thank you!

martgil:
@sosnovsky Ready for review. Thank you.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -299,7 +302,12 @@ export class GmailElementReplacer implements WebmailElementReplacer {
         if (!convoUpperIcons.is('.appended') || convoUpperIcons.find('.use_secure_reply').length) {
           // either not appended, or appended icon is outdated (convo switched to encrypted)
           this.addfcConvoIcon(convoUpperIcons, this.factory.btnWithoutFc(), '.show_original_conversation', () => {
-            convoUpperIcons.find('.gZ').trigger('click');
+            const linkWithGmKeyId = $(this.sel.linkWithGmIdKey).attr('data-recorded-src') || '';
+            const dataMsgId = $(this.sel.dataMsgId).attr('data-message-id')?.replace('#', '') || '';
+            const tokenParam = new URLSearchParams(linkWithGmKeyId).get('token') || '';
+            const gmIdKey = JSON.parse(tokenParam)[1];
+            const webmailUrl = Google.webmailUrl(this.acctEmail) + `/?ik=${gmIdKey}&permmsgid=${dataMsgId}&view=lg`;
+            window.open(webmailUrl, '_blank', `height=680, width=620, top=30, left=${(screen.width - 620) / 2}`);
```
sosnovsky, line 310: It seems we overcomplicated this functionality - I checked previous code and there we just triggered click on the button which opened full thread in a separate window. And now we show only original of the last message.

So let's make it work as before - clicking `see original` should just trigger click on this button:

Screenshot 2023-10-23 at 13 00 20

martgil: Ok, the one I initially wrote requires too much. That's why we used `.trigger('click');` to call that one. I have fixed it and now working well :)

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -228,7 +231,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
 
   private addfcConvoIcon = (containerSel: JQueryEl, iconHtml: string, iconSel: string, onClick: () => void) => {
     containerSel.addClass('appended').children('.use_secure_reply, .show_original_conversation').remove(); // remove previous FlowCrypt buttons, if any
-    Xss.sanitizeAppend(containerSel, iconHtml)
+    Xss.sanitizePrepend(containerSel, iconHtml)
```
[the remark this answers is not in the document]
martgil: Thank you - I have fixed the this one as well.

martgil:
@sosnovsky This is now working well. thank you!

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -294,12 +295,12 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
+    if (convoUpperIconsContainer.length) {
       if (useEncryptionInThisConvo) {
-        if (!convoUpperIcons.is('.appended') || convoUpperIcons.find('.use_secure_reply').length) {
-          // either not appended, or appended icon is outdated (convo switched to encrypted)
-          this.addfcConvoIcon(convoUpperIcons, this.factory.btnWithoutFc(), '.show_original_conversation', () => {
-            convoUpperIcons.find('.gZ').trigger('click');
+        if (!convoUpperIconsContainer.is('.appended') || convoUpperIconsContainer.find(convoUpperIcons).length) {
+          this.addfcConvoIcon(convoUpperIconsContainer, this.factory.btnWithoutFc(), '.show_original_conversation', () => {
+            const openOriginalInNewTabIndex = convoUpperIcons.length - 1; // the open in new tab icon button has always been in the last item
+            convoUpperIconsContainer.find(convoUpperIcons).eq(openOriginalInNewTabIndex).trigger('click');
```
sosnovsky, line 303: probably we can just use just `.find(convoUpperIcons).last()` instead of calculating index

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -294,12 +295,12 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
     // conversation top-right icon buttons
-    if (convoUpperIcons.length) {
+    if (convoUpperIconsContainer.length) {
       if (useEncryptionInThisConvo) {
-        if (!convoUpperIcons.is('.appended') || convoUpperIcons.find('.use_secure_reply').length) {
-          // either not appended, or appended icon is outdated (convo switched to encrypted)
-          this.addfcConvoIcon(convoUpperIcons, this.factory.btnWithoutFc(), '.show_original_conversation', () => {
-            convoUpperIcons.find('.gZ').trigger('click');
+        if (!convoUpperIconsContainer.is('.appended') || convoUpperIconsContainer.find(convoUpperIcons).length) {
+          this.addfcConvoIcon(convoUpperIconsContainer, this.factory.btnWithoutFc(), '.show_original_conversation', () => {
```
sosnovsky, line 301: `see original` button is hiding/showing when you perform some actions with thread, like starring messages, expanding/hiding messages etc.
But it should stay there constantly after it was added:

https://github.com/FlowCrypt/flowcrypt-browser/assets/1062093/8b0dd3aa-cd40-4d92-8281-b7883affdf01

sosnovsky (commented):
Works well now 👍

extension/js/content_scripts/webmail/gmail-element-replacer.ts
martgil: Thank you, Roma. Will apply this instead.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
martgil: Thank you -- I found the problem of why such behavior occurs. When performing actions, the upper icon buttons container from Gmail gets re-injected, and then the 'see original' button gets injected as well. I'm having difficulty with troubleshooting it as those updates happen every sec like a "heartbeat" in Gmail. We could test such behavior by going to `Console` and manipulating the upper icon container with `$('div.hj').append('<span>test</span>')` and clicking the star icon will remove this as well.

I have fixed all this by finding another element (but it could not be so perfect).

sosnovsky: It's not flickering, but doesn't move along with other element, so it stays on the same place and overflows other elements, like this:

Screenshot 2023-10-26 at 15 57 06

It seems to be more complicated than I expected, then let's leave this flickering as it was before your latest change.

Also let's add check for not re-appending this button when it's already inserted - https://github.com/FlowCrypt/flowcrypt-browser/blob/d8b4dc6ccea9482405ebbbb5b38fb2c00aa74a09/extension/js/content_scripts/webmail/gmail-element-replacer.ts#L230
Here we remove `.show_original_conversation` element and then append it again. Need to add check in the beginning of `addfcConvoIcon`:
```js
if ($(containerSel).find(iconSel).length) { return }
```

martgil: Yes, I also find it complicated. I'll revert the changes now and implement the suggested fix for the element not to remove/re-append. thank you!

martgil:
@sosnovsky I have reverted the changes and also, and the repeated append has been fixed as well. Thank you!

sosnovsky (approved):
Looks good and works well 👍

Issue #5410, opened 2023-09-13
Parsing of attached public key fails sometimes


sosnovsky:
I sent encrypted email with attached public key, but key attachment parsing fails with error:

<img width="200" alt="Screenshot 2023-09-13 at 12 27 10" src="https://github.com/FlowCrypt/flowcrypt-browser/assets/1062093/3e079452-30b2-44a0-82b1-336134861393">

Console log:
```
webmail_bundle.js:74 BrowserMsg(ajaxGmailAttachmentGetChunk) (no status text): 200 when GET-ing https://gmail.googleapis.com/gmail/v1/users/me/messages/18a8dd7fd57aa685/attachments/ANGjdJ--2kq-UIIBb0g1YDRg4XiCGTFRxAKqx7XfrxtLgsRpbagJ75PXFVQqSR8NeTJ97leYcZBKkumv8by5Z8gGkQruwXB98lDt0rQMEPJ3W6ZBiT1Fv9Z-LumXICMnUNyPH3DUSpEDW1tu6GDGwykKbIfb971dEfOUAam_jlEG4uQKH_aneURGhKbcHWfFqEVd4BjOOsUyL9Vb3ybXmkKKxsf9tsB-xNUVWSU8WU04JhCtQKtOnK2wNlhWT3c1IP9bfFazKenFJERjJP2zyiWi2U9ZcCLDhQpJUpnsE9mu7WJxj1Huvmz0THkTTb-_G6DE91TulqlNubs4xFX5L4ZR2WD52mi7s6g7g5V6c9Bi0TNIRwhYanwshBNQUsE (no body):  -> (no standard err msg)


[callerStack]
    at BrowserMsg.sendRaw (chrome-extension://ahfaejcjbnfgkcmifnehflgfbbejigbf/js/content_scripts/webmail_bundle.js:9982:22)
    at BrowserMsg.sendAwait (chrome-extension://ahfaejcjbnfgkcmifnehflgfbbejigbf/js/content_scripts/webmail_bundle.js:9968:33)
    at Object.ajaxGmailAttachmentGetChunk (chrome-extension://ahfaejcjbnfgkcmifnehflgfbbejigbf/js/content_scripts/webmail_bundle.js:9722:61)
    at Gmail.attachmentGetChunk (chrome-extension://ahfaejcjbnfgkcmifnehflgfbbejigbf/js/content_scripts/webmail_bundle.js:8957:70)
[...]

PR #5472, opened 2023-10-30
#5459 Fix visible 'noname' email attachment file

### LINKED ISSUE #5459: 'noname' attachment not hidden in password-protected message
After merging https://github.com/FlowCrypt/flowcrypt-browser/pull/5445 `noname` attachments are not hidden in password-protected messages and `2 attachments` label is visible, but it should be hidden too.

Need also to add ui test checking that `noname` attachment and this label is hidden in password-protected messages.

before
Screenshot 2023-10-17 at 08 20 52

after 
Screenshot 2023-10-17 at 09 42 31

### LINKED ISSUE #5460: Incorrect label with attachments count
When message has attached public key and other attachment, UI shows `2 Attachments` label, while it should be `1 Attachment` as public key attachment isn't shown:

Screenshot 2023-10-18 at 22 59 36

For example you can check `PDF attached` thread in `flowcrypt.compatibility` account

martgil:
This PR fixes the visible 'noname' attachment file.

Current behavior:
image

Proposed fix:
image

close #5459
close #5460

- Tests added or updated

### To be filled by reviewers


martgil:

martgil:
I noticed that Live Gmail test is somehow skipped. I'm not sure how I could re-run that particular test.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -457,6 +457,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
     }
     const loaderContext = new GmailLoaderContext(this.factory, msgEl, attachmentsContainerInner);
     attachmentsContainerInner = $(attachmentsContainerInner);
+    attachmentsContainerInner.parent().find(this.sel.numberOfAttachments).hide();
```
sosnovsky, line 460: Attachments count label should be visible (as per https://github.com/FlowCrypt/flowcrypt-browser/issues/5460#issuecomment-1784910402)

sosnovsky:
> I noticed that Live Gmail test is somehow skipped. I'm not sure how I could re-run that particular test.

We have condition in semaphore.yml to run live tests only in `master` or when branch name ends with `live-test`: https://github.com/FlowCrypt/flowcrypt-browser/blob/ee8fbcc785f0c663f7d738751849d94cb1f57032/.semaphore/semaphore.yml#L62

So live tests don't run in each branch, as then they will interfere between different PRs and fail often.
Another method - comment mentioned line in `semaphore.yml` and live tests will run in this branch.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
martgil: Got it - thank you! I understand better why the email `PDF attached` counts 2, because the public key is also part of the count in addition to the pdf attachment.

martgil: We do have the code to show the attachment count label only when the numbers of rendered rendered attachments are more than 2 items. 

I guess this is all good just the count needs to ensure to be correct? It will be fixed then with the later changes.

martgil: @sosnovsky I have fixed the over all issue for `noname` attachment, incorrect number of attachments, and showing the label for attachments. Only the Gmail test, not working quite well.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -473,21 +482,27 @@ export class GmailElementReplacer implements WebmailElementReplacer {
...
-      // Aligned with Gmail, the label is shown only if there are 2 or more attachments
-      attachmentsContainerInner.parent().find(this.sel.numberOfAttachmentsDigit).text(nRenderedAttachments);
-      attachmentsContainerInner.parent().find(this.sel.numberOfAttachments).show();
-    }
     if (nRenderedAttachments === 0) {
       attachmentsContainerInner.parents(this.sel.attachmentsContainerOuter).first().hide();
+    } else {
+      const elementsToClone = ['span.a2H', 'div.a2b'];
+      const scannedByGmailLabel = $(elementsToClone[0]).first().clone();
+      const scannedByGmailPopover = $(elementsToClone[1]).first().clone(true);
```
sosnovsky, line 498: maybe we can just update text in existing html elements, instead of cloning them and then re-appending?

sosnovsky (commented):
Works much better now, well done, I just noticed a small bug - please check `flowcrypt-browser #5410 attached ecc public key` thread in `flowcrypt.compatibility` inbox, it doesn't hide attachments count label:

Screenshot 2023-11-06 at 14 12 27

extension/js/content_scripts/webmail/gmail-element-replacer.ts
martgil: >  I just noticed a small bug 

Thank you - I'll look into it in and will fix that as well.

> maybe we can just update text in existing html elements, instead of cloning them and then re-appending?

Actually, that was my first attempt in fixing the issue but the HTML structuring in Gmail is not scraping-friendly, I mean, the text node and html elements kinda mixed up in a single span so simple text content changing for example via `.text()` don't work very great. It's a bit complex for example, the word/label "One attachment" appears for unmodified number of attachments, wherein with modified number of attachments - they will be defined in a numerical representation which conflicts with the HTML structuring so I came up with element cloning so I can have the better HTML structure I wanted to have + the uniformity of the number of attachments is pleasant compared to modifying "One" to "1" is again complex because the word "One" is a text node and "1" however, is nicely wrapped in an element. 

The live gmail test is not passing at all. Was the test in the web extension page maybe sufficient enough as I could add alternative for testing the expected numbers?

sosnovsky: > Actually, that was my first attempt in fixing the issue but the HTML structuring in Gmail is not scraping-friendly, I mean, the text node and html elements kinda mixed up in a single span so simple text content changing for example via .text() don't work very great. It's a bit complex for example, the word/label "One attachment" appears for unmodified number of attachments, wherein with modified number of attachments - they will be defined in a numerical representation which conflicts with the HTML structuring so I came up with element cloning so I can have the better HTML structure I wanted to have + the uniformity of the number of attachments is pleasant compared to modifying "One" to "1" is again complex because the word "One" is a text node and "1" however, is nicely wrapped in an element.

Got it, thanks for detailed explanation, then let's use your current implementation 👍

> The live gmail test is not passing at all. Was the test in the web extension page maybe sufficient enough as I could add alternative for testing the expected numbers?

Having live gmail test will be useful for detecting if there will be any changes in Google's html markup, which can lead to breaking of our html parsing. We can use existing tests, instead of creating a separate one:
- `mail.google.com - attachments which contain emoji in filename are rendered correctly` - here let's check that page has label `2 Attachments  •  Scanned by Gmail`
- `mail.google.com - pubkey file gets rendered` - here check that attachments count label is hidden

martgil: > Works much better now, well done, I just noticed a small bug

Thank you @sosnovsky. It is now fixed.

 I have made changes that would satisfy all this. Test are all passing now.

https://flowcrypt.semaphoreci.com/workflows/5ebecf93-c789-40ad-888d-ce64a720642e?pipeline_id=2fa93122-d051-44d3-855d-94cd5319d44b

martgil:

sosnovsky (commented):
I checked `flowcrypt-browser #5410 attached ecc public key` thread and attachments count label is hidden now, but `import public key` block is hidden too - it should be visible, as we should hide only attachments count label in this case.

Also please check `Encrypted email with public key attached` thread in `flowcrypt.compatibility` account, it has the same issue as reported in previous review - `import public key` and attachments count label are both visible:

Screenshot 2023-11-09 at 11 19 40


Both mentioned threads should show `import public key` block and hide attachments count label.

martgil:
> Also please check Encrypted email with public key attached thread in flowcrypt.compatibility account

It working well for me - no label shown. Only the import public key button is missing.

image

sosnovsky:
> It working well for me - no label shown.

Seems like it's a different thread - on my screenshot sender is `e2e.enterprise.test`(thread id is `FMfcgzGtwzspbBBstksRxMDLgCHFjHhS`), while on your screenshot - `flowcrypt.compatibility`:

Screenshot 2023-11-10 at 13 22 11

sosnovsky:
> I'll be checking it once again. it is actually an encrypted public key, so it was once detected as regular encrypted attachment then later been recognized and rendered as public key from which the process of hiding labels comes first. So i would need to find a way to recognized it as encrypted public key before hand so when I reached the test nRenderedattachments, it will let the rendering of the public key ui.

It works well now, I just noticed that top separator isn't hidden on `public key` block:

Screenshot 2023-11-10 at 13 25 12

martgil:
@sosnovsky, I have also addressed the last reported issues. Could you please take another look?

sosnovsky (commented):
> It works well now, I just noticed that top separator isn't hidden on public key block:

This issue is fixed ✅

> Seems like it's a different thread - on my screenshot sender is e2e.enterprise.test(thread id is FMfcgzGtwzspbBBstksRxMDLgCHFjHhS), while on your screenshot - flowcrypt.compatibility:

This issue is still presented - `1 Attachment  •  Scanned by Gmail` label isn't hidden in `FMfcgzGtwzspbBBstksRxMDLgCHFjHhS` thread (checked in Chrome and Firefox)

martgil:
> This issue is still presented

I have fixed it now and also fixed this case in pgp_block.

Before:
image

After:
image

sosnovsky (approved):
All good now, thanks for fixes!

martgil:
> All good now, thanks for fixes!

Thank you too for all the help.

PR #5495, opened 2023-11-24
#5494 Added test for #5497 hotfix

### LINKED ISSUE #5494: Secure reply/Secure forward disappears from Gmail
### Description
When receiving an encrypted email reply, the option to reply/forward securely has disappeared. 

image

### Steps to reproduce

1. User 1 sends flowcrypt encrypted email to User 2.
2. User 2 received the encrypted email and responded to it.
3. User 1 receives the encrypted reply but has no option to reply/forward the email securely (encrypted).
4. From User 2's view, the option to reply/forward encrypted is rendered in Gmail correctly.

Reference:
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGwHpShPkWdxBZFXMBHZtTxNdvW
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGwHpRXzjZFRnMhWmZkMLQKtfsq

martgil:
This PR added a test for a hotfix for the original issue https://github.com/FlowCrypt/flowcrypt-browser/pull/5497.

<strike>This PR updates the class name that `getConvoRootEl()` of `gmail-element-replacer.ts` updates in order to render action buttons for received encrypted emails.</strike>

close #5494 

- Test added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -578,7 +578,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   };
 
   private getConvoRootEl = (anyInnerElement: HTMLElement) => {
-    return $(anyInnerElement).closest('div.Bk, div.if, td.Bu').first();
+    return $(anyInnerElement).closest('div.if, div.aHU, td.Bu').first();
   };
 
   private insertEncryptedReplyBox = (messageContainer: JQuery<HTMLElement>) => {
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -578,7 +578,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   };
 
   private getConvoRootEl = (anyInnerElement: HTMLElement) => {
-    return $(anyInnerElement).closest('div.if, td.Bu').first();
+    return $(anyInnerElement).closest('div.Bk, td.Bu').first();
   };
 
   private insertEncryptedReplyBox = (messageContainer: JQuery<HTMLElement>) => {
--- a/extension/js/content_scripts/webmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail-element-replacer.ts
@@ -578,7 +578,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   };
 
   private getConvoRootEl = (anyInnerElement: HTMLElement) => {
-    return $(anyInnerElement).closest('div.Bk, td.Bu').first();
+    return $(anyInnerElement).closest('div.Bk, div.if, td.Bu').first();
   };
 
   private insertEncryptedReplyBox = (messageContainer: JQuery<HTMLElement>) => {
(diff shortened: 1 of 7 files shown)


extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -578,7 +578,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   };
 
   private getConvoRootEl = (anyInnerElement: HTMLElement) => {
-    return $(anyInnerElement).closest('div.if, td.Bu').first();
+    return $(anyInnerElement).closest('div.Bk, td.Bu').first();
```
sosnovsky, line 581: let's leave old class too, as it's possible that html code was updated not for all Gmail users, so we should keep older classes too

martgil: Got it - I'll add it back. thank you!

sosnovsky (commented):
All good now, but it's strange that our live tests don't fail after such Gmail HTML change.
I ran them after we noticed this issue and they all passed - https://flowcrypt.semaphoreci.com/jobs/bb91c8f6-b5fd-4465-814a-ac8d083aef77

Can you please also add live test which will check that secure reply buttons are added to conversation view page.

sosnovsky:
Probably no need in a separate test, can add this check to some existing live test.

extension/js/content_scripts/webmail/gmail-element-replacer.ts
```diff
@@ -578,7 +578,7 @@ export class GmailElementReplacer implements WebmailElementReplacer {
   };
 
   private getConvoRootEl = (anyInnerElement: HTMLElement) => {
-    return $(anyInnerElement).closest('div.if, td.Bu').first();
+    return $(anyInnerElement).closest('div.Bk, div.if, td.Bu').first();
```
sosnovsky, line 581: I've just noticed that it'll be better to use `div.aHU` selector, as it's used by root conversation element, while `div.Bk` is used by each message

martgil: I just merged the hotfix from the `master` branch.

martgil:

test/source/tests/gmail.ts
```diff
@@ -307,6 +307,7 @@ export const defineGmailTests = (testVariant: TestVariant, testWithBrowser: Test
           await gotoGmailPage(gmailPage, `/${threadId}`); // go to encrypted convo
           t.timeout(minutes(4)); // extend ava's timeout
           await GmailPageRecipe.trimConvo(gmailPage, threadId);
+          expect(await gmailPage.isElementPresent('div.aHU')).to.equal(true); // @getConvoRootEl() target element for rendering encrypted reply or forward buttons
```
sosnovsky, line 310: Hmm, I wanted to comment that it should be better to check if we have `compose.htm` iframe on the page (which means secure reply buttons were loaded successfully) than checking for exact selector, but on the next lines we already check for iframe presence.

It was strange for me that this test didn't fail, I ran it locally and found that for `ci.tests.gmail` account `td.Bu` is still present in Gmail code, so reply buttons work well even without our fix.

So we already have check for secure reply buttons presence and test didn't fail just because Gmail code changes affected not all users.

sosnovsky:
@martgil looks like we already have test which checks presence of secure reply buttons, which didn't fail because it works well even without fix for our live test account, so no new tests were needed. So I'll close current PR, thanks for working on it and your quick response on Friday issue👍

martgil:
@sosnovsky I understand completely. I would like to be more involved in problem discovery and finding solutions to various issues we may encounter. Thank you!

PR #5514, opened 2023-12-05
#5466 Fix unencrypted reply rendering

### LINKED ISSUE #5466: unencrypted reply to an encrypted message makes sent reply unreadable with flowcrypt
Hello, 

we are using flowcrypt 8.5.1 with Firefox, and when we reply with the normal "reply" Gmail button to an encrypted message, we initially see the reply correctly in the conversation at the bottom (while the initially encrypted message is not loading, but we see it above)

screenshot: 
Screenshot_20231025_114643

However, when we go back to our reply (e.g. F5), it gets displayed wrongly: we see the unencrypted message, but the previously encrypted message GPG ASCII is added immediately afterwards, making the email hard to read (or in case of images/attachments: they are not available anymore). see first screenshot. The same issue is visible at the receiver of the reply when using flowcrypt.

screenshot: 
Screenshot_20231025_115029

When we try this with e.g. a combination of protonmail and a user who doesn't have a public key (simply adding it manually to the reply), it works at the "non-flowcrypt" side: meaning the GPG armor is displayed in its encrypted form and below the message, so the unencrypted reply is easy to read

screenshot: 
Screenshot_20231025_115301

### LINKED ISSUE #5486: PGP message block mixed with plain text email thread
### Description

The PGP message block of the previous conversion has been mixed up with the plain text email. 

### Steps to reproduce

1. Open the email `Double --` with the thread id `FMfcgzGtwgjqMHdhvWnxfbTwVQfbWsfk` from the `flowcrypt compatibility` account.
2. Notice that the PGP block is mixed up with the plain text email, which somehow breaks the readability of the email.
Optional: Disable the FlowCrypt extension from the browser, and the plain text will start rendering correctly.  

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGwHfqVxRbwzrpTLnHRRvPXwXnH

sosnovsky:
This PR fixes broken rendering of email message which contains unencrypted reply to encrypted message.

close #5466
close #5486

before

Screenshot 2023-12-05 at 21 03 39

after

Screenshot 2023-12-05 at 21 03 55

- Tests added or updated

### To be filled by reviewers


sosnovsky:
Hi @martgil, Ioan will be unavailable for some time, so I assigned you as a reviewer for this PR

martgil (approved):
looks good to me. thanks!

PR #5536, opened 2023-12-18
#5529 Fix non-visible ‘noname’ attachment

### LINKED ISSUE #5529: 'noname' attachment isn't rendered in plain message
I received non-encrypted message with `noname` attachment, it was hidden by FlowCrypt, but should be rendered:

Screenshot 2023-12-15 at 15 00 49

sosnovsky:
This PR fixes non-visible `noname` attachment in non-encrypted message

close #5529

- Tests added or updated

### To be filled by reviewers


sosnovsky:
@martgil I changed your code for handling `noname` attachments a bit, as previously it would hide `noname` attachments even in non-encrypted messages, which was an unexpected behavior.

martgil (approved):

martgil:
> @martgil I changed your code for handling `noname` attachments a bit, as previously it would hide `noname` attachments even in non-encrypted messages, which was an unexpected behavior.

By reviewing the new changes, I have learned where I went wrong. When I tried to fix the "noname" issue before, I focused on the public static `class Attachment` and how it handles "noname" files. However, I should have made changes to the message-renderer. I'm sorry.

sosnovsky:
> By reviewing the new changes, I have learned where I went wrong. When I tried to fix the "noname" issue before, I focused on the public static `class Attachment` and how it handles "noname" files. However, I should have made changes to the message-renderer. I'm sorry.

Hmm, I think you updated `message-renderer` too, as I checked changes history and the latest changes were from your PR about `noname` attachments. We've just missed there that some plain email messages can include `noname` attachments, which are expected to be rendered. So I've added additional check to hide `noname` attachments only when original message has empty body or includes `encrypted.asc` attachment.

PR #5547, opened 2023-12-28
#5537 Update PGP/MIME message structure

### LINKED ISSUE #5537: Update PGP/MIME message format according to standards
> Yeah, looks like we should update code for PGP/MIME message composing in browser extension 
> But maybe it will be a simple fix for Android, as I opened the same message on iOS and it was rendered there (maybe no name for image causes problem on Android):

Yes, I can add a fix. But this will be a non-standard way, and it will significantly complicate the code.

The current browser realization doesn't follow the standard. Let me add more details here

The browser extension generates the following:

```
------sinikael-?=_1-16998840995020.4112130840361884
Content-Type: application/pgp-encrypted; name=
Content-Description: PGP/MIME version identification
Content-Disposition: attachment
X-Attachment-Id: f_mxDfixWgPOEcEygnCeJiNRdWbtvpqe@flowcrypt
Content-Id: <f_mxDfixWgPOEcEygnCeJiNRdWbtvpqe@flowcrypt>
Content-Transfer-Encoding: 7bit

Version: 1
```

Based on https://proton.me/support/pgp-mime-pgp-inline it should be something like this one:

image

Also, based on https://datatracker.ietf.org/doc/html/rfc3156#autoid-4

```
The multipart/encrypted MIME body MUST consist of exactly two body
   parts, the first with content type "application/pgp-encrypted".  This
   body contains the control information.  A message complying with this
   standard MUST contain a "Version: 1" field in this body.  Since the
   OpenPGP packet format contains all other information necessary for
   decrypting, NO OTHER INFORMATION IS REQUIRED HERE.
```

image

All of that means we have to generate the following:

```
------sinikael-?=_1-16998840995020.4112130840361884
Content-Type: application/pgp-encrypted
Content-Description: PGP/MIME version identification
Content-Transfer-Encoding: 7bit

Version: 1
```
###########################################################

For now, on my side **name=** is a problem. JavaMail recognizes it as an error. If a parameter is present it should have a value(or `quoted-string` for the empty value like `name=""`). More details can be found [here](https://datatracker.ietf.org/doc/html/rfc2045#section-5.1)

image

###########################################################
> I tried to send similar message from protonmail and it's rendered in Android app.

It is not `PGP/MIME` encrypted message. It's `PGP/Inline`

image

Like 

image

_Originally posted by @DenBond7 in https://github.com/FlowCrypt/flowcrypt-android/issues/2527#issuecomment-1862745204_

sosnovsky:
This PR updates PGP/MIME message structure according to https://datatracker.ietf.org/doc/html/rfc3156#section-4:
- removed `name` attribute from PGP/MIME version identification attachment
- removed `X-Attachment-Id` and `Content-ID` headers

close #5537

- Tests added or updated

### To be filled by reviewers


sosnovsky:
@martgil this one ready for review, thanks!

martgil:
LGMT!

martgil (approved):

PR #5549, opened 2023-12-30
#5548 Improve error handling in updatePrivateKeyHandler()

### LINKED ISSUE #5548: Improve error handling in `updatePrivateKeyHandler()`
The Pull Request #5545 is not well defined in terms of error handling, resulting in options for fixing the private key UI not appearing. The correct way to properly handle the error is to bring back the check for the private key's identities and then throw an identifiable error message such as `throw new KeyCanBeFixed(encryptedPrv)` which can then be handled in the `handlePrivateKeyError()` accurately.

martgil:
This PR completely fixes the error handling in `updatePrivateKeyHandler()` where it doesn't allow the user to fix those "fixable" private keys when updating their private keys.

close #5548 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/settings/modules/my_key_update.htm
+++ b/extension/chrome/settings/modules/my_key_update.htm
@@ -16,7 +16,7 @@
   </head>
 
   <body id="settings">
-    <div id="content" class="my_key page compatibility_fix_container display_none" style="text-align: left">
+    <div id="my_key_update_container" class="my_key page display_none" style="text-align: left">
       <div class="line"><b>Update Private Key</b> with Fingerprint: <span class="fingerprint good"></span></div>
       <div class="line">
         You can use other OpenPGP compatible software to update your private key (change expiration, change user ids) and load updated key below.
@@ -49,6 +49,7 @@
       </div>
       <div class="line"><a href="#" class="action_show_public_key">Back to Public Key</a></div>
     </div>
+    <div id="compatibility_fix" class="compatibility_fix_container"></div>
 
     <script src="/lib/purify.js"></script>
     <script src="/lib/fine-uploader.js"></script>
--- a/extension/chrome/elements/shared/key_errors.ts
+++ b/extension/chrome/elements/shared/key_errors.ts
@@ -55,10 +55,10 @@ export class KeyErrors {
 
   private toggleCompatibilityView = (visible: boolean) => {
     if (visible) {
-      $('#add_key_container').hide();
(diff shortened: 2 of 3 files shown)


martgil:
@sosnovsky Ready for review. This is the appropriate one. 

Happy New Year.
-Mart

extension/chrome/settings/modules/my_key_update.htm
```diff
@@ -16,7 +16,7 @@
   </head>
 
   <body id="settings">
-    <div id="content" class="my_key page compatibility_fix_container display_none" style="text-align: left">
+    <div id="my_key_update_container" class="my_key page display_none" style="text-align: left">
```
sosnovsky, line 19: I think this change can affect some styles, as we have many css rules for `body#settings > div#content`, it'll be better to leave `id="content"` and use css class here instead

test/source/tests/settings.ts
```diff
@@ -770,7 +770,9 @@ export const defineSettingsTests = (testVariant: TestVariant, testWithBrowser: T
         await myKeyPage.waitAndType('@input-prv-key', keyCanBeFixedTestKey);
         await myKeyPage.waitAndType('@input-passphrase', 'hard to guess passphrase');
         await myKeyPage.waitAndClick('@action-update-key');
-        await myKeyPage.waitAndRespondToModal('confirm', 'confirm', 'Public and private key updated locally');
+        await myKeyPage.waitAll('@input-compatibility-fix-expire-years', { timeout: 30 });
+        await myKeyPage.selectOption('@input-compatibility-fix-expire-years', '1');
+        await myKeyPage.waitAndClick('@action-fix-and-import-key');
```
sosnovsky, line 775: let's also check that key was successfully fixed, as currently we just click `fix` button and test will pass even if some error will be shown after this click. so we should also check imported key

sosnovsky:
> @sosnovsky Ready for review. This is the appropriate one.
> 
> Happy New Year. -Mart

Happy new year to you too!

extension/chrome/settings/modules/my_key_update.htm
martgil: updated. thank you!

test/source/tests/settings.ts
martgil: I have added a simple check for the expiry date ensuring that its value was updated correctly as expected.

sosnovsky (approved):
Works well 👍

PR #5592, opened 2024-02-05
#5498 Upgrade to puppeteer v22

### LINKED ISSUE #5396: Debug html attachments on Semaphore not contain screenshots
Previously debug html attachments had screenshots and information about reason for test failure, but currently they have only mock API logs (example - https://flowcrypt.semaphoreci.com/jobs/0be03fb0-c584-4943-b6d5-4164e2a94ce3/artifacts)

Issue probably appeared after updating to Puppeteer 21.0.0, as `testFailSingleAttemptDebugHtml` fails with error `Error: Requesting main frame too early!`. It works well in Puppeteer 20.9.0

### LINKED ISSUE #5498: Upgrade puppeteer to v22
After upgrading to puppeteer v21 some tests are failing - https://flowcrypt.semaphoreci.com/jobs/9c365e5a-a511-40c7-b9df-87d1da3b6fd0/summary?report_id=4ae71336-e44b-39bf-b9d2-752e234818a5

sosnovsky:
This PR updates test code to be compatible with puppeteer v22

close #5498
close #5396

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-render-module.ts
```diff
@@ -192,7 +193,7 @@ export class ComposeRenderModule extends ViewModule<ComposeView> {
   };
 
   public closeMsg = () => {
-    $('body').attr('data-test-state', 'closed'); // used by automated tests
+    document.querySelector('body')?.setAttribute('data-test-state', 'closed'); // used by automated tests
```
ioanmo226, line 196: @sosnovsky What was the reason to use javascript query selector instead of jquery?
Was there some problem?

sosnovsky: I've tried it as possible solution for puppeteer not detecting correct value of `data-test-state` value, but it didn't help. However I left `querySelector` as for such simple selectors we can use native js functions, instead of jquery selectors.

ioanmo226: I see

sosnovsky: Initially I thought that it's possible to migrate from jquery to using native js methods, but we have a lot of jquery related code and it'll be too time-consuming. For now we can just start using native js selectors, to not produce more jquery dependent code.

ioanmo226:
Hmmm...
Currently stuck with not finding element inside iframe for tests like `decrypt - not an armored public key in a file that looked like a public key`
Seems like strange issue.
Checking deeply now.

sosnovsky:
Yeah, it was update in some chrome version (which affected `puppeteer` too) where it changed iframe behavior, making some iframes content inaccessible (wasn't able to find what exactly it causes). For example, adding `.reverse()` here fixed an issue with inaccessible iframe element, because of changed order of iframes rendering:
https://github.com/FlowCrypt/flowcrypt-browser/pull/5592/files#diff-7d582ec344027d9d2e8423c4dfafb1d130c549a2e5f4c3163e75b986c20c29f2R508

sosnovsky:
@ioanmo226 there were a new `22.8.0` puppeteer update - https://github.com/puppeteer/puppeteer/blob/main/packages/puppeteer-core/CHANGELOG.md#2280-2024-05-06 which includes `disable IsolateSandboxedIframes trial to prevent flakiness`, maybe it'll work for our failing tests

ioanmo226:
Thank you for your info.
let me check

ioanmo226:
Sure, here's the revised text:

---

Currently, only three tests are failing.
For example, in the `mail.google.com - success notification after setup, click hides it, does not re-appear + offers to reauth` test, we set the Google access token and refresh token to `test_bad_access_token` in the code snippet below.
However, as shown in the image, the request uses a previously cached authorization header, causing the app to treat it as a valid token.

I believe this is an issue with Puppeteer/Chromium and should be addressed there.
Interestingly, this caching issue only occurs on our Gmail mock page and not on our extension pages.

Let me know your thoughts.

I added a `console.log` for the `sendAs` API call's authorization header.
image
Here's the request screenshot.
image

https://github.com/FlowCrypt/flowcrypt-browser/blob/2054795f84a9be29ab37a560411ef7d54ca31c6a/test/source/tests/setup.ts#L649-L650

sosnovsky:
@ioanmo226 it seems to be not a bug, but a feature :)

[page.setExtraHTTPHeaders](https://github.com/FlowCrypt/flowcrypt-browser/blob/3f9039533f71895e82bdf08abd5ec188fd99c17f/test/source/browser/browser-handle.ts#L30) is used for some tests, which automatically adds headers to all outgoing requests.
I noticed similar issue when fixing https://github.com/FlowCrypt/flowcrypt-browser/issues/5715 - outgoing test request shouldn't have `Authorization` header, but it's still present.
To fix this issue, `extraHeaders` needs be removed from specific test, but it can lead to some `missing header` errors. Currently trying to make it work for https://github.com/FlowCrypt/flowcrypt-browser/issues/5715

ioanmo226:
Hmm... I think setExtraHttpHeaders function should be used to set header for page itself.
But currently it seems to override api requests(Like above `sendAs` api).

ioanmo226:
LIke if we set header to `Autorization: Bearer test-bad-token` and if we try to call `sendAs` api with header `Authorziation: Bearer {my_token}` then it seems to override to  `Autorization: Bearer test-bad-token`  headert

ioanmo226:
As you can see, I set a custom `testData` header, but it was overridden in the sendAs API, which should not have happened.

image

sosnovsky:
Yes, `setExtraHTTPHeaders` seems to overwrite existing headers if they're present, but strange that the same test passes in puppeteer v20, as it works the same there.

ioanmo226:
I mean it's overriding api (fetch/xmlhttpreqest) headers. I think this is wrong behavior. It's overriding all api call headers.

Below is screenshot from puppeteer v20 and you can see that there is no `testData` header in here which is in v22.

image

ioanmo226:
@sosnovsky What do you think about above?

sosnovsky:
I tried to fix failing tests by creating new pages instead of reloading existing ones (so they get updated `setExtraHTTPHeaders`) in https://github.com/FlowCrypt/flowcrypt-browser/tree/5498-puppeteer-v22, and it seems to work well, but some others tests are broken. Will check them later this week
I moved this issue to the next milestone - want to prepare a new extension build and send it to Chrome Web Store review today, so we'll have some time before manifest v2 deprecation in June if they'll find any issues in our migration.

sosnovsky:
@ioanmo226 this one is finally ready for review

ioanmo226 (approved):
👍

PR #5607, opened 2024-02-19
#1509 Migrate to Manifest V3

### LINKED ISSUE #1509: google chrome manifest v3 web extensions proposal [June 2024]

sosnovsky:
This PR migrates extension code to Manifest V3

close #1509

- Tests added or updated

### To be filled by reviewers


sosnovsky:
@ioanmo226 here is my current changes for manifest v3 migration, you can continue working on this code, feel free to ask questions:
- the biggest change in manifest v3 is usage of service workers instead of background pages - https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers. Previously background pages were constantly running in background, while service workers are alive only when we call them from extension.
- it breaks our `ExpirationCache`, as with current implementation it's not possible to use it for storing cache values for predefined period of time. I renamed it to `SimpleExpirationCache`, and tried to use `chrome.storage.session` for storing cache values
- also `window.` is not available in service worker, it led to many `window is undefined` errors. I replaced `window.location.href` with `location.href`, `window.document.body` with `document.body` etc and those errors are gone, but not sure if it breaks this code functionality
- I also copied code from `extension/js/background_page` to `extension/js/service_worker` with some changes to make it work with manifest v3

Per updated Google timeline - in June 2024 they'll start disabling manifest v2 extensions in pre-release versions of Chrome, so by that time we should have finished and tested migration to manifest v3.
It'll be great to have fully functional build in the beginning of April, so we can test it to find any issues during April and release final manifest v3 build in the beginning of May.

If some functionality can't be fully migrated to manifest v3 - we can simplify it for now and work on full functionality later. Google regularly adds new features to manifest v3, maybe some new feature will be helpful for us and make implementation easier.

P.S. Also Firefox has a different implementation for manifest v3 - https://extensionworkshop.com/documentation/develop/manifest-v3-migration-guide/. They continue support background pages, but they are non-persistent in manifest v3, need to adapt code for them too.

ioanatflowcrypt:
@sosnovsky Thank you for your detailed response and code update.

I have now finished checking your code and manfest v3 update guide.

Could you list TODO list (to update to v3) you have in mind? It would help me complete task faster.

PS: I lost 2FA code for my previuos github account and submitted recovery action. Might be recovered in 1-2 days.

sosnovsky:
> Could you list TODO list (to update to v3) you had in mind? It would help me complete task faster.

For now we should start with these issues, as implementing them will make possible to run UI tests which will show broken functionality, and then it'll be easier to understand what needs to be fixed:
- [ ] insert "Secure compose" button on each gmail.com page load (currently it's inserted only on first load, I think, when service worker is initialized)
- [ ] implement replacement for current `ExpirationCache`, as with manifest v3 it's not possible to store something in memory for predefined amount in time

Initially I thought it'll be possible to create manifest v2 and v3 builds from the same codebase, then we could merge manifest v3 changes to `master` one by one, but it seems to be quite complicated, so we'll need to have separate branch for manifest v3 until it's ready to be released.

ioanmo226:
Thank you for your answer. Let me implement above

ioanmo226:
@sosnovsky Could you please check if `Setup Flowrypt` initial screen works fine in your side?
When I tap `Sign in with google` and then select account and click next, it just do nothing and console displays `Uncaught (in promise) Error: No tab with id: 1745224200.
`

sosnovsky:
> @sosnovsky Could you please check if Setup Flowrypt initial screen works fine in your side?
When I tap Sign in with google and then select account and click next, it just do nothing and console displays Uncaught (in promise) Error: No tab with id: 1745224200.

It probably doesn't work, as I've tested only already configured extension and didn't go through setup process.

> Also Secure compose button is not even inserted on first load in my side.
> Could you check if you pushed latest changes?

Yes, it's my latest changes, probably one of my last changes broke compose button insert even on the first load :)
Maybe it's related to `inMemoryStorage`, need to check if token is successfully fetched there

ioanmo226:
Got it. Thank you for your answer/

sosnovsky:
> Got it. Thank you for your answer/

Feel free to message me if you'll have any questions about manifest v3 migration, as it's quite complicated issue and probably second opinion will be useful for making some implementation decisions.

ioanmo226:
Which second option do you mean?
> probably second opinion will be useful for making some implementation decisions

ioanmo226:
Aha, I see

ioanmo226:
@sosnovsky This one is one is now fixed.

> @sosnovsky Could you please check if Setup Flowrypt initial screen works fine in your side?
> When I tap Sign in with google and then select account and click next, it just do nothing and console displays Uncaught (in promise) Error: No tab with id: 1745224200.
> 

However `InMemoryStore.getUntilAvailable` function does not perform as expected; it only invokes the `InMemoryStore.get` method once, even if the value remains undefined.
This issue likely arises because the service worker remains active only during use and is terminated afterwards, preventing the use of `Timer.sleep`.
Given the necessity of `setInterval` or `setTimeout` in various parts of the extension, I recommend utilizing `chrome.alarms` instead. 

I have implemented this solution in the latest commit. I welcome your feedback and am open to any suggestions you may have. Additionally, please inform me if you consent to adding the `alarms` permission.

https://github.com/FlowCrypt/flowcrypt-browser/blob/e3b3e63850a11fcd221ce385cfb0938f79b6d52d/extension/js/common/platform/store/in-memory-store.ts#L23

https://github.com/FlowCrypt/flowcrypt-browser/pull/5607/commits/3abf9f9b2aff708d8d667d3be06c7f7304be77b4

sosnovsky:
> Given the necessity of setInterval or setTimeout in various parts of the extension, I recommend utilizing chrome.alarms instead.
> I have implemented this solution in the latest commit. I welcome your feedback and am open to any suggestions you may have. Additionally, please inform me if you consent to adding the alarms permission.

Yes, we'll need to use `alarms` permission, as it's required for running timers in service worker (https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers#alarms), so you can add it to `manifest.json`. As I remember, adding `alarms` permission shouldn't show additional permissions alert for users.

By the way, when working on manifest v3 update I found tool for testing extension update - https://github.com/GoogleChromeLabs/extension-update-testing-tool. It'll show how update to the new version will look for users, so you can check if any additional permission alerts will be shown (we should try to migrate without such additional alerts, as some users can decline them and then some functionality will be broken))

> PS: I thought there was logic in Catch.setHandledTimeout but I can see that it does nothing.
> Maybe we need to implement setHandledTimeout and setHandledInterval with chrome.alarms api?

I commented initial logic and temporary replaced it with `return ms`, as extension failed to build because of `window is undefined` error. We'll need to reimplement it using `chrome.alarms` API, you're right

ioanmo226:
Got it. Thank you for your detailed comment

ioanmo226:
Got it. Thank you for your detailed comment

extension/js/common/platform/catch.ts
```diff
@@ -243,16 +244,36 @@ export class Catch {
...
+  public static setHandledInterval = (cb: () => void | Promise<void>, ms: number): string => {
+    // Couldn't set chrome.alrams here because chrome.alrams becomes undefined in gmail.com content script
+    const alarmName = `interval_${new Date().getTime()}_${ms}`;
+    BrowserMsg.send.bg.createAlarmWithDelay({ alarmName, ms });
+    BrowserMsg.addListener('perform_timeout_action', async ({ alarmName: intervalAlarmName }: Bm.PerformTimeoutAction) => {
+      if (alarmName === intervalAlarmName) {
+        Catch.try(cb)();
+      }
+    });
+    return alarmName;
```
ioanmo226, line 260: @sosnovsky As you can see, I've implemented the `setInterval` function as shown above. I had to create an alarm on the service worker side and then receive an event for the `perform_action`. As you know, we cannot send a callback (function) using `chrome.runtime.sendMessage`.

However, this approach doesn't seem ideal for a few reasons:
1. The `BrowserMsg.addListener('perform_timeout_action')` part won't be triggered upon completion of the `Catch.setHandledInterval` call or when the receiving end doesn't exist.
2. The code becomes somewhat cluttered.

Of course, we need to implement logic to use `setInterval` and `setTimeout` for normal pages (not service workers). But for service workers, I'm not sure what the best solution would be to implement `setInterval` and `setTimeout`.

Do you have better suggestion for implementing this?

extension/chrome/elements/compose-modules/compose-draft-module.ts
```diff
@@ -28,7 +28,7 @@ export class ComposeDraftModule extends ViewModule<ComposeView> {
 
   private currentlySavingDraft = false;
   private disableSendingDrafts = false;
-  private saveDraftInterval?: number;
+  private saveDraftAlarm?: string;
```
sosnovsky, line 31: I think we still can use `setInterval` and `setTimeout` in content script code, `chrome.alarms` should be used only for background timers in service worker (for example for deleting expired items in in-memory storage - `Catch.setHandledInterval(() => inMemoryStore.deleteExpired(), 60000);`)

extension/js/background_page/background_page.ts
```diff
@@ -59,8 +59,11 @@ console.info('background_process.js starting');
   BrowserMsg.bgAddListener('ajaxGmailAttachmentGetChunk', BgHandlers.ajaxGmailAttachmentGetChunkHandler);
   BrowserMsg.bgAddListener('settings', BgHandlers.openSettingsPageHandler);
   BrowserMsg.bgAddListener('update_uninstall_url', BgHandlers.updateUninstallUrl);
+  BrowserMsg.bgAddListener('create_alarm_with_delay', BgHandlers.createAlarmWithDelay);
+  BrowserMsg.bgAddListener('clear_alarm', BgHandlers.clearAlarm);
```
sosnovsky, line 63: this file isn't used in manifest v3, such listeners should be added in https://github.com/FlowCrypt/flowcrypt-browser/blob/05ca61543f39dca62997b3554adbba9e1559a182/extension/js/service_worker/background.ts

extension/js/background_page/bg-handlers.ts
```diff
@@ -45,6 +45,15 @@ export class BgHandlers {
...
 
+  public static createAlarmWithDelay = async ({ alarmName, ms }: Bm.CreateAlarmWithDelay) => {
+    // Create the alarm with delay
+    void chrome.alarms.create(alarmName, { when: Date.now() + ms });
+  };
+
+  public static clearAlarm = async ({ alarmName }: Bm.ClearAlarm) => {
+    void chrome.alarms.clear(alarmName);
+  };
+
```
sosnovsky, line 56: this should be in https://github.com/FlowCrypt/flowcrypt-browser/blob/05ca61543f39dca62997b3554adbba9e1559a182/extension/js/service_worker/bg-handlers.ts

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -215,9 +215,9 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
...
           r.send();
           let status: number;
-          const responsePollInterval = Catch.setHandledInterval(() => {
+          const responsePollAlarm = Catch.setHandledInterval(() => {
             if (status >= 200 && status <= 299 && (r.responseText.length >= minBytes || treatAs === 'publicKey')) {
-              window.clearInterval(responsePollInterval);
+              Catch.clearAlarm(responsePollAlarm);
               processChunkAndResolve(r.responseText);
               r.abort();
             }
```
sosnovsky, line 223: probably can use usual `setInterval` instead of `chrome.alarms` here too

extension/js/common/platform/catch.ts
ioanmo226: How about this one? @sosnovsky

sosnovsky: Probably we can move `chrome.alarms` code to service worker?
As I remember, we don't need to set background timers from content script code

sosnovsky: I mean here we can use old code with `setInterval`, and background timers using `chrome.alarms` will be set only in service worker, as they're initiated there

ioanmo226: Yep!
I thought too much.
It should be much easier

extension/js/background_page/background_page.htm
```diff
@@ -1,21 +0,0 @@
-<html>
-  <head>
-    <meta content="text/html;charset=utf-8" http-equiv="Content-Type" />
-  </head>
-
-  <body>
-    <script src="/lib/trusted-types.js"></script>
-    <script src="/lib/purify.js"></script>
-    <script src="/lib/jquery.min.js"></script>
-    <script src="/lib/openpgp.js"></script>
```
ioanmo226, line 10: @sosnovsky  As we removed background_page, now opgp becomes null for below code.
I'm thinking about those 3 solutions.

1. Import ES module in service worker (background.ts). Maybe it's openpgp.min.mjs?. We can't just import openpgp.js (by use `import lib/openpgp.js` as openpgp.js code is `something like var openpgp = (function (exports) {....})()`)
2. Use webpack or other solution to combine libraries like openpgp in backgroun.ts
3. Use `[importScripts](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts)` to load library (AFAIK, it is deprecated)?

What's your take on this?
Please let me know your suggestion.

https://github.com/FlowCrypt/flowcrypt-browser/blob/4edbf1836cff75b6262bacc1757f45382661520c/extension/js/common/core/crypto/pgp/openpgp-key.ts#L39

sosnovsky: > Use [importScripts](https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts) to load library (AFAIK, it is deprecated)?

I didn't find any information about it's deprecation, and it's recommended at https://developer.chrome.com/docs/extensions/develop/concepts/service-workers/basics, so we can try to use `importScripts` for importing `openpgp`

ioanmo226: I found it from stackoverflow I think but couldn't find link.
Here's alternative link
https://www.mediawiki.org/wiki/Topic:Ufathwjhyszsis17

sosnovsky: It seems this wiki article is specific for mediawiki, while for chrome extensions `importScripts` is a recommended solution for importing js files, let's try to use it

ioanmo226: Okay, Got it

ioanmo226: You know we are using servie worker as a module therefore we can't use `importScripts`. If we remove `module`, then `import` statements won't work in `background.ts` file

> Uncaught TypeError: Failed to execute 'importScripts' on 'WorkerGlobalScope': Module scripts don't support importScripts().

Therefore there are 2 options left from my suggestion?

> - Import ES module in service worker (background.ts). Maybe it's openpgp.min.mjs?. We can't just import openpgp.js (by use import lib/openpgp.js as openpgp.js code is something like var openpgp = (function (exports) {....})())
> - Use webpack or other solution to combine libraries like openpgp in backgroun.ts
> 

https://github.com/FlowCrypt/flowcrypt-browser/blob/056d979c3704e60ca05b4a04adf9b5aaa6dc3e73/extension/manifest.json#L70

ioanmo226: Another solution I'm thinking is make our custom `importScripts` function which fetches library first and then execute it with `eval` or something like that. But it might come with security risk

sosnovsky: Yeah, it's better to use some ready solutions than write custom functions, which can be broken after some browser update.

So let's try to import ES module in service worker, and see if it works for our case.

ioanmo226:
Just put this link for reference for firefox manifest v3
https://github.com/mozilla/web-ext/issues/2532#issuecomment-1657258201

ioanmo226:
@sosnovsky Importing openpgp library(ES Module) feature is now implemented and all basic functions now work fine.
But now we have `forge` problem.
`forge` doesn't support/provide ES Module.

Maybe we should consider this? We can't guarantee every library can support ESM.
> Another solution I'm thinking is make our custom importScripts function which fetches library first and then execute it with eval or something like that. But it might come with security risk

ioanmo226:
What's your opinion about this?
Could you let me know your suggestion to this issue?

sosnovsky:
From mentioned options I would prefer using webpack, as we already use it for packing `@openpgp/web-stream-tools` for content script.

But probably it's possible to just remove `openpgp` and `node-forge` from service worker?
Some time ago we worked on moving openpgp.js from background page to content script (https://github.com/FlowCrypt/flowcrypt-browser/issues/4906), and as it's now included in content script maybe we can now remove it from service worker and move remaining functionality to content script?
Same for `node-forge` - can try to move it's calls to content script.

ioanmo226:
Ok, let me try to move from service worker to content script for openpgp and forge

ioanmo226:
@sosnovsky I checked deeply but I think we couldn't get openpgp completely out of service worker?
As you can see, ContactStore depends on database which is opened from service worker.
If we try to reference this database from content script, there should be n calls to open database when there are n iframes. (Also currently it returns error though)
I tried to reference `db` variable from service worker and tried to use it from ContactStore but it also didn't succeed

What's your suggestion?

https://github.com/FlowCrypt/flowcrypt-browser/commit/c5cf84d80ec394606a8e03e65bcb4937cae07874
https://github.com/FlowCrypt/flowcrypt-browser/commit/2ab84bd8bd371031eec26137168528900de34bbf

sosnovsky:
Yeah, looks like it won't be easy to remove `openpgp` from service worker (probably can work on it after finishing manifest v3 migration), but it seems to work well with your current code.
And what about `node-forge`, can it be removed from service worker?

ioanmo226:
Haven't checked yet. let me check

ioanmo226:
@sosnovsky As I checked, it won't be easy to move `forge` either.
It's not forge is alone used in service worker but forge was called from `ContactStore` which should be depend on service worker because of IDBDatabase.
I think the only option we have is use `webpack` to pack libraries into `background.js`

Please correct me if I'm wrong and let me know your suggestion.

https://github.com/FlowCrypt/flowcrypt-browser/commit/d2aff54c7ae41ccde25acf927cf26179968b8f04

sosnovsky:
> I think the only option we have is use webpack to pack libraries into background.js

Ok, let's go with webpack solution for now

ioanmo226:
@sosnovsky  Hmm.. it's becoming more complex.
When I try to import forge with `import * as forge from 'node-forge` (already decleared path in tsconfig) it threw `window is not defined error`.

Do we have to fix this issue in forge and submit PR?
Or find a way to get forge out of service worker? (Might be difficult though)
Or do you have other suggestions?

https://github.com/FlowCrypt/flowcrypt-browser/commit/5834c1a403db0aadac158e0f84a097172009803c

sosnovsky:
> Do we have to fix this issue in forge and submit PR?

It seems `node-forge` isn't actively developed now, last release was a couple years ago.
Probably we can apply fixes to `node-forge` when building extension, like we currently do for `openpgp` and `web-stream-tools`, what do you think? https://github.com/FlowCrypt/flowcrypt-browser/blob/2bbe51655d67d9ae5100d22428ff0f089edb1de9/scripts/build.sh#L124

ioanmo226:
I agree but not sure if it can be fixed with just regex replace or something like that.
Let me try.

ioanmo226:
@sosnovsky I tried to bundle background.js with webpack but I'm stuck at this. :-(
It still displays forge.pkcs7, forge.asn1, etc are undefined.

Could you take a look at this issue? Would be much appreciated.
I'll also look at this issue too though.

sosnovsky:
> Could you take a look at this issue? Would be much appreciated. I'll also look at this issue too though.

For sure, I'll check it later today or tomorrow

ioanmo226:
@sosnovsky Have you checked?

sosnovsky:
> @sosnovsky Have you checked?

Yes, I'm checking it - tried to build without webpack and noticed that `forge.js` isn't included in `lib` folder for service worker (only `openpgp` and `streams` files are there). I think that's why building with `webpack` doesn't help, as importing `forge` doesn't correctly references `forge.js`, checking it further now

Screenshot 2024-03-14 at 10 31 30

ioanmo226:
Hmm.. for me even without webpack, forge is included in lib folder.
image

ioanmo226:
Maybe you didn't remove this `comment` in tsconfig.json?

image

sosnovsky:
> Maybe you didn't remove this `comment` in tsconfig.json?

Yeah, you're right, after removing `forge.js` appeared in service worker, thanks for the tip!

ioanmo226:
@sosnovsky I tried to build forge mjs file from forge repository with rollup library.
It seems to work fine. What do you think about my solution?

sosnovsky:
> @sosnovsky I tried to build forge mjs file from forge repository with rollup bundler. It seems to work fine. What do you think about my solution?

For me it doesn't work - `secure compose` button isn't added to Gmail page and there are errors in console:
Screenshot 2024-03-15 at 10 54 02

Does it work for you?

ioanmo226:
OHhh. yep
It doesn't work in gmail page.
Just tried extension pages and mock ui tests.

sosnovsky:
@ioanmo226 did you try to use [offscreen document](https://developer.chrome.com/docs/extensions/develop/migrate/to-service-workers#move-dom-and-window) for fixing `openpgp` and `forge` imports? Here is another article - https://developer.chrome.com/blog/Offscreen-Documents-in-Manifest-v3
Not sure if it can be used for this, but can be worth checking.

ioanmo226:
Haven't tried yet. But I heard using offscreen object for referencing variable is not good idea and should not be used.
I think from online article or something like that.
Let me dig deep though

extension/js/common/core/crypto/smime/smime-key.ts
```diff
@@ -1,18 +1,23 @@
...
 export const DATA_OID = '1.2.840.113549.1.7.1';
 export const SIGNED_DATA_OID = '1.2.840.113549.1.7.2';
 export const ENVELOPED_DATA_OID = '1.2.840.113549.1.7.3';
 export class SmimeKey {
+  private static get forgeCustom(): typeof forge {
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    return forge.default ? forge.default : forge;
+  }
```
ioanmo226, line 20: @sosnovsky How about now? What do you think about solution?
When I import like above `import * as forge` then i have to call `forge.default.asn1` instead of `forge.asn1`. Therefore I added above logic.
Maybe there could be simple solution for this.

Let me know your suggestion.

ioanmo226: Ok, I'm currently fixing `test re-auth after updating chrome extension` test (It fails because OAuth2.js coudln't get window object because it was running from background). Fixing it to pass requried screen dimensions to service worker and pass it to OAuth2.

Let me have a look at  `ajaxGmailAttachmentGetChunk` after current one

ioanmo226: I have a question. Why don't we just use `fetch` in `ajaxGmailAttachmentGetChunk`?

ioanmo226: Also I think we couldn't move `ajaxGmailAttachmentGetChunk` to content script because of below reason?
https://github.com/FlowCrypt/flowcrypt-browser/blob/cb1423a3bcbeecb3a8471c457c94a8e76280d9fd/extension/js/common/api/email-provider/gmail/gmail.ts#L160

ioanmo226: Okay, Got it.

ioanmo226: Hmm.. This manifest v3 migration really has tons of problems.

`URL.createObjectURL` is now removed in service worker and therefore we can't use it.

Currently finding solution now.

https://github.com/FlowCrypt/flowcrypt-browser/blob/a852f8779ff0d921e49fa5d79b07196d62abdd65/extension/js/common/browser/browser-msg.ts#L556
https://github.com/FlowCrypt/flowcrypt-browser/blob/4edbf1836cff75b6262bacc1757f45382661520c/extension/js/common/browser/browser.ts#L16
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/u0NH7L3v9L4?pli=1#:~:text=Some%20workarounds%3A,use%C2%A0URL.createObjectURL.

ioanmo226: Do you agree using offscreen api?

sosnovsky: > Do you agree using offscreen api?

Yes, let's use it.

> Hmm.. This manifest v3 migration really has tons of problems.

Yep, that's why we deferred this migration, but now there is no choice as manifest v2 will be fully deprecated this summer :(

ioanmo226: @sosnovsky I got a bit confused. We tried to create object url (URL.createObjectURL) because browser message can't send a lot of data. But if we use offscreen document we also need to send message to offscreen blob data.
Isn't this what we wanted to avoid with createObjectURL?

Please correct me if I'm wrong.

https://github.com/FlowCrypt/flowcrypt-browser/blob/a852f8779ff0d921e49fa5d79b07196d62abdd65/extension/js/common/browser/browser-msg.ts#L556
https://github.com/FlowCrypt/flowcrypt-browser/blob/a852f8779ff0d921e49fa5d79b07196d62abdd65/extension/js/common/browser/browser-msg.ts#L545

sosnovsky: It seems there is no strict limit on maximum size of browser message - it'll probably depend on available system memory. Here we'll have maximum size of 10-15mb (as it's the max for message attachments), so let's try to remove usages of `URL.createObjectURL` and send original objects instead.

ioanmo226: Ok

ioanmo226: Ok

ioanmo226: I removed `URL.createObjectURL` usage but `decrypt - public key is rendered minimized for outgoing messages` test is failing because of that. I tried to run this test in master branch with `URL.createObjectURL` usage removed and it also failed.
I'm digging more.

sosnovsky: Are there any errors in console? I think public key size shouldn't be too large for browser msg

ioanmo226: No errors in console

ioanmo226: No errors in console

ioanmo226: I found reason and it's because when message is returned from `ajaxGmailGetChunk`, with the `createObjectUrl` usage removed `object` is returned instead of `Buf`.
This is a temp fix and it fixed issue.
https://github.com/FlowCrypt/flowcrypt-browser/pull/5607/commits/4262aeeed49e11f09caf0a31413b7515a63c28b7

ioanmo226: I found reason and it's because when message is returned from `ajaxGmailGetChunk`, with the `createObjectUrl` usage removed `object` is returned instead of `Buf`.
This is a temp fix and it fixed issue.
https://github.com/FlowCrypt/flowcrypt-browser/pull/5607/commits/4262aeeed49e11f09caf0a31413b7515a63c28b7

ioanmo226: Also FYI created new PR as we also have to test live test with new branch name

sosnovsky: > I found reason and it's because when message is returned from `ajaxGmailGetChunk`, with the `createObjectUrl` usage removed `object` is returned instead of `Buf`. This is a temp fix and it fixed issue. [4262aee](https://github.com/FlowCrypt/flowcrypt-browser/commit/4262aeeed49e11f09caf0a31413b7515a63c28b7)

Ok, let's just add `// todo: it's a temporary fix, should be removed` comment, so we won't forget about it

ioanmo226: Okay

PR #5613, opened 2024-02-21
#5610 fix: windows build issue 

### LINKED ISSUE #5610: Build issue in windows
When we run `npm run build` or `npm run build-incremental` in git bash or something like that, it throws below error.

image

ioanmo226:
This PR fixed build issue in windows system.

close #5610 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ $ cd some/folder/flowcrypt-browser
 $ npm install
 ```
 
-2. To build the project (currently available for Linux and Mac only), run the following command:
+2. To build the project, run the following command:
 
 ```bash
 $ npm run build
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ $ npm install
 $ npm run build
 ```
 
+- For building project on Windows we recommend to use [Cygwin](https://www.cygwin.com/) with rsync package.
+
 After executing the build command, you can find your built project in the `build/chrome-consumer` and `build/firefox-consumer` folders.
 
 > ℹ️ Note: The `build` folder also contains other versions of the browser extension.
--- a/package.json
+++ b/package.json
@@ -61,9 +61,9 @@
     "zxcvbn": "4.4.2"
   },
   "scripts": {
-    "build": "./scripts/build.sh",
-    "build-incremental": "./scripts/build.sh --incremental",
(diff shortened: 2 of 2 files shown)


ioanmo226:
Hmm... This one seems complex than I've imagined.
when we use sh in linux, it doesn't recognize `set -euxo pipefail`
when we use bash in windows, it throws below error in cygwin


> scripts/build.sh: line 2: $'\r': command not found
> scripts/build.sh: line 7: $'\r': command not found
> : invalid option namee 8: set: pipefail
>

ioanmo226:
Also in `git bash` `sh build.sh` works but in `cygwin` `sh build.sh` throws error.
Maybe shell version is different?

ioanmo226:
This cross-platform build cannot be easily developed or maintained with a bash script.
Therefore, my suggestion is to rewrite the script using Node.js. By doing so, the script will function properly across all platforms, independent of bash versions and platform differences.

@sosnovsky What do you think?
Do you think this one is worth doing it? If so when? right now or low priority?

sosnovsky:
> This cross-platform build cannot be easily developed or maintained with a bash script. Therefore, my suggestion is to rewrite the script using Node.js. By doing so, the script will function properly across all platforms, independent of bash versions and platform differences.
> 
> @sosnovsky What do you think? Do you think this one is worth doing it? If so when? right now or low priority?

I think it'll require too much time as our build script is quite complicated and it won't be easy to rewrite it using node.js.
Our main dev environments are macOS and linux, so it'll be ok to support only them for now.

> Also in git bash sh build.sh works but in cygwin sh build.sh throws error.

Maybe it fails because of the first line in `build.sh` - `#!/usr/bin/env bash`, as it's specific for *nix systems.

If you need to build extension on Windows, it'll probably be better to create copy of `build.sh` and name it `build-windows.sh`, then update it's code to work in cygwin, and add new script to `package.json` - `"build-windows": "sh ./scripts/build-windows.sh"`

ioanmo226:
I'm also using Mac and Linux so it's not a problem.
Do you wanna me to take a look at this one first or manifest v3 update?

sosnovsky:
let's work on manifest v3 update, as currently there is no need to fix build on windows

PR #5676, opened 2024-04-13
#5664 Pull email's received time directly on Gmail

### LINKED ISSUE #5664: Time in print-preview does not follow/match the time format from Gmail
Description:

The time for the received email shown in print-previewing a secure email does not match the time-format Gmail used to display in the email thread view.

Feedback from the user:

> 4/8/2024 - When I look at a flowcrypt message I sent in gmail, it shows the correct time sent in the normal 12 hour format, but when I print out that same email it shows a 24 hour clock format, and the time is at least 4 hours different.  How can I correct that please?

Reference:
https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzGxSbrvkWfcfvtQmDLBbTLgfMXt

martgil:
The goal of this PR is to pull the exact time an email was received from Gmail. This also resolves issues with accurately retrieving the time, whether it's in a 12 or 24-hour format, since it's already provided within the pulled data from Gmail.

close #5664 

Screenshots:
Opening the same email using different time format 

12 hour format:

image
image

24 hour format:

image
image

To change time format in Gmail:

1.) Go to Gmail Settings -> See all settings.
2.) Change language from 'English US' to 'English UK'.
3.) Save and open the desired email.

- Tests added or updated

### To be filled by reviewers


martgil:
Hi @sosnovsky, I hope you're doing well. I'd like to check if you have any preferences regarding the date and time format, or if the one proposed in this PR is sufficient?

extension/js/common/message-renderer.ts
```diff
@@ -430,8 +430,8 @@ export class MessageRenderer {
   };
 
   private getMessageInfo = async (fullMsg: GmailRes.GmailMsg): Promise<MessageInfo> => {
-    const sentDate = GmailParser.findHeader(fullMsg, 'date');
-    const sentDateStr = sentDate ? Str.fromDate(new Date(sentDate)).replace(' ', ' at ') : '';
+    const sentDate = $('div.gK span[title]').attr('title');
```
sosnovsky, line 433: let's also add check that this value is a valid date (for example, if Google will change it's markup in the future and `div.gK` will have another content).

and if it's value is invalid or empty, then we should have fallback by using value parsed from `fullMsg` (`GmailParser.findHeader(fullMsg, 'date');`). it can also be converted to user's timezone, so print info will show correct time for user.

martgil: Hi @sosnovsky the `.toISOString()` seem not to be one we're looking for for a function that is aware of user's timezone and so `.toLocaleString()` be my option to use.

That being said this would be inappropriate now for this PR but would be interested to hear your feedback before I will place it as a new PR and close this one since we would no longer be needed to pull the date and time from Gmail web UI.

sosnovsky: Looks clean 👍 But need to check if it doesn't break anything, as several tests are failing now.

martgil: Yes, Roma — I'll do it. I've come up with a different idea to improve them. Hopefully, it works. Be right back shortly.

martgil: Hello @sosnovsky I ended up adding a fallback option as the original suggestion suggested. I think it would be the best option to go with.

extension/js/common/message-renderer.ts
```diff
@@ -431,7 +431,13 @@ export class MessageRenderer {
...
   private getMessageInfo = async (fullMsg: GmailRes.GmailMsg): Promise<MessageInfo> => {
     const sentDate = GmailParser.findHeader(fullMsg, 'date');
-    const sentDateStr = sentDate ? Str.fromDate(new Date(sentDate)).replace(' ', ' at ') : '';
+    const gmailDateReceived = $('div.gK span[title]').attr('title');
+    let sentDateStr: string;
+    if (gmailDateReceived !== undefined && !isNaN(new Date(gmailDateReceived).getTime())) {
+      sentDateStr = gmailDateReceived.replace(', ', ' at ');
+    } else {
+      sentDateStr = sentDate ? Str.fromDate(new Date(sentDate)).replace(' ', ' at ') : '';
+    }
```
sosnovsky, line 440: Why did you decide not to use `toLocaleString`? It seemed to work well - produced correctly formatted time strings.

Current implementation can lead to incorrect string for some users, as we don't exactly know which format they have in Gmail UI. For example how it formats date for me:

gmail ui:
Screenshot 2024-04-26 at 12 10 02

print view (`Mar 14 at 2024` looks strange):
Screenshot 2024-04-26 at 12 10 24

martgil: Hello @sosnovsky, thanks for checking it. Here are my comments.

> Why did you decide not to use toLocaleString? It seemed to work well - produced correctly formatted time strings.

I haven't gathered more information about why the original devs come up to using "toLocaleString()" so I thought that although I have used it correctly, there might be reason why they used it -- I have that in mind which give me hesitation -- but I'm looking into adding it now as I check it -- I definitely return correct formatted time.

> Current implementation can lead to incorrect string for some users, as we don't exactly know which format they have in Gmail UI. For example how it formats date for me:

This is a bit of a problem. Some date formats that Gmail shows isn't correctly recognized by javascript's "Date" constructor that leads to inaccurate dates. Would it be good if we used the dates from Gmail as is (safest option) or should be parse them? Parsing with other language could also have some drawbacks.

martgil: I'll find a way... I believe I can.

sosnovsky: > I haven't gathered more information about why the original devs come up to using "toLocaleString()" so I thought that although I have used it correctly, there might be reason why they used it 

Probably `toISOString` is used for consistent formatting across different locales, but we've missed to apply timezone offset for it to have correct time. Here is possible solution - https://stackoverflow.com/a/28149561/3091318, let's add it to `Str.fromDate`

> This is a bit of a problem. Some date formats that Gmail shows isn't correctly recognized by javascript's "Date" constructor that leads to inaccurate dates. Would it be good if we used the dates from Gmail as is (safest option) or should be parse them? Parsing with other language could also have some drawbacks.

I think the most stable solution will be to use date from `sentDate` (like it was before), but for print string use `toLocaleString` instead of `Str.fromDate`

martgil: Hello @sosnovsky , In my previous commit, I used "toLocaleString" inside "Str.fromDate" and now I have both issues fixed at the same time. So please let me know if you would the current over to the suggestion above?

martgil:
Hi @sosnovsky I have re-run the test but the failing one is not related to my PR. This should be good to go.

extension/js/common/core/common.ts
```diff
@@ -256,11 +256,16 @@ export class Str {
...
+  public static fromDate = (date: Date, ishour12Format = false) => {
+    const formatOptions: Intl.DateTimeFormatOptions = {
+      year: 'numeric',
+      month: '2-digit',
+      day: '2-digit',
+      hour: '2-digit',
+      minute: '2-digit',
+      hour12: ishour12Format,
+    };
+    return date.toLocaleString('en-CA', formatOptions).replace(/,/g, ''); // en-CA produces yyyy-dd-mm, hh:mm format
```
sosnovsky, line 268: What about using current user locale like `date.toLocaleString([], formatOptions)`, as for US date locale day precedes month, so it's `dd-mm` instead of `mm-dd`. By using user's locale we'll have date formatting usual for them

extension/js/common/message-renderer.ts
```diff
@@ -431,7 +431,10 @@ export class MessageRenderer {
 
   private getMessageInfo = async (fullMsg: GmailRes.GmailMsg): Promise<MessageInfo> => {
     const sentDate = GmailParser.findHeader(fullMsg, 'date');
-    const sentDateStr = sentDate ? Str.fromDate(new Date(sentDate)).replace(' ', ' at ') : '';
+    const gmailDateReceived = $('div.gK span[title]').attr('title');
+    // detects current timezone by checking if Gmail's dateTime includes trailing AM/PM otherwise its 24hour format
+    const ishour12Format = (gmailDateReceived !== undefined && gmailDateReceived.endsWith('M') ? true : false) || false;
```
sosnovsky, line 436: it can be simplified to `const isHour12Format = gmailDateReceived?.endsWith('M')`

martgil: thanks!

extension/js/common/core/common.ts
martgil: The usage of "en-CA" produces yyyy-dd-mm format with a dash separator itself compared to using `[]` or `undefined` value passed into the toLocaleString. its the most concise option in my opinion compared to using `[]` then replacing '/' with '-' not to mention that it will result to dd/mm/yyyy format instead of the preferred yyyy-dd-mm... But please let me know if you prefer the "dd/mm/yyyy" format over the original "yyyy-dd-mm" and will change it immediately.

sosnovsky: > its the most concise option in my opinion compared to using [] then replacing '/' with '-' not to mention that it will result to dd/mm/yyyy format instead of the preferred yyyy-dd-mm

`date.toLocaleString([], formatOptions)` will produce date format specified by user browser locale (so it'll be usual for user), and there is no need to replace `/` with `-`. For example, user in Germany will get `27.04.2024` date string, US user - `04/27/2024` and so on

martgil: Hello @sosnovsky, I've encountered an issue. I noticed that if I passed `[]` as the defined Locale format for "toLocaleString", the test results would differ for each of us. For example, if I make changes and use the correct static time for myself, such as in the "forward - pgp/mime signed-only" test where it passed for me, the time returned in the CI test environment would be different. I believe we need to establish a standardized date format to follow, considering the diverse formats dates can offer.

image

On CI test:
image

sosnovsky: Can we then use `toLocaleString` only for the print view and for `fromDate` we should just fix incorrect timezone by adding current user timezone offset like this:
```js
var timeZoneOffset = (new Date()).getTimezoneOffset() * 60000; //offset in milliseconds
var localTime = new Date(date.getTime() + timeZoneOffset);
```

martgil: Hello @Roma I got the same output just like the ".toISOString"... I think we have just concluded that the toISOString produces inaccurate output when its not as I check the original email with this code snippet and the code we had before (.toISOString) and produces the same time... So I think its just looks wrong since its not displayed in user locale format in print...

Could we let use the .toLocaleString only to print as we'll get different date and time outputs if we use toLocaleString across tests.

sosnovsky: > Could we let use the .toLocaleString only to print as we'll get different date and time outputs if we use toLocaleString across tests.

Sounds good, let's use it only for print view

martgil:
Hi @sosnovsky does this look good to you? I also plan to share a build coming from this PR to double-check if the user would get his desired date and time when print-previewing email -- it should work fine.

extension/js/common/message-renderer.ts
```diff
@@ -431,7 +431,21 @@ export class MessageRenderer {
...
+    const ishour12Format = (gmailDateReceived?.endsWith('M') ? true : false) || false;
+    const formatOptions: Intl.DateTimeFormatOptions = {
+      year: 'numeric',
+      month: '2-digit',
+      day: '2-digit',
+      hour: '2-digit',
+      minute: '2-digit',
+      hour12: ishour12Format,
+    };
+    const sentDateStr = sentDate ? new Date(sentDate).toLocaleString([], formatOptions).replace(', ', ' at ') : '';
```
sosnovsky, line 448: current implementation looks quite complicated, I think it'll be better to use your original version (where you parsed message date from `div.gK span[title]`) + add fallback using `.toLocaleString()`, so it'll be like this:
```ts
let messageDate = $('div.gK span[title]').attr('title');
if (!messageDate || isNaN(Date.parse(messageDate))) {
  const sentDate = GmailParser.findHeader(fullMsg, 'date');
  messageDate = sentDate ? new Date(sentDate).toLocaleString() : '';
}

Comment (martgil) [reply extension/js/common/message-renderer.ts]: Thanks @sosnovsky - could you please check if the changes now is good to you? The only way to test the toLocaleString() is by checking its epoch date version.

File: test/source/tests/decrypt.ts [hunk 6]
```diff
@@ -1158,6 +1158,7 @@ XZ8r4OC6sguP/yozWlkG+7dDxsgKQVBENeG6Lw==
           await printPage.waitForContent('@print-to', 'To: flowcrypt.compatibility@gmail.com');
           await printPage.waitForContent('@print-cc', 'ci.tests.gmail@flowcrypt.dev');
           await printPage.waitForContent('@print-content', 'Test print message');
+          expect(!isNaN(Date.parse(await printPage.readHtml('@print-date')))).to.equal(true);
```
sosnovsky, line 1161: here you just check that `print-date` contains valid date, but don't check it's value.
let's also check for correct string - parse sent date of message with id `18ecbf57e1dfb9b5` and check if it's `.toLocaleString()` value equals to `print-date` value:
```ts
const expectedPrintDateString = (new Date(1712818847000)).toLocaleString();
await printPage.waitForContent('@print-content', expectedPrintDateString);
```

martgil: Hello @sosnovsky, I've done similar with this and the reason I have updated it with the current one is that. The locale that the test environment and the locale for our local host machine are different so we've got different results. I honestly not sure how to achieve it for both it will have same result for CI and local host environment - I'm sorry.

sosnovsky: But in mentioned code we convert unix timestamp `1712818847000` to locale string on the running machine (which simulates behavior of our browser extension), so it should give correct results in both environments - locally and when running CI tests. Can you please re-try it? Thanks

test/source/tests/decrypt.ts
```diff
@@ -1151,13 +1151,16 @@ XZ8r4OC6sguP/yozWlkG+7dDxsgKQVBENeG6Lw==
...
+          const expectedPrintDateString = new Date(1712818847000).toLocaleString();
           const printPage = await browser.newPageTriggeredBy(t, () => pgpBlock.click('@action-print'));
+          await printPage.waitForContent('@print-date', expectedPrintDateString);
           await printPage.waitForContent('@print-user-email', 'First Last <flowcrypt.compatibility@gmail.com>');
           await printPage.waitForContent('@print-subject', '<b><h1>Test print dialog');
           await printPage.waitForContent('@print-from', 'From: sender@domain.com');
           await printPage.waitForContent('@print-to', 'To: flowcrypt.compatibility@gmail.com');
           await printPage.waitForContent('@print-cc', 'ci.tests.gmail@flowcrypt.dev');
           await printPage.waitForContent('@print-content', 'Test print message');
+          expect(!isNaN(Date.parse(await printPage.readHtml('@print-date')))).to.equal(true);
```
sosnovsky, line 1163: this line isn't needed anymore, as we're checking for correct date in `print-date` earlier in this test

sosnovsky (commented):
Looks good now, you can send test build to reporter for checking

test/source/tests/decrypt.ts
martgil: Agree. I have removed it away from the other test.

sosnovsky (approved):

sosnovsky:
@martgil I merged this PR, to include it in the upcoming release.

PR #5716, opened 2024-05-16
#5681 Update ossar-analysis.yml

### LINKED ISSUE #5681: Fix failing OSSAR static analysis reporting
From Roma:

> By the way, we have also some automated scanning configured in flowcrypt-browser repo, can also check opened issues at https://github.com/FlowCrypt/flowcrypt-browser/security/code-scanning and find why ESLint fails there.

Reference: https://github.com/FlowCrypt/flowcrypt-security/issues/277

martgil:
This PR updates the ossar-analysis.yml file in `.github/workflows/` by updating the github action to its most recent version.

Reference: https://github.com/FlowCrypt/flowcrypt-browser/new/master?filename=.github%2Fworkflows%2Fossar.yml&workflow_template=code-scanning%2Fossar

close #5681 

### To be filled by reviewers

### DIFF
--- a/.github/workflows/ossar-analysis.yml
+++ b/.github/workflows/ossar-analysis.yml
@@ -4,7 +4,6 @@
 name: OSSAR
 
 on:
-  push:
   pull_request:
 
 jobs:
--- a/.github/workflows/ossar-analysis.yml
+++ b/.github/workflows/ossar-analysis.yml
@@ -27,7 +27,7 @@ jobs:
     - run: git checkout HEAD^2
       if: ${{ github.event_name == 'pull_request' }}
 
-    - run: git checkout {{ github.sha }}
+    - run: git checkout ${{ github.sha }}
       if: ${{ github.event_name == 'push' }}
     # Ensure a compatible version of dotnet is installed.
     # The [Microsoft Security Code Analysis CLI](https://aka.ms/mscadocs) is built with dotnet v3.1.201.
--- a/.github/workflows/ossar-analysis.yml
+++ b/.github/workflows/ossar-analysis.yml
@@ -1,3 +1,8 @@
+# This workflow uses actions that are not certified by GitHub.
+# They are provided by a third-party and are governed by
+# separate terms of service, privacy policy, and support
+# documentation.
+
 # This workflow integrates a collection of open source static analysis tools
 # with GitHub code scanning. For documentation, or to provide feedback, visit
 # https://github.com/github/ossar-action
@@ -7,14 +12,21 @@ on:
   push:
   pull_request:
 
+
+permissions:
+  contents: read
+
 jobs:
   OSSAR-Scan:
     # OSSAR runs on windows-latest.
     # ubuntu-latest and macos-latest support coming soon
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+      actions: read # only required for a private repository by github/codeql-action/upload-sarif to get the Action run status
     runs-on: windows-latest
 
     steps:
-      # Checkout your code repository to scan
     - name: Checkout repository
       uses: actions/checkout@v4
       with:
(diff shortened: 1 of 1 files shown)


martgil:
Hello @sosnovsky, this is now ready for review. It's difficult to tackle a place without a map but I'm glad I have found the right root-cause.

You can double-check and ensure that this PR fixes the issue by noticing the time it takes for the OSSAR's push vs pull_request and it both finishes in 2 minutes mark.

image

Compare to other PR's OSSAR test results:

image

I believe its also better to limit the push checks to master for consistency. As for example, I do not see any codeQL security checks that runs on a pull request for both pull_request and push events at the time.

martgil:
Ready for review. thank you!

.github/workflows/ossar-analysis.yml
```diff
@@ -1,20 +1,33 @@
...
+
 # This workflow integrates a collection of open source static analysis tools
 # with GitHub code scanning. For documentation, or to provide feedback, visit
 # https://github.com/github/ossar-action
 name: OSSAR
 
 on:
   push:
+    branches: [ "master" ]
   pull_request:
```
sosnovsky, line 14: let's use parameters from example workflow at https://github.com/actions/starter-workflows/blob/main/code-scanning/ossar.yml:
```yml
on:
  push:
    branches: [ $default-branch, $protected-branches ]
  pull_request:
    # The branches below must be a subset of the branches above
    branches: [ $default-branch ]
```

.github/workflows/ossar-analysis.yml
```diff
@@ -44,6 +56,6 @@ jobs:
 
       # Upload results to the Security tab
```
sosnovsky, line 57: here we still use `ossar-action` v1, while your fix is included in v2.
`ossar-action` config should be updated to:
```yml
- name: Run OSSAR
  uses: github/ossar-action@v2.0.0
  id: ossar
```

.github/workflows/ossar-analysis.yml
```diff
@@ -44,6 +56,6 @@ jobs:
...
     - name: Upload OSSAR results
-      uses: github/codeql-action/upload-sarif@v3
+      uses: github/codeql-action/upload-sarif@v2
```
sosnovsky, line 59: why we started using older action version here?

.github/workflows/ossar-analysis.yml
martgil: Thank you - I have updated the GitHub action to use the given variables.

.github/workflows/ossar-analysis.yml
martgil: Updated thanks - I apologize, I came late to realize about the versioning and so for the upload sarif's version. I can see just now that about the release's updated version.

.github/workflows/ossar-analysis.yml
martgil: I thought it would have to use the version 2 to work in Security Tab. I checked the CodeQL GitHub repo and they mentioned about the V3 but I didn't realize that the version could have applies to its child GitHub actions - https://github.com/github/codeql-action. I appreciate your help in leading me on the right track.

.github/workflows/ossar-analysis.yml
```diff
@@ -1,20 +1,33 @@
...
 
 jobs:
   OSSAR-Scan:
     # OSSAR runs on windows-latest.
     # ubuntu-latest and macos-latest support coming soon
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+      actions: read # only required for a private repository by github/codeql-action/upload-sarif to get the Action run status
     runs-on: windows-latest
```
sosnovsky, line 28: I also noticed that `ossar-action` repository has sample file for `ubuntu-latest` - https://github.com/github/ossar-action/blob/main/.github/workflows/sample-workflow-ubuntu-latest.yml

I tried to replace `window-latest` with `ubuntu-latest` in our config, and it runs much faster - [2m 11s](https://github.com/FlowCrypt/flowcrypt-browser/actions/runs/9174491542/job/25225465955) on `windows` vs [23s](https://github.com/FlowCrypt/flowcrypt-browser/actions/runs/9174852536/job/25226581446) on `ubuntu`.

So let's switch to `ubuntu-latest` here

martgil: Thanks Roma - I will check that and switch to the `ubuntu-latest` as the target machine.

martgil:
Hello @sosnovsky - I have update this and the OSSAR checks is much finished faster.

.github/workflows/ossar-analysis.yml
martgil: I think I just need to update this as it's not correct to use the "$default-branch" value as they where just a placeholders for example if someone started to create OSSAR through github and this doesn't seem to apply on workflow files made manually.

sosnovsky (approved):
All good now, thanks!

PR #5717, opened 2024-05-17
#5715 Fix id token fetch for exception reports

### LINKED ISSUE #5715: Error reports send unauthorized requests from content script
When sending error report from content script, `idToken` is `undefined` which causes `401 Unauthorized` error on FES.

It happens because in [doSendErrorToSharedTenantFes](https://github.com/FlowCrypt/flowcrypt-browser/blob/cfe8276dac2189f3a8167e7db9e060114d63fbc0/extension/js/common/platform/catch.ts#L312) we try to parse `acctEmail` from current url, but there is no such url parameter in content script, need to fetch current `acctEmail` from storage.

sosnovsky:
This PR fixes id token fetch for sending error reports from content script.

\+ replaces `{ line: 0, col: 0 }` with `{ line: 1, col: 1}` for reports with unknown error location (as backend expects positive number, and throws error when line or col is 0)

close #5715

- Tests will be added later (https://github.com/FlowCrypt/flowcrypt-browser/issues/5718)

### To be filled by reviewers

### DIFF
--- a/extension/js/common/platform/catch.ts
+++ b/extension/js/common/platform/catch.ts
@@ -312,13 +312,8 @@ export class Catch {
 
   private static async doSendErrorToSharedTenantFes(errorReport: ErrorReport) {
     try {
-      let acctEmail = Url.parse(['acctEmail']).acctEmail;
-      if (acctEmail) {
-        acctEmail = String(acctEmail);
-      } else {
-        const acctEmails = await GlobalStore.acctEmailsGet();
-        acctEmail = acctEmails?.length ? acctEmails[0] : undefined;
-      }
+      const { acctEmail: parsedEmail } = Url.parse(['acctEmail']);
+      const acctEmail = parsedEmail ? String(parsedEmail) : (await GlobalStore.acctEmailsGet())?.[0];
       if (!acctEmail) {
         console.error('Not reporting error because user is not logged in');
         return;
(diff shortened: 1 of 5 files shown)


extension/js/common/platform/catch.ts
```diff
@@ -311,8 +312,13 @@ export class Catch {
...
     try {
-      const uncheckedUrlParams = Url.parse(['acctEmail']);
-      const acctEmail = String(uncheckedUrlParams.acctEmail);
+      let acctEmail = Url.parse(['acctEmail']).acctEmail;
+      if (acctEmail) {
+        acctEmail = String(acctEmail);
+      } else {
+        const acctEmails = await GlobalStore.acctEmailsGet();
+        acctEmail = acctEmails?.length ? acctEmails[0] : undefined;
+      }
```
ioanmo226, line 321: How about write like below?

```ts
const { acctEmail: parsedEmail } = Url.parse(['acctEmail']);
let acctEmail = parsedEmail 
    ? String(parsedEmail) 
    : (await GlobalStore.acctEmailsGet())?.[0];

```

ioanmo226: I think that is why we added safe casting (?) before accessing first element?

> ?.[0]

sosnovsky: Yeah, you're right - I thought `?` checks only for `undefined` or `null` values (similar to Swift behavior), but it works for empty arrays too. I updated code according to your suggestion, thanks!

ioanmo226 (approved):
👍

Issue #4954, opened 2023-02-13
Add ui test for id tokens refresh


sosnovsky:
original issue - https://github.com/FlowCrypt/flowcrypt-browser/issues/4844

we should try to reproduce it with our mock apis:
- create id token which expires in 1-2 seconds
- wait 2 seconds
- reload gmail page
- check if no warning shown (which means id token was refreshed successfully)

PR #5731, opened 2024-05-23
#5691 Refresh id token if it expires

### LINKED ISSUE #5691: Check Google id token expiration before usage
Sometimes in browser extension console logs I see error:

```
webmail_bundle.js:49 BrowserMsg(ajax) Unauthorized: 401 when GET-ing https://..../keys/private (no body):  -> No matching authenticator found for Authorization starting with Bearer eyJhbGciOiJSUzI1NiIsImt...


[callerStack]
    at BrowserMsg.sendRaw (chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:10292:26)
    at BrowserMsg.sendAwait (chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:10280:37)
    at Object.ajax (chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:10435:38)
    at Api.ajax (chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:7334:55)
    at Api.apiCall (chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:7642:34)
    at KeyManager.getPrivateKeys (chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:7879:34)
    at chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:16058:78
    at chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/content_scripts/webmail_bundle.js:93:27
```

I checked backend logs and found reason for mentioned error

```
Client error: 401 Unauthorized GET http://[127.0.0.1]/v1/keys/private | id: a1bc06e10319 | message: authentication:
The JWT is no longer valid - the evaluation time NumericDate{1714160438 -> Apr 26, 2024, 7:40:38 PM UTC} is on or after the Expiration Time (exp=NumericDate{1714159800 -> Apr 26, 2024, 7:30:00 PM UTC}) claim value (even when providing 30 seconds of leeway to account for clock skew).
```

Currently we don't have check for id token expiration time, so let's check if id token has expired before using it and expired tokens should be refreshed before using.

ioanmo226:
This PR implemented logic to refresh id token if it expires

close #5691 // if this PR closes an issue

### To be filled by reviewers


extension/js/common/api/shared/api.ts
```diff
@@ -412,7 +412,24 @@ export class Api {
...
     } else {
-      return await Api.ajax(req, resFmt);
+      try {
+        return await Api.ajax(req, resFmt);
+      } catch (firstAttemptErr) {
+        const idToken = headers?.authorization?.split('Bearer ')?.[1];
+        if (ApiErr.isAuthErr(firstAttemptErr) && idToken) {
+          // Needed authorization from the service worker side to avoid circular dependency injection errors
+          // that occur when importing GoogleAuth directly.
+          const authorization = await BrowserMsg.send.bg.await.getGoogleApiAuthorization({ idToken });
```
sosnovsky, line 422: maybe it's possible to fix circular dependency without using service worker?
as it's better to keep service worker lightweight, and use it only for tasks which are difficult or not possible to be implemented in the content script.

sosnovsky (commented):
let's also add test for this situation when id token is expired - I think we already have mock function for generating id token. so test should be like this:
- generate id token with 5 seconds expiration
- wait for 5 seconds
- try to perform request to external service
- check if id token was refreshed and request succeeded

extension/js/common/api/shared/api.ts
ioanmo226: I attempted to resolve the circular dependency issue by creating a base service, such as `HttpService`, which contains only the essential common functions. I then had the `GoogleOAuth` and `Api` classes extend this base service. To circumvent the circular dependency, I used the `GoogleOAuth.googleApiAuthHeader` function in a `catch` block, thinking this separation would resolve the issue since the `GoogleOAuth` and `Api` classes wouldn't be directly related. However, the circular dependency persisted because the `Api` class still utilizes the `GoogleOAuth` class, and `GoogleOAuth` imports the `AccountServer` class, which, in turn, extends the `Api` class.

In my opinion, solving this circular dependency issue would require significant changes, which I find a bit overwhelming and am uncertain if the effort would be justified.

ioanmo226: Let me know your thoughts

sosnovsky: How can I reproduce such circular dependency error?
I tried to build extension from your commit before circular dependency fix, builds passed without any errors and extension seems to work well.

sosnovsky: Was able to reproduce by running failed tests, will check it

sosnovsky (approved):
Yeah, seems like there is no easy way to make it work for all api calls without using service worker. So let's try this solution.
There is an existing issue for adding ui test - https://github.com/FlowCrypt/flowcrypt-browser/issues/4954

PR #5732, opened 2024-05-23
#5730 Fix error when sending blank signed message

### LINKED ISSUE #5730: Error when trying to send empty signed-only message
> I noticed just now that when trying to send a blank signed message, there was this unhandled error message that occurred.

image

_Originally posted by @martgil in https://github.com/FlowCrypt/flowcrypt-browser/issues/5599#issuecomment-1940538245_

martgil:
This PR improves the case where sending an empty signed message using FlowCrypt returns unhandled error. This is due to `createCleartextMessage()` not accepting empty message. So passing a newline that would serve as empty message would satify the condition when user choose to send blank signed email when asked by the FlowCrypt browser extension. 

close #5730

### To be filled by reviewers

### DIFF
--- a/extension/js/common/core/crypto/pgp/openpgp-key.ts
+++ b/extension/js/common/core/crypto/pgp/openpgp-key.ts
@@ -268,6 +268,7 @@ export class OpenPGPKey {
       const message = await opgp.createMessage({ text });
       return await opgp.sign({ message, format: 'armored', signingKeys: [signingPrv], detached });
     }
+    text = text ? text : '\n';
     const message = await opgp.createCleartextMessage({ text });
     return await opgp.sign({ message, signingKeys: [signingPrv] });
   }


martgil:
Hi @sosnovsky, I think this particular one does not need to test for such a small changes. Would you agree on that or should I make another test for sending blank signed message?

extension/js/common/core/crypto/pgp/openpgp-key.ts
```diff
@@ -268,6 +268,7 @@ export class OpenPGPKey {
       const message = await opgp.createMessage({ text });
       return await opgp.sign({ message, format: 'armored', signingKeys: [signingPrv], detached });
     }
+    text = text ? text : '\n';
```
sosnovsky, line 271: Original issue happens because empty string ` ` is equal to `false` in js.
But if we'll create empty string as `String(text)` then it'll be `true`.
So let's just convert original `text` parameter to string (as `String(data)`) at https://github.com/FlowCrypt/flowcrypt-browser/blob/75621567b2ef423461d5363776e556921ff4c86a/extension/js/common/core/crypto/pgp/msg-util.ts#L162

martgil: Hello @sosnovsky -  That would work too because if `text` is still doesn't hold any string value -- it will still failed to pass the truthy value test check - https://github.com/openpgpjs/openpgpjs/blob/026b348cf87988cf95896dd9c5dd1a0c8129ddb1/src/cleartext.js#L210.

So I think a way to solve this is to put something "invisible" in replacement when the text is completely blank to pass the checks.

sosnovsky: Yeah, you're right - I initially tested `new String(text)` with raw openpgp.js, but it doesn't work when used in FlowCrypt code. So let's use your solution ✅

sosnovsky (approved):
👍

extension/js/common/core/crypto/pgp/openpgp-key.ts
martgil: Thanks - Roma, I understand your side now. The way you test it out is also my go to too --- just poping my test enviroment with raw openpgp.js. Thanks for considering the suggested solution.

PR #5744, opened 2024-05-28
#5741 Refactor IGNORE_ERR_MSG for pattern matching on welcome-message endpoint

### LINKED ISSUE #5741: Don't report "400 when POST-ing https://flowcrypt.com/attester/welcome-message" error
We already have code for skipping this error - https://github.com/FlowCrypt/flowcrypt-browser/pull/5284, but these errors are reported again after manifest v3:
```
Bad Request: 400 when POST-ing https://flowcrypt.com/attester/welcome-message string: email,pubkey -> This key does not appear valid(Error: Could not find valid self-signature in key b4a63b803788e67e)
```

Maybe error text has changed and it's different from value in `IGNORE_ERR_MSG` array.

martgil:
This PR refactors the `IGNORE_ERR_MSG` pattern for the welcome-message endpoint to prevent reporting errors.

close #5741 

### To be filled by reviewers


martgil:
@sosnovsky Ready for review. Thank you!

martgil:
Hi @sosnovsky

Out of topic - I think we should reconfigure the "Security" tab settings for this repo to remove the OSSAR GitHub action and some others with no configurations found: `Code scanning results / ESLint Completed in 2s — 1 configuration not found`.

sosnovsky:
> Hi @sosnovsky
> 
> Out of topic - I think we should reconfigure the "Security" tab settings for this repo to remove the OSSAR GitHub action and some others with no configurations found: `Code scanning results / ESLint Completed in 2s — 1 configuration not found`.

Right, thanks for noticing! I removed OSSAR from `Security` tab

extension/js/common/platform/catch.ts
```diff
@@ -44,7 +44,7 @@ export class Catch {
     // benign error https://github.com/WICG/ResizeObserver/issues/38#issuecomment-422126006 https://stackoverflow.com/questions/49384120/resizeobserver-loop-limit-exceeded
     'ResizeObserver loop limit exceeded',
     // https://github.com/FlowCrypt/flowcrypt-browser/issues/5280
-    '400 when POST-ing https://flowcrypt.com/attester/welcome-message string: email,pubkey -> This key does not appear valid',
+    '400 when POST-ing https://flowcrypt.com/attester/welcome-message string: email,pubkey',
```
sosnovsky, line 47: This change won't fix the issue, as in `catch.ts` we have check https://github.com/FlowCrypt/flowcrypt-browser/blob/4252b8b706bf6b857be1ad270d1efdca8ef1ac50/extension/js/common/platform/catch.ts#L86

But original error message is `Bad Request: 400 when POST-ing https://flowcrypt.com/attester/welcome-message string: email,pubkey -> This key does not appear valid(Error: Could not find valid self-signature in key b4a63b803788e67e)` which won't be equal to message from `IGNORE_ERR_MSG`

I think there were some updates which caused error message to be changed from original `400 when POST-ing https://flowcrypt.com/attester/welcome-message string: email,pubkey -> This key does not appear valid` to the current longer version. Need to find this change and fix it to have the same error message as before

sosnovsky:
Hi @martgil, can you please work on finishing this PR before working on other tasks, so it can be included in the next browser extension release (in the end of this week, or in the beginning of the next one). Thanks!

extension/js/common/platform/catch.ts
martgil: Thanks @sosnovsky I understand what you mean by now - i've been struggling to find out the exact place. What i have observed is that, the request status codes where prepended to error messsage, so format have been changed with status code, for example "Bad Request: 400 when POST-ing..." is now the new format +  the details are getting appended at the end...

I'll do my best to fix it in timely manner.

sosnovsky: Probably we can just remove details in the end of message, so `(Error: Could not find valid self-signature in key b4a63b803788e67e)` will be removed and full message will look like `Bad Request: 400 when POST-ing https://flowcrypt.com/attester/welcome-message string: email,pubkey -> This key does not appear valid`, is it easy to do?

martgil: Hello @sosnovsky - upon checking, the details in the error message is part of the attester-ts response - https://github.com/FlowCrypt/flowcrypt-attester-ts/blob/f7f37fef465b93807406934af965cf1895138b89/source/modules/api-public/common/abstract-pub-handler.ts#L162.

So basically, it was all part of the resMsg - https://github.com/FlowCrypt/flowcrypt-browser/blob/9bb6396ed57f316760fa859660bbcbab67159033/extension/js/common/api/shared/api-error.ts#L101.

Your suggestion is feasible to trim down that part. But also, I think it will also be an option to just adjust our checking for which errors should not be reported based on their exception message. In this case, I have adjusted the checks so it enables us to ignore some error based on the error message excerpt and not require us to match the whole entire error message.

One good reason for this is that, only that error message in attester-ts has that error message pattern where it includes the details. where it is helpful in any other case other than the error reporting.

sosnovsky: Sounds good, nice suggestion, should work well 👍

sosnovsky (approved):
👍

PR #5750, opened 2024-05-31
#5736 Group similar error reports

### LINKED ISSUE #5736: Group similar error reports
There are many error reports which are sent in [doSendErrorToSharedTenantFes](https://github.com/FlowCrypt/flowcrypt-browser/blob/d4c6c45fadccc244e3de726427533c3c4dec6345/extension/js/common/platform/catch.ts#L313) with similar message like `429 when GET-ing https://www.googleapis.com/gmail/v1/users/me/threads/18fa6109b0f0d121 (no body)`, which has the only difference - thread id.

For such cases we should just replace such parameters with `THREAD_ID`, so similar reports will be grouped on our backend.

ioanmo226:
This PR implemented feature to group similar error reports

close #5736 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/js/common/platform/catch.ts
+++ b/extension/js/common/platform/catch.ts
@@ -82,6 +82,7 @@ export class Catch {
     originalErr: unknown,
     isManuallyCalled: boolean
   ): boolean {
+    url = url.replace(/chrome-extension:\/\/[^\/]+\//, 'chrome-extension://EXTENSION_ID/');
     const exception = Catch.formExceptionFromThrown(originalErr, errMsg, url, line, col, isManuallyCalled);
     if (Catch.IGNORE_ERR_MSG.indexOf(exception.message) !== -1 || (errMsg && Catch.IGNORE_ERR_MSG.indexOf(errMsg) !== -1)) {
       return false;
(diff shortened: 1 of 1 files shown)


ioanmo226:
@sosnovsky Please check

extension/js/common/platform/catch.ts
```diff
@@ -281,6 +281,14 @@ export class Catch {
     }
   }
 
+  private static groupSimilarReports(value: string): string {
+    return value
+      .replace(/chrome-extension:\/\/[^\/]+\//, 'chrome-extension://EXTENSION_ID/')
+      .replace(/https:\/\/www\.googleapis\.com\/gmail\/v1\/users\/me\/threads\/[^\/]+\//, 'https://www.googleapis.com/gmail/v1/users/me/threads/THREAD_ID/')
+      .replace(/https:\/\/www\.googleapis\.com\/gmail\/v1\/users\/me\/messages\/[^\/]+\//, 'https://www.googleapis.com/gmail/v1/users/me/messages/MESSAGE_ID/')
+      .replace(/https:\/\/www\.googleapis\.com\/gmail\/v1\/users\/me\/drafts\/[^\/]+\//, 'https://www.googleapis.com/gmail/v1/users/me/drafts/DRAFT_ID/');
+  }
```
sosnovsky, line 290: currently only `EXTENSION_ID` replacement works, in other regexes trailing slashes should be removed - as urls look like https://www.googleapis.com/gmail/v1/users/me/messages/12079801d03027dc

ioanmo226: Ohhh. misssed it.
Just fixed it

sosnovsky (approved):
Works well now, thanks for quick fix!

Issue #5713, opened 2024-05-14
Update to ESLint v9


sosnovsky:
Migration guide - https://eslint.org/docs/latest/use/migrate-to-9.0.0

sosnovsky:
some tips on migrating to v9 https://x.com/karlhorky/status/1792500811307622631

it also mentions that it's easier to migrate to ESLint flat config before starting migration to v9, I filed separate issue for config migration - https://github.com/FlowCrypt/flowcrypt-browser/issues/5728

ioanmo226:
@sosnovsky When upgrading to ESLint v9, I encountered two major issues with packages:

1. **[eslint-plugin-import](https://github.com/import-js/eslint-plugin-import)**: This package doesn't yet support ESLint v9 and flat config. The solutions could be:
   - Wait for their official release.
   - Submit a PR (though this takes time).
   - Remove the `import/order` rule temporarily, which seems to be the simplest solution until their official release.
   - https://github.com/FlowCrypt/flowcrypt-browser/blob/2b2c2599c9258457aa2ac8da7fff8ce46cb8ed9f/eslint.config.mjs#L122

2. **[typescript-eslint](https://github.com/typescript-eslint/typescript-eslint)**: This package doesn't support ESLint v9. However, there is an alpha version (latest is ^8.0.0-alpha.30) that supports ESLint v9. We might consider using this version even though it's not yet stable.
   - [Official ESLint v9 support notice](https://github.com/typescript-eslint/typescript-eslint/issues/8211)

ioanmo226:
Let me know your thoughts

sosnovsky:
Oh, I thought `typescript-eslint` added support for eslint v9 in their current release, then let's wait before `typescript-eslint` update

ioanmo226:
You mean postpone our ESLint v9 update until `typescript-eslint` stable release and `eslint-plugin-import` release for ESLint v9 support?

sosnovsky:
Yes, let's wait until support in stable typescript-eslint version. Maybe import plugin will add support too by that time, but it's not so critical.

sosnovsky:
By the way, it looks like currently we don't use any rules from `eslint-plugin-import` - eslint config mentions only `'import/order'` rule, but it's `off`. I haven't found any other rules from this package in our config, so it should be ok to remove it.

ioanmo226:
Sounds good!

ioanmo226:
[...]

PR #5756, opened 2024-06-03
5728 switch to eslint flat config

### LINKED ISSUE #5728: Switch to ESLint flat config
Before migrating to ESLint v9 we should start using flat config:
https://eslint.org/docs/latest/use/configure/migration-guide

ioanmo226:
This PR switches eslint to flat config

close #5728 // if this PR closes an issue

### To be filled by reviewers


.vscode/settings.json
```diff
@@ -1 +1,3 @@
-{}
+{
+  "eslint.experimental.useFlatConfig": true
```
sosnovsky, line 2: there is a pre-release version of `vscode-eslint` which supports flat config - https://github.com/microsoft/vscode-eslint?tab=readme-ov-file#version-305---pre-release. it can be enabled in `vscode-eslint` extension settings in vs code

extension/chrome/elements/pgp_block_modules/pgp-block-attachments-module.ts
```diff
@@ -15,7 +15,7 @@ import { Str } from '../../../js/common/core/common.js';
 import { AttachmentWarnings } from '../shared/attachment_warnings.js';
 import { MsgUtil } from '../../../js/common/core/crypto/pgp/msg-util.js';
 
-declare const filesize: { filesize: Function }; // eslint-disable-line @typescript-eslint/ban-types
+declare const filesize: { filesize: (size: number) => number }; // eslint-disable-line @typescript-eslint/ban-types
```
sosnovsky, line 18: `eslint-disable-line` not needed now

extension/chrome/elements/compose-modules/compose-render-module.ts
```diff
@@ -485,6 +495,7 @@ export class ComposeRenderModule extends ViewModule<ComposeView> {
 
   private renderReplySuccessMimeAttachments = (attachmentsFilenames: string[]) => {
     const attachments = $('<div id="attachments"></div>');
+    // eslint-disable-next-line @typescript-eslint/no-for-in-array
     for (const index in attachmentsFilenames) {
       if (attachmentsFilenames.hasOwnProperty(index)) {
```
sosnovsky, line 500: probably we can just replace `for in` with `for (const [index, filename] of attachmentsFilenames.entries()) {`

sosnovsky (commented):
Well done 👍

extension/chrome/elements/compose-modules/compose-render-module.ts
ioanmo226: Yeah, just missed it.

ioanmo226:
Thank you for your comments. Fixed reviews and pushed. Please check

sosnovsky (approved):
All good now, thanks 👍
Next part will be migrating to eslint v9 - https://github.com/FlowCrypt/flowcrypt-browser/issues/5713, it'll probably require less changes after switching to flat config.

PR #5759, opened 2024-06-06
#5666 Add compose button on Thunderbird

### LINKED ISSUE #5666: [thunderbird] Add compose button into Thunderbird UI

martgil:
This PR adds a functional compose button on top of Thunderbird mail client.

close #5666

- Difficult to test - (changes are not readily completed for tests in Thunderbird Mail environment)

### To be filled by reviewers


martgil:
Hi @sosnovsky - I hope you are well.

I would like to hear your suggestion with my progress and approach on how I enable secure compose on Thunderbird mail. What i did is that, I re-use the pop-up we have and injected them on messageDisplay and composeDisplay as a popup, that when clicked, will redirect then to the web extension page where they can directly compose their secure email.

I come up with this idea because, an extension can only inject a single option_ui so by having a way to  compose email (encrypted, signed or both) its so flexible to so in MailExtension view. This is pretty much it for this PR.

To test it out:

1. Install and it up the Thunderbird extension build on a Thunderbird email client.
2. Open any message and click on "FlowCrypt" injected option ui -> choose secure compose -> choose which inbox you want to use.
3. Upon clicking the account, you'll be greeter with secure compose - which would let you send secure email (encrypted, signed or combination of both).

I look forward to your feedback.

Thanks!

package.json
```diff
@@ -45,6 +45,7 @@
   },
   "dependencies": {
     "@flowcrypt/fine-uploader": "5.16.4",
+    "@types/thunderbird-webext-browser": "^109.0.4",
```
sosnovsky, line 48: seems like these types are not used anywhere, dependency can be removed

extension/js/common/browser/browser.ts
```diff
@@ -53,21 +57,28 @@ export class Browser {
...
   };
 
-  public static openSettingsPage = async (path = 'index.htm', acctEmail?: string, page = '', rawPageUrlParams?: Dict<UrlParam>, addNewAcct = false) => {
+  public static openSettingsPage = async (
+    path = 'index.htm',
+    acctEmail?: string,
+    page = '',
+    rawPageUrlParams?: Dict<UrlParam>,
+    addNewAcct = false,
+    currentPageParams?: Dict<UrlParam>
```
sosnovsky, line 66: maybe we can use `rawPageUrlParams` for `useFullScreenCompose` parameter?

extension/chrome/popups/select_account.ts
```diff
@@ -56,8 +56,9 @@ View.run(
     };
 
     private actionChooseAcctHandler = async (clickedElement: HTMLElement) => {
+      const useFullScreenSecureCompose = Browser.isThunderbirdMail() ? true : false;
```
sosnovsky, line 59: it can be simplified to `const useFullScreenSecureCompose = Browser.isThunderbirdMail()`

sosnovsky (commented):
I think we should have different actions for different buttons:

- FlowCrypt button in toolbar should open `Encrypted inbox` (also let's replace it's label text with just `FlowCrypt`, so it will use less space in the toolbar):
Screenshot 2024-06-07 at 21 20 59

- FlowCrypt button in compose window should open `Secure compose`
Screenshot 2024-06-07 at 21 23 14

extension/chrome/popups/select_account.ts
martgil: will do - thanks!

martgil:
Hi @sosnovsky - Just a quick question for "FlowCrypt button in compose window should open Secure compose". Do you mean by this is that, upon clicking the FlowCrypt button on the compose display, it should directly go to the first account's inbox without the ability to choose for which inbox the user to use? Or would mean only the text. 

For example, if the placement of the FlowCrypt button was on the toolbar - it should say "Encrypted inbox", then when accessed from the compose display should say "Secure compose" but without showing the setting button?

sosnovsky:
> Hi @sosnovsky - Just a quick question for "FlowCrypt button in compose window should open Secure compose". Do you mean by this is that, upon clicking the FlowCrypt button on the compose display, it should directly go to the first account's inbox without the ability to choose for which inbox the user to use? Or would mean only the text.

Here we can use the same behavior as the current one - show list of accounts and then open `encrypted compose` for chosen account. As it's probably quite complicated to get current account from Thunderbird UI. Also for simplification we can automatically open `encrypted compose` if it's only one account is configured.

> For example, if the placement of the FlowCrypt button was on the toolbar - it should say "Encrypted inbox", then when accessed from the compose display should say "Secure compose" but without showing the setting button?

I like this idea - it'll make more clear what will be opened after button click.

martgil:
> Also for simplification we can automatically open encrypted compose if it's only one account is configured.

I agree on everything else - sounds like a plan. I'm on it now.

package.json
martgil: Ah yes - I pre-added this so that wouldn't miss them later one. This is pretty much useful when I started calling the `browser` object known by the Thunderbird API.

I'll check and remove it when unused. thanks!

extension/js/common/browser/browser.ts
martgil: Thank you - I have re-used the `rawPageUrlParams` parameter. I initially though it was only for iframe parameters but it was all clear to me know.

martgil:
This PR is almost done.

> For example, if the placement of the FlowCrypt button was on the toolbar - it should say "Encrypted inbox", then when accessed from the compose display should say "Secure compose" but without showing the setting button?

>> I like this idea - it'll make more clear what will be opened after button click.

Just need to show "Secure compose" instead of "Encrypted Inbox" when accessing in an composeAction Button in thunder bird and hide the settings page.

martgil:
Hi @sosnovsky - I believe this one is ready for a review. Thanks!

To test this PR manually, there would be two cases: 1.) using secure compose with single Gmail account 2.) using secure compose with two user accounts from which one account will be selected out of the accounts.

tooling/build-types-and-manifests.ts
```diff
@@ -53,20 +53,13 @@ addManifest(
...
   manifest => {
     manifest.browser_specific_settings.strict_min_version = '102.0';
-    manifest.browser_action.default_title = 'FlowCrypt Encryption for Thunderbird';
+    manifest.browser_action.default_title = 'FlowCrypt';
     manifest.name = 'FlowCrypt Encryption for Thunderbird';
-    manifest.description = 'Secure end-to-end encryption with FlowCrypt'; // needs to updated later
+    manifest.description = 'Simple end-to-end encryption to secure email and attachments on Thunderbird';
+    manifest.permissions = [...manifest.permissions, 'compose'];
     manifest.compose_action = {
       default_title: 'FlowCrypt', // eslint-disable-line @typescript-eslint/naming-convention
```
sosnovsky, line 61: let's call it `Secure compose`, so it'll be the same as in the browser extension

extension/js/common/inject.ts
```diff
@@ -62,12 +62,39 @@ export class Injector {
       .append(this.factory.metaStylesheet('webmail') + this.factory.metaNotificationContainer()); // xss-safe-factory
   };
 
-  public openComposeWin = (draftId?: string): boolean => {
+  public openComposeWin = (
+    draftId?: string,
+    fullscreen = false,
+    composeMsgDetails?: { subject?: string; recipients?: { to?: string[]; cc?: string[]; bcc?: string[] } } | undefined
```
sosnovsky, line 68: let's create a separate type for it, you can check https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/chrome/elements/compose-modules/compose-types.ts, there are similar types there

extension/js/common/browser/browser.ts
```diff
@@ -53,21 +57,28 @@ export class Browser {
...
     const pageUrlParams = rawPageUrlParams ? JSON.stringify(rawPageUrlParams) : undefined;
     if (acctEmail || path === 'fatal.htm') {
-      await Browser.openExtensionTab(Url.create(basePath, { acctEmail, page, pageUrlParams }));
+      await Browser.openExtensionTab(Url.create(basePath, { acctEmail, page, pageUrlParams, ...currentPageParams }));
     } else if (addNewAcct) {
       await Browser.openExtensionTab(Url.create(basePath, { addNewAcct }));
     } else {
       const acctEmails = await GlobalStore.acctEmailsGet();
-      await Browser.openExtensionTab(Url.create(basePath, { acctEmail: acctEmails[0], page, pageUrlParams }));
+      await Browser.openExtensionTab(Url.create(basePath, { acctEmail: acctEmails[0], page, pageUrlParams, ...currentPageParams }));
```
sosnovsky, line 76: probably `...currentPageParams` can be removed, as unused now?

extension/js/common/browser/browser.ts
```diff
@@ -16,6 +16,10 @@ export class Browser {
     return URL.createObjectURL(new Blob([content], { type: 'application/octet-stream' }));
   };
 
+  public static isThunderbirdMail = () => {
+    return Catch.browser().name === 'thunderbird';
+  };
```
sosnovsky, line 21: let's move it to https://github.com/FlowCrypt/flowcrypt-browser/blob/master/extension/js/common/platform/catch.ts, we have similar `isFirefox` method there

martgil: Updated - thank you!

extension/js/common/browser/browser.ts
martgil: Yes, I will remove it. thanks!

extension/js/common/inject.ts
martgil: Thanks for the hint - I have created a type definition for that in the appropriate place it should be. Thank you!

extension/chrome/elements/compose-modules/compose-recipients-module.ts
```diff
@@ -144,10 +144,15 @@ export class ComposeRecipientsModule extends ViewModule<ComposeView> {
   public validateEmails = (uncheckedEmails: string[]): { valid: EmailParts[]; invalid: string[] } => {
     const valid: EmailParts[] = [];
     const invalid: string[] = [];
+    const uniqueEmails = new Set<string>();
```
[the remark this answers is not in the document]
martgil: Thanks, Roma - Upon further investigation, It wouldn't be possible to have duplicate emails, since Gmail itself prevents that.

I dig deeper into the code to answer my question why I got duplicate emails is that a draft were being made right before the recipients where added programmatically which produces such issue - so there is also that.

Later on, I have fixed the issue by moving the `startDraftTimer()` to which I believe where it should belong and that is on the `onRecipientAdded()` which requires web UI interaction, that way, I wouldn't get a blank draft with duplicate recipients when in fact opening a secure compose for the first time.

martgil:
Hi @sosnovsky I still can't commit and push changes. My changes and commit message are not empty - I'm not so sure why `lint-staged` sees it as empty.

image

sosnovsky:
Hi @martgil, probably here is explanation of your situation - https://stackoverflow.com/questions/71420124/how-do-i-solve-this-empty-git-commit-warning/77799678#77799678

Your commit includes only reverted `prettier` changes, but `lint-staged` automatically runs `prettier` and re-applies reverted changes. As a result - your commit becomes empty.

martgil:
Thanks @sosnovsky for the help. I'm now able to push new changes again :)

martgil:
Hi @sosnovsky - this one is ready for review.

To check the rendered quoted text from compose display, simply open any emails from Thunderbird -> click on reply button -> click on secure compose and there it will autoload the previous email.

One thing to note is that, email body cannot be obtained simply on a messageDisplay view, its the one that is first shown where the second secure compose is injected. I think it was on purpose so the rendering of quoted email is very limited when accessing the secure compose on composeDisplay view. Same thing to test, it works for single and multi account.

martgil:
Hi @sosnovsky - Its ready for review. Thank you!

At this point, you may check the following:

Quoted email rendering if you wish to reply from a compose display. Eg. You opened an email  from Thunderbird -> click reply -> open Secure Compose automatically if single account or Choose account depending on how many user account configured in FlowCrypt -> opens Secure Compose with quoted email rendered.

I've tested out that its not possible to obtain the quoted email if the user came from the messageDisplay as the message object doesn't return an object similar to what composeDisplay returns which includes complete email details.

sosnovsky:
Hi @martgil, thanks for detailed explanation of Thunderbird API! In this case, I think, it'll be more convenient to have ability to open current thread in `encrypted inbox`, so user will be able to see decrypted version and then reply to it.
As currently when user receives encrypted message it's not possible to see decrypted content, the only way is to click `flowcrypt` button on toolbar, open encrypted inbox and then search for this message there.

I propose to rename `Secure compose` button to `FlowCrypt` and clicking on it will open current thread in `encrypted inbox` (it should be possible by sending `threadId` parameter). What do you think, should it be possible to implement it this way?

Screenshot 2024-06-21 at 14 53 20

martgil:
Hi @sosnovsky - Goodnews - I have out that its possible for us to obtain the threadId by querying an email's `Message-Id` using Gmail API - I'm currently exploring it out.

For my reference: https://stackoverflow.com/questions/33938816/how-to-get-thread-id-from-gmail-for-a-certain-email

martgil:
Just a little bit more... I've implement almost everything and able to render the message inbox through `threadId` parameter after parsing the actual threadId from the `message-id` header. This can be tested out by simply clicking the secure compose on Thunderbird's message Display and should bring you to the thread message.

The things I would still be including is the account cross checking making sure that the account can only interact with their emails and not from other accounts' emails.

sosnovsky:
@martgil does clicking on `secure compose` on message display works well for you? For me message decryption stucks in `Loading...` state:

Screenshot 2024-06-25 at 19 48 53

However messages decrypted correctly when opening directly from `encrypted inbox`.
Seems to be related to https://github.com/FlowCrypt/flowcrypt-browser/issues/5770, as thunderbird and firefox probably use the same web engine.

martgil:
Hi @sosnovsky,

When opening an encrypted email with Thunderbird's secure compose, I also experience the same issue. One thing I have observed is that there was an error during a process we might have tried to perform in Thunderbird. The specific error says, "Unchecked lastError value: Error: DataCloneError: Promise object could not be cloned." I will check that now and let you know everything I can gather.

martgil:
Hi @sosnovsky - I have found a minor flaw of how we planned to implement the ability to reply to an email. Assume there was 3 accounts in thunderbird name a, b, c... with two of them have their FlowCrypt setup let say a and b set up FlowCrypt. In this current PR, assuming you are acting as user inbox a and u tried to reply securely with it with user b... user b wont have user b's thread id to that email of user a.

So to cut the story short, I believe we no longer need to list the FlowCrypt account list in secure reply at all since we do not need to put extra efforts for cross account checking which should not exists at all. So I propose, when secure compose is clicked if in 1.) compose Display - will show full screen secure compose with quoted text present and if 2.) in message Display will load up the email thread for that account. Would you agree on that?

sosnovsky:
> So to cut the story short, I believe we no longer need to list the FlowCrypt account list in secure reply at all since we do not need to put extra efforts for cross account checking which should not exists at all. So I propose, when secure compose is clicked if in 1.) compose Display - will show full screen secure compose with quoted text present and if 2.) in message Display will load up the email thread for that account. Would you agree on that?

Sounds good, will be quicker and more convenient than manually choosing account from list 👍

martgil:
> Sounds good, will be quicker and more convenient than manually choosing account from list 

Thank you - I have also trim down tremendous unnecessary code. I'll update you really soon.

extension/js/common/browser/browser-msg.ts
```diff
@@ -446,6 +451,40 @@ export class BrowserMsg {
     });
   }
 
+  public static thunderbirdSecureComposeHandler() {
+    const handleClickEvent = async (tabId: number, acctEmail: string, thunderbirdMsgId: number, composeMethod?: browser.compose._ComposeDetailsType) => {
+      const accountEmails = (await GlobalStore.get(['account_emails'])).account_emails;
+      const useFullScreenSecureCompose = (await browser.windows.getCurrent()).type === 'messageCompose';
```
martgil, line 457: Hi @sosnovsky I think I have pretty much everything in this PR. Just one thing for me to do, I need a better way of detecting the account that has FlowCrypt successfully setup more than the account that is already connected with FlowCrypt. I have checked that the "await GlobalStore.get(['account_emails'])" just lists all the account connected to FlowCrypt - that includes previously setup account but then decided to abandon setup such as performing account reset doesnt seem to deduce that array.

Other than that, it can now be reviewed in terms of usage. Please observe, quote rendering, visuals, ability to identify reply or forward, rendering of email based on threadId.

Thank you for help.

extension/chrome/elements/compose.ts
```diff
@@ -260,6 +272,57 @@ export class ComposeView extends View {
   };
 
   public isCustomerUrlFesUsed = () => Boolean(this.fesUrl);
+
+  private preParseEmailRecipientsIfNeeded = async () => {
+    if (Catch.isThunderbirdMail() && this.thunderbirdMsgId) {
+      const thunderbirdMimeMsg = await browser.messages.getFull(this.thunderbirdMsgId);
```
sosnovsky, line 278: I think we can replace it with `const { headers, parts }  = await browser.messages.getFull(this.thunderbirdMsgId);`
And then instead of using `thunderbirdMimeMsg.headers...` and `thunderbirdMimeMsg.parts...` we can just type `headers...` and `parts...`, so no need for duplicating `thunderbirdMimeMsg`

extension/chrome/elements/compose.ts
```diff
@@ -41,9 +41,12 @@ export class ComposeView extends View {
   public readonly removeAfterClose: boolean;
   public readonly disableDraftSaving: boolean;
   public readonly debug: boolean;
+  public readonly useFullScreenSecureCompose: boolean;
   public readonly isReplyBox: boolean;
   public readonly replyMsgId: string;
   public readonly replyPubkeyMismatch: boolean;
+  public readonly composeMethod: 'reply' | 'forward' | undefined;
+  public readonly thunderbirdMsgId: number;
   public replyOption?: ReplyOption;
```
sosnovsky, line 50: there is already `replyOption` parameter which is similar to `composeMethod`, probably we can use it without adding `composeMethod` param?

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -72,14 +87,21 @@ export class InboxView extends View {
...
         $('body').text('Not supported for ' + emailProvider);
       } else {
         await this.inboxMenuModule.render();
+        const parsedThreadId = await this.parseThreadIdFromHeaderMessageId();
         if (this.threadId) {
           await this.inboxActiveThreadModule.render(this.threadId);
+        } else if (parsedThreadId) {
+          await this.inboxActiveThreadModule.render(parsedThreadId);
         } else {
           await this.inboxListThreadsModule.render(this.labelId);
```
sosnovsky, line 106: no need to parse thread id when we already have existing `threadId`, so `const parsedThreadId = await this.parseThreadIdFromHeaderMessageId();` should be moved to `else` case:
```ts
if (this.threadId) {
  await this.inboxActiveThreadModule.render(this.threadId);
} else {
  const parsedThreadId = await this.parseThreadIdFromHeaderMessageId();
  if (parsedThreadId) {
    await this.inboxActiveThreadModule.render(parsedThreadId);
  } else {
    await this.inboxListThreadsModule.render(this.labelId);
  }
}
```

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -132,6 +154,29 @@ export class InboxView extends View {
...
+  private preRenderSecureComposeInFullScreen = () => {
+    if (this.useFullScreenSecureCompose && this.thunderbirdMsgId) {
+      this.injector.openComposeWin(undefined, true, this.thunderbirdMsgId, this.composeMethod);
+    }
+  };
+
+  private parseThreadIdFromHeaderMessageId = async () => {
+    let threadId;
+    const tbMsgId = Number(this.thunderbirdMsgId);
+    if (Catch.isThunderbirdMail() && tbMsgId && !this.useFullScreenSecureCompose) {
```
sosnovsky, line 166: no need to convert `this.thunderbirdMsgId` to `number`, as it's already has `number` type. we can just check if it's defined in `if`:
```ts
if (Catch.isThunderbirdMail() && this.thunderbirdMsgId && !this.useFullScreenSecureCompose) {
```

extension/chrome/settings/inbox/inbox.ts
```diff
@@ -132,6 +154,29 @@ export class InboxView extends View {
...
+      const thunderbirdFullMsg = await browser.messages.getFull(tbMsgId);
+      if (thunderbirdFullMsg.headers) {
+        const messageId = thunderbirdFullMsg.headers['message-id'][0];
+        if (messageId) {
+          const gmail = new Gmail(this.acctEmail);
+          const gmailRes = await gmail.threadIdGet(messageId);
+          threadId = gmailRes.messages[0].threadId;
```
sosnovsky, line 173: we already have `new Gmail` initialized, can just use it:
```ts
const gmailRes = await this.gmail.threadIdGet(messageId);
```

extension/css/cryptup.css
```diff
@@ -1,4 +1,8 @@
 /* © 2016-2018 FlowCrypt Limited. Limitations apply. Contact human@flowcrypt.com */
+.height-0 {
+  height: 0;
+}
+
```
sosnovsky, line 5: looks like it's not used anywhere, can be removed

extension/js/common/api/email-provider/gmail/gmail-parser.ts
```diff
@@ -65,6 +65,13 @@ export namespace GmailRes {
     nextPageToken: string;
     resultSizeEstimate: number;
   };
+  export type GmailThreadIdQuery = {
+    messages: {
+      id: string;
+      threadId: string;
+    }[];
+    resultSizeEstimate: number;
+  };
```
sosnovsky, line 74: we have `GmailMsgList` type which is the same as this one, let's just use it

extension/js/common/api/email-provider/gmail/gmail.ts
```diff
@@ -33,6 +33,10 @@ export class Gmail extends EmailProviderApi implements EmailProviderInterface {
     }
   };
 
+  public threadIdGet = async (messageId: string): Promise<GmailRes.GmailThreadIdQuery> => {
+    return await Google.gmailCall<GmailRes.GmailThreadIdQuery>(this.acctEmail, 'messages', { method: 'GET', data: { q: `rfc822msgid:${messageId}` } });
+  };
```
sosnovsky, line 38: we already have `msgList` which performs the same api call, let's just use it with `q` param set to `rfc822msgid:${messageId}`

sosnovsky (commented):
overall integration works great 👍
just need to wait for https://github.com/FlowCrypt/flowcrypt-browser/issues/5770 fix, as currently it's not possible to test some functionality because of the message stuck

martgil:
>  just need to wait for #5770 fix, as currently it's not possible to test some functionality because of the message stuck

Ah, yes. I'm settle all this requested change while were are looking into #5770 root case. The root cause is not so obvious as sometimes it works well quite randomly when I investigated about the issue.

extension/chrome/elements/compose.ts
```diff
@@ -260,6 +272,57 @@ export class ComposeView extends View {
...
+      const from = thunderbirdMimeMsg.headers?.from[0];
+      const to = thunderbirdMimeMsg.headers?.to || [];
+      const cc = thunderbirdMimeMsg.headers?.cc || [];
+      const bcc = thunderbirdMimeMsg.headers?.bcc || [];
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const subject = thunderbirdMimeMsg.headers!.subject[0];
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const date = thunderbirdMimeMsg.headers!.date[0];
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const msgId = thunderbirdMimeMsg.headers?.['message-id'][0] || '';
```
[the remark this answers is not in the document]
martgil: Will do - thanks!

extension/chrome/elements/compose.ts
martgil: Thank you - I will update it accordingly.

extension/chrome/settings/inbox/inbox.ts
martgil: Ah yes, I see. It's illogical to always run the threadId parsing process when its known. I'll update it accordingly.

extension/chrome/settings/inbox/inbox.ts
martgil: Okay - will update. thanks!

extension/css/cryptup.css
martgil: yes, can remove now.

extension/chrome/settings/inbox/inbox.ts
martgil: Ah yes, will do it as as follows.

extension/js/common/api/email-provider/gmail/gmail.ts
martgil: Thanks - I'll remove threadIdGet now and the related typings i made and I'll use the one that already exists.

extension/js/common/api/email-provider/gmail/gmail-parser.ts
martgil: Thanks - I'll use  the `GmailMsgList`.

extension/chrome/elements/compose.ts
martgil: Thanks for the suggestion - I reviewed this case and I think I can just use `replyOption` for me not to add yet another url parameter to parse. Please kindly check if i'll do it the right way.

martgil: Upon checking, the typings would not align to `replyOption` when the moment I will need to pass it to the `addTripleDotQuoteExpandFooterAndQuoteBtn`. the typings...`'reply'|'forward'`... of the thunderbird seems to align very well with `addTripleDotQuoteExpandFooterAndQuoteBtn` so even if I had removed the requirement of passing URL parameter in the compose.ts, the replyOption accepts `'a_reply'|'a_forward'` so there is that restriction that I see.

martgil:
Needs to wait for https://github.com/FlowCrypt/flowcrypt-browser/issues/5770.

martgil:
Hi @sosnovsky - I have found a reliable way for you to test message decryption on FlowCrypt inbox without us having to wait for the official fix for #5770.

On thunderbird, open encrypted inbox and wait for at least 10 seconds for the page to load completely. Then after that, click on the encrypted email which would open the email thread.

That method works for me:

image

The same method of waiting for 10s works for me with my Firefox build of the extension. The message stuck on Gmail's extension page dont work with that waiting method.

tooling/build-types-and-manifests.ts
```diff
@@ -58,23 +58,24 @@ addManifest('firefox-consumer', manifest => {
...
+      default_title: 'Secure Compose', // eslint-disable-line @typescript-eslint/naming-convention
       default_icon: '/img/logo/flowcrypt-logo-64-64.png', // eslint-disable-line @typescript-eslint/naming-convention
-      // default_popup will be updated later
-      default_popup: '/chrome/popups/default.htm', // eslint-disable-line @typescript-eslint/naming-convention
+    };
+    (manifest.browser_specific_settings as browser._manifest.FirefoxSpecificProperties).strict_min_version = '102.0';
+    delete manifest.background;
+    manifest.background = {
+      type: 'module',
+      scripts: ['/js/service_worker/background.js'],
```
sosnovsky, line 78: no need for `delete manifest.background;` as it's re-assigned on the next line

extension/js/service_worker/background.ts
```diff
@@ -67,4 +67,7 @@ console.info('background.js service worker starting');
   BrowserMsg.alarmListen();
   await BgHandlers.updateUninstallUrl({});
   injectFcIntoWebmail();
+
+  // Thunderbird event handlers
+  BrowserMsg.thunderbirdSecureComposeHandler();
```
sosnovsky, line 72: it should be performed only when extension is opened in Thunderbird, as currently in Chrome it causes errors:
```
catch.js:64 ReferenceError: browser is not defined
    at BrowserMsg.thunderbirdSecureComposeHandler (chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/common/browser/browser-msg.js:284:9)
    at chrome-extension://pjohdcpiapjhagmakkjpgjmodmgfmkkd/js/service_worker/background.js:82:20
    at <async> <anonymous> (/extension/js/service_worker/background.ts:20:2)
```

extension/js/common/browser/browser-msg.ts
```diff
@@ -446,6 +451,40 @@ export class BrowserMsg {
...
+      }
+    };
+    browser.composeAction.onClicked.addListener(async tab => {
+      const messageDetails = await browser.compose.getComposeDetails(Number(tab.id));
+      const composeMethod = messageDetails.type;
+      const msgId = Number(messageDetails.relatedMessageId);
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const acctEmail = Str.parseEmail(messageDetails.from as string).email!;
+      await handleClickEvent(Number(tab.id), acctEmail, msgId, composeMethod);
+    });
```
sosnovsky, line 475: when I click `secure compose` from new message window it just opens `encrypted inbox`, but not encrypted compose modal
Screenshot 2024-07-02 at 15 34 00

martgil: Thanks @sosnovsky - can you please clarify: Does it open the email thread in FlowCrypt inbox page for that email?

For "secure compose" clicked in message display: the threadId will be loaded and would load that email.
For "secure compose" clicked in compose display - eg.via Thunderbird reply -> secure compose or Thunderbird forward -> secure compose: It will open the compose modal with quoted email rendered for forwarded or reply.

extension/js/service_worker/background.ts
martgil: checks added - thank you!

extension/js/common/browser/browser-msg.ts
sosnovsky: Here I mean creating new thread by clicking `New message` button in thunderbird and then `secure compose` button in compose modal, it opens `encrypted inbox`, but without `secure compose` modal:

Screenshot 2024-07-03 at 15 20 39

I noticed that opened url is moz-extension://fa4f4d23-0b95-4696-a57e-f4852bf94f8e/chrome/settings/inbox/inbox.htm?acctEmail=...&useFullScreenSecureCompose=___cu_true___&thunderbirdMsgId=0, probably `thunderbirdMsgId=0` causes `secure compose` modal not appearing

martgil: Oh ok got it - i has been corrected as of now. Tested most features and they should work fine now in this PR.

martgil:
Hi @sosnovsky - I retested this PR again and most of them works well now.

extension/chrome/elements/compose.ts
```diff
@@ -260,6 +273,60 @@ export class ComposeView extends View {
   };
 
   public isCustomerUrlFesUsed = () => Boolean(this.fesUrl);
+
+  private preParseEmailRecipientsIfNeeded = async () => {
+    if (Catch.isThunderbirdMail() && this.thunderbirdMsgId && this.thunderbirdMsgId !== 0) {
+      const { headers, parts } = await browser.messages.getFull(this.thunderbirdMsgId);
```
sosnovsky, line 279: I noticed a couple issues when replying of forwarding message:
- when I replied to email message, it created a new thread instead of adding reply message to existing thread
- when I tried to forward message - recipients list was filled with recipients of original message, while it should be empty when forwarding.

I think we can simplify current implementation and just re-use existing code for populating data for reply and forward:
- there is `replyMsgId` url param which then used in `fetchReplyMeta` method for fetching reply data. You found a way how to get `msgId` from `thunderbirdMsgId` by using `headers?.['message-id'][0] || '';`, what about parsing it and using as `replyMsgId` param?
- there is `replyOption` url param which defines how recipients list should be populated, we can just convert `reply` and `forward` to `a_reply` and `a_forward`

martgil: Hi @sosnovsky - thank you for raising this.

We have this limitation where the "message-id", in headers object only available to obtain when the secure compose click invoked from the `messageDisplay` and not from the `composeDisplay`. The place to identify if its part of `composeDisplay` is when the compose pop-up window from Thunderbird appears. To be very specific, the object message-id from headers is not present from `composeDisplay` which is sensible -- so there is that limitations with us where we can't rely on messageId to fetch threadId -> replyMsgId.

However, I can correct the reported behaviors and use the `replyOption` way of determining pre-filled data. I will proceed in this and let me know if the implementation I would be taking look good to you.

> when I replied to email message, it created a new thread instead of adding reply message to existing thread
> when I tried to forward message - recipients list was filled with recipients of original message, while it should be empty when forwarding.

martgil: Will check and do the necessary changes to correct these. Thank you for your help.

extension/js/common/browser/browser-msg.ts
```diff
@@ -449,6 +454,40 @@ export class BrowserMsg {
...
+      if (accountEmails.length !== 0) {
+        await BgUtils.openExtensionTab(
+          Url.create('/chrome/settings/inbox/inbox.htm', { acctEmail, useFullScreenSecureCompose, thunderbirdMsgId, composeMethod })
+        );
+        await browser.tabs.remove(tabId);
+      } else {
+        await BgUtils.openExtensionTab(Url.create('/chrome/settings/initial.htm', {}));
+      }
+    };
+    browser.composeAction.onClicked.addListener(async tab => {
```
sosnovsky, line 471: I also noticed that it's recommended to use `messenger...` instead of `browser...` in thunderbird environment - https://webextension-api.thunderbird.net/en/128-esr-mv3/#getting-started, I think it'll be useful for differentiating browser and thunderbird actions

martgil: Thanks @sosnovsky - I have settled this one.

extension/chrome/elements/compose.ts
martgil: > when I replied to email message, it created a new thread instead of adding reply message to existing thread

@sosnovsky - I think I understand this now - when reply is clicked where an existing thread (previous email exchanges) are already made, then it should load and quote all of the conversation instead of just the current email where user replied to. I'll get it fixed.

sosnovsky: > when I replied to email message, it created a new thread instead of adding reply message to existing thread

To reproduce:
- open some thread in thunderbird
- click `reply`
- switch to `encrypted reply`
- compose and send reply from FlowCrypt encrypted compose
- then on recipient side sent reply will be on separate thread (not in the existing one)

martgil: Got it - I see the bigger picture now on how it works. Will try to fix this and let you know.

martgil: Thanks @sosnovsky - Its fixed. I think I have fixed it. Please feel free to perform additional checks as I might've them during my checkings.

sosnovsky:
@martgil I noticed a small issue - after finishing extension setup, clicking `Compose encrypted email` will open Gmail page in external browser. I think it'll be better to open `encrypted inbox` page in Thunderbird:

Screenshot 2024-07-16 at 15 43 31

martgil:
Ah yes, I missed that one. I'll be fixing that too.

sosnovsky (commented):
Well done, almost completed it 👍

Another small issue:
- open encrypted message in thunderbird
- click `reply` to open thunderbird compose window
- in thunderbird compose window click `secure compose` to open FlowCrypt compose window
- FlowCrypt compose window is opened, but in the bottom there is stuck `decrypting...` label:
Screenshot 2024-07-16 at 21 28 41

I checked console messages, and there was several `Unchecked lastError value: Error: Promised response from onMessage listener went out of scope` error messages, probably they're related to mentioned issue.

Screenshot 2024-07-16 at 21 31 35

martgil:
> I checked console messages, and there was several Unchecked lastError value: Error: Promised response from onMessage listener went out of scope error messages, probably they're related to mentioned issue.

Thanks @sosnovsky, I'm still figuring it out - its quite difficult to debug this one. I also ran into some problem with the previous implementation that produces UI bug for example, I can't close the secure_compose_window, and the trash folder is clicked while the secure compose is in full_window, it seems to act as a secure reply box due to `replyMsgId` but it's not awful, just need to figure it out very precisely.

martgil:
Hi @sosnovsky - I have fixed the last two issues raised earlier. Please check as you might've noticed some other issues. Thanks

sosnovsky:
Hi @martgil, currently testing your latest changes - previous issues seems to be fixed, but I noticed another one:
- open encrypted message in thunderbird
- click `forward` (it happens only for forwarding, not replying) to open thunderbird compose window
- in thunderbird compose window click `secure compose` to open FlowCrypt compose window
- FlowCrypt encrypted inbox shows error `FlowCrypt encountered an error with unknown cause. TypeError: this.popoverItems[option] is undefined`:

Screenshot 2024-07-22 at 13 49 56

clicking `retry` doesn't help

martgil:
> Hi @martgil, currently testing your latest changes - previous issues seems to be fixed, but I noticed another one:

Thanks Roma - I'll check this one and get it fixed.

extension/chrome/elements/compose-modules/compose-quote-module.ts
```diff
@@ -177,6 +187,9 @@ export class ComposeQuoteModule extends ViewModule<ComposeView> {
     if (decryptRes.success) {
       return decryptRes.content.toUtfStr();
     } else if (decryptRes.error && decryptRes.error.type === DecryptErrTypes.needPassphrase) {
+      if (Catch.isThunderbirdMail() && this.view.useFullScreenSecureCompose) {
+        await Time.sleep(2300);
+      }
```
sosnovsky, line 192: what's usage for this `time.sleep()`, is it needed?
I tried to delete it and didn't notice any issues

sosnovsky:
> > Hi @martgil, currently testing your latest changes - previous issues seems to be fixed, but I noticed another one:
> 
> Thanks Roma - I'll check this one and get it fixed.

It can also be reproduced with these steps:
- click `new message` button in thunderbird to open compose window
- click `secure compose` button in compose window
- FlowCrypt extension opens with the same `TypeError: this.popoverItems[option] is undefined` error

extension/chrome/elements/compose-modules/compose-quote-module.ts
martgil: Time.sleep() seems to fix the issue where "decrypting..." phase stucks for decrypting quoted email when replying to encrypted email. Background scripts seems to comunicate to the page but the page is not ready yet.

extension/chrome/elements/compose-modules/compose-quote-module.ts
```diff
@@ -215,7 +228,8 @@ export class ComposeQuoteModule extends ViewModule<ComposeView> {
     if (method === 'reply') {
       const header = `<div ${dirAttr}>On ${dateStr}, ${from ?? ''} wrote:</div>`;
       const sanitizedQuote = Xss.htmlSanitize(header + escapedText);
-      return `<blockquote ${dirAttr}>${sanitizedQuote}</blockquote>`;
+      const thunderbirdClass = Catch.isThunderbirdMail() ? 'class="height-0"' : '';
+      return `<blockquote ${thunderbirdClass} ${dirAttr}>${sanitizedQuote}</blockquote>`;
```
sosnovsky, line 232: also for this `height-0` class - I tried to remove it, and didn't notice any changes in the previous message quote rendering. maybe it's not needed?

extension/chrome/elements/compose-modules/compose-quote-module.ts
sosnovsky: Seems like some other change fixed this issue, as decrypting works well for me even after removing this `Time.sleep` part. Please check on your side

extension/chrome/elements/compose-modules/compose-quote-module.ts
martgil: It fixes the blockquote scroll when in fullscreen where without it, it will make the secure compose window extremely long when quoted email contains a lot of newlines . eg long emails. You could open long HTML emails and you'll notice the UI bug.

extension/chrome/elements/compose-modules/compose-quote-module.ts
martgil: Ok, will check tomorrow morning.

extension/chrome/elements/compose-modules/compose-quote-module.ts
sosnovsky: Ah, I see, thanks for clarification!

extension/chrome/elements/compose-modules/compose-quote-module.ts
martgil: Hi @sosnovsky - Thanks for raising this. Upon checking, the issue still exists on my end after removing this grace period allowance `Time.sleep()`. For example, the issue will reproduce on the recent encrypted conversion. I think this is just bare minimum, otherwise, it will be stuck rendering "decrypting..." with no way the passphrase dialog not appearing.

extension/chrome/elements/compose-modules/compose-quote-module.ts
martgil: No worries at all. thank you for checking this extensively.

extension/chrome/elements/compose-modules/compose-quote-module.ts
sosnovsky: Got it 👍

martgil:
> It can also be reproduced with these steps:
>
> - click new message button in thunderbird to open compose window
> - click secure compose button in compose window
> - FlowCrypt extension opens with the same TypeError: this.popoverItems[option] is undefined error

Hi @sosnovsky Greetings! I have fixed this one together with other issues I can spot. Can you please take another look please? Thanks.

sosnovsky (approved):
Well done, all found issues seem to be fixed, thanks!

martgil:
Thanks @sosnovsky - it's time now to learn my way into message decryption within the Thunderbird mail client.

PR #5765, opened 2024-06-10
#5764 Fix long ‘environment’ string

### LINKED ISSUE #5764: Error when trying to send exception from content script
In backend logs there are errors saying `Error processing POST api/v1/log-collector/exception | id: 7d39a4a33f96 | message: ERROR: value too long for type VARCHAR(20)`. I noticed that it's possible to have longer string for `environment` property when it has `ex:script:gmail` value: https://github.com/FlowCrypt/flowcrypt-browser/blob/e6fe45d270efdfabe16631bcb4b4c22c0a9bd0be/extension/js/common/platform/catch.ts#L196

sosnovsky:
This PR makes `environment` string shorter to be valid for `log-collector/exception` endpoint

close #5764

- Not worth testing

### To be filled by reviewers

### DIFF
--- a/extension/js/common/platform/catch.ts
+++ b/extension/js/common/platform/catch.ts
@@ -193,7 +193,7 @@ export class Catch {
     } else if (origin === 'https://flowcrypt.com') {
       env = 'web:prod';
     } else if (origin === 'https://mail.google.com') {
-      env = 'ex:script:gmail';
+      env = 'ex:s:gmail';
     }
     return browserName + ':' + env;
   }


extension/js/common/platform/catch.ts
```diff
@@ -193,7 +193,7 @@ export class Catch {
     } else if (origin === 'https://flowcrypt.com') {
       env = 'web:prod';
     } else if (origin === 'https://mail.google.com') {
-      env = 'ex:script:gmail';
+      env = 'ex:s:gmail';
     }
```
martgil, line 197: Hi @sosnovsky - Do you think it's necessary to add comments explaining "env" if it stands for "script"?

sosnovsky: I think it's ok to leave it without comment, as currently we use this property only for differentiating environments where exception happened.

martgil: Oh, okay - thanks!

martgil (approved):

PR #5795, opened 2024-07-24
#5444 ConfiguredIDPOAuth shows authentication popup and save JWT to storage

### LINKED ISSUE #5444: ConfiguredIdpOAuth should produce an oauth window + save JWT to storage
part of #5311

ioanmo226:
This PR implemented feature to let ConfiguredIDPOAuth shows authentication popup and save JWT to storage

close #5444// if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


test/source/mock/lib/oauth.ts
```diff
@@ -8,6 +8,8 @@ import { Str } from '../../core/common';
 const authURL = 'https://localhost:8001';
 
 export class OauthMock {
+  public static customIDPClientId = 'ZFfdspDb9Oz1vnFdA4NrCyNaA9N3jYDl';
```
tomholub, line 11: could this value be something like "custom-test-idp-client-id" ? Or is it using an actual external idp?

test/source/tests/tooling/browser-recipe.ts
```diff
@@ -46,12 +46,8 @@ export class BrowserRecipe {
     const settingsPage = await browser.newExtensionSettingsPage(t, acctEmail);
     const oauthPopup = await browser.newPageTriggeredBy(t, () => settingsPage.waitAndClick('@action-connect-to-gmail'));
     await OauthPageRecipe.google(t, oauthPopup, acctEmail, 'approve');
-    if (checkForConfiguredIdPOAuth)
-      await settingsPage.waitAndRespondToModal(
-        'warning',
-        'confirm',
-        'Custom IdP is configured on this domain, but it is not supported on browser extension yet.'
-      );
+    if (checkForConfiguredIdPOAuth) await settingsPage.page.waitForNavigation({ waitUntil: 'networkidle2' });
```
tomholub, line 49: I would hope ideally here it would explicitly write OauthPageRecipe.customIdp(...) under the .google() line

tomholub (commented):
Some comments for code clarity, just a quick review

test/source/mock/lib/oauth.ts
ioanmo226: It's actual test external idp(my auth0 test idp).
It can also be `custom-test-idp-client-id` but I think setting actual external idp won't be secret issue?

test/source/tests/tooling/browser-recipe.ts
ioanmo226: You're absolutely right. I also thought about this. The custom IDP popup appears automatically after the user completes Google authentication, without any user action. The challenge is getting the custom IDP popup instance since there's no user interaction. Need to think about a solution for this.

test/source/mock/lib/oauth.ts
tomholub: Since we have Auth0 too, could you work with Roma to have similar one set up on our Auth0 account?

In that case could name the new one `custom-idp-client-id-for-automated-testing`

I don't see any security issues here at all, the IdP doesn't give access to any actual resources. As long as it's properly named as a test in the Auth0 settings as well.

test/source/tests/tooling/browser-recipe.ts
tomholub: If it makes it easier for you, you could do `OauthPageRecipe.googleFollowedByCustomIdp` so you're still reusing the instance. My goal is that usually it only does `OauthPageRecipe.google()` and in a few cases where we test for this, we do it explicitly

test/source/mock/lib/oauth.ts
ioanmo226: Okay, Got it

test/source/tests/tooling/browser-recipe.ts
ioanmo226: Sorry, I don't quite get it.
Do you mean I should put the `OauthPageRecipe.google()` call and the `await settingsPage.page.waitForNavigation({ waitUntil: 'networkidle2' });` logic into `OauthPageRecipe.googleFollowedByCustomIdp` and then call that when `checkForConfiguredIdPOAuth` is true?

ioanmo226: What you mentioned is correct.

For testing, the custom IDP is not an actual IDP; it is just a mock. When the Google IDP mock is finished, it opens the mock `/o/oauth2/auth `URL and then automatically redirects to the success page (https://google.localhost:${port}/robots.txt). This page is then used to retrieve the auth code and get the custom IDP ID token.

https://github.com/FlowCrypt/flowcrypt-browser/blob/3f10b9ef5595428a3d34d07b6f728e6109bb48e8/test/source/mock/google/google-endpoints.ts#L137

> - user clicks to log in
> - Google IdP pops up, user interacts with Google IdP, Google IdP goes away
> - custom IdP pops up, user interacts with custom IdP, custom IdP goes away
> - success
> -

ioanmo226: FYI, I set `customIDPClientId` to the actual custom IDP client ID, but it can be any string, like `custom-test-idp-client-id`, and it doesn't matter at all.

`public static customIDPClientId = 'ZFfdspDb9Oz1vnFdA4NrCyNaA9N3jYDl';`

ioanmo226: For the Google IDP mock, after the user taps the `Sign in with Google` button, it opens the `/o/oauth2/auth` URL (without `proceed=true`). Therefore, we need to redirect the user to `/o/oauth2/auth?proceed=true` at `oauth-page-recipe.ts#L38`.

However, for the custom IDP mock, there's no check to see if `proceed=true`. Consequently, after the Google mock IDP is finished and the custom IDP mock window opens, it automatically redirects to `robots.txt` which is then used to get the authorization code and ID token, and the popup closes automatically.

It's a good idea to check if `proceed=true` in the custom IDP mock as well. However, as I mentioned, it's challenging to get a reference to the custom mock IDP window.

Let me know if this makes sense.

https://github.com/FlowCrypt/flowcrypt-browser/blob/8770025a965a3ad13cd8fa81d7c3f92fe8117499/test/source/tests/page-recipe/oauth-page-recipe.ts#L38


https://github.com/FlowCrypt/flowcrypt-browser/blob/3f10b9ef5595428a3d34d07b6f728e6109bb48e8/test/source/mock/google/google-endpoints.ts#L132

https://github.com/FlowCrypt/flowcrypt-browser/blob/3f10b9ef5595428a3d34d07b6f728e6109bb48e8/test/source/mock/google/google-endpoints.ts#L137

tomholub: Okay, i get it now. 

Actually, from user perspective, I think there should probably be a button anyway. So the second the window doesn't show up on its own. Instead an alert (for simplicity?) shows up "Please also log in with company credentials". and a "Log in" button

Then you can click that in ".customIdp()" in tests, and also wait for network on that method. 

Then when we expect custom login, you do ".gmail()" followed by ".customIdp()" and it's clear it's tested for. 

also in this new method, please copy what you said here about the lack of user interaction, as a comment. So it's clear why so different from .gmail

thanks!

ioanmo226: That sounds great.
Thank you for your clarification.

test/source/tests/setup.ts
```diff
@@ -2525,18 +2526,25 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
       testWithBrowser(async (t, browser) => {
+        const port = t.context.urls?.port ?? '';
         const oauthConfig = {
-          clientId: 'your-client-id',
-          clientSecret: 'your-client-secret',
-          redirectUrl: 'https://example.com/redirect',
-          authCodeUrl: 'https://example.com/auth',
-          tokensUrl: 'https://example.com/tokens',
+          clientId: OauthMock.customIDPClientId,
+          clientSecret: OauthMock.customIDPClientId,
```
tomholub, line 2532: Here supposed to have the secret, not id

test/source/tests/setup.ts
```diff
@@ -2525,18 +2526,25 @@ AN8G3r5Htj8olot+jm9mIa5XLXWzMNUZgg==
...
+          redirectUrl: `https://google.localhost:${port}/robots.txt`,
```
tomholub, line 2533: Okay, there is a bit of a difficulty here. We used `robots.txt` for Google IdP because it's the same org, so sharing auth result with them is safe.

But robots.txt should not be used for real authentication of 3rd party IdPs. I know this is just a mock but we should adjust it closer to what we expect it to be in prod.

What happens with real Auth0? Does it get sent to robots too? It should have its own redirect url provided by Auth0. When we use it, what happens? Is it a problem that our content script doesn't run there, or does it work fine?

test/source/tests/tooling/browser-recipe.ts
tomholub: > That sounds great.
> Thank you for your clarification.

In this case, `checkForConfiguredIdPOAuth` should be called `expectCustomIdp`

test/source/tests/setup.ts
ioanmo226: Well, I (IDP providers should also) set redirect_url to robots.txt in auth0 configuration.
Was not sure which redirect url to set and thought it would be safe to set redirect_url to robots.txt.
Maybe you prefer to set redirect url to our chrome extension url?(like chrome-extension://{id}/auth.js or something like that)?

ioanmo226: When customers set their custom redirect_url, it's challanging to get authorization code from this redirect_url as we can't inject content script in every site

tomholub: > When customers set their custom redirect_url, it's challanging to get authorization code from this redirect_url as we can't inject content script in every site

Right

> Maybe you prefer to set redirect url to our chrome extension url?(like chrome-extension://{id}/auth.js or something like that)?

IIRC Google IdP didn't allow this, causing us to use the hacky method with robots.txt. 

If Auth0 allows this then absolutely yes, we should use this method. Please see if it works as expected. The file could be called `oauth-redirect-page.htm` or similar, unless we already have such page, in which case we can reuse whatever we already have.

tomholub: > If we set the redirect URL to oauth-redirect-page.htm, it shows a page blocked error. I think we need to add chrome-extension://{id} to host-permissions in manifest.json which I think will trigger permission warning.

You're probably trying to set up content scripts at  oauth-redirect-page.htm which doesn't make sense. It's an extension page, you can just add `<script>` tags like any other extension page. See `attachment.htm` page or similar.

> Also, will the client be happy to set the redirect URL like chrome-extension://{id}? Maybe they prefer to set their own redirect_url?

No problem here, they'll set whatever we tell them to set, if it points to the installed app (addon or mobile app) and not to a web page we control.

tomholub (commented):
Other than the redirect page in progress, my comments were addressed 👍

sosnovsky (commented):
Looks good for me 👍 Just need to fix redirect part mentioned by Tom.
Probably we can use [launchWebAuthFlow](https://developer.chrome.com/docs/extensions/reference/api/identity#method-launchWebAuthFlow) method, which enables auth flows for non-Google IdP. There is also [getRedirectURL](https://developer.chrome.com/docs/extensions/reference/api/identity#method-getRedirectURL) method for generating correct redirect URL.

ioanmo226:
I previously decided to use our own oauth solution instead of launchwebAuthFlow because of below reason.
Also  we need to add identity permission in manifest.

https://github.com/FlowCrypt/flowcrypt-browser/pull/4660#discussion_r961695736

> Looks good for me 👍 Just need to fix redirect part mentioned by Tom. Probably we can use [launchWebAuthFlow](https://developer.chrome.com/docs/extensions/reference/api/identity#method-launchWebAuthFlow) method, which enables auth flows for non-Google IdP. There is also [getRedirectURL](https://developer.chrome.com/docs/extensions/reference/api/identity#method-getRedirectURL) method for generating correct redirect URL.

test/source/tests/setup.ts
ioanmo226: No I didn't. Just set redirect url to `oauth-redirect-page.htm` and after user finishes custom IDP authentication, it redirects to this page and above is result. (ERR_BLOCKED_BY_CLIENT)

>You're probably trying to set up content scripts at oauth-redirect-page.htm which doesn't make sense. It's an extension page, you can just add <script> tags like any other extension page. See attachment.htm page or similar.

ioanmo226:
The biggest problem with launchWebAuthflow is we have limited control for UI testing.

extension/js/common/api/authentication/configured-idp-oauth.ts
```diff
@@ -2,21 +2,155 @@
...
-      return;
+    if (storage?.authentication?.oauth?.clientId && storage.authentication.oauth.clientId !== this.GOOGLE_OAUTH_CONFIG.client_id) {
+      await Ui.modal.info('Google login succeeded. Now, please log in with your company credentials as well.');
+      return await this.newAuthPopup(acctEmail, { oauth: storage.authentication.oauth });
     }
+    return authRes;
   };
+
+  public static async newAuthPopup(acctEmail: string, authConf: AuthenticationConfiguration): Promise<AuthRes> {
+    acctEmail = acctEmail?.toLowerCase();
```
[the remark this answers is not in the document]
ioanmo226: Yep!

test/source/tests/setup.ts
ioanmo226: Maybe this one might be correct? It's chatgpt answer though.

> Chrome Security Policies: Chrome has strict security policies that might block certain behaviors that seem unusual or insecure, such as redirecting to a chrome-extension:// URL from a web context, even if it's part of an OAuth flow.

sosnovsky:
> Also we need to add identity permission in manifest.

Maybe it can be added without any additional confirmation popup, like `scripting` permission - it was added with manifest v3 migration and didn't require user confirmation.

> The biggest problem with launchWebAuthflow is we have limited control for UI testing.

Probably it'll work the same as our current mock implementation - it'll open mock authorization endpoint and will wait until it receives auth token. Also a couple years passed since your initial comment about poor documentation, there should be more guides about it's implementation. I think it's worth to check these methods for our auth process, and if it's still broken - then try to use `https://EXTENSION_ID.chromiumapp.org/oauth` as redirect url (such urls are generated by `getRedirectURL` method)

ioanmo226:
@tomholub and @sosnovsky,

I suggest updating our OAuth mechanism. Currently, we inject `oauth2_inject.js` in `robots.txt` to obtain the authorization code. Instead, we could set the redirect URI to `https://{extensionId}.chromiumapp.org/`, listen for `chrome.tabs.onUpdated`, and check if the tab ID matches the newly created auth window's tab ID and if the tab URL matches the redirect URI.
Then, we can retrieve the authorization code directly.
This approach mirrors what the `launchWebAuthFlow` function does.

Let me know your thoughts!

PS: We cannot rely on a content script solution for handling OAuth redirects. When OAuth redirects to our Chrome extension page (or to https://{extensionId}.chromiumapp.org), Chrome prevents the content from loading due to security concerns. As a result, the content script does not load, making it unusable for capturing the authorization code in this scenario.

ioanmo226:
I agree.
Let me try with your suggested solution and then try with my above solution if it doesn't work.

> > Also we need to add identity permission in manifest.
> 
> Maybe it can be added without any additional confirmation popup, like `scripting` permission - it was added with manifest v3 migration and didn't require user confirmation.
> 
> > The biggest problem with launchWebAuthflow is we have limited control for UI testing.
> 
> Probably it'll work the same as our current mock implementation - it'll open mock authorization endpoint and will wait until it receives auth token. Also a couple years passed since your initial comment about poor documentation, there should be more guides about it's implementation. I think it's worth to check these methods for our auth process, and if it's still broken - then try to use `https://EXTENSION_ID.chromiumapp.org/oauth` as redirect url (such urls are generated by `getRedirectURL` method)

sosnovsky:
> Let me try with your suggested solution and then try with my above solution if it doesn't work.
Sounds good, let's try it this way. Usually such native solutions work better in Chrome, and it'll simplify our code to something like this:
```ts
chrome.identity.launchWebAuthFlow({
  'url': authConf.oauth.authCodeUrl + '?client_id=' + clientId + '&response_type=token',
  'interactive': true
}, function(redirectUri) {
  // Extract token from redirectUri.
});
```

tomholub:
> > Also we need to add identity permission in manifest.
> 
> 
> 
> Maybe it can be added without any additional confirmation popup, like `scripting` permission - it was added with manifest v3 migration and didn't require user confirmation.
> 
> 
> 
> > The biggest problem with launchWebAuthflow is we have limited control for UI testing.
> 
> 
> 
> Probably it'll work the same as our current mock implementation - it'll open mock authorization endpoint and will wait until it receives auth token. Also a couple years passed since your initial comment about poor documentation, there should be more guides about it's implementation. I think it's worth to check these methods for our auth process, and if it's still broken - then try to use `https://EXTENSION_ID.chromiumapp.org/oauth` as redirect url (such urls are generated by `getRedirectURL` method)

The problem from what I renege was that the identity mechanism only allowed to authenticate with the account that is connected to your browser. So no multi user to my knowledge

I wouldn't start working on replacing that unless we run into a roadblock with our existing mechanisms. They work well imo

ioanmo226:
Maybe we could only use launchWebAuthFlow for custom IDP authentication and keep using our customized OAuth solution for google IDP.
What's your thouht?

ioanmo226:
Ohhh. just tested and this problem is now past story.
I can now login to any account. We can use launchWebAuthFlow solution for google IDP too I think
image


> The problem from what I renege was that the identity mechanism only allowed to authenticate with the account that is connected to your browser

tomholub:
> @tomholub and @sosnovsky,
> 
> 
> 
> I suggest updating our OAuth mechanism. Currently, we inject `oauth2_inject.js` in `robots.txt` to obtain the authorization code. Instead, we could set the redirect URI to `https://{extensionId}.chromiumapp.org/`, listen for `chrome.tabs.onUpdated`, and check if the tab ID matches the newly created auth window's tab ID and if the tab URL matches the redirect URI.
> 
> Then, we can retrieve the authorization code directly.
> 
> This approach mirrors what the `launchWebAuthFlow` function does.
> 
> 
> 
> Let me know your thoughts!
> 
> 
> 
> PS: We cannot rely on a content script solution for handling OAuth redirects. When OAuth redirects to our Chrome extension page (or to https://{extensionId}.chromiumapp.org), Chrome prevents the content from loading due to security concerns. As a result, the content script does not load, making it unusable for capturing the authorization code in this scenario.

I would like to, but last I tried it didn't work. I think we are using web application type of oauth credentials in oauth settings on google, which doesn't allow setting this type of redirect url meant due extensions. While when trying to use the extension type, it had some other limitations making it unusable for our purposes (I think no pffline_mode or maybe something else). I wouldn't try to change anything unless we absolutely have to. If the current gmail mechanism works, I don't think we have to. 

Of course, if I'm wrong and it's just a matter of adding this redirect url and a small code change, then sure. But from my experience, you'll run into a lot of dead ends.

tomholub:
> Ohhh. just tested and this problem is now past story.
> 
> I can now login to any account. We can use launchWebAuthFlow solution for google IDP too I think
> 
> image
> 
> 
> 
> 
> 
> > The problem from what I renege was that the identity mechanism only allowed to authenticate with the account that is connected to your browser

If it seems easy to do and simplifies things then sure. Does it allow offline_mode? Also, does it work the same on firefox?

tomholub:
But how about we leave existing google mechanism untouched in this pr if it works. Try out the new identity mechanism on 3rd party logins. That way no risk. We gain some experience with it in a low impact scenario. 

Then we can consider changing google login as well in a separate PR in future version.

sosnovsky:
Yeah, it should be safe to start using new identity flow for 3-rd party IdPs, and if no issues found - migrate Google IdP to it later.

ioanmo226:
Sounds good!. Thank you

ioanmo226:
Yeah, it works same on Firefox
>If it seems easy to do and simplifies things then sure. Does it allow offline_mode? Also, does it work the same on firefox?

tomholub (commented):
code looks good on my end, didn't test it

sosnovsky:
@ioanmo226 do you have active test instance of auth0? I tried to use credentials which you previously shared with me, but got error `Authorization page could not be loaded.`

ioanmo226:
How did you test? With which client_id, client_secret did you test. And also did you test with local FES?

ioanmo226:
FYI, redirect_url in `authentication.oauth` is not necessary/required in client configuration as redirect_url needs to be same for all oauth (custom IDP) configurations. (https://EXTENSION_ID.chromiumapp.org/oauth)

sosnovsky:
I just implicitly set authentication configuration in `fetchAndSaveClientConfiguration` method, without local FES.
For redirect URL I used `chrome.identity.getRedirectURL('oauth')`, but it still gives me `Authorization page could not be loaded.` error.
My extension id is the same as I previously shared with you for redirect URL (`pjohdcpiapjhagmakkjpgjmodmgfmkkd`).

tomholub:
> FYI, redirect_url in `authentication.oauth` is useless/not required in client configuration as redirect_url needs to be same for all oauth (custom IDP) configurations. (https://EXTENSION_ID.chromiumapp.org/oauth)

Probably also true on android and ios, but it could be useful if we ever do authentication on the web. Once we figure this out, we'll update the docs.

ioanmo226:
I haven't added bnjglocicdkmhmoohhfkfkbbkejdhdgc redirect_url before. Just added to my auth0 config. Please check again.

sosnovsky (approved):
All good now, works great, thanks!

PR #5796, opened 2024-07-24
#5667 Added message decryption on Thunderbird

### LINKED ISSUE #5667: [thunderbird] Recognize encrypted emails and decrypt them

martgil:
This PR added message decryption and its necessary UI improvements on Thunderbird email client. 

close #5667 

- Difficult to test (cannot find a reliable way to perform tests in a Thunderbird mail environment)

### To be filled by reviewers


martgil:
Update: Embedding `iframe` within the messageDisplay of the Thunderbird Mail client is impossible. So as a reliable option for us is to just embed the CSS and images within the document.body and format it to as close as what we have in PGP_block page.

image

Another problem I have is that, opgp object from `openpgpjs-custom.js` reference is `undefined` but when I include it, there is error in Object from which error details is conceal from Thunderbird console with no way to obtain relevant information, but I think its because the `openpgp.js` file is simply not injected so I'll check if I can find a workaround to it otherwise, message decryption is definitely possible.

@sosnovsky, just sharing this with you. There's no need to check it for now, but I'll let you know when help is needed.

sosnovsky:
@martgil thanks for update! In the case if it's not possible to directly decrypt message inside Thunderbird UI, we can add `decrypt message` button above message content which will open and decrypt this message inside FlowCrypt extension.

martgil:
> @martgil thanks for update! In the case if it's not possible to directly decrypt message inside Thunderbird UI, we can add `decrypt message` button above message content which will open and decrypt this message inside FlowCrypt extension.

Thanks for the feedback and suggestion. I will follow your suggestion, as of now, I'm getting positive result from previous problem I got. Right now, I'm getting an error about key's being passphrase protected, and so im in the process of popping the passphrase prompt inside the Thunderbird mail message display UI.

martgil:
Hi @sosnovsky -  I have almost finished this one where in its current state, general message decryption and signed message verification is working. However, that's when the private key has already been unlocked. Right now, the only way it works is by unlocking the passphrase for the first time by unlocking any secure message from the FlowCrypt browser extension page and then going to the Thunderbird mail messageDisplay afterwards will result to successful message decryption.

There is several ways I think it can be further improved such as:

- Unlock the private key right upon user successfully recovers his private key via setup or if its generated from the extension and save its private key unlock state (only exclusive in the Thunderbird extension).
- Detect `needs passphrase` decryption error and if it occurs, then show passphrase through the FlowCrypt browser extension page, unlocks the keys there and redirect user back to the Thunderbird mail extension.

As for now, it can be tested as follow:

1. Build the project from this PR.
2. Setup this PR in a Thunderbird Mail extension and set it up for your account.
3. Open FlowCrypt browser extension page in  Thunderbird and open any encrypted message that requires you to unlock with a passphrase.
4. Once that is done, go back to Thunderbird and open the same email where it should automatically rendered the decrypted message.

extension/js/service_worker/bg-handlers.ts
```diff
@@ -112,4 +113,66 @@ export class BgHandlers {
...
+        const accountId = messageDetails?.folder?.accountId || '';
+        const acctEmail = (await messenger.accounts.get(accountId))?.name || '';
+        await handleClickEvent(tabId, acctEmail, msgId);
+      }
+    });
+  };
+
+  public static thunderbirdContentScriptRegistration = async () => {
+    await messenger.messageDisplayScripts.register({
+      js: [{ file: './js/content_scripts/thunderbird-content-script.js' }],
```
sosnovsky, line 153: after build changes all extension versions (not only thunderbird), include additional `thunderbird-content-script.js` which is almost 5mb:

Screenshot 2024-08-26 at 20 16 37

for browsers we use [injectFcIntoWebmail](https://github.com/FlowCrypt/flowcrypt-browser/blob/a84810e6d1598d1a38a07881a1a98d56427a779f/extension/js/service_worker/inject.ts#L7) method which adds all needed files to each gmail tab.

do you think it'll be possible to adapt this method for thunderbird?

sosnovsky:
Decryption works great, very smoothly, looks like native Thunderbird functionality 👏

By the way, is it possible to remove this warning about  secret key when message was successfully decrypted by FlowCrypt?
Screenshot 2024-08-26 at 20 26 09

sosnovsky:
> Detect needs passphrase decryption error and if it occurs, then show passphrase through the FlowCrypt browser extension page, unlocks the keys there and redirect user back to the Thunderbird mail extension.

This will be a better solution, as storing unlocked private key will make it more vulnerable for possible attacks. Let's go with this one

extension/js/service_worker/bg-handlers.ts
martgil: > after build changes all extension versions (not only thunderbird), include additional thunderbird-content-script.js which is almost 5mb:

Ah, yes - sorry about that. I'll find a way a viable alternative solution to it from which `thunderbird-content-script.js` should only be created for the Thunderbird build.

martgil:
> By the way, is it possible to remove this warning about secret key when message was successfully decrypted by FlowCrypt?

@sosnovsky Unfortunately, I've spend times working on that and that part of the Thunderbird isn't really changable. I mean, it is changable in a way user would be importing custom css file in their machine to get rid of it. Some says, that, it can be removable through custom CSS, like editing and overriding the `userChrome.css` and import it in Thunderbird, so there is that but not programmatically changable in some extent.

> This will be a better solution, as storing unlocked private key will make it more vulnerable for possible attacks. Let's go with this one

Thanks for confirming that. I'll take a look and let you know.

extension/js/service_worker/bg-handlers.ts
sosnovsky: let's start with trying to adapt [injectFcIntoWebmail](https://github.com/FlowCrypt/flowcrypt-browser/blob/a84810e6d1598d1a38a07881a1a98d56427a779f/extension/js/service_worker/inject.ts#L7) method to inject all needed scripts into thunderbird, so it'll work the same in browsers and thunderbird.
and if it doesn't work - then use `thunderbird-content-script.js`

martgil: Hi @sosnovsky - I tried to adapt using injectFcIntoWebmail and there are incompatibility that prevents me from somehow make it work as close as how it works on browser. For example, in Thunderbird, registration for content script is done within `messageDisplayScript.register`, from which it only accept one of the two parameter (file or code). And then several files doesnt seem to work when defined in `.register()` - https://webextension-api.thunderbird.net/en/91/messageDisplayScripts.html.

However, I have done my best to achieve the best result while still complying with the requirements, which I believe I have accomplished in this PR. Please review and test it out. From my perspective, the results are extremely positive, and I'm excited to hear your feedback. :)

Thanks for your help.

martgil:
Hello @sosnovsky - this one is ready for review. Thank you!

sosnovsky:
Hi @martgil, sorry for the long review - I was just checking how script injection works in Thunderbird, added some additional logging and noticed that `injectContentScripts` isn't performed there.

After some investigation I've found that our script didn't get correct list of tabs in Thunderbird, because we expect tab URL to start with `https://mail.google.com`, while in Thunderbird URLs are different.

I'll try to make injection work for Thunderbird, to make it's logic similar to browser extension, so it'll work the same for different platforms.

Or maybe you've already tried to change expected tab URLs for Thunderbird and it didn't work?

martgil:
Hello @sosnovsky - Thank you for thoroughly reviewing this.

>  I was just checking how script injection works in Thunderbird, added some additional logging and noticed that injectContentScripts isn't performed there.

Ah yes, I didn’t change this within injectContentScripts because, as far as I know, content scripts are slightly different in Thunderbird compared to what I understand from its development. In Thunderbird, we use messageDisplayScript and [register](https://webextension-api.thunderbird.net/en/91/messageDisplayScripts.html#register-messagedisplayscriptoptions) it. I’m not entirely sure how exactly content script injection works in Thunderbird, especially since it appears to be applied per tab. I reviewed this during the last review and found that it might not be usable for Thunderbird, given the restrictions I understand.

However, I might have been confused and not fully understood the entire concept of content script injection, including which specific scripts need to be injected.

> After some investigation I've found that our script didn't get correct list of tabs in Thunderbird, because we expect tab URL to start with https://mail.google.com, while in Thunderbird URLs are different.

Yes, to be exact, it is in the imap:// protocol of some sort, if I’m not mistaken from my last check. Origin can be checked via Thunderbird's dev tools if I understood it correctly.

> I'll try to make injection work for Thunderbird, to make it's logic similar to browser extension, so it'll work the same for different platforms.

I understand, Roma. I apologize if I haven’t been reliable on this matter. But I'm here to help if you need me check some other things.

> Or maybe you've already tried to change expected tab URLs for Thunderbird and it didn't work?

I actually do use CSP, but the imap:// protocol doesn’t seem to work with it. For example, `window.open()` is blocked by the `imap://` origin. Unfortunately, it’s been hard to find resources on this. To be clear, the changes and tests I conducted are not related to `injectContentScripts()`.

sosnovsky:
> I understand, Roma. I apologize if I haven’t been reliable on this matter. But I'm here to help if you need me check some other things.

You've done a great job, initially I didn't expect that it'll be possible to show decrypted messages directly in the Thunderbird message UI, but you did and it works well. Current solution is good, I just want to check Thunderbird injection logic to find possibility make it work like on browsers, so we won't need to support 2 different injection approaches in the future.

sosnovsky:
Hi @martgil, it seems `messenger.messageDisplayScripts.register` can register multiple files, so there is no need to compile all js files into `thunderbird-content-script.js`.

I updated `thunderbirdContentScriptRegistration` this way and it injects all needed files:

```ts
public static thunderbirdContentScriptRegistration = async () => {
    const contentScriptGroups = chrome.runtime.getManifest().content_scripts ?? []; // we know it's in the manifest
    const files = contentScriptGroups[0].js?.map(url => url.replace(/moz-extension:\/\/[^/]+\//, './')) ?? [];
    await messenger.messageDisplayScripts.register({
      js: files.map(file => ({ file })),
      css: [{ file: './css/cryptup.css' }],
    });
  };
```

But looks like `sweetalert2.js` has some incompabilities with Thunderbird, as I was receiving `uncaught exception: Object` error in console until I removed `sweetalert2.js` from `manifest.json`. Will check it further.

sosnovsky:
I'll try to find possible fix for using `sweetalert` lib in Thunderbird, so we won't need to remake our popups functionality.
For now, I've pushed my changes based on this PR to [5667-thunderbird-fix](https://github.com/FlowCrypt/flowcrypt-browser/tree/5667-thunderbird-fix) branch

martgil:
> I'll try to find possible fix for using `sweetalert` lib in Thunderbird, so we won't need to remake our popups functionality. For now, I've pushed my changes based on this PR to [5667-thunderbird-fix](https://github.com/FlowCrypt/flowcrypt-browser/tree/5667-thunderbird-fix) branch

I understand - thanks! I'll help you with this after I've finished all the GH issue assigned to me in the enterprise-server repo.

sosnovsky:
> > I'll try to find possible fix for using `sweetalert` lib in Thunderbird, so we won't need to remake our popups functionality. For now, I've pushed my changes based on this PR to [5667-thunderbird-fix](https://github.com/FlowCrypt/flowcrypt-browser/tree/5667-thunderbird-fix) branch
> 
> I understand - thanks! I'll help you with this after I've finished all the GH issue assigned to me in the enterprise-server repo.

Thanks, will keep you updated!

sosnovsky:
finished in https://github.com/FlowCrypt/flowcrypt-browser/pull/5824

martgil:
superseded by https://github.com/FlowCrypt/flowcrypt-browser/pull/5824

PR #5816, opened 2024-08-20
#5615 Fix: change signature when switching different send mail address

### LINKED ISSUE #5615: email signature does not change when switching between different "send mail as identities"
Hello,

we are using latest 8.5.3 on Firefox and we have some Gmail accounts with multiple "Identities" which use their own email signature (under Settings -> Account -> "Send mail as" we have multiple email addresses and they are configured to each use a specific email signature).

Switching between different "from" addresses in the normal Gmail composer updates the email signature in the email body. 

However, this is not happening when using secure compose: I can switch the "from" address, but the signature will stay the same (see screenshot)
image

 I found https://github.com/FlowCrypt/flowcrypt-browser/issues/61 but I'm not sure if this is the same issue ?

kind regards,

ioanmo226:
This PR implemented a fix to change signature when switching different send mail address.

close #5615 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


extension/chrome/elements/compose-modules/compose-sender-module.ts
```diff
@@ -67,5 +67,9 @@ export class ComposeSenderModule extends ViewModule<ComposeView> {
   private actionInputFromChangeHandler = async () => {
     await this.view.recipientsModule.reEvaluateRecipients(this.view.recipientsModule.getValidRecipients());
     this.view.footerModule.onFooterUpdated(await this.view.footerModule.getFooterFromStorage(this.view.senderModule.getSender()));
+    const signaturePattern = /--\s*<br\s*\/?>[\s\S]*$/i;
```
sosnovsky, line 70: this regex works well for new messages, but when you compose reply message and switch send mail address - it removes older messages content too.

need to update this regex to leave older messages text (starting from `On ... wrote:...`)

ioanmo226: Nice catch :+1: 
Implemented a fix and pushed

sosnovsky (commented):
It works better now, but doesn't persist new line between signature and older messages content:

Screenshot 2024-08-21 at 12 06 05

ioanmo226:
Oops!. Nice catch :+1:
Tested with reply/forward/secure compose but didn't try with draft.
Let me fix

sosnovsky (approved):
All good now, thanks!

PR #5820, opened 2024-08-30
#5704 Enable google contact search in firefox

### LINKED ISSUE #5704: Enable Google Contact Search -> Your platform is not supported
image

> Your platform is not supported: browser does not support extension windows

nothing shows in console

`Version 8.5.4`

ioanmo226:
This PR fixed enable google contact search issue in firefox

close #5704 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/js/common/browser/env.ts
+++ b/extension/js/common/browser/env.ts
@@ -41,7 +41,12 @@ export class Env {
   }
 
   // Check if the current context is a Service Worker
-  public static isBackgroundPage() {
+  public static async isBackgroundPage() {
+    // In firefox, window&document is not null even in background page
+    if (typeof browser !== 'undefined' && typeof browser.runtime?.getBackgroundPage !== 'undefined') {
+      const backgroundPage = await browser.runtime.getBackgroundPage();
+      return backgroundPage === window;
+    }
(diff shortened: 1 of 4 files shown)


extension/js/common/browser/env.ts
```diff
@@ -41,7 +41,12 @@ export class Env {
   }
 
   // Check if the current context is a Service Worker
-  public static isBackgroundPage() {
+  public static async isBackgroundPage() {
+    // In firefox, window&document is not null even in background page
+    if (typeof browser !== 'undefined' && typeof browser.runtime?.getBackgroundPage !== 'undefined') {
```
sosnovsky, line 46: this check probably can be simplified to just `if (browser?.runtime?.getBackgroundPage !== undefined)`

sosnovsky (approved):
Works well now, thanks!

extension/js/common/browser/env.ts
ioanmo226: It would throw `Uncaught ReferenceError: browser is not defined` in chrome.

sosnovsky: Ah, I see, then all good 👍

Issue #5668, opened 2024-04-10
Recognize and support decrypting and downloading encrypted attachments


martgil:

martgil:
For me: Explore option to recognize, render, decrypt and download encrypted attachments. Signed messages with detached signature should also be covered in this issue.

PR #5828, opened 2024-09-12
#5668 Add Thunderbird attachment decryption

martgil:
This PR adds Thunderbird port the ability to render, recognize attachment and download decrypted attachment (when possible).  

#5668

- Difficult to test - Pathways for performing tests in Thunderbird it not yet know. 

### To be filled by reviewers


martgil:
There is some limitation I've encountered for attachment's from which we tried to decrypt and render in case they match any of encrypteMsgNames here: https://github.com/FlowCrypt/flowcrypt-browser/blob/a6c8f5919efcaacbc68545642b1f31a6e6918617/extension/js/common/core/attachment.ts#L43.

Problem is that, the Thunderbird client tries to decrypt the attachment named of those before the extension resulting to conflict and irretrievable attachments so we cant proceed to do decryption for ourselves.

cc: @sosnovsky 

I'll check if this is the same case for detached signature.

sosnovsky:
> Problem is that, the Thunderbird client tries to decrypt the attachment named of those before the extension resulting to conflict and irretrievable attachments so we cant proceed to do decryption for ourselves.

So there is no way to get original encrypted data from these attachments?
Maybe we can use Gmail API for fetching attachment data by it's id?

martgil:
> > Problem is that, the Thunderbird client tries to decrypt the attachment named of those before the extension resulting to conflict and irretrievable attachments so we cant proceed to do decryption for ourselves.
> 
> So there is no way to get original encrypted data from these attachments? 

I have explored the issue further. The attachment object gets only accessible if the attachment were actually decrypted by thunderbird, that's the only chance the extension can access the attachment object if they we're already decrypted by Thunderbird itself. I had accessed the attachment after I imported my private key into my thunderbird's account settings.

We can somehow add this as one of the know limitation of the FlowCrypt for Thunderbird extension - what do you think?

> Maybe we can use Gmail API for fetching attachment data by it's id?

Unfortunately, that's the problem. We wont be able to get any information about attachment without the thunderbird decrypt the message.asc first.

sosnovsky:
Thanks for exploration, let's use your current solution with private key import and maybe later we'll find some way to make it work directly in FlowCrypt extension.

martgil:
Hi @sosnovsky - I'd like to report my progress in this PR. For testing, I've prepared every test cases in flowcrypt.compatibility account that we'll be using to test the implemented features in Thunderbird build.

Below are the test cases that needs to be tested. The thunderbird tests are all compiled in the "thunderbird_client_test"

Needs to be manually tested are (for reviewer):

- [ ] Attachment decryption
- [ ] Render signature verification for detached signed messages. (This has very strange issue wherein in the extension, verifying this same detached signed message wont work but works on openpgp.js signature verification. I do not re-invent the wheel by any means as I only used existing functions.).
- [ ] Attachment decryption for detected encrytedMsg as an attachments: https://github.com/FlowCrypt/flowcrypt-browser/blob/a6c8f5919efcaacbc68545642b1f31a6e6918617/extension/js/common/core/attachment.ts#L43-L51
For this case, the FlowCrypt from private key must be loaded into the Thunderbird client for the client to let us access the encryptedMsg attachment. To do this, export the private keys from the FlowCrypt browser extension by accessing it through the Thunderbird. Then once the private keys are downloaded, import it to Thunderbird by going to **Thunderbird's Account Settings** -> **End-to-end encryption** -> import keys.

martgil:
As of now, the only one I have problems with in-general is the signature verification for detached signatures.

This one fails on FlowCrypt thunderbird version but not on web browser version -- verifying the same signed message on openpgp.js raw code works as it passes the verification:
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/KtbxLxgGDrtCHDtsVZWgJFtGfFkcSjGtGV

Meanwhile, this works on thunderbird but not on web -- the `rawSignedContent` seemed to have significant amount of data from raw mimeContent instead of just the signed text hence the signature verification passes. In comparison, I'm not sure why my verification result returns empty "signerLongids" despite i've already imported the correct public key into the FlowCrypt on my thunderbird:
https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#label/thunderbird_client_test/FMfcgzGlkPTsgVjRjdGFZpxRqHhqGrQW

extension/js/service_worker/bg-handlers.ts
```diff
@@ -162,6 +163,57 @@ export class BgHandlers {
...
   };
 
+  public static thunderbirdGetDownloadableAttachment = async (): Promise<Bm.Res.ThunderbirdGetDownloadableAttachment> => {
+    const processableAttachments: ThunderbirdAttachment[] = [];
+    const [tab] = await messenger.mailTabs.query({ active: true, currentWindow: true });
+    const message = await messenger.messageDisplay.getDisplayedMessage(tab.id);
+    let from = '';
+    if (tab.id && message?.id) {
+      from = Str.parseEmail(message.author).email || '';
+      const mimeMsg = await messenger.messages.getFull(message.id);
```
sosnovsky, line 173: Recently Thunderbird introduced new param `decrypt` which enables/disables message decryption for full message fetch - https://webextension-api.thunderbird.net/en/128-esr-mv2/changes/128.html#id12.

Probably setting it to `false` should help to handle case described in https://github.com/FlowCrypt/flowcrypt-browser/pull/5828#issuecomment-2422154817

extension/js/content_scripts/webmail/thunderbird/thunderbird-element-replacer.ts
```diff
@@ -19,71 +21,112 @@ export class ThunderbirdElementReplacer extends WebmailElementReplacer {
...
+    return [{ interval: 2000, handler: () => this.handleThunderbirdMessageParsing() }];
   };
 
-  public replaceThunderbirdMsgPane = async () => {
+  public handleThunderbirdMessageParsing = async () => {
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    this.acctEmail = (await BrowserMsg.send.bg.await.thunderbirdGetCurrentUser())!;
+    const emailBodyToParse = $('div.moz-text-plain').text().trim() || $('div.moz-text-html').text().trim() || $('div.moz-text-flowed').text().trim();
+    const { processableAttachments: fcAttachments, from: signerEmail } = await BrowserMsg.send.bg.await.thunderbirdGetDownloadableAttachment();
     if (Catch.isThunderbirdMail()) {
```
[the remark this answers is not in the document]
martgil: Thanks - I'll remove unneeded checks.

extension/js/service_worker/bg-handlers.ts
martgil: Thanks @sosnovsky - After trying to add the `decrypt` option set to `false`, it doesn't allow me to access the encrypted "message.asc" though I was really hoping that to work.

Upon reviewing, the main issue I think relies on listAttachments() as in that part the error raises when I did try another debugging. This leads me to wrote a support help via Thunderbird forum - https://support.mozilla.org/en-US/questions/1474480.

I'll further check if there's any workaround I can share with you.

sosnovsky: What about using [messages.getRaw](https://webextension-api.thunderbird.net/en/latest/messages.html#getraw-messageid-options) for this case?
I tried to run `getRaw` with `decrypt: false` param and it returns full content of `message.asc` attachment.

martgil: Thanks @sosnovsky - you are right. the `.getRaw()` seemed to be the way to go. I tried to use "mailparser" but I can't get it to work in background script. I mean, I have settled the issue in typescript types through content script tsconfig.json but the usage for the actual js does not work. I think the 'mailparser' js in not injected in content script so I'm getting undefined when calling `simpleParser`: https://nodemailer.com/extras/mailparser/#simpleparser

Sorry, do you know how I could check if mailparser's simpleParser gets injected into TB?

martgil: Please kindly ignore my last question regarding "mailparser", im doing it incorrectly. I'm not trying different approach with PostalMime which said to be usable for webpack compiled project: https://www.npmjs.com/package/postal-mime

martgil: Hi @sosnovsky, after quite some time trying, I didn't find a way to use the ['postal-mime'](https://www.npmjs.com/package/postal-mime) module even after I tried to bundle it from source using webpack.

I am think that's the best module that can assist us with the current code I have and for the situation given where attachments will be acessible that way.

sosnovsky: Hi @martgil, I tried to check this functionality, but seems like FlowCrypt functionality isn't working in the latest Thunderbird release 128.5.0 - encrypted messages are not decrypted there.
Can you please check if it works for you?

martgil: Hello @sosnovsky, the extension does not work on 128.5.0 for the rest of the features involves content-script as im testing it :( The errors happens in content scripts in Thunderbird really lacks of context, and console will just says "uncaught exception: Object" with no way to review the Object details.

It's the same exact error when I tried to setup the content scripts in decryption process.

martgil: Could be but I tested some other builds where I used in my development phase exactly using Thunderbird ESR 128 4.x.x but not it does not work now too.. I'll be rechecking it, again it should be some result of the other PRs possibly.

martgil: This PR might've change a few related to content scripts https://github.com/FlowCrypt/flowcrypt-browser/commit/aa27485fa1229beeb255b426d6ec9ef20c807b73

extension/js/content_scripts/webmail/thunderbird/thunderbird-element-replacer.ts
```diff
@@ -98,48 +141,66 @@ export class ThunderbirdElementReplacer extends WebmailElementReplacer {
...
+      .append($('<img>').attr('src', messenger.runtime.getURL('/img/svgs/download-link.svg'))); // xss-safe-value
+    attachmentHtmlRoot.append(attachmentFileTypeIcon); // xss-escaped
+    attachmentHtmlRoot.append(attachmentFilename); // xss-safe-value
+    attachmentHtmlRoot.append(attachmentDownloadBtn); // xss-safe-value
+    return attachmentHtmlRoot;
   };
 
-  private resemblesCleartextMsg = (body: string) => {
+  private downloadThunderbirdAttachmentHandler = async (decryptedFileName: string, encryptedData: Buf) => {
+    if (encryptedData) {
```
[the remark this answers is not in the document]
martgil: agreed, i've removed this checks.

extension/js/service_worker/bg-handlers.ts
martgil: Unfortunately, this is difficult to debug really because of the lack of context from Thunderbird so I'm experiecing the issue were I receieve "Unexpected exception: Object" exactly as https://github.com/FlowCrypt/flowcrypt-browser/issues/5796#issuecomment-2328302980 where it occurs within the process of injecting content script, but its way deeper perhaps because one of the content script there has issue. possibly "streams_web.js".

martgil: Hi @sosnovsky, our speculation regarding content script injection might have been correct, but we can’t guarantee it due to the lack of context in the error message. I’m concerned about this PR and the reliability of the Thunderbird build at this point, as even a minor issue could result in an unusable extension. The functionality will continue to work once the content script issue is resolved.

I'd like to fix the issue as much as I'd like since this almost completed the requirements for attachment decryption purpose solely. Perhaps after we've found and fix the content script issue, we can later then work on other improvements on attachments supports even further.

Do you have any other suggestion or idea for us to completely fix the content script issue based on the recently merged branches?

sosnovsky: Hi @martgil, in mentioned PR Ioan made 2 separate builds for `web-stream-tools` lib -`streams_web.js` is used in browser, while `streams_common.js` is used in tests, because of some import issues, as I remember.

Thunderbird errors can be caused by similar issue, let's try to use `streams_common.js` in Thunderbird build.

martgil: @sosnovsky I've tried several things including using `streams_common.js` in Thunderbird build and it doesn't work and I'm still getting the generic error `uncaught exception: Object` from the Thunderbird email client.

martgil: Hi @sosnovsky sorry for pinging you so often. I still really need help for this one. Also, I'm now started to wonder that this issue in Thunderbird build will experience the same issue more often if it still in the same branch like in the main.

I'm thinking of ways to solve that as well so that it won't be affected in case of any unexpected change from the content script. Do you think it will be solve by having it a separate branch of its own?

I'm a bit stuck with this PR, I can't fix it on my own. Sorry about that.

sosnovsky: Hi @martgil, sorry for the late reply!

On this week we've had a high priority task - to add password-protected messages compliance for all FlowCrypt apps (https://github.com/FlowCrypt/flowcrypt-browser/pull/5879), so I didn't have enough time to deeply check this issue with Thunderbird content scripts.

> I'm thinking of ways to solve that as well so that it won't be affected in case of any unexpected change from the content script. Do you think it will be solve by having it a separate branch of its own?

I think it’s possible to fix this and maintain a single `main` branch, as managing a separate branch for Thunderbird builds could become more complicated in the future. I’ll give it a try next week and let you know if I find a way to resolve the issue.

martgil: Hello Roma, no worries - I understand completely.

> On this week we've had a high priority task - to add password-protected messages compliance for all FlowCrypt apps (#5879), so I didn't have enough time to deeply check this issue with Thunderbird content scripts.

I understand. If there is anything I can do to try and fix this, please feel free to provide instructions, and I'll follow them.

> I think it’s possible to fix this and maintain a single main branch, as managing a separate branch for Thunderbird builds could become more complicated in the future. I’ll give it a try next week and let you know if I find a way to resolve the issue.

martgil: Hello @sosnovsky,

I hope you’re doing well. I just wanted to follow up and kindly ask if there have been any updates on this. Have you been able to find a solution for the content script issue?

Thank you for your time and help!

sosnovsky:
Hi @ioanmo226, earlier @martgil started working on Thunderbird version of FlowCrypt extension and it was able to successfully decrypt messages. But some Thunderbird changes between releases broke FlowCrypt implementation and message decryption fails because of some generic `Unexpected exception: Object` error.

Can you please identify the origin of this error and fix the message decryption in Thunderbird.
It can be done in this PR, where Mart started working on attachment decryption, or in a separate PR.

Feel free to reach out to me or Mart if you have any Thunderbird implementation questions.

martgil:
@sosnovsky - thanks for the heads up on this. hopefully we can recover this one.

ioanatflowcrypt:
Hi @sosnovsky,

I’ve dug deeper into the Thunderbird issue and pinpointed the root cause.
After our upgrade to OpenPGP.js v6, Thunderbird is throwing an `Unexpected exception: Object` error because:

- **OpenPGP.js v6 drops its pure-JS fallback.** It now immediately throws unless it can find a working `crypto.subtle`.  
- **Thunderbird’s `messageDisplayScripts` context isn’t a secure origin**, so `window.crypto.subtle` (and `ReadableStream`) are unavailable, triggering the exception.  
- To trace this, I wrapped every library call in try/catch and commented them out one by one, due to Thunderbird’s not proper error reporting.

> You’ll recall that on **iOS** we worked around the same restriction by serving our WebView bundle through a local HTTP server—this forces a “secure” context so that `crypto.subtle` is defined.

---

### Potential solutions

1. **Bundle polyfills into `webmail_bundle.js`**  
   - Prepend UMD/ponyfill builds of Web Streams and WebCrypto in our build script.  

2. **Move crypto to the background script**  
   - Background/service-worker contexts in Thunderbird are secure and expose the native WebCrypto API. We could proxy all crypto calls there—but it would require significant refactoring.

4. **Add a Node-crypto fallback via Webpack**  

5. **Revert to OpenPGP.js v5**  
   - As a last resort, roll back to v5 and re-enable its pure-JS fallback:

---

Let me know which approach you’d prefer, or if you have other ideas!

sosnovsky:
Hi @ioanatflowcrypt, thanks for checking it! Let's use a simpler solution here, since Thunderbird isn't a high-priority platform. Adding a polyfill specifically for the Thunderbird build should be ok.

PR #5837, opened 2024-09-18
#5831 Remove chrome.tabs.onUpdated listener in firefox

### LINKED ISSUE #5831: Check if chrome.tabs.onUpdated still needed for Firefox scripts injection
We use [chrome.tabs.onUpdated](https://github.com/FlowCrypt/flowcrypt-browser/blob/ac59d1f0f7e13aee66d37c2c2e704d91f96b3641/extension/js/service_worker/inject.ts#L20) listener for injecting content script to Firefox. It was added several years ago because standard way of scripts loading didn't work in Firefox.

Need to check if this listener can be removed and we can use our standard way of script injection in Firefox too.

ioanmo226:
This PR removed chrome.tabs.onUpdated listener in firefox

close #5831 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/js/service_worker/inject.ts
+++ b/extension/js/service_worker/inject.ts
@@ -2,8 +2,6 @@
 
 'use strict';
 
-import { Catch } from '../common/platform/catch.js';
-
 export const injectFcIntoWebmail = () => {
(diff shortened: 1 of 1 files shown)


ioanmo226:
Seems to work well in firefox even after removing chrome.tabs.onUpdated listener

PR #5845, opened 2024-09-26
#5838 Added checks for RichText() before initializing squire in Secure Reply message box

### LINKED ISSUE #5838: Incorrect newlines when composing reply message
- go to `flowcrypt.compatibility` inbox
- open `#5466 unencrypted reply` thread
- expand first message
- click `secure reply` on first message toolbar
- type "Hello<kbd>Enter</kbd>World"
- you'll get
```
World
Hello
```

@martgil you already fixed same issue in https://github.com/FlowCrypt/flowcrypt-browser/pull/5729, maybe similar fix will work here too.

martgil:
This PR fixes squire initialization for secure compose message box where newlines gets injected above the current line when `Enter` is pressed.

close #5838 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/compose-input-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-input-module.ts
@@ -40,7 +40,9 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
       'click',
       this.view.setHandler(el => this.actionAddIntroHandler(el), this.view.errModule.handle(`add intro`))
     );
-    this.initSquire(this.isRichText());
+    if (this.isRichText()) {
+      this.initSquire(true);
+    }
     // Set lastDraftBody to current empty squire content ex: <div><br></div>)
     // https://github.com/FlowCrypt/flowcrypt-browser/issues/5184
     this.view.draftModule.setLastDraftBody(this.squire.getHTML());
--- a/extension/chrome/elements/compose-modules/compose-input-module.ts
+++ b/extension/chrome/elements/compose-modules/compose-input-module.ts
@@ -36,13 +36,11 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
   }
 
   public setHandlers = () => {
+    this.toggleRichTextFormatting();
     this.view.S.cached('add_intro').on(
       'click',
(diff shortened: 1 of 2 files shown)


martgil:
Hi @sosnovsky - this one is ready for review. Thank you!

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -40,7 +40,9 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
       'click',
       this.view.setHandler(el => this.actionAddIntroHandler(el), this.view.errModule.handle(`add intro`))
     );
-    this.initSquire(this.isRichText());
+    if (this.isRichText()) {
+      this.initSquire(true);
+    }
```
sosnovsky, line 45: in this case Squire probably won't be re-initialized when user turns off rich text mode?

I think it'll be better to use `removeRichTextFormatting` here, which checks if `isRichText` is enabled before re-initializing squire:
```ts
this.isRichText() ? this.addRichTextFormatting() : this.removeRichTextFormatting();
```

martgil: Hi @sosnovsky - I missed this scenario but I have corrected it now.

martgil:
Hi @sosnovsky - Greetings. May I request for another review. thank you!

extension/chrome/elements/compose-modules/compose-input-module.ts
```diff
@@ -36,11 +36,11 @@ export class ComposeInputModule extends ViewModule<ComposeView> {
   }
 
   public setHandlers = () => {
+    this.toggleRichTextFormatting();
```
sosnovsky, line 39: `toggleRichTextFormatting` sounds like we change rich text property here, but we just configure textarea according to the existing property value. 
`configureRichTextFormatting` is more correct naming, as for me.

martgil: Ah, I see. I agree with your suggestion and have changed the name as recommended. Thanks!

sosnovsky (approved):
All good now, thanks!

martgil:
thank you, too.

PR #5855, opened 2024-10-27
# 5576 Add TrustedHTML flag in DomPurify

### LINKED ISSUE #5576: Use Trusted Types in DOMPurify
I think it is also a good idea to update some of the few options for the DOMPurify configuration option to return trusted type by adding the `RETURN_TRUSTED_TYPE` property set to true when `DOMPurify.sanitize()` is used:

https://github.com/FlowCrypt/flowcrypt-browser/blob/a40396649d6670f3c9c1ecbc66486d56d4a9a20e/extension/types/purify.d.ts#L18

Reference:
https://github.com/cure53/DOMPurify?tab=readme-ov-file#what-about-dompurify-and-trusted-types
https://web.dev/articles/trusted-types#use_a_library

_Originally posted by @martgil in https://github.com/FlowCrypt/flowcrypt-browser/pull/5575#pullrequestreview-1838071922_

martgil:
This PR added `RETURN_TRUSTED_TYPE` flag in `DomPurfify.sanitize()` applicable functions.

close #5576

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/pgp_block_modules/pgp-block-print-module.ts
+++ b/extension/chrome/elements/pgp_block_modules/pgp-block-print-module.ts
@@ -2,6 +2,7 @@
 
 'use strict';
 
+import { TrustedHTML, TrustedTypesWindow } from 'types/trusted-types.js';
 import { Time } from '../../../js/common/browser/time.js';
 import { Catch } from '../../../js/common/platform/catch.js';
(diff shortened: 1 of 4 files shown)


martgil:
browser support and some javascript libraries do not work support TrustedTypes - https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API

PR #5873, opened 2024-11-25
#5863 Use built in dompurify type

### LINKED ISSUE #5863: Use built-in DOMPurify types
Currently we have custom DOMPurify types (https://github.com/FlowCrypt/flowcrypt-browser/blob/b8aa4bebd85f33d2fd15b3682db5da8d4a74002b/extension/types/purify.d.ts), which are probably from some older version and can be outdated. DOMPurify provides built-in types for the actual version - https://github.com/cure53/DOMPurify/blob/main/dist/purify.cjs.d.ts, let's switch to using it.

ioanmo226:
This PR used built in dompurify type

close #5863 // if this PR closes an issue

### To be filled by reviewers


ioanmo226:
Hmm...  
For the `decrypt - not an armored public key in a file that looked like a public key` test, any operations on `attachmentFrame` don't seem to work at all for some reason.  
Even `attachmentFrame.waitForAny('@download-attachment')` doesn't respond.  
Interestingly, it works fine for the PGP block frame.  
FYI this only happens on gmail mock page.
Can't figure out the reason yet.  
Still digging.


https://github.com/FlowCrypt/flowcrypt-browser/blob/aa27485fa1229beeb255b426d6ec9ef20c807b73/test/source/tests/decrypt.ts#L1636

ioanmo226:
PS: happens on master too

sosnovsky:
Maybe was caused by puppeteer update (or underlying chrome version)?

ioanmo226:
Maybe because of underlying chrome version?

sosnovsky:
Yes, recently puppeteer switched to chrome v131, you can try to rollback puppeteer version to check if test passes with older chrome

ioanmo226:
Hmm... Still fails with older puppeteer version.
Still far, not sure what the reason is.

sosnovsky:
> It fails randomly, succeeding only 1-3 times out of 5 tests.
> The exact cause is unclear. Should we consider skipping this test for stability until we identify the root cause?

Sorry, missed it - yes, we can temporarily disable it in the another PR

PR #5876, opened 2024-12-02
#5864 Use built in linkify html type

### LINKED ISSUE #5864: Use built-in linkify-html types
Instead of https://github.com/FlowCrypt/flowcrypt-browser/blob/b8aa4bebd85f33d2fd15b3682db5da8d4a74002b/extension/types/linkifyHtml.d.ts let's use built-in `linkify-html` types.

ioanmo226:
This PR uses built in linkify html type

close #5864 // if this PR closes an issue

### To be filled by reviewers


package.json
```diff
@@ -12,6 +12,7 @@
     "@types/chrome": "0.0.284",
     "@types/dompurify": "3.2.0",
     "@types/jquery": "3.5.32",
+    "@types/linkifyjs": "^2.1.7",
```
sosnovsky, line 15: here we don't need separate types package, as starting from v3 `linkifyjs` includes ts types in the main package - https://github.com/nfrasser/linkifyjs/releases/tag/v3.0.0, they're available at `node_modules/linkifyjs/dist/` or `node_modules/linkify-html/dist/`

ioanmo226: Yep. you are right!
Fixed

sosnovsky (dismissed):
Works well 👍

sosnovsky (approved):
👍

PR #5879, opened 2024-12-06
#5878 Password protected messages compliance

### LINKED ISSUE #5878: Password-protected messages compliance
To ensure password-protected messages comply with security rules, client apps must check the subjects of composed password-protected email messages for specified strings. If any configured terms are found, the application should display an error message indicating that the password encryption method is incompatible with the composed message.

Client configuration will have 2 new optional properties (subject check should be performed only if both properties are set):
- `disallow_password_messages_for_terms` (_string[]_) - specifies an array of strings to check against the subject of the composed password-protected message. If any string in this array is found in the subject, an error alert must be displayed.
- `disallow_password_messages_error_text` (_string_) - defines the text to be displayed in the error alert.

related https://github.com/FlowCrypt/flowcrypt-ios/issues/2644

ioanmo226:
This PR implemented password protected messages compliance

close #5878 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


test/source/tests/compose.ts
```diff
@@ -236,6 +236,42 @@ export const defineComposeTests = (testVariant: TestVariant, testWithBrowser: Te
       })
     );
 
+    test(
+      'user@disallow-password-message-terms.flowcrypt.test - disallow password  protected message terms',
+      testWithBrowser(async (t, browser) => {
+        const acct = 'user@disallow-password-message-terms.flowcrypt.test';
+        const rules = getKeyManagerAutogenRules(t.context.urls!.port!);
+        const disallowedPasswordMessageErrorText = 'Password-protected messages are disabled';
```
sosnovsky, line 244: let's add support for links here too, forgot to mention it in the issue description, thanks!

ioanmo226: Will the error text only contain links that the app should automatically detect and replace with <a> tags?
Or will it already include HTML tags (like <a href="...">)?

sosnovsky: We should auto-detect links, probably using `linkifyjs` will make it simple

extension/chrome/elements/compose-modules/compose-err-module.ts
```diff
@@ -160,6 +161,15 @@ export class ComposeErrModule extends ViewModule<ComposeView> {
...
 
   public throwIfEncryptionPasswordInvalidOrDisabled = async ({ subject, pwd }: { subject: string; pwd?: string }) => {
+    const disallowedPasswordMessageTerms = this.view.clientConfiguration.getDisallowPasswordMessagesForTerms();
+    const disallowedPasswordMessageErrorText = this.view.clientConfiguration.getDisallowPasswordMessagesErrorText();
+    if (
+      disallowedPasswordMessageErrorText &&
+      disallowedPasswordMessageTerms &&
+      subject.split(' ').some(term => disallowedPasswordMessageTerms.includes(term))
+    ) {
+      throw new ComposerUserError(linkifyHtml(disallowedPasswordMessageErrorText));
```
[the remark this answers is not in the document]
ioanmo226: Sounds good. Just fixed

extension/js/common/client-configuration.ts
```diff
@@ -110,6 +112,21 @@ export class ClientConfiguration {
     return this.clientConfigurationJson.enforce_keygen_expire_months;
   };
 
+  /**
+   * An array of strings to check against the subject of the composed password-protected message.
+   * If any string in this array is found in the subject, an error alert must be displayed.
+   */
+  public getDisallowPasswordMessagesForTerms = (): string[] | undefined => {
+    return this.clientConfigurationJson.disallow_password_messages_for_terms;
```
sosnovsky, line 120: let's also filter non-empty strings here, to avoid some unexpected behavior (as noticed by Den at https://github.com/FlowCrypt/flowcrypt-android/issues/2905#issuecomment-2528260144)

extension/chrome/elements/compose-modules/compose-err-module.ts
```diff
@@ -160,6 +161,15 @@ export class ComposeErrModule extends ViewModule<ComposeView> {
   };
 
   public throwIfEncryptionPasswordInvalidOrDisabled = async ({ subject, pwd }: { subject: string; pwd?: string }) => {
+    const disallowedPasswordMessageTerms = this.view.clientConfiguration.getDisallowPasswordMessagesForTerms();
+    const disallowedPasswordMessageErrorText = this.view.clientConfiguration.getDisallowPasswordMessagesErrorText();
+    if (
+      disallowedPasswordMessageErrorText &&
+      disallowedPasswordMessageTerms &&
+      subject.split(' ').some(term => disallowedPasswordMessageTerms.includes(term))
```
sosnovsky, line 169: I missed that here we search only for separate words, as it's possible to have multiple words in a single term (like `internal plan`), and it'll also fail for cases when word is surrounded by brackets like `[INTERNAL]`.
So let's search for terms in full subject string.

sosnovsky: We've also got a bit more details from customer (https://github.com/FlowCrypt/flowcrypt-android/issues/2905#issuecomment-2529418537):
- search should be case-insensitive
- possible string values are `['ABDU', '[Classification: Data Control: Internal Data Control]']`

extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -306,6 +306,33 @@ export class MsgUtil {
     return diagnosis;
   }
 
+  public static isPasswordMesageEnabled(subject: string, disallowTerms: string[]) {
```
sosnovsky, line 309: `isPasswordMesageEnabled` -> `isPasswordMessageEnabled`

extension/js/common/core/crypto/pgp/msg-util.ts
```diff
@@ -306,6 +306,33 @@ export class MsgUtil {
...
+      // Check for exact matches (full-term match within the subject)
+      if (
+        lowerCaseSubject.includes(lowerCaseTerm) &&
+        (lowerCaseSubject === lowerCaseTerm ||
+          lowerCaseSubject.startsWith(lowerCaseTerm + ' ') ||
+          lowerCaseSubject.endsWith(' ' + lowerCaseTerm) ||
+          lowerCaseSubject.includes(' ' + lowerCaseTerm + ' '))
+      ) {
+        return false;
+      }
```
sosnovsky, line 330: disallowed terms can be surrounded not only by spaces.
for example, term can be `internal data` and subject string `[internal data] year results` should trigger error message, as it contains full `internal data` string.
probably using regex should make it simpler to detect such cases

test/source/tests/unit-node.ts
```diff
@@ -904,6 +904,30 @@ ${testConstants.smimeCert}`),
...
+        'Conference information [Classification: Data Control: Internal Data Control]': false,
+        'Classification: Data Control: Internal Data Control - Tomorrow meeting': true,
+        'Internal Data Control - Finance monitoring': true,
+        // term check should work only for exact matches - if we have droid in the list of strings,
+        // password-protected messages shouldn't be disabled for subjects with Android word
+        'Android phone update': true,
+        'droid phone': false,
+        // Check for case insensitive
+        'DROiD phone': false,
+      };
```
sosnovsky, line 921: well done with this test 👍
let's also add mentioned case when disallowed term is surrounded by brackets

extension/js/common/core/crypto/pgp/msg-util.ts
ioanmo226: Nice catch 👍 
Let me take a look

ioanmo226:
@sosnovsky Please check current implementation

sosnovsky (approved):
Works great now and well-tested, thanks!

PR #5884, opened 2024-12-20
Release 8.5.8

sosnovsky:
This PR updates version to 8.5.8 and adds changelog

### To be filled by reviewers

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "flowcrypt-browser",
-  "version": "8.5.7",
+  "version": "8.5.8",
   "description": "Simple end-to-end encryption to secure email and attachments on Google.",
   "resolutions": {
     "graceful-fs": "4.1.13"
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -37,7 +37,7 @@
     "typescript": "5.7.2",
     "typescript-eslint": "^8.18.0",
     "undici-types": "^7.0.0",
-    "web-ext": "8.3.0",
+    "web-ext": "7.12.0",
```
ioanmo226, line 40: Sorry but could you tell me the reason we have to downgrade web-ext?

sosnovsky: Yeah, our build server currently uses nodejs v16 which doesn't has built-in `Blob` support, but it's needed for `web-ext v8`. So I've temporarily downgraded it to v7 for current release.

Before next release I'll check if we should use some `blob-polyfill` or maybe we'll update node on build server to the newer version.

ioanmo226: Aha, i see
Thank you for your explanation.

ioanmo226 (approved):

PR #5887, opened 2024-12-24
#4953 Use test replaceable port

### LINKED ISSUE #4953: stop adding then replacing port 8001 in mock builds
I suppose it could.. we would still need to do two rounds of replacing? The step that normally replaces `[BUILD_REPLACEABLE_....]` and currently puts the `:8001` would still need to insert some more maybe `[TEST_REPLACEABLE_MOCK_PORT]` and then that would be replaced here.

It's true that as things are now, there is no meaning to the `:8001` in particular. So that should not be put it, to not be confusing. So `[TEST_REPLACEABLE_MOCK_PORT]` or similar would be a better choice here.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/pull/4928#discussion_r1104285058_

ioanmo226:
This PR used test replaceable port

close #4953 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


tooling/build-types-and-manifests.ts
```diff
@@ -1,6 +1,7 @@
 /* ©️ 2016 - present FlowCrypt a.s. Limitations apply. Contact human@flowcrypt.com */
 import { readFileSync, writeFileSync } from 'fs';
 import { execSync as exec } from 'child_process';
+const MOCK_PORT = '[TEST_REPLACEABLE_MOCK_PORT]';
```
ioanmo226, line 4: We can't import the `MOCK_PORT` variable here as we do in other places (e.g., `import { MOCK_PORT } from './const.js';`) because this is a separate TypeScript project. 

To address this, I inserted `[TEST_REPLACEABLE_MOCK_PORT]` and replaced it later in `config-mock-build.sh`.

However, I feel that the current implementation has made the code a bit more complex. Let me know if this aligns with your expectations or if you have any other suggestions.

ioanmo226: @sosnovsky

test/source/mock/fes/customer-url-fes-endpoints.ts
```diff
@@ -80,7 +80,7 @@ export const getMockCustomerUrlFesEndpoints = (config: FesConfig | undefined): H
     '/api/v1/message/FES-MOCK-EXTERNAL-ID/gateway': async ({ body }, req) => {
       const port = parsePort(req);
       if (parseAuthority(req) === standardFesUrl(port) && req.method === 'POST') {
-        // test: `compose - user@standardsubdomainfes.localhost:8001 - PWD encrypted message with FES web portal`
+        // test: `compose - user@standardsubdomainfes.localhost:{port} - PWD encrypted message with FES web portal`
```
sosnovsky, line 83: let's just remove `:{port}` from tests names, so it'll be `compose - user@standardsubdomainfes.localhost - PWD encrypted message with FES web portal`, same for other tests

test/source/tests/browser-unit-tests/unit-Sks.js
```diff
@@ -23,7 +23,7 @@
 BROWSER_UNIT_TEST_NAME(`Sks lookup pubkey - trailing slash`);
 (async () => {
   const email = 'john.doe@example.com';
-  const sks = new Sks('https://localhost:8001/');
+  const sks = new Sks(`https://localhost:${MOCK_PORT}/`);
```
sosnovsky, line 26: for unit tests we should leave fixed `:8001`, as different ports are used only in ui tests, so we can run multiple ui tests on different localhost ports

tooling/build-types-and-manifests.ts
sosnovsky: Yes, using separate `MOCK_PORT` variable adds some complexity, but it also removes a bit of confusion from generated original mock code where added `:8001` should be replaced later again.
Your implementation looks good, let's continue with this approach.
Thanks for asking!

test/source/mock/fes/customer-url-fes-endpoints.ts
ioanmo226: Sounds good!

ioanmo226:
Hmm.. for some reason, tests fail and some cases tests succeed.
Might be because of using global MOCK_PORT variable. not sure yet though.
Will dig more

ioanmo226:
Hmm.. It now succeesed everytime.

ioanmo226:
Enterprise mock tests(15 tests) fail after 4 consequent succees checks.
Error message says can't get extension id. Not sure why that happens sometimes.

ioanmo226:
Digging more

ioanmo226:
Seems like above one fixed issue.
Please review

test/source/test.ts
```diff
@@ -106,7 +107,11 @@ const startMockApiAndCopyBuild = async (t: AvaContext) => {
   const address = mockApi.server.address();
   if (typeof address === 'object' && address) {
     const result = await asyncExec(`sh ./scripts/config-mock-build.sh ${buildDir} ${address.port}`);
+    const filepath = `${buildDir}/js/common/core/const.js`;
 
+    const fileContent = await readFile(filepath, { encoding: 'utf-8' });
+    const updatedCode = fileContent.replace(/const (MOCK_PORT) = [^;]+;/g, `const $1 = '${address.port}';`);
+    await writeFile(filepath, updatedCode);
```
sosnovsky, line 114: probably we can perform `MOCK_PORT` replacement in `config-mock-build.sh` script, like we do for `TEST_REPLACEABLE_MOCK_PORT`?

extension/js/common/core/common.ts
```diff
@@ -159,7 +160,7 @@ export class Str {
     if (email.includes(' ')) {
       return false;
     }
-    email = email.replace(/\:8001$/, ''); // for MOCK tests, todo: remove from production
+    email = email.replace(new RegExp(`:${MOCK_PORT}$`), ''); // for MOCK tests, todo: remove from production
```
sosnovsky, line 163: is it still used or we can remove this replacement?

ioanmo226: I thought it is still used but let me check again

test/source/test.ts
ioanmo226: Let me check

extension/js/common/core/common.ts
ioanmo226: Checked and this logic is still used

scripts/config-mock-build.sh
```diff
@@ -2,5 +2,7 @@
 BUILD_PATH=./build/test/mock-builds/port-$2-$(openssl rand -hex 12)
 mkdir -p $BUILD_PATH
 cp -r ./$1/* $BUILD_PATH
-grep ":8001" $BUILD_PATH -lr | xargs sed -i.bak -e "s/\:8001/\:$2/g"
+grep "\[TEST_REPLACEABLE_MOCK_PORT\]" $BUILD_PATH -lr | xargs sed -i.bak -e "s/\[TEST_REPLACEABLE_MOCK_PORT\]/$2/g"
+# Replace export const MOCK_PORT line with the provided mock port
+grep "export const MOCK_PORT = '\[TEST_REPLACEABLE_MOCK_PORT\]';" $BUILD_PATH -lr | xargs sed -i.bak -e "s/export const MOCK_PORT = '\[TEST_REPLACEABLE_MOCK_PORT\]';/export const MOCK_PORT = '$2';/"
```
sosnovsky, line 7: does second `xargs sed ...` replace anything?

as first `xargs sed ...` at line 5 finds all occurrences of `[TEST_REPLACEABLE_MOCK_PORT]` and replaces them with provided port address.

then `grep "export const MOCK_PORT = '\[TEST_REPLACEABLE_MOCK_PORT\]';"` probably won't find anything, as all `[TEST_REPLACEABLE_MOCK_PORT]` were already replaced.

ioanmo226: Yep!
You are right. Thank you.
I thought too much!

sosnovsky (approved):
All good now 👍

PR #5894, opened 2025-01-07
#5032 Show clearer error message why key is not usable

### LINKED ISSUE #5032: clearer error message as to why key is not usable
User is getting `This OpenPGP key is not usable.` which doesn't say much, but we do seem to have access to the actual problem as visible from compatibility test:

> **(14) PK 0 > Encrypt/Decrypt test: Got error performing encryption/decryption test: Error: Error encrypting message: Could not find valid encryption key packet in key 87e87b1cf73b7e44: elgamal keys are considered too weak.:**

The error message could  be clearer, so we could say eg `Could not find valid encryption key packet in key 87e87b1cf73b7e44: elgamal keys are considered too weak.` instead of `This OpenPGP key is not usable.` when importing (or wherever that error was originally encountered.

_Originally posted by @tomholub in https://github.com/FlowCrypt/flowcrypt-browser/issues/5030#issuecomment-1479323251_

ioanmo226:
This PR implemented fix to show clearer error message why key is not usable

close #5032 // if this PR closes an issue

image

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/pgp_pubkey.ts
+++ b/extension/chrome/elements/pgp_pubkey.ts
@@ -13,6 +13,8 @@ import { Url } from '../../js/common/core/common.js';
 import { View } from '../../js/common/view.js';
 import { Xss } from '../../js/common/platform/xss.js';
 import { ContactStore } from '../../js/common/platform/store/contact-store.js';
+import { Buf } from '../../js/common/core/buf.js';
+import { OpenPGPKey } from '../../js/common/core/crypto/pgp/openpgp-key.js';
 
 // todo - this should use KeyImportUI for consistency.
 View.run(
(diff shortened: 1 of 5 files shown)


extension/chrome/elements/pgp_pubkey.ts
```diff
@@ -167,8 +184,9 @@ View.run(
       }
     };
 
-    private showKeyNotUsableError = () => {
+    private showKeyNotUsableError = async () => {
       $('.error_container').removeClass('hidden');
+      $('.error_introduce_label').text(`This OpenPGP key is not usable.\n${await this.getErrorText()}`);
```
sosnovsky, line 189: let's use `<small>` html tag for error description, it will take less space than normal text size.
also `error: error encrypting message:` part can't be removed, so error message will begin directly from error reason:


and incorrect key row is misaligned with other rows, probably because of smaller `view key` button. let's make it the same width as `import key` button.

ioanmo226: Sounds good!

sosnovsky (approved):
looks good now 👍

Issue #5896, opened 2025-01-09
Add a hotfix for error (err: bad recipient) issue


martgil:
This has been reported in details through the following GitHub issue below. Due to the number of reports, the demand of having a hotfix would be considerably be a safer decision.

issue https://github.com/FlowCrypt/enterprise-server/issues/6264

martgil:
This is already fixed in #5897

PR #5897, opened 2025-01-09
#5896 Propose hotfix in extension when sending password messages

martgil:
This PR proposed a hotfix in the FlowCrypt browser extension for an issue that exists when sending password messages.

#5896

### To be filled by reviewers


martgil:
Hello @sosnovsky - please kindly check if this hotfix looks good to you. I'll be sending the screen recording via email. Thanks!

sosnovsky (commented):
current solution affects all sent messages, while issue happens only for password-protected messages with accented characters in recipient name.
let's apply fix only for such cases.

in `webPortalMessageUpload` we use [formatEmailWithOptionalName](https://github.com/FlowCrypt/flowcrypt-browser/blob/7adc496827e9b6e3e2d5aef59c267f250583b59b/extension/js/common/api/account-servers/external-service.ts#L149) method for formatting recipient emails before uploading message. there we can also transform accented characters to their non-accented versions (`Š` will become `S`, `ř` -> `r` etc).

then implemented solution won't affect other messages and make it possible to send password-protected messages to all recipients.

martgil:
> current solution affects all sent messages, while issue happens only for password-protected messages with accented characters in recipient name. let's apply fix only for such cases.
> 
> in `webPortalMessageUpload` we use [formatEmailWithOptionalName](https://github.com/FlowCrypt/flowcrypt-browser/blob/7adc496827e9b6e3e2d5aef59c267f250583b59b/extension/js/common/api/account-servers/external-service.ts#L149) method for formatting recipient emails before uploading message. there we can also transform accented characters to their non-accented versions (`Š` will become `S`, `ř` -> `r` etc).
> 
> then implemented solution won't affect other messages and make it possible to send password-protected messages to all recipients.

You’re right. I’ve followed your instructions, and this has helped me improve in providing a fix. Unfortunately, adding a test doesn’t seem feasible since the transformation of text only occurs and becomes visible right when the request is sent. I verified the fix by inspecting the request sent by the browser extension to shared-tenant-fes. Indeed, my changes correctly replace accented characters with their non-accented versions.

One can verify this change by adding a recipient on contacts.google.com with accented characters. Then, send a password-protected message to them using the extension while monitoring the network traffic. Specifically, watch the request sent to shared-tenant-fes, more precisely to POST https://flowcrypt.com/shared-tenant-fes/api/v1/message.

extension/js/common/core/common.ts
```diff
@@ -283,7 +283,12 @@ export class Str {
   };
   private static formatEmailWithOptionalNameEx = ({ email, name }: EmailParts, forceBrackets?: boolean): string => {
     if (name) {
-      return `${Str.rmSpecialCharsKeepUtf(name, 'ALLOW-SOME')} <${email}>`;
+      /*
+       * hotfix for issue https://github.com/FlowCrypt/enterprise-server/issues/6264
+       * this replaces accented characters with their unaccented counterparts
+       */
+      const normalizedName = name.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
+      return `${Str.rmSpecialCharsKeepUtf(normalizedName, 'ALLOW-SOME')} <${email}>`;
```
sosnovsky, line 291: `formatEmailWithOptionalNameEx` used for common encrypted emails too.

to replace accented characters only for password-protected messages let's create a separate method `Str.replaceAccentedChars` and use it in `webPortalMessageUpload` as `to: (recipients.to || []).map(Str.formatEmailWithOptionalName).map(Xss.stripEmojis).map(Str.replaceAccentedChars)`

martgil: Thanks, Roma, I notice that it is indeed incorrect. I'll be fixing it.

martgil:
@sosnovsky This one is ready for a review again. I've also manually tested this and it works as it should. The accented characters where being replaced leads to `200 OK` when sending the upload to FES.

sosnovsky (approved):
Nice temporary fix 👍

martgil:
Yes, it wouldn't be possible without your help. thank you, too.

PR #5901, opened 2025-01-17
#5900 Fix delay issue when adding secure messaging action buttons on Gmail context menu

### LINKED ISSUE #5900: Fix delay in message context menu buttons added by FlowCrypt
**Description:**

The context menu entries added by the FlowCrypt browser extension in Gmail does have a few second/miliseconds delay before getting added for the first time.

Image

Users may accidentally click the secure message actions by FlowCrypt when they delay occurs instead of the regular reply button - not very user friendly there. Let's find a solution to fix the delay so that added buttons in the context menu be present there upon user clicks the context menu button.

**Steps to reproduce:**

1. Open any FlowCrypt extension recognizable email.
2. Click on the context menu button on the right side corner of the email thread.
3. Observe the context menu and the addition of FlowCrypt Secure action buttons with slight delay before getting injected into the menu.

Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzQZSjbvlFqBsVKKzJtBQMwWFlbh

martgil:
This PR fixes the delay issue when the extension tried to add the secure messaging action buttons on Gmail context menu.

close #5900

Screen recording of the propose fix:

https://github.com/user-attachments/assets/14f69919-c81c-4b58-bba5-f7f3b950cd41

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
@@ -291,10 +291,12 @@ export class GmailElementReplacer extends WebmailElementReplacer {
   };
 
   private replaceActionsMenu = () => {
-    if ($('.action_menu_message_button').length <= 0) {
-      this.addMenuButton('reply', '#r');
-      this.addMenuButton('forward', '#r3');
-    }
+    $(this.sel.msgActionsBtn).on('click', () => {
+      if ($('.action_menu_message_button').length <= 0) {
+        this.addMenuButton('reply', '#r');
+        this.addMenuButton('forward', '#r3');
+      }
+    });
   };
 
   private replaceConvoBtns = (force = false) => {


martgil:
Hi @sosnovsky, this one is ready for a review. Thank you!

martgil:
Hello @sosnovsky - This one is ready for a review. Thank you!

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
```diff
@@ -291,10 +291,12 @@ export class GmailElementReplacer extends WebmailElementReplacer {
...
-    if ($('.action_menu_message_button').length <= 0) {
-      this.addMenuButton('reply', '#r');
-      this.addMenuButton('forward', '#r3');
-    }
+    $(this.sel.msgActionsBtn).on('click', () => {
+      if ($('.action_menu_message_button').length <= 0) {
+        this.addMenuButton('reply', '#r');
+        this.addMenuButton('forward', '#r3');
+      }
+    });
```
sosnovsky, line 299: this replacement will search for `this.sel.msgActionsBtn` in document DOM each second, which can lead to performance issues.

it'll be better to use `$(document).on('click', this.sel.msgActionsBtn, () => { ...` - it can be performed only 1 time and will catch all clicks at `this.sel.msgActionsBtn`.

martgil: Thanks @sosnovsky - I've been finding a working alternative to this. You are right the statement that using $(document).on('click',.... but due to Gmail complexity, this isn't as simple as targeting the `this.sel.msgActionsBtn` but I highly agree with the concept your trying to teach to. I can't say it precisely why it doesn't work but after doing quite more debugging, I found a solution that is similar to what you've suggested. The result was significantly noticeable and love it showed up nicely without any delays with the recent changes.

martgil:
Good afternoon, @sosnovsky,

I believe this is now ready for your review. I would also like to request that you test it to ensure it meets your expectations. From my side, the result looks excellent, and I love how it works now compared to the initial implementation.

Looking forward to your feedback!

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
```diff
@@ -291,10 +291,13 @@ export class GmailElementReplacer extends WebmailElementReplacer {
...
-      this.addMenuButton('forward', '#r3');
-    }
+    $(document).on('click', 'div.aHU.hx', event => {
+      const $actionsBtn = $(event.currentTarget).find(this.sel.msgActionsBtn);
+      if ($actionsBtn.length && !$('.action_menu_message_button').length) {
+        this.addMenuButton('reply', '#r');
+        this.addMenuButton('forward', '#r3');
+      }
+    });
   };
```
sosnovsky, line 301: with current implementation `replaceActionsMenu` will be executed each second, adding more and more `$(document).on('click')` handlers which can lead to performance issues.

it'll be more correct to add this handler in [GmailWebMailStartup](https://github.com/FlowCrypt/flowcrypt-browser/blob/27b8b376642d92f4efb1491aeb1412c9ef7daa9e/extension/js/content_scripts/webmail/gmail/gmail-webmail-startup.ts#L43), as it'll be executed only once and handle all future clicks:

```ts
private addSecureActionsToMessageMenu = () => {
  $('body').on('click', 'div.aHU.hx', event => {
    this.replacer.addSecureActionsToMessageMenu(event.currentTarget)
  });
}
```

sosnovsky (commented):
Yeah, works much faster now 👍

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
martgil: Thank you @sosnovsky for your constant guidance. I see what you mean by now, its the heartbeat like injection secure compose which may lead to performance issue. I have done the necessary changes and the result is good.

sosnovsky (approved):
Works great 👍

PR #5909, opened 2025-01-24
#5905 Add Secure Reply-All to Gmail context menu

### LINKED ISSUE #5905: Add "secure reply to all" button to message actions menu
Adding `Secure Reply To All` button to this menu will be a convenient addition, so all reply options will have their secure alternatives:

Image

martgil:
This PR adds the FlowCrypt's "Secure Reply All" option to Gmail's context menu.

close #5905

- Tests added or updated

### To be filled by reviewers


martgil:
Hi @sosnovsky, we do have some problem with the recent change. While it works, it may not works best occasionally.

https://github.com/user-attachments/assets/62ea185a-a515-4217-be65-efdadcb273a7

That's what I recently noticed when constantly checking it here. The reason why this is very tricky DOM manipulation is that, the actionMenu aka the Gmail context menu is not hidden or present in the document.body not unless a single click on the actionMenuBtn is made which would then programmatically added the Gmail's reply,forward buttons, etc. That makes me solve it by immediately binding the click event and modify the actionMenu right after the click - this solves the bug in delayed injection of secure actions.

However, now that we move it on gmail replace, the time it works may vary now since they're maybe some latency plus the fact that if i click on the actionMenuBtn, it is not what the $(document).click('... is actual targetting but its parent `div.aHU.hx` and then checks for `$(event.currentTarget).find(this.sel.msgActionsBtn).length` may not work for some unknown reason. Not to mention that, I may not be clicking the the `div.aHU.hx`...

Is there any alternative left for us? What if the the function for adding secure action buttons on context menu works "reliably" in `this.replacer.runIntervalFunctionsPeriodically()` or maybe we could find some efficient way... It's too buggy in a way its unreliable, it showed up when clicked but may occationally failed unlike in `runIntervalFunctionsPeriodically()` where it had the chance pretty quickly to get it back while still not populating the context menu.

martgil:
I have been consistently testing this, and it seems like a bug is working against me. It's no joke. When I test the same scenario with the JavaScript console or web developer console enabled, the UI bug does not occur. However, as soon as I close the web developer console and refresh the page, the issue reappears.

sosnovsky:
I think it happens because on document load `div.b7.J-M` is empty and `addMenuButton` fails to insert secure buttons, as it doesn't find requested selectors. Probably in `$(document).on('click', 'div.aHU.hx', event => {` we can add check for `$(this.sel.msgActionsMenu).is(':empty')`, and if it's empty - retry to add buttons in 100ms.

sosnovsky:
Hi @martgil, do you think we can finalize this PR before the end of the week to publish new extension release?

martgil:
Yes @sosnovsky - I believe thats completely possible. I'll try your suggestion but if that doesn't work. Could you please let me know your thoughts on my latest change for the time being? It tried to address the constant addition of `click()` event handler.

https://github.com/FlowCrypt/flowcrypt-browser/pull/5909/commits/5c0118640544ffa33dfd5eb863f8b21c9a1ca6d4

martgil:
> I think it happens because on document load `div.b7.J-M` is empty and `addMenuButton` fails to insert secure buttons, as it doesn't find requested selectors. Probably in `$(document).on('click', 'div.aHU.hx', event => {` we can add check for `$(this.sel.msgActionsMenu).is(':empty')`, and if it's empty - retry to add buttons in 100ms.

Thanks for sharing your insight. I'll try it right away.

sosnovsky:
> Yes @sosnovsky - I believe thats completely possible. I'll try your suggestion but if that doesn't work. Could you please let me know your thoughts on my latest change for the time being? It tried to address the constant addition of `click()` event handler. Just in case i need to look for another way to address the issue.
> 
> [5c01186](https://github.com/FlowCrypt/flowcrypt-browser/commit/5c0118640544ffa33dfd5eb863f8b21c9a1ca6d4)

I think it would still affect performance as binding/unbinding happens each second in this case.
Finding way to bind it only once, should be the most effective way of handling this, but we can have your current solution as a backup option.

martgil:
> > Yes @sosnovsky - I believe thats completely possible. I'll try your suggestion but if that doesn't work. Could you please let me know your thoughts on my latest change for the time being? It tried to address the constant addition of `click()` event handler. Just in case i need to look for another way to address the issue.
> > [5c01186](https://github.com/FlowCrypt/flowcrypt-browser/commit/5c0118640544ffa33dfd5eb863f8b21c9a1ca6d4)
> 
> I think it would still affect performance as binding/unbinding happens each second in this case. Finding way to bind it only once, should be the most effective way of handling this, but we can have your current solution as a backup option.

Thanks @sosnovsky - I have good news. My recent change with the help of your guidance, has worked flawlessly. I'm working on fixing the failing test.

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
```diff
@@ -149,13 +149,21 @@ export class GmailElementReplacer extends WebmailElementReplacer {
...
+        Catch.setHandledTimeout(addButtons, retryInterval);
+      } else {
+        $(document).on('click', 'div.aHU.hx', () => {
+          if (!$('.action_menu_message_button').length) {
+            this.addMenuButton('a_reply', '#r');
+            this.addMenuButton('a_reply_all', '#r2');
+            this.addMenuButton('a_forward', '#r3');
+          }
+        });
       }
```
sosnovsky, line 164: here timeout should be added inside `$(document).on('click')`, as in some cases `msgActionsMenu` element will be empty and `addMenuButton` won't find correct element:

```ts
$(document).on('click', 'div.aHU.hx', () => {
           if ($('.action_menu_message_button').length) { return }
           if ($(this.sel.msgActionsMenu).is(':empty')) {
              Catch.setHandledTimeout(addButtons, retryInterval);
           } else {
               addButtons();
           }
         });
```

martgil: I have tried to use this but this also gives undesirable result. I'll be continuing to work on this again by tomorrow. Problem is that sometimes it completely doesn't retry, and what happened is that the event happens right after the click which result to secure actions to be added after the user tried to open the context menu for the second time.

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
```diff
@@ -149,13 +149,21 @@ export class GmailElementReplacer extends WebmailElementReplacer {
...
-        this.addMenuButton('forward', '#r3');
+    const retryInterval = 100;
+    const addButtons = () => {
+      if (!$(this.sel.msgActionsBtn).length) {
+        Catch.setHandledTimeout(addButtons, retryInterval);
+      } else {
+        $(document).on('click', 'div.aHU.hx', () => {
+          if (!$('.action_menu_message_button').length) {
+            this.addMenuButton('a_reply', '#r');
+            this.addMenuButton('a_reply_all', '#r2');
```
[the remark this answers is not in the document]
martgil: Thanks for noticing this @sosnovsky!

Ah, we should also consider that. I dont exactly know how to switch from a thread with mulitple recipients to covert it in single recipient, could you please guide me on how to do that?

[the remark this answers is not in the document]
sosnovsky: Probably it can be done just by checking if plain `reply all` button exists (`#r2` selector)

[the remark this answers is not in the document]
martgil: Thanks - it should be fixed by now.

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
martgil: being dependent to $(document) and then choosing 'div.aHU.hx' is definitely not the way to get. It has some latency issue because the user may not exactly interacting with div.aHU where the button for opening context menu is located. For example, the event gets trigger right after the context menu gets closed and not when the button for opening the context menu in Gmail gets clicked.

sosnovsky:
Hi @martgil, what issues left to be fixed in this task?
Let's finish it today-tomorrow to be able to publish new release on Wednesday.

martgil:
Hello @sosnovsky - I'm trying to do my best in fixing this one. If we're using $(document)... I'd get an unreliable results... like i've shared in the above screen recording - sometimes it showed up sometime its not showing at all. so I'm just working now on its reliability since its very tricky.

martgil:
I would like to see our opportunity to use the previous approach @sosnovsky since that's we're getting reliable results compared to using $(document).on('click', selector) since HTML structure is obscure in Gmail with that approach.

Is there anyhting I could test just to see how much performance loss were worrying about? if its not that much, maybe we could allow it?

sosnovsky:
> Is there anyhting I could test just to see how much performance loss were worrying about? if its not that much, maybe we could allow it?

Ok, please try to use previous approach, and I'll investigate `$(document)` usage.
So we'll have at least 1 working solution, and then can decide on better approach.

martgil:
Got it, thank you. I'll try to finish this today so there will be new tasks for tomorrow.

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
```diff
@@ -291,13 +282,14 @@ export class GmailElementReplacer extends WebmailElementReplacer {
...
 
-  private addMenuButton = (action: 'reply' | 'forward', selector: string) => {
-    const gmailActionsMenuContainer = $(this.sel.msgActionsMenu).find(selector);
-    const button = $(this.factory.actionsMenuBtn(action)).insertAfter(gmailActionsMenuContainer); // xss-safe-factory
-    button.on(
-      'click',
-      Ui.event.handle((el, ev: JQuery.Event) => this.actionActivateSecureReplyHandler(el, ev))
-    );
+  private addMenuButton = (replyOption: ReplyOption, gmailContextMenuBtn: string) => {
+    if ($(gmailContextMenuBtn).css('display') === 'block') {
```
sosnovsky, line 286: this can be replaced by `if ($(gmailContextMenuBtn).is(':visible')) {`

martgil: Got it - @sosnovsky, I'll change it as suggested. Btw, could you please confirm if it works for you as you expected it to work. For me, it works very well now, very unnoticeable to almost no latency.

martgil: It works almost perfect now, one thing left is for the test to pass. If you noticed issues along the way, please let me know.

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
```diff
@@ -922,4 +921,17 @@ export class GmailElementReplacer extends WebmailElementReplacer {
...
+      if ($(this.sel.msgActionsMenu).is(':visible') && !$('.action_menu_message_button').length) {
+        this.addMenuButton('a_reply', '#r');
+        this.addMenuButton('a_reply_all', '#r2');
+        this.addMenuButton('a_forward', '#r3');
+      } else {
+        Catch.setHandledTimeout(addSecureOptionsToGmail, 100);
+      }
+    };
+    addSecureOptionsToGmail();
+  };
```
sosnovsky, line 936: Button insertion works well, but the current implementation runs `addSecureOptionsToGmail` several hundred times per second.
I added logging to track calls to `addSecureActionsToMessageMenu` from the `everything` function and each execution of `addSecureOptionsToGmail`:
Screenshot 2025-02-03 at 22 55 06

With Gmail open for less than a minute, `addSecureOptionsToGmail` was called approximately 800 times per second, causing memory usage to grow from ~300MB to ~600MB. This happens because each `addSecureOptionsToGmail` call adds another `Catch.setHandledTimeout` when `msgActionsMenu` is not visible, leading to an increasing number of scheduled `addSecureOptionsToGmail` executions.

To fix it, I tried to merge both of our solutions by handling message menu button clicks while keeping the existing interval function in `everything` - https://github.com/FlowCrypt/flowcrypt-browser/commit/791ed66aa4f51fe63fd4846a23ebaaa0a56f083b
It seems to work well for me - please check it too.
If it looks good to you, let's add it to the current PR.

sosnovsky:
@martgil live tests probably fail because of current 45 minutes execution limit - https://github.com/FlowCrypt/flowcrypt-browser/blob/cbc16870e7aa3765f79ff10c7c1c99e6d62849d3/.semaphore/semaphore.yml#L64
Let's increase it to 60 minutes for now, but later will try to merge some live tests, so they'll be executed faster.

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
martgil: @sosnovsky - I have evaluated the PR issue-5905-test-fix but unfortunately, it doesn't meet our expectation and there's much more noticable latency when it comes to adding our secure buttons. Meanwhile, I think I found the best solution ever for this one and it should looks like:

```
  public setupSecureActionsOnMenuButtonClick = () => {
    $(document).ready(() => {
      const targetSelector = '.b7.J-M';
      const observer = new MutationObserver(() => {
        const $element = $(targetSelector);
        if ($element.length && $element.is(':visible')) {
          this.addSecureActionsToMessageMenu();
        }
      });
      observer.observe(document.body, { childList: true, subtree: true });
    });
  };
  ```
  
  I have it on my code now and its result is acceptable and achieve better result compared to using the previous implementation - except that it doesnt run in a loop.

**Comment by martgil** [comment]
I'm just checking any issues on the tests. Meanwhile, @sosnovsky please try the latest change and it should be the most acceptable solution that we can find.

**Comment by martgil** [comment]
@sosnovsky are you preparing the new FlowCrypt browser extension release? This is already a good PR but the tests are not yet passing but I'm hoping to get this added.

**Comment by martgil** [comment]
Thanks for your answer, Roma - please feel free to test this one. Meanwhile, its just the unrelated tests that are failing, but the test dedicated for this change is passing without any issues.

File: extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts [hunk 5]
```diff
@@ -148,13 +149,15 @@ export class GmailElementReplacer extends WebmailElementReplacer {
...
+  public setupSecureActionsOnGmailMenu = () => {
+    $(() => {
+      const observer = new MutationObserver(() => {
+        const gmailActionsMenu = $(this.sel.msgActionsMenu);
+        if (gmailActionsMenu.is(':visible')) {
+          this.addSecureActionsToMessageMenu();
+        }
+      });
+      observer.observe(document.body, { childList: true, subtree: true });
     });
```
sosnovsky, line 161: great solution with `MutatuonObserver` - works very consistent, didn't notice issues with it.
just a couple improvements:
- `$(() => {` isn't needed, as our code is executed after document is loaded
- instead of jquery `gmailActionsMenu.is(':visible')` let's use native js check `gmailActionsMenu.offsetParent !== undefined`, it's more lightweight but gives the same result.

then `setupSecureActionsOnGmailMenu` will look like this:
```ts
public setupSecureActionsOnGmailMenu = () => {
    const observer = new MutationObserver(() => {
      const gmailActionsMenu = document.querySelector(this.sel.msgActionsMenu);
      if (gmailActionsMenu && (gmailActionsMenu as HTMLElement).offsetParent !== undefined) {
        this.addSecureActionsToMessageMenu();
      }
    });
    observer.observe(document.body, { childList: true, subtree: true });
  };
```

sosnovsky:
Hi @martgil, let's work on finishing this one today, so everything will be ready for release.
I pushed my proposed changes to https://github.com/FlowCrypt/flowcrypt-browser/commit/4ddea2eb07d723885690bb226e8053e678a662a5, mock tests pass there, currently waiting for live tests results.

martgil:
Yes @sosnovsky - I'll finish this one today and do things as you suggested.

extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
martgil: Thanks @sosnovsky - I've already change the necessary code that needs to be updated. I'm currently waiting for live tests for this one to pass. I'm not getting so lucky to have passing tests in a single go unlike yours in https://github.com/FlowCrypt/flowcrypt-browser/commit/4ddea2eb07d723885690bb226e8053e678a662a5.

sosnovsky (approved):
👍

martgil:
Hi @sosnovsky - the live tests are now passing :) Could we merge this one?

martgil:
Thank you, Roma!

sosnovsky:
Thanks for finding a well-working solution 👍

PR #5914, opened 2025-01-28
#5906 Keep original reply message when switching to secure mode

### LINKED ISSUE #5906: Keep original reply message when switching to secure mode
- open thread with multiple messages (including encrypted). for example, `#5004 cryptup_reply div not hidden issue` thread in `flowcrypt.compatibility` inbox
- click `reply` button in the toolbar of the first message
- compose box appears with red banner
- click `Switch to encrypted reply`
- secure compose box is loaded
- click `...` to show quoted content of the previous messages
- it'll include content of all thread messages, while it's expected to have quoted content only of the first message as reply is composed to it

probably worked well in 8.5.6

\+ implement ui test

ioanmo226:
This PR fixed keep original reply message when switching to secure mode issue

close #5906// if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
+++ b/extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
@@ -618,8 +618,8 @@ export class GmailElementReplacer extends WebmailElementReplacer {
     return from ? Str.parseEmail(from) : undefined;
   };
 
-  private getLastMsgReplyParams = (convoRootEl: JQuery, replyOption?: ReplyOption): FactoryReplyParams => {
-    return { replyMsgId: this.determineMsgId($(convoRootEl).find(this.sel.msgInner).last()), replyOption };
+  private getLastMsgReplyParams = (convoRootEl: JQuery, replyOption?: ReplyOption, replyBoxMessageId?: string): FactoryReplyParams => {
+    return { replyMsgId: replyBoxMessageId ?? this.determineMsgId($(convoRootEl).find(this.sel.msgInner).last()), replyOption };
   };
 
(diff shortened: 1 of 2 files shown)


extension/js/content_scripts/webmail/gmail/gmail-element-replacer.ts
```diff
@@ -643,9 +643,12 @@ export class GmailElementReplacer extends WebmailElementReplacer {
     const newReplyBoxes = $('div.nr.tMHS5d, td.amr > div.nr, div.gA td.I5').not('.reply_message_evaluated').filter(':visible').get();
     if (newReplyBoxes.length) {
       this.replyOption = undefined;
+      // Try to get message id from plain reply box
+      // https://github.com/FlowCrypt/flowcrypt-browser/issues/5906
+      const replyBoxMessageId = $(newReplyBoxes[0]).find('[name="lts"]').val() as string;
```
sosnovsky, line 648: `[name="lts"]` seems to have id of the first thread message, as I've tested composing encrypted reply to the third thread message and quoted text included only first message content.

but there is another hidden input `[name="rm"]`, and looks like it have id of correct reply message. tried to use it in the code and reply message contained correct quoted text.

sosnovsky: Hmm, interesting - after I've replaced `[name="lts"]` with `[name="rm"]` switching to encrypted reply works correctly in some threads, while in others quote text is just `undefined`, probably because of error you've mentioned.

ioanmo226:
Fixed and ready for review.

PR #5919, opened 2025-01-31
#5913 Add encrypted attachment to password protected message

### LINKED ISSUE #4702: attachments not sent through email provider for fes message

### LINKED ISSUE #5913: Attachment is not visible in sent password-protected message
- create encrypted message with attachment and 2 recipients - recipient A which uses encryption and recipient B which don't use encryption (so FlowCrypt will propose to send password-protected message)
- send message
- go to `Sent` folder
- there will be 2 sent messages:
   - encrypted to recipient A - attachment is shown here
   - password-protected to recipient B - attachment is not visible here

Expected behavior - attachment is visible in both sent emails

ioanmo226:
This PR adds encrypted attachment to password protected messages so that attachment can be seen in password protected messages.

close #5913 // if this PR closes an issue
close #4702 // if this PR closes an issue

- Difficult to test(The current mock Gmail environment doesn't support decrypting sent messages. In a live Gmail test, we would need to use a real non-public key email recipient, which could be inconvenient as it requires sending actual messages to that recipient.)

### To be filled by reviewers

### DIFF
--- a/extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
+++ b/extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
@@ -85,7 +85,7 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
       );
     }
     // rich text: PGP/MIME - https://tools.ietf.org/html/rfc3156#section-4
-    const attachments = await this.formatEncryptedMimeDataAsPgpMimeMetaAttachments(encrypted, pubkeys);
+    const attachments = this.formatEncryptedMimeDataAsPgpMimeMetaAttachments(encrypted);
     return await SendableMsg.createPgpMime(this.acctEmail, this.headers(newMsg), attachments, {
       isDraft: this.isDraft,
     });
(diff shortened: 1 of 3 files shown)


extension/chrome/elements/compose-modules/formatters/encrypted-mail-msg-formatter.ts
```diff
@@ -212,8 +212,8 @@ export class EncryptedMsgMailFormatter extends BaseMailFormatter {
     });
   };
 
-  private formatEncryptedMimeDataAsPgpMimeMetaAttachments = (data: Uint8Array) => {
-    const attachments: Attachment[] = [];
+  private formatEncryptedMimeDataAsPgpMimeMetaAttachments = async (data: Uint8Array, pubkeys: PubkeyResult[]) => {
+    const attachments: Attachment[] = await this.view.attachmentsModule.attachment.collectEncryptAttachments(pubkeys);
```
sosnovsky, line 216: Seems like with such change we break standardized PGP/MIME message structure, which expects only 2 attachments (https://datatracker.ietf.org/doc/html/rfc3156#section-4).

Let's move attachments in password-protected messages into `encrypted.asc`, like in usual PGP/MIME messages.

It can be done on Monday, no rush to finish it on this week, as new browser extension release is expected on the next week.
Thanks for fixing all reported issues in time, well done!

ioanmo226: Yep!
You are right.
Let me implement as you suggested

ioanmo226: Fixed. please check

sosnovsky (approved):
Works great, thanks!

PR #5943, opened 2025-02-21
#5940 Clean up unused alarms

### LINKED ISSUE #5940: Cleanup Unused Alarms to Prevent Exceeding Limit
I encountered the following error on the extension errors page:

> Uncaught (in promise) Error: An extension cannot have more than 500 active alarms.

It looks like we need to clean up alarms, specifically those used for delete_expired, as they may be accumulating unnecessarily.
This could be causing the extension to exceed the 500 active alarms limit.

ioanmo226:
This PR implemented feature to clean up unused alarms

close #5940 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/js/common/browser/browser-msg.ts
+++ b/extension/js/common/browser/browser-msg.ts
@@ -429,10 +429,19 @@ export class BrowserMsg {
     BrowserMsg.HANDLERS_REGISTERED_BACKGROUND[name] = handler;
   }
 
-  public static createIntervalAlarm(action: string, periodInMinutes: number) {
+  public static async clearAllDeleteExpiredAlarms() {
+    const alarms = await chrome.alarms.getAll();
+    for (const alarm of alarms) {
+      if (alarm.name.includes('delete_expired_interval_')) {
+        await chrome.alarms.clear(alarm.name);
+      }
+    }
+  }
+
+  public static async createIntervalAlarm(action: string, periodInMinutes: number) {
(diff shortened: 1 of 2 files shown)


extension/js/common/browser/browser-msg.ts
```diff
@@ -429,10 +429,19 @@ export class BrowserMsg {
...
+        await chrome.alarms.clear(alarm.name);
+      }
+    }
+  }
+
+  public static async createIntervalAlarm(action: string, periodInMinutes: number) {
     const alarmName = `${action}_interval_${Date.now()}`;
 
-    void chrome.alarms.create(alarmName, { periodInMinutes });
+    await chrome.alarms.create(alarmName, { periodInMinutes });
```
sosnovsky, line 444: `chrome.alarms.create` documentation says `If there is another alarm with the same name (or no name if none is specified), it will be cancelled and replaced by this alarm.`.

What about using `${action}_interval` as name (without timestamp), so existing alarm will be automatically cancelled and replaced with new one?

extension/js/common/browser/browser-msg.ts
```diff
@@ -513,6 +522,7 @@ export class BrowserMsg {
 
   public static alarmListen() {
     const alarmListener = (alarm: { name: string }) => {
+      console.log(alarm.name);
```
sosnovsky, line 525: can be removed

sosnovsky (approved):
looks good 👍

PR #5950, opened 2025-03-13
#5949 Fix: invalid checksum mismatch error in signed messages with public key attached

### LINKED ISSUE #5912: PGP invalid checksum appears on forwarded signed message
Description:

I noticed that invalid checksum appears on forwarded signed message. Meanwhile the signed message verification works.

Image

There's also a typographical error which makes "}" as part of the rendered content but should be omitted.

https://github.com/FlowCrypt/flowcrypt-browser/blob/aeda798bbdb30449791812042975ccc6766b09c1/extension/chrome/elements/pgp_block_modules/pgp-block-render-module.ts#L52

Step to reproduce:

1. User 1 send signed message User 2.
2. User forwards the signed message as plain text using (Gmail forward) to User 3.
3. User 3 sees the following message above using FlowCrypt browser.

### LINKED ISSUE #5949: First time sender of signed message with pubkey attached received checksum mismatch error
Hello team,

Good day!

I noticed that when I reply `signed` with public key included for a plain email for the first time, I get the following error message about checksum mismatch error.

Image

This kindof confusing as the message says the message I just sent sounds like it has been altered in transit when in fact its not.

I think there would be check that missed and should not have assumption about checksum mismatch of my newly sent email?

_Reference: https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzQZTgMmjGcKRKrTstGQhxVNZBSF_

ioanmo226:
This PR fixed invalid checksum mismatch error in signed messages with public key attached

close #5949 // if this PR closes an issue
close #5912 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers


sosnovsky (approved):
Works well 👍

PR #5951, opened 2025-03-13
#5947 Add info modal for key with newer expiry than existing key

### LINKED ISSUE #5947: [Suggested Improvement] Offer private update when imported an updated version of existing key
Some users tried to load the updated version of their keys via "Add Key" of FlowCrypt Settings -> Additional Settings.

While it is technically working correctly and the FlowCrypt extensions rejects it and tells the user that it already exists. It doesn't mean that the two version where identical and doesn't have any further information whether which one is newer and vise versa.

IMHO, we can improve it by detecting which one is more recently updated. And in case where the one they tried to import was a newer one, we can them inform them to update their record via `Public Key` -> In Private key section, they will click `update` and update their keys.

Or even better is to just redirect them to update private key frame by telling some like `This key is a newer version of xzy, please click here if you want to update your keys`.

martgil:
This PR adds a check to determine if the user is importing a private key with a newer expiry date than the existing one and notifies the user accordingly.

close #5947

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/ui/key-import-ui.ts
+++ b/extension/js/common/ui/key-import-ui.ts
@@ -356,7 +356,19 @@ export class KeyImportUi {
       const keyinfos = await KeyStore.get(acctEmail);
       const privateKeysIds = keyinfos.map(ki => ki.fingerprints[0]);
       if (privateKeysIds.includes(k.id)) {
-        throw new UserAlert('This is one of your current keys, try another one.');
+        const key = keyinfos.find(ki => ki.id === k.id);
+        /* eslint-disable @typescript-eslint/no-non-null-assertion */
+        const existingKey = await KeyUtil.parse(key!.public);
+        const hasNewerExpiration = !!(k.expiration && existingKey.expiration && k.expiration > existingKey.expiration);
+        if (hasNewerExpiration) {
+          const updateKeyMessage =
+            "The key you're trying to import is a newer version of one you already have, based on its expiry date.\n" +
+            'Follow the <a href="https://flowcrypt.com/docs/getting-started/common-tasks/update-private-key.html">guide</a> \n' +
+            'to update it or contact your FlowCrypt admin.';
+          await Ui.modal.info(updateKeyMessage, true);
+        } else {
+          throw new UserAlert('This is one of your current keys, try another one.');
+        }
       }
     }
   };


martgil:
wip: currently adding ui test

martgil:

extension/js/common/ui/key-import-ui.ts
```diff
@@ -356,7 +356,19 @@ export class KeyImportUi {
       const keyinfos = await KeyStore.get(acctEmail);
       const privateKeysIds = keyinfos.map(ki => ki.fingerprints[0]);
       if (privateKeysIds.includes(k.id)) {
-        throw new UserAlert('This is one of your current keys, try another one.');
+        const key = keyinfos.find(ki => ki.id === k.id);
+        /* eslint-disable @typescript-eslint/no-non-null-assertion */
+        const existingKey = await KeyUtil.parse(key!.public);
```
sosnovsky, line 361: `eslint-disable` here will disable `no-non-null-assertion` rule for all lines until the end of the file. `eslint-disable-next-line` will be more correct to use here.

but it'll be even better to add additional check than use `eslint-disable`:
```ts
const key = keyinfos.find(ki => ki.id === k.id);
if (key) {
  const existingKey = await KeyUtil.parse(key.public);
  ...
```

extension/js/common/ui/key-import-ui.ts
```diff
@@ -356,7 +356,19 @@ export class KeyImportUi {
...
+        const hasNewerExpiration = !!(k.expiration && existingKey.expiration && k.expiration > existingKey.expiration);
+        if (hasNewerExpiration) {
+          const updateKeyMessage =
+            "The key you're trying to import is a newer version of one you already have, based on its expiry date.\n" +
+            'Follow the <a href="https://flowcrypt.com/docs/getting-started/common-tasks/update-private-key.html">guide</a> \n' +
+            'to update it or contact your FlowCrypt admin.';
+          await Ui.modal.info(updateKeyMessage, true);
```
sosnovsky, line 368: Is it possible to add modal button here which will navigate user directly to `update private key` modal:
Screenshot 2025-03-23 at 21 37 31

extension/js/common/ui/key-import-ui.ts
martgil: Thanks, I'll be correcting this.

martgil:
Thanks for your recommendation and for making it possible.

PR #5969, opened 2025-04-03
#5966 Add reliable checks for ambigious file extensions

### LINKED ISSUE #5966: Public key detection doesn't work reliably on certain public key attachment
There is this very specific public key attachment that doesn't invoke FlowCrypt automatic public key import prompt for quick import public key as attachments.

The email support below contains a .eml file from which it contains public key attachments that although has `application/pgp-keys` has this name: `abc.def@ghi.com.asc` (not their real email add).

https://mail.google.com/mail/u/human@flowcrypt.com/#inbox/FMfcgzQZTzVdDXmCqkbJVqxxBNRtsQSM

The reporting user however had proceed to manually importing the public key which works just fine as the public key itself is valid and usable.

martgil:
This PR does fix .asc public key attachment not recognized.

Fix:

image

Current behavior: 

image

Closes #5966 

### To be filled by reviewers


martgil:
Hi @sosnovsky, this one is ready for a review. thanks!

extension/js/common/core/attachment.ts
```diff
@@ -204,7 +204,7 @@ export class Attachment {
     } else if (this.name === 'message' && isBodyEmpty) {
       // treat message as encryptedMsg when empty body for the 'message' attachment
       return 'encryptedMsg';
-    } else if (this.name.match(/(\.pgp$)|(\.gpg$)|(\.[a-zA-Z0-9]{3,4}\.asc$)/g)) {
+    } else if (this.name.match(/(\.pgp$)|(\.gpg$)|(\.[a-zA-Z0-9]{3,4}\.asc$)/g) && !this.isPublicKey() && !this.isPrivateKey()) {
```
sosnovsky, line 207: what about adding test for this case? probably shouldn't be too complicated, but will help to not break this solution in the future

martgil: Thank you for the remarks. You're right, and I missed that opportunity as I initially thought it wouldn't need testing. I'll definitely need to add a test for this.

martgil:
Hi again @sosnovsky, I've added a small test for this change. Do you think it will be sufficient?

sosnovsky (approved):
all good now, thanks!

PR #5980, opened 2025-04-21
#5979 Better detection for emails with noname attachments

### LINKED ISSUE #5979: HTML email with noname attachment mistakenly identified as encryptedMsg
### Description

The FlowCrypt browser extension mistakenly identifies an HTML email with a noname (image attachment) as encryptedMsg which result to not readable email when FlowCrypt is installed.

Original email is HTML + a `noname` image:

Image

FlowCrypt recognizes it as encryptedMsg and displays the following:

Image

### Steps to reproduce:

FlowCrypt browser extension version: 8.5.10

Test email that replicates the issue:

_https://mail.google.com/mail/u/flowcrypt.compatibility@gmail.com/#inbox/FMfcgzQZVKKDhBwhZCPtzmcBHhGCBJNt_

- Open the email and you'll be able to replicate the screen from the screenshot above.

### Workaround

Use FlowCrypt's "see original" to read the HTML email while we are working on a fix.

martgil:
This PR add an in-between detection before concluding that the detected `noname` file is a possible `encryptedMsg`.

close #5979

- Tests added or updated

### To be filled by reviewers


martgil:
We'd normally, wrote test for such PR but for such case, its a customer email and I cannot wrote such that specific test email with a "noname" file that has "application/pdf" type.

I think the most feasible and right way to do this is by splitting the test `decrypt - an ambiguous file "noname" should not be recognized as an encrypted message and should be hidden in encrypted message` into two, as the newer change should also fit for email `const threadId1 = '18adb91ebf3ba7b9'; // email attachment "noname" with type img/<image-extension>` and I believe it would not be part of that tests.

martgil:
Hi @sosnovsky - this one is ready for a review. Thank you.

sosnovsky:
> We'd normally, wrote test for such PR but for such case, its a customer email and I cannot wrote such that specific test email with a "noname" file that has "application/pdf" type.

It'll be useful to have such test, so in the future we won't break correct `noname` attachments parsing.
It should be possible to create such message with `noname` .pdf attachment by using [Header Tools Lite](https://addons.thunderbird.net/en-US/thunderbird/addon/header-tools-lite/?src=search) extension for Thunderbird, it allows to edit message headers before sending.
Can you please try to create such message for test by removing `name` property from sending attachment?

martgil:
Hi @sosnovsky, I tried your suggestion and it works flawlessly by using https://addons.thunderbird.net/en-US/thunderbird/addon/header-tools-lite/?src=search. This tool definitely helps a lot.

martgil:
Hello @sosnovsky, I've added a test by crafting one. Please let me know if this looks good to you.

sosnovsky (approved):
all good now, thanks!

PR #5987, opened 2025-05-09
#5977 Use original eslint header plugin

### LINKED ISSUE #5977: there is simpler solution for using default `header` plugin, please check https://github.com/microsoft/vscode/blob/54b12e5ec877a5f45ebc927b055fed259a195f1f/eslint.config.js#L16
there is simpler solution for using default `header` plugin, please check https://github.com/microsoft/vscode/blob/54b12e5ec877a5f45ebc927b055fed259a195f1f/eslint.config.js#L16

adding `pluginHeader.rules.header.meta.schema = false;` to eslint config should make it compatible with the latest eslint. let's use it, instead of our fork

_Originally posted by @sosnovsky in https://github.com/FlowCrypt/enterprise-server/pull/6360#discussion_r2050468683_

ioanmo226:
This PR changed to default eslint header plugin

close #5977 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -2,6 +2,7 @@
 import tseslint from 'typescript-eslint';
 import noOnlyTestsPlugin from 'eslint-plugin-no-only-tests';
 import headerPlugin from 'eslint-plugin-header';
+headerPlugin.rules.header.meta.schema = false;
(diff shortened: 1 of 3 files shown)


sosnovsky (approved):
👍

PR #6016, opened 2025-06-09
#6013 Use Clipboard API instead of ClipboardJS

### LINKED ISSUE #6013: Use Clipboard API instead of ClipboardJS
We currently use ClipboardJS library to copy text to the clipboard, but this can now be done using the native Clipboard API - https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText

ioanmo226:
This PR uses Clipboard API instead of ClipboardJS

close #6013 // if this PR closes an issue

### To be filled by reviewers


sosnovsky (approved):
works well 👍

PR #6020, opened 2025-06-12
#6018 Migrate zxcvbn to zxcvbn-ts

### LINKED ISSUE #6018: Migrate to zxcvbn-ts
We currently use `zxcvbn` which wasn't updated for 8 years, but there is modern alternative which is actively maintained - https://github.com/zxcvbn-ts/zxcvbn.
Let's migrate to it, here is migration guide - https://zxcvbn-ts.github.io/zxcvbn/guide/migration/

ioanmo226:
This PR migrated zxcvbn to zxcvbn-ts

close #6018 // if this PR closes an issue

### To be filled by reviewers


extension/types/zxcvbn.d.ts
sosnovsky, line 1: what about using built-in `zxcvbn-ts` types? so we won't need to manually update them in the future

package-lock.json
sosnovsky, line 1: `package-lock.json` is not synchronized with `package.json` as still includes old `"zxcvbn": "4.4.2"`, running `npm install` removes this package record

sosnovsky (commented):
works well 👍

extension/types/zxcvbn.d.ts
ioanmo226: I couldn’t use the built-in `zxcvbn-ts` types because I didn’t use the ES module version.
The library doesn’t provide a single bundled `.esm` file like `zxcvbn-ts.mjs`, so to use ES modules, we’d have to copy the entire `dist` folder and include it manually.
I assumed using the CommonJS version would be sufficient, but let me know if you’d prefer I switch to ES modules.

PR #6108, opened 2025-10-28
#6105 Replace 'iso-8859-2' with plain JS

### LINKED ISSUE #6105: replace 'iso-8859-2' with plain JS
It seems we use `iso-8859-2` only at https://github.com/FlowCrypt/flowcrypt-browser/blob/71b033be7162df5924b09bb826d3613e66e08dae/extension/js/common/core/mime.ts#L485

But now this functionality can be implemented with plain js, using [TextDecoder](https://developer.mozilla.org/en-US/docs/Web/API/TextDecoder), like this:
```js
const decoder = new TextDecoder('iso-8859-2');
return decoder.decode(node.rawContent!);
```

ioanatflowcrypt:
This PR replaced 'iso-8859-2' with plain JS

close #6105 // if this PR closes an issue

### To be filled by reviewers


sosnovsky (approved):
well done 👍

PR #6116, opened 2025-11-19
#6115 Fix: Incorrect date issue when printing message

### LINKED ISSUE #6115: Incorrect date when printing message
Customer reported that in some cases message date is incorrect when using print function, as it displays date of the first message in the thread instead of the selected message date.

I was able to reproduce this issue:
- open thread with multiple encrypted messages
- click `expand all` button:

Image

- click `print` on the second message in the thread
- print page will use date of the first message, instead of the second

This bug isn't reproducible when first message isn't expanded - print page correctly uses date of the selected message in this case.
So seems like this bug in parsing message date happens when the first thread message is visible.

ioanatflowcrypt:
This PR fixed incorrect date issue when printing message

close #6115 // if this PR closes an issue

### To be filled by reviewers

### DIFF
--- a/extension/js/common/message-renderer.ts
+++ b/extension/js/common/message-renderer.ts
@@ -461,7 +461,18 @@ export class MessageRenderer {
 
   private getMessageInfo = async (fullMsg: GmailRes.GmailMsg): Promise<MessageInfo> => {
     const sentDate = GmailParser.findHeader(fullMsg, 'date');
-    const sentDateStr = sentDate ? new Date(sentDate).toLocaleString() : '';
+    let sentDateStr: string | undefined;
+
+    // Look for the date element within the specific message container
+    const msgContainer = $(`div.adn[data-legacy-message-id="${fullMsg.id}"]`);
+    if (msgContainer.length) {
+      sentDateStr = msgContainer.find('.gK span[title]').first().attr('title');
+    }
+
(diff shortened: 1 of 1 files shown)


extension/js/common/message-renderer.ts
```diff
@@ -461,10 +461,7 @@ export class MessageRenderer {
 
   private getMessageInfo = async (fullMsg: GmailRes.GmailMsg): Promise<MessageInfo> => {
     const sentDate = GmailParser.findHeader(fullMsg, 'date');
-    let sentDateStr = $('div.gK span[title]').attr('title');
-    if (!sentDateStr || isNaN(Date.parse(sentDateStr))) {
-      sentDateStr = sentDate ? new Date(sentDate).toLocaleString() : '';
-    }
+    const sentDateStr = sentDate ? new Date(sentDate).toLocaleString() : '';
```
sosnovsky: earlier we've used similar code, but got user feedback about incorrect format of date when printing messages, so we've switched to parsing message date directly from Gmail page (https://github.com/FlowCrypt/flowcrypt-browser/pull/5676)

so in this issue it'll be better to stick with webpage date parsing, but fix it to get correct message date

ioanatflowcrypt:
@sosnovsky Seems like you need to enable auto-merge

PR #6131, opened 2025-12-08
#6128 Use stripped url for exception

### LINKED ISSUE #6128: Use stripped URL for exceptions from mail.google.com
Currently we send full URL with exception reports (like https://mail.google.com/mail/u/0/#inbox/WhctKLbvMNLndrHSj..., https://mail.google.com/mail/u/0/#sent/KtbxLzFrMS..., https://mail.google.com/mail/u/1/#inbox/rtjXfHsNNgrJVZL..., https://mail.google.com/mail/u/1/#search/MSCGwzQrb...) which causes each report to appear as a separate exception in our dashboard because URLs are different.

Let's strip the URLs down to just https://mail.google.com/mail/ in these cases.

ioanatflowcrypt:
This PR uses stripped url for exception

close #6128 // if this PR closes an issue

### To be filled by reviewers


sosnovsky (commented):
not sure that this solution will work - it splits URLs using `?` as in the previous implementation, while this issue is for URLs which doesn't have `?`:
https://mail.google.com/mail/u/0/#inbox/WhctKLbvMNLndrHSj...
https://mail.google.com/mail/u/0/#sent/KtbxLzFrMS...
https://mail.google.com/mail/u/1/#inbox/rtjXfHsNNgrJVZL...
https://mail.google.com/mail/u/1/#search/MSCGwzQrb...

and let's also add test for such type of URLs, to be sure that they're grouped correctly.

extension/js/common/platform/catch.ts
```diff
@@ -262,12 +262,14 @@ export class Catch {
...
+    const reportUrl = location.href.split('?')[0];
     return {
       name: exception.name.substring(0, 50),
       message: Catch.groupSimilarReports(exception.message.substring(0, 200)),
-      // Use https://mail.google.com/mail as URL for content script errors
+      // Strip Gmail URLs to group similar errors from different threads/messages
       // https://github.com/FlowCrypt/flowcrypt-browser/issues/6031
-      url: Catch.RUNTIME_ENVIRONMENT === 'ex:s:gmail' ? 'https://mail.google.com/mail' : Catch.groupSimilarReports(location.href.split('?')[0]),
+      // https://github.com/FlowCrypt/flowcrypt-browser/issues/6128
+      url: Catch.RUNTIME_ENVIRONMENT.endsWith(':ex:s:gmail') ? 'https://mail.google.com/mail/' : Catch.groupSimilarReports(reportUrl),
```
ioanatflowcrypt: `Catch.RUNTIME_ENVIRONMENT` is initialized by the `Catch.environment()` function, which checks the origin of `location.href`. If the origin is `https://mail.google.com`, it sets the environment to `ex:s:gmail` and therefore url should be `https://mail.google.com/mail/`.

So I think this should work.

By the way, I totally agree that we should add a test for this case.


```ts
public static environment(url = location.href): string {
    const browserName = Catch.browser().name;
    const origin = new URL(location.href).origin;
... 
else if (origin === 'https://mail.google.com') {
      env = 'ex:s:gmail';
    }
```

PR #6166, opened 2026-02-06
#6165 Use presigned s3 urls for password protected messages

### LINKED ISSUE #6165: Use pre-signed S3 URLs for password-protected messages
Currently password-protected messages attachments have maximum size ~5mb. To support larger attachments, we implemented new API in FES - it uses pre-signed S3 URLs, so messages are uploaded directly from browser extension.

This change requires some code changes:
- instead of using `/api/v1/message/new-reply-token`, we should use new `POST /api/v1/messages/allocation` endpoint. It returns response `{ "storageFileName": string, "replyToken": string, "uploadUrl": string }`
- then browser extension should upload generated password-protected message content to S3 (using `storageFileName` as file name) by sending PUT request to `uploadUrl`
- after successful message upload, extension should send `POST /api/v1/messages` request with structure similar to current `POST /api/v1/message` request in [webMessageUpload](https://github.com/FlowCrypt/flowcrypt-browser/blob/a60803421865af9dd0b00265d21223f35b4f2c74/extension/js/common/api/account-servers/external-service.ts#L143), but instead of including message content as `encrypted.asc` file, request should use `storageFileName` property with value received from `POST /api/v1/messages/allocation` request

Also, let's leave ability to use our existing api - so both these implementations will be available in browser extension. By default, new flow with pre-signed S3 URL should be used, but if client configuration has `DISABLE_FES_PRESIGNED_URLS` property - extension should use current flow with message upload to FES.

For now, this functionality is enabled only for `...@flowcrypt.com` users, so for testing please use `...@flowcrypt.com` email address. If you'll have any questions or find some issues - please mention me or @Jerbell.

ioanatflowcrypt:
This PR implemented feature to use presigned s3 urls for password protected messages

close #6165 // if this PR closes an issue

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/test/source/mock/fes/customer-url-fes-endpoints.ts
+++ b/test/source/mock/fes/customer-url-fes-endpoints.ts
@@ -6,13 +6,8 @@ import { HandlersDefinition } from '../all-apis-mock';
 import { HttpClientErr, Status } from '../lib/api';
 import { messageIdRegex, parseAuthority, parsePort } from '../lib/mock-util';
 import { MockJwt } from '../lib/oauth';
-import {
-  FesConfig,
-  MessageCreateBody,
-  validateMessageCreateBody,
-  createMockBodyForValidator,
-  generateEmailToExternalIdAndUrl,
-} from './shared-tenant-fes-endpoints';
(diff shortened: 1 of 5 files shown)


sosnovsky (approved):
works great 👍

PR #6219, opened 2026-05-07
#6218 Use extension-specific oauth redirect url

### LINKED ISSUE #6218: Use extension-specific OAuth redirect URL instead of Google robots.txt
Reference: https://github.com/FlowCrypt/flowcrypt-security/issues/322#issuecomment-4387006511

martgil:
This PR replaces the Google robots.txt redirect URL to extension specific oauth redirect URL.

Closes #6218

### To be filled by reviewers


sosnovsky (commented):
Hi @martgil, a lot of tests are failing because local dev extension builds use dynamic extension id. Maybe there is some way to use fixed id for local builds, so we can add this id to the list of allowed redirect URIs for our OAuth app.

martgil:
You're correct, Roma. I'll look for some other alternative.

martgil:
Hello @sosnovsky - Just a quick question. The Mozilla Firefox extension id has also this dynamically generated id, does this means, the extension id for both builds should matches exactly identical? Or does the Mozilla firefox build has to had the static extension id too?

sosnovsky:
`identity.getRedirectURL()` in firefox uses id from `browser_specific_settings` property of `manifest.json` (https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity/getRedirectURL), so probably it shouldn't change between builds and remain static

sosnovsky:
also firefox uses `browser.identity.getRedirectURL()` instead of `chrome.identity.getRedirectURL()`

martgil:
Got it - thanks!

martgil:
Hello @sosnovsky - I have managed to make a static extension id from which the current one we have on this PR is `nnlimfkjgafpgmilbnaabplbokhpicbo`. I have certificate where the key is originated from. It it technically a test file but should I keep it uploaded here somewhere in `test/`?

sosnovsky:
is this certificate required for creating local builds? if no, then there is no need to put it in the repository
also your current solution adds `key` property directly to `manifest.json` which will probably affect our production builds. I think we should have a separate `chrome-consumer-local` build and inject `key` property only for this build type

martgil:
> is this certificate required for creating local builds? if no, then there is no need to put it in the repository

Edit: No, it is not required for creating the local builds. Thanks for your answer.

> I think we should have a separate chrome-consumer-local build and inject key property only for this build type

Got it - will do.

martgil:
Hi @sosnovsky - May I here your feedback on this one, please? I may have missed important ones.

extension/manifest.json
```diff
@@ -58,7 +58,7 @@
       ]
     },
     {
-      "matches": ["https://www.google.com/robots.txt*"],
+      "matches": ["https://www.google.com/oauth2/callback*"],
```
sosnovsky: it'll probably won't work, as we now use extension id specific redirect URIs, and https://www.google.com/oauth2/callback won't be called causing auth process stuck.

I also re-checked `launchWebAuthFlow` documentation, and looks like it should work with our current OAuth Client ID, so users won't need to re-authenticate and their current refresh tokens will stay valid. Also switching to `launchWebAuthFlow` will allow us to remove a lot of our custom OAuth code (like `oauth2_inject.ts`, `oauth2_finish.ts` etc).

So let's try to migrate auth to `launchWebAuthFlow`, hopefully it won't require a lot of code changes and we already have it implemented in `ConfiguredIdpOAuth`, so you can re-use some code from it.

martgil: Thanks Roma. I have followed your suggestions. It doesn't seem to have required alot of code changes. Can you please give it another check?

extension/js/common/api/authentication/google/google-oauth.ts
```diff
@@ -211,24 +191,34 @@ export class GoogleOAuth extends OAuth {
...
     requestedScopes: string[];
     expectedState: string;
-    authWindowResult: Bm.AuthWindowResult;
+    authUrl: string;
   }): Promise<AuthRes> {
     /* eslint-disable @typescript-eslint/naming-convention */
     try {
-      if (!authWindowResult.url) {
-        return { acctEmail, result: 'Denied', error: 'Invalid response url', id_token: undefined };
+      const redirectUri = await chrome.identity.launchWebAuthFlow({ url: authUrl, interactive: true });
```
sosnovsky, line 204: it works great in chrome, but fails in firefox which uses another format for OAuth redirect URIs - `http://127.0.0.1/mozoauth2/[subdomain of URL returned by identity.getRedirectURL()]` (https://stackoverflow.com/a/72968217/3091318).
please fix it so it works in Firefox too

extension/js/common/api/authentication/google/google-oauth.ts
```diff
@@ -211,24 +191,34 @@ export class GoogleOAuth extends OAuth {
...
+        if (
+          errorMsg.toLowerCase().includes('user') ||
+          errorMsg.toLowerCase().includes('cancel') ||
+          errorMsg.toLowerCase().includes('deny') ||
+          errorMsg.toLowerCase().includes('denied') ||
+          errorMsg.toLowerCase().includes('close')
+        ) {
+          return { acctEmail, result: 'Closed', error: errorMsg, id_token: undefined };
+        }
+        return { acctEmail, result: 'Denied', error: `Failed to launch web auth flow: ${errorMsg}`, id_token: undefined };
```
sosnovsky, line 212: let's simplify this part:
```ts
const errorMsg = chrome.runtime.lastError?.message || 'access_denied';
const normalizedErrorMsg = errorMsg.toLowerCase();

const userCancelled = ['user', 'cancel', 'deny', 'denied', 'close'].some((keyword) =>
    normalizedErrorMsg.includes(keyword)
);

if (userCancelled) {
   return { acctEmail, result: 'Closed', error: errorMsg, id_token: undefined };
}

...
```

sosnovsky (commented):
well done 👍

extension/js/common/api/authentication/google/google-oauth.ts
martgil: @sosnovsky Thanks for testing this out on the Firefox. Do you think we should have a dedicated test for the Mozilla firefox browser for this? Or would it be sufficient to just implement a fix for the firefox built itself?

extension/js/common/api/authentication/google/google-oauth.ts
martgil: Thanks @sosnovsky - I'll simplify that part accordingly.

extension/js/common/api/authentication/google/google-oauth.ts
martgil: @sosnovsky is it correct that you'll then be needing to add such url to Google: `https://[extension-id].extensions.mozilla.org/` for Firefox?

sosnovsky: I added http://127.0.0.1/mozoauth2/[extension-id] to allowed redirect URLs, as it's not possible to add domains like `mozilla.org`, because Google requires domain ownership in such cases. http://127.0.0.1/mozoauth2/[extension-id] is a solution from Firefox to have extension-specific redirect URLs.

martgil: Thanks @sosnovsky ! This should confirm the fix we applied. The failing test is unrelated to the changes I've made. It should be good to go.

sosnovsky: >  Do you think we should have a dedicated test for the Mozilla firefox browser for this? Or would it be sufficient to just implement a fix for the firefox built itself?

No need to add dedicated test, as previously we tried to add firefox-specific tests but it required too much effort, so let's just implement a fix and test it manually.

martgil:
Hi @sosnovsky - I have manually tested it on Firefox and it works well now. Then tests for the chrome build have passed too. Feel free to re-check at your convenience.

extension/js/common/api/authentication/generic/oauth.ts
```diff
@@ -65,7 +65,13 @@ export class OAuth {
     if (!chrome?.identity?.getRedirectURL) {
       throw new Error('chrome.identity.getRedirectURL is not available in this context');
     }
-    return chrome.identity.getRedirectURL('oauth');
+    const redirectUri = chrome.identity.getRedirectURL('oauth');
+    if (navigator.userAgent.includes('Firefox')) {
```
sosnovsky: let's use existing `Catch.isFirefox()` check

sosnovsky (commented):
works great now, well done 👍

extension/js/common/api/authentication/generic/oauth.ts
martgil: Done - Thank you, Roma :+1:

sosnovsky (approved):
all good now, thanks!

PR #6227, opened 2026-05-20
#6226 Add test case for handling executable file downloads in decrypt tests

### LINKED ISSUE #6226: Improve exectuable file attachment detection
Reference: https://github.com/FlowCrypt/flowcrypt-security/issues/331

martgil:
This PR adds a new test for executable file downloads, handles the dangerous download confirmation modal, and verifies the downloaded file format and filename.

Closes #6226 

- Tests added or updated

### To be filled by reviewers


extension/js/common/core/attachment.ts
```diff
@@ -294,7 +294,7 @@ export class Attachment {
       'wsf',
       'wsh',
       'xll',
-    ].some(exeFileExtension => this.name.endsWith('.' + exeFileExtension));
+    ].some(exeFileExtension => this.name.toLocaleLowerCase().endsWith('.' + exeFileExtension));
```
sosnovsky: we should use locale-independent `toLowerCase` here, as `toLocaleLowerCase` can produce some unexpected results. for example, in Turkish locale uppercase `I` will be converted to dotless `ı`, causing `disk.ISO` attachment to pass this check.

martgil: Thanks @sosnovsky 👍  - I'll be applying your recommendation.

extension/js/common/core/attachment.ts
```diff
@@ -257,10 +257,13 @@ export class Attachment {
       'diagpack',
```
sosnovsky: Google's blocked extensions list has `diagpkg` instead of `diagpack`, probably there is a typo in our extensions list? there is also `mjs` extension in their list, let's add it here too

extension/js/common/core/attachment.ts
```diff
@@ -289,12 +297,16 @@ export class Attachment {
...
       'vxd',
       'wsc',
       'wsf',
       'wsh',
       'xll',
-    ].some(exeFileExtension => this.name.endsWith('.' + exeFileExtension));
+      'xlsm',
+      '7z',
+      'ace',
+    ].some(exeFileExtension => this.name.toLowerCase().endsWith('.' + exeFileExtension));
```
sosnovsky, line 311: our current warning message says `This executable file was not checked for viruses, and may be dangerous to download or run`, but not all file types are executable in this list (like `7z` or `pub`).
let's also update warning message to `This attachment can be unsafe and was not checked for viruses. Only proceed if you trust the sender and expected this file.`

extension/js/common/core/attachment.ts
martgil: Thanks @sosnovsky. I also ensured that all file extensions blocked by google were also blocked in this list.

extension/js/common/core/attachment.ts
martgil: Thanks, that is more general and less alarming.

sosnovsky (approved):
looks good now, thanks!

PR #6234, opened 2026-05-29
#6233 Improve css validation

### LINKED ISSUE #6233: Improve CSS validation
Reference: https://github.com/FlowCrypt/flowcrypt-security/issues/335

martgil:
This PR improves currently implemented css validation on rendered elements in `pgp_block`.

Closes #6233 

### To be filled by reviewers


martgil:

extension/js/common/platform/xss.ts
```diff
@@ -118,6 +118,7 @@ export class Xss {
         const style = node.getAttribute('style')?.toLowerCase();
         if (style && (style.includes('url(') || style.includes('@import'))) {
           node.removeAttribute('style'); // don't want any leaks through css url()
+          return; // stop processing: do not re-add any part of this style attribute
```
sosnovsky: if `return` will be executed here, then all next checks will be skipped?

so if `style` property includes some URL, then other checks for `node.tagName === 'IMG'` and `node.tagName === 'A'` won't be executed?
for example, `` will return `` - `style` property removed, but remote image is loaded

martgil: Yes, but I realize I should formulate even better logic to fix this rather than what's current. BRB.

martgil:
Hi @sosnovsky - This one is ready for a review. Thanks!

extension/js/common/platform/xss.ts
```diff
@@ -115,14 +116,16 @@ export class Xss {
...
         // mitigation rather than a fix, which will involve updating CSP, see https://github.com/FlowCrypt/flowcrypt-browser/issues/2648
-        const style = node.getAttribute('style')?.toLowerCase();
-        if (style && (style.includes('url(') || style.includes('@import'))) {
-          node.removeAttribute('style'); // don't want any leaks through css url()
-        }
-        // strip css styles that could use to overlap with the extension UI
+        let style = node.getAttribute('style') || '';
+        style = Xss.sanitizeCssStyle(style);
         if (style && Xss.FORBID_CSS_STYLE.test(style)) {
           const updatedStyle = style.replace(Xss.FORBID_CSS_STYLE, '');
```
sosnovsky, line 122: in this implementation, our current regex from `FORBID_CSS_STYLE` will replace just first found property prefix, while leaving other dangerous properties as is:

> "position:absolute;z-index:999999;top:0;left:0;width:100%;height:100%;background:white".replace(/(?:^|;)\s*(?:z-index|position|display|visibility|opacity|transform|clip-path|clip|top|left|right|bottom|pointer-events|font-size|line-height|width|height|text-indent|filter)\s*:/i, '');
> // will return "absolute;z-index:999999;top:0;left:0;width:100%;height:100%;background:white"

martgil: Done. Now matches all, not just first.

martgil:
Hi @sosnovsky - This one is ready for review. Thank you!

extension/js/common/platform/xss.ts
```diff
@@ -49,7 +49,8 @@ export class Xss {
   private static ADD_ATTR = ['email', 'page', 'addurltext', 'longid', 'index', 'target', 'fingerprint', 'cryptup-data'];
   private static FORBID_ATTR = ['background'];
   private static HREF_REGEX_CACHE: RegExp | undefined;
-  private static FORBID_CSS_STYLE = /z-index:[^;]+;|position:[^;]+;|background[^;]+;/g;
+  private static FORBID_CSS_STYLE =
+    /(?:^|;)\s*(?:z-index|position|display|visibility|opacity|transform|clip-path|clip|top|left|right|bottom|pointer-events|font-size|line-height|width|height|text-indent|filter)\s*:[^;]*;?/gi;
```
sosnovsky, line 53: looks like this regex fully removes only first statement, and for next ones it removes just `;`:

> "position:absolute;z-index:999999;top:0;left:0;width:100%;height:100%;background:white".replace(/(?:^|;)\s*(?:z-index|position|display|visibility|opacity|transform|clip-path|clip|top|left|right|bottom|pointer-events|font-size|line-height|width|height|text-indent|filter)\s*:[^;]*;?/gi, ''); // returns "z-index:999999left:0height:100%;background:white"

fixed regex should look like `/(^|;)\s*(?:z-index|position|display|visibility|opacity|transform|clip-path|clip|top|left|right|bottom|pointer-events|font-size|line-height|width|height|text-indent|filter)\s*:[^;]*(?=;|$)/gi`

and let's also add test for such cases, to be sure that our implementation works correctly, thanks!

extension/js/common/platform/xss.ts
```diff
@@ -274,6 +277,33 @@ export class Xss {
...
+    while ((match = urlRegex.exec(cleaned)) !== null) {
+      const fullMatch = match[0];
+      const url = match[2];
+      // Only allow data: and cid: schemes
+      const isSafe = /^(data:|cid:)/i.test(url);
+      if (!isSafe) {
+        // Remove the unsafe url(...) token completely
+        cleaned = cleaned.replace(fullMatch, '');
+      }
+    }
```
sosnovsky: let's replace this `while` loop with a simpler single `replace` statement:

```ts
cleaned = cleaned.replace(/url\(\s*(["']?)(.*?)\1\s*\)/gi, (fullMatch: string, _, url: string) => {
      // Only allow data: and cid: schemes
      const isSafe = /^(data:|cid:)/i.test(url);
      return isSafe ? fullMatch : '';
    });

Comment (martgil) [reply extension/js/common/platform/xss.ts]: Got it - this is much more cleaner

File: extension/js/common/platform/xss.ts (hunk 3 above)
Comment (martgil) [reply extension/js/common/platform/xss.ts:53]: Thank you - will craft a test email and then write a test.

**Comment by martgil** [comment]
The failing test is not related to this PR but from a fragile test.

File: extension/js/common/platform/xss.ts [hunk 5]
```diff
@@ -274,6 +277,26 @@ export class Xss {
     }
   };
 
+  /**
+   * Remove @import rules and any url(...) that would cause an out‑of‑band request.
+   * Only data: and cid: URLs are allowed.
+   */
+  private static sanitizeCssStyle = (css: string): string => {
+    return css
+      .split(';')
```
sosnovsky, line 303: new regex works well, but I noticed one more possible passthrough - if attacker adds css comment before the property name, it won’t be caught by our regex. for example, `style="/*x*/position:absolute;/*x*/top:0;/*x*/left:0;/*x*/z-index:999"` survives our sanitization.

to fix it let's remove all css comments before processing css. it can be done by adding `.replace(/\/\*[\s\S]*?\*\//g, '')` before `.split(';')` line

martgil: Thanks @sosnovsky - this is a must. I'll add it.

martgil:
Hi @sosnovsky - Thanks for your suggestions, I think this one is ready to merge.

extension/js/common/platform/xss.ts
```diff
@@ -274,6 +277,27 @@ export class Xss {
...
+   * Only data: and cid: URLs are allowed.
+   */
+  private static sanitizeCssStyle = (css: string): string => {
+    return css
+      .replace(/\/\*[\s\S]*?\*\//g, '')
+      .split(';')
+      .map(part => part.trim())
+      .filter(part => {
+        return !/^(z-index|position|display|visibility|opacity|transform|clip-path|clip|top|left|right|bottom|pointer-events|font-size|line-height|width|height|text-indent|filter)\s*:/i.test(
+          part
```
sosnovsky, line 307: I found one more possible way to bypass our current checks - attacker can use escaped symbols to bypass this regex. for example, `style="background-image: u\72l(https://attacker.example/pixel.png);"` will pass our check and then converted to `style="background-image: url(https://attacker.example/pixel.png);"`.
so let's additionally normalize all escaped symbols before performing regex checks:
```ts
private static normalizeCssEscapes = (css: string): string => {
    return css.replace(/\\(?:\r\n|[\n\r\f])|\\([0-9a-fA-F]{1,6}\s?|.)/g, (_match: string, escaped: string | undefined) => {
      if (typeof escaped === 'undefined') {
        return '';
      }
      if (/^[0-9a-fA-F]/.test(escaped)) {
        const codePoint = Number.parseInt(escaped.trim(), 16);
        return codePoint > 0 && codePoint <= 0x10ffff ? String.fromCodePoint(codePoint) : '';
      }
      return escaped;
    });
  };
```

then we can replace `.map(part => part.trim())` with `.map(part => normalizeCssEscapes(part.trim()))`

martgil: Thanks, @sosnovsky. I will implement this additional countermeasure.

sosnovsky (approved):
perfect, thanks for the fixes!

PR #6237, opened 2026-06-05
#6236 Improve URL paths handling

### LINKED ISSUE #6236: Encode dynamic URL paths and query parameters
In some places, we build URL strings with raw interpolated values, including Attester public key lookup paths and query strings. Dynamic path segments should use `encodeURIComponent`, and query strings should use `URLSearchParams` or `Url.create` to prevent encoded traversal or parameter injection from changing the intended request target.

sosnovsky:
This PR adds `encodeURIComponent` for dynamic path segments + uses `URLSearchParams` and `Url.create` for query strings

close #6236

- Tests added or updated

### To be filled by reviewers


martgil (approved):
LGTM - Thanks for the fix!

PR #6270, opened 2026-08-05
#6269 improve img srcset sanitation checks

### LINKED ISSUE #6269: Sanitize srcset attribute on images
Reference: https://github.com/FlowCrypt/flowcrypt-security/issues/361

martgil:
This PR improves img srcset sanitation checks on remote images loaded in content script.

close #6269 

- Tests added or updated

### To be filled by reviewers

### DIFF
--- a/extension/js/common/platform/xss.ts
+++ b/extension/js/common/platform/xss.ts
@@ -196,14 +196,24 @@ export class Xss {
       if (node.tagName === 'IMG') {
         const img = node as HTMLImageElement; // Narrow type to HTMLImageElement
         const src = img.getAttribute('src');
+        const srcset = img.getAttribute('srcset');
+        // an image is remote when `src` or any `srcset` candidate references an external URL; `srcset`
+        // (and protocol-relative `//` URLs) must be covered too, otherwise they could be used to trigger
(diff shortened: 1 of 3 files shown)


martgil:

extension/js/common/platform/xss.ts
```diff
@@ -196,14 +196,24 @@ export class Xss {
       if (node.tagName === 'IMG') {
         const img = node as HTMLImageElement; // Narrow type to HTMLImageElement
         const src = img.getAttribute('src');
+        const srcset = img.getAttribute('srcset');
```
sosnovsky, line 199: I think better solution will be to ignore `srcset` property by removing it before parsing with `img.removeAttribute('srcset');`, since basic `src` support is enough for loading images and `srcset` just adds support for responsive images. Gmail doesn't support `srcset` property too, as noted at https://www.caniemail.com/features/html-srcset/

However, we should apply similar URL check for `src` property, since attacker can use the same `/Logout` link there too.

martgil: Thanks - I'll try to apply your proposed solution. I also realized that sanitation wont work this time, in addition to the fact that Gmail doesn't support srcset. Thanks again for this insight.

PR #6277, opened 2026-08-21
#6276 Improve message parsing for signed message

### LINKED ISSUE #6276: Client slowdown when parsing recursive multipart/signed messages
Reference: https://github.com/FlowCrypt/flowcrypt-security/issues/360

martgil:
This PR improve message parsing for signed message and prevents recursive parsing.

close #6276 

- Tests added or updated

### To be filled by reviewers


martgil:
Hi @sosnovsky - This one is ready for a review. Thanks!

extension/js/common/core/mime.ts
```diff
@@ -356,30 +357,70 @@ export class Mime {
...
+
+  private static getSubjectFromNode(node: MimeParserNode): string | undefined {
+    let current = node;
+    let fallbackSubject: string | undefined;
+    for (let depth = 0; depth <= Mime.MAX_SIGNED_CONTENT_DEPTH; depth++) {
+      const currentSubject = current.headers.subject?.[0]?.value;
+      if (!Mime.isSignedContentNode(current)) {
+        return currentSubject || fallbackSubject;
+      }
+      fallbackSubject = currentSubject;
```
sosnovsky: I think it'll be better to use `fallbackSubject = currentSubject || fallbackSubject` for cases when `currentSubject` is undefined, so we can use previous non-null value.

extension/js/common/core/mime.ts
```diff
@@ -356,30 +357,70 @@ export class Mime {
...
+      /* eslint-disable no-underscore-dangle */
+      const childNodes = current._childNodes;
+      if (childNodes === false || !childNodes.length) {
+        return currentSubject || fallbackSubject;
+      }
+      current = childNodes[0];
+      /* eslint-enable no-underscore-dangle */
     }
-    /* eslint-enable no-underscore-dangle */
     return undefined;
```
sosnovsky: also here we should return `fallbackSubject` instead of `undefined`, for cases when in nested nodes some deep node is `undefined`, but there was some non-null node in previous nodes

extension/js/common/core/mime.ts
```diff
@@ -60,6 +60,7 @@ export type MimeProccesedMsg = {
 type SendingType = 'to' | 'cc' | 'bcc';
 
 export class Mime {
+  private static readonly MAX_SIGNED_CONTENT_DEPTH = 50;
```
sosnovsky, line 63: there is similar recursive call in https://github.com/FlowCrypt/flowcrypt-browser/blob/78f34e3c4ba7bb23f28faee544653cc118bed71b/extension/js/common/message-renderer.ts#L618

which can cause extension lag for nested signed content. we'll need to implement similar fix there, but let's leave it for separate PR

sosnovsky (commented):
works well, just a couple minor improvements 👍

extension/js/common/core/mime.ts
martgil: Got it - Will remember this one and fix it on separate PR after this one.

extension/js/common/core/mime.ts
martgil: Thanks Roma, I have tried to correct this one.

extension/js/common/core/mime.ts
martgil: Got it - thanks!

martgil:

extension/js/common/core/mime.ts
```diff
@@ -356,33 +357,73 @@ export class Mime {
...
+
+  private static getSubjectFromNode(node: MimeParserNode): string | undefined {
+    let current = node;
+    let fallbackSubject: string | undefined;
+    for (let depth = 0; depth <= Mime.MAX_SIGNED_CONTENT_DEPTH; depth++) {
+      const currentSubject = current.headers.subject?.[0]?.value;
+      if (!Mime.isSignedContentNode(current)) {
+        return currentSubject || fallbackSubject;
+      }
+      fallbackSubject = currentSubject || fallbackSubject;
```
sosnovsky: it's possible for `multipart/mixed` content to include both signed and unsigned parts. current code stops when it encounters a non-signed node, so it can miss a nested signed node. let’s perform search for another signed node in this case:
```ts
const currentSubject = current.headers.subject?.[0]?.value;
fallbackSubject = currentSubject || fallbackSubject;
if (!Mime.isSignedContentNode(current)) {
    const nestedSignedContentNode = Mime.retrieveSignedContentNode([current]);
    if (!nestedSignedContentNode) {
          return fallbackSubject;
    }
    current = nestedSignedContentNode;
    continue;
}
```

martgil: I apologize, I missed that part. I have rechecked it again and fix the said behavior.

martgil:
Hi @sosnovsky - This one is ready for a review. Thanks!

sosnovsky (approved):
all good now, thanks!

PR #6298, opened 2026-09-22
#6297 Improve promises handling in ExpirationCache

### LINKED ISSUE #6297: Improve ExpirationCache promises cache implementation
`ExpirationCache` currently attempts to store `Promise` objects in `chrome.storage.session`, which only supports serializable values (https://developer.chrome.com/docs/extensions/reference/api/storage). Rejected promises also bypass the current try/catch because they are returned without `await`.
Additionally, `downloader.ts` evicts full-message requests from `rawMessages` instead of `fullMessages`.

We should add in-memory `Map<string, Promise<T>>` for storing pending requests and use `storage.session` caching only for serializable values.

sosnovsky:
This PR fixes promise caching in `ExpirationCache` and corrects full-message cache eviction.

close #6297

- Tests added or updated

### To be filled by reviewers


martgil (approved):
LGTM. Nice cleanup 👍

PR #6300, opened 2026-09-22
#6299 Use native browser APIs for timeout promises

### LINKED ISSUE #6299: Simplify fetch progress and timeout handling
Extension creates many timeout promises that are never cleared:
https://github.com/FlowCrypt/flowcrypt-browser/blob/b652839a68207270ee00615d030909e4ea6255e7/extension/js/common/api/shared/api.ts#L129

We should replace manual timeout promises and response stream pumping with native browser APIs. Use `AbortSignal.timeout()` for request timeouts and `ReadableStream.pipeThrough()` with a `TransformStream` for download progress. Continue using XHR for upload progress until request streams are reliable.

sosnovsky:
This PR replaces manual timeout promises and response streaming with native browser APIs

close #6299

### To be filled by reviewers


extension/js/common/api/shared/api.ts
```diff
@@ -178,29 +134,29 @@ export class Api {
...
-    const abortController = new AbortController();
-    const requestInit: RequestInit & { duplex?: 'half' } = {
+    const requestInit: RequestInit = {
       method: req.method,
       headers: headersInit,
       body,
-      duplex,
       mode: 'cors',
-      signal: abortController.signal,
+      signal: AbortSignal.timeout(req.timeout ?? 20000),
```
martgil: This changes download timeout from a rolling idle timeout (timer resets per chunk) to a hard AbortSignal.timeout(req.timeout ?? 20000). Gmail attachment downloads (gmail.ts:171) have no custom timeout, so a large attachment streaming for >20s now fails where it previously succeeded. Note ajaxWithJquery deliberately sets timeout: undefined for progress requests (api.ts:280).

Maybe we could give download-progress requests a generous ceiling instead of the tight default: 

signal: AbortSignal.timeout(req.timeout ?? (req.progress?.download ? 5 * 60 * 1000 : 20000)),

sosnovsky: thanks for noticing! I updated code to use similar per-chunk timeout reset, like we had before

martgil (approved):
LGTM now 👍

